PR #310, opened 2015-01-07
feature: af_array_index

9prady9:
* af_array_index lets you use array to index another array
* cpp wrapper for this works only along 0th dimension as of now


jeremysalwen:
I have a use for array indexing along the second dimension from the C++ wrapper.  

For now I am Transposing before and after as a workaround.

pavanky:
@jeremysalwen You can us [lookup](https://github.com/arrayfire/arrayfire/blob/devel/include/af/index.h#L36) to do this right now. This merge does not add the C++ support yet.

PR #317, opened 2015-01-08
Split sort_by_key instantiation into multiple files for CUDA backend

shehzan10:
* Compilation times out on Tegra K1


shehzan10:
I have built this on the tegra a few times and each time it passed. So it can be merged and this should fix time out issues.

PR #344, opened 2015-01-16
Added pi definition to fix ORB on Windows

pentschev:


shehzan10:
Pi is defined in utils.h. You can use that one.

pavanky:
Merging it for now. Fix this in a later commit.

PR #354, opened 2015-01-22
Fixes ORB for cases where no features exist

pentschev:


gpryor:
Nice - - this is supported on CPU, too?

If so I'll need to try it out on my ARM box. Do you have the standard test imagery posted anywhere?

pentschev:
@gpryor yes, it is supported on all backends, but CPU might not be fully optimized yet. There's no test imagery as of now, as I am still working on further requirements (e.g., Hamming matcher). I hope to have some more interesting examples once we have graphics support.

gpryor:
Fantastic. I'll play with the ORB code on my arm box and let you know how it goes. I'll donate some test imagery, too.

pentschev:
@gpryor that would be awesome. There are actually two test images for unit tests only, but not to show ORB at work.

PR #359, opened 2015-01-28
CUDA cannot be linked against libstdc++ on OSX 

jramapuram:
Resolve issue with CUDA 7.0 causing link errors with thrust. 
CUDA 7.0 onwards does not support linking against libstdc++


umar456:
This block will not work in CUDA 6.5 because it only supports clang with libstdc++ on OSX. We need to check which toolkit is being used before we set the standard library to libc++.

pavanky:
While this is true, we may need to have a hard dependency on CUDA 7 because of cuda solver library. 

@bkloppenborg Will probably not like this happening though.

bkloppenborg:
We can use the `CUDA_COMPUTE_CAPABILITY` CMake variable that comes from the `CMakeModules/CUDACheckCompute.cmake` script to detect the compute version. From this we can switch to `libc++` for the appropriate compute level.

Concerning the prospect of a hard dependency on Compute >= 7.0 I am not a fan as even Ubuntu 14.10 only has CUDA 6.0 in its standard repositories. I have found many researchers in my field do almost anything to avoid updating their machines because they have lot of custom-compiled code that can take days to setup again after a system update or there are other administrative barriers to upgrading (e.g. security reviews).

jramapuram:
Used ${CUDA_VERSION_MAJOR} to do this.

pavanky:
:+1:

pavanky:
@bkloppenborg If we do support older versions of CUDA, we will be limiting the users to a subset of the entire ArrayFire functionality.

This is something we can do.

src/backend/cuda/CMakeLists.txt
```diff
@@ -129,9 +129,15 @@ CL_KERNEL_TO_H(
 IF("${APPLE}")
     ADD_DEFINITIONS(-D__STRICT_ANSI__)
     IF(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
-        ADD_DEFINITIONS(-stdlib=libstdc++)
-        SET(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -stdlib=libstdc++")
-        SET(CMAKE_STATIC_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -stdlib=libstdc++")
+        IF(${CUDA_VERSION_MAJOR} VERSION_LESS 7)
+            SET(STD_LIB_BINDING, "-stdlib=libstdc++")
```
umar456: SET's arguments are separated by spaces not commas.

jramapuram: Yup, weird that it worked though. Fixed that in the new commit. Thanks

PR #379, opened 2015-02-16
BUGFIX Fixed sobel output types

shehzan10:
* Test was failing for char on tegra
* All types other than double now return type float (double returns double)


9prady9:
:+1:

PR #386, opened 2015-02-17
Indexed min and max operations for all backends

### LINKED ISSUE #385: FEAT: Indexed min and max for all backends.

pavanky:
- Fixes #385


9prady9:
:+1: for tests

PR #403, opened 2015-02-25
Fixes memory leaks on FAST and ORB

pentschev:


pavanky:
Going to solve this differently. Look at #406

PR #407, opened 2015-02-26
Changes to backend API

### LINKED ISSUE #109: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](http://www.parashift.com/c++-faq/delete-handles-null.html "Do I need to check for NULL before delete p?") like the following.
* [opencl::copyData](https://github.com/arrayfire/arrayfire/blob/58d104342ea7a03db7e7c08561041ba1a7e8a4ca/src/backend/opencl/copy.cpp#L21 "opencl::copyData function")
* [string::resize](https://github.com/arrayfire/arrayfire/blob/58d104342ea7a03db7e7c08561041ba1a7e8a4ca/src/backend/opencl/cl.hpp#L535 "string::resize function")

### LINKED ISSUE #406: Buffers not being deleted when going out of scope
This mostly happens on empty buffers. This can result in the memory manager searching through a large number of buffers eventually.

This can lead to a slow down because the memory manager is trying to garbage collect over and over again.

pavanky:
- All backend functions now return `Array<T>` instead of `Array<T> *`
- This works now because raw data is stored in a shared_ptr

- Fixes #403 
- Fixes #406 
- Fixes #109 (No longer relevant)

### DIFF
--- a/src/api/c/approx.cpp
+++ b/src/api/c/approx.cpp
@@ -23,14 +23,14 @@ template<typename Ty, typename Tp>
 static inline af_array approx1(const af_array in, const af_array pos,
                                const af_interp_type method, const float offGrid)
 {
-    return getHandle(*approx1<Ty>(getArray<Ty>(in), getArray<Tp>(pos), method, offGrid));
+    return getHandle(approx1<Ty>(getArray<Ty>(in), getArray<Tp>(pos), method, offGrid));
 }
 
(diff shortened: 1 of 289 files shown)


pavanky:
Merge this after tests pass in our ci.

9prady9:
:+1:

PR #415, opened 2015-02-26
Remove C++11 conditional from src/api/c.

umar456:
There was a std::contitional in src/api/c. Removed it because
CUDA on OSX cannot compile with C++11 on CUDA framework < 7.0


9prady9:
:+1:

gpryor:
That's a shame. It's amazing because LLVM 3.6 is imminent and I believe that they are due to be supporting C++14.

PR #424, opened 2015-02-28
Use the chromium repository to build gtest.

umar456:
Updated the project so that we are using GTest from the Chromium
project

* GTest is downloaded as a submodule
* CMake builds gtest just as before
* Files are not redownloaded after you destroy the build dir
* Updated GTest to the current master. (No need for Windows patch)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "test/data"]
 	path = test/data
 	url = https://github.com/arrayfire/arrayfire_data
+[submodule "test/gtest"]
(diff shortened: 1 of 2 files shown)


shehzan10:
:+1:

pavanky:
:+1:  This solves so many irritating issues. Thanks!

shehzan10:
I guess we can update the wiki instructions. I don't think subversion is a requirement anymore.

9prady9:
shouldn't be, svn is used only in gtest to update it.

Issue #378, opened 2015-02-13
colorspace conversion feature


9prady9:
Possible API 
```c
af_err af_colorspace(af_array *out, const af_array image, af_cspace_t to, af_cspace_t from)
```

```cpp
array colorspace (const array &image, CSpace to, CSpace from)
```

9prady9:
Progress update: Upon discussion with @pavanky and @shehzan10 , we decided that we shall finish independent functions for each type of conversion and then write a wrapper for those functions which will look similar to the API suggested above.

9prady9:
Finished the following individual functions:
* `rgb2gray`
* `gray2rgb`
*  `hsv2rgb`
*  `rgb2hsv`

We will proceed to `colorspace` now.

PR #427, opened 2015-03-01
FEATURE: hsv to rgb and vice versa conversion functions

9prady9:
Related to #378


9prady9:
@pavanky This is ready for merge.

Issue #429, opened 2015-03-01
Reduce compile times for ArrayFire


pavanky:
Template instantiations are useful to improve the runtime performance, but the time taken to build the libraries is much longer.

This issue is open so people can provide suggestions on how to reduce the compile times.

Some suggestions already brought up internally at arrayfire:

- The number of templates can be reduced in the OpenCL backend. This can be achieved by generating strings from the would be template parameters and using them as keys that would map to a kernel. This would require a lot of rework in the OpenCL backend, but fixing the major ones (such as sort, reduce, ireduce, convolve) can be done fairly easily.

- cl.hpp is a highly templated header only library that is included in practically every where in the OpenCL backend. The suggestion here is to used a pre-compiled header. This may be compiler specific and will require further study. Alternatively we can wrap the cl.hpp so that various classes and functions are instantiated only once.

- Reduce the number of template instantiations. For example convolve2 instantiates 17x17 different kernels. The instantiations can be made for square kernels and non square ones use a fall back version. This has the cons of reducing performance for non critical cases, and increasing code size. But the number of instantiations can be reduced by an order of magnitude, thereby increasing performance.

If anyone has more ideas, please feel free to comment.

umar456:
We can reduce the number of instantiation based off of a compile flag if performance is a concern. I attempted to do that the other day on the CUDA backend but refactoring it would require a little more thought if we don't want to repeat a ton of code.

All of the compilers we are using support pre-compiled headers, but there isn't a standard method to do this in CMake. 

The only reason to have templates in the OpenCL backend is for type safety. Everything else can be passed as a parameter. We are compiling the OpenCL kernels at runtime anyway.

Moving templates who's instantiations are known to the cpp file would also reduce compile times. There are a few functions which fall in this category but aren't a major contributor to the slow builds.

gpryor:
Just my two cents, but I believe it is best to prioritize ease of development for us and other contributors rather than deal with slow first time compilation speed.

A way to increase first compilation speed would be to provide downloads from Jenkins with pre-compiled .o files and other binaries - - this might already be a reality. A simple note would suffice to new users to download from Jenkins rather than clone from github if 15 minutes of initial compile time is a worry.

LLVM is a big chunk to compile and I don't mind the first compile time as I typically spin things up on Amazon on a monster machine and then back the instance off after the first build. Still very fast.

All things in moderation.

pavanky:
[...]

PR #432, opened 2015-03-02
Reduced convolution compilation time

9prady9:
Removed template instantiations for larger kernel sizes
so that we can use fft based convolution for those sizes.


pavanky:
Related to #429

Issue #453, opened 2015-03-05
build opencl backend dependencies with cmake


glehmann:
opencl backend dependencies — clFFT, clBLAS and boost compute — must be built by hand at this time.
This is a task that could be done with cmake's external projects.

shehzan10:
Fixed by #552

PR #454, opened 2015-03-05
download and build the opencl dependencies with cmake

glehmann:
clBLAS, clFFT and boost compute were used to be configure and built manually.
They are now built using cmake's external projects, and linked statically.
The system installed packages can also be used by setting the corresponding
USE_SYSTEM_<package> option to ON.
This also fix a possible conflict with clBLAS and clFFT when installing these
project shared libraries from arrayfire

see #453


glehmann:
Please do not merge yet: this depends on https://github.com/clMathLibraries/clBLAS/pull/67

pavanky:
@glehmann We need to be a bit mindful about how it affects Windows builds. Do you have a windows machine handy that you can test this on ?

glehmann:
@pavanky I only have a VM with windows 7, and no opencl installed on it. Is it possible to install opencl in a VM?

pavanky:
@glehmann It is fine. We'll test it out once the Pull requests for clMath go through.

9prady9:
:+1: good learning for me about CMake. thanks @glehmann

kknox:
I merged the PR's mentioned above by @glehmann.  gtg.

pavanky:
@glehmann We are doing a beta (binary) release today. We'll be merging this after the release is done.

pavanky:
@glehmann I am getting the following error when trying to build a fresh clone of your branch

```
CMake Error at src/backend/opencl/CMakeLists.txt:17 (FIND_PACKAGE):
  By not providing "FindclBLAS.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "clBLAS", but
  CMake did not find one.

  Could not find a package configuration file provided by "clBLAS" with any
  of the following names:

    clBLASConfig.cmake
    clblas-config.cmake

  Add the installation prefix of "clBLAS" to CMAKE_PREFIX_PATH or set
  "clBLAS_DIR" to a directory containing one of the above files.  If "clBLAS"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
```

pavanky:
The problem is with USE_SYSTEM_* being set to ON by default. Can you set it to off ?

Also when set to ON, it would be better if user had to specify the root paths manually rather than CMake trying to find "FindclBLAS.cmake"

glehmann:
I'll set it to OFF by default. Having it to ON by default helps to identify the dependencies for the packagers, but I understand that this is much more convenient for the users and the developers when set to OFF by default.

The message is automatically generated by `find_package` and I have no control over it. I would also prefer if a less obvious reference to the FindclBLAS.cmake…

And the user indeed has to set clFFT_DIR and clBLAS_DIR by hand when the USE_SYSTEM_* are set to ON and clBLAS and clFFT can't be found automatically.

glehmann:
I've updated the pull request. Please try it on windows before merging — I didn't have a chance to test it on this platform.

pavanky:
@glehmann The examples for clFFT were added recently and they don't seem to compile with the static libraries on windows and linux. I reported the issue here: https://github.com/clMathLibraries/clFFT/issues/74

I may have a quick solution for Linux but the windows one will take some time to figure out.

CMakeModules/build_clBLAS.cmake
```diff
@@ -0,0 +1,39 @@
...
+  message(WARNING "Building clBLAS with Ninja has known issues with CMake older than 3.2")
+  set(byproducts)
+else()
+  set(byproducts BYPRODUCTS ${clBLAS_location})
+endif()
+
+ExternalProject_Add(
+  clBLAS-external
+  GIT_REPOSITORY https://github.com/clMathLibraries/clBLAS.git
+  GIT_TAG 47662a6ac1186c756508109d7fef8827efab4504
```
shehzan10: Is this GIT_TAG necessary or can we simply use HEAD always? We would prefer having HEAD.
Also, can you change the repository URL to point to https://github.com/arrayfire/clBLAS.git

Same changes for clFFT.

glehmann: I don't know if the tag is necessary, but it is a good idea to keep it in my opinion: it ensures that the version used is the expected one at the time the code was written, a bit like having clBLAS and clFFT directly in the repository or in a submodule.

glehmann:
I've updated the PR to use arrayfire's repositories for clFFT and clBLAS

src/backend/opencl/CMakeLists.txt
```diff
@@ -12,46 +12,34 @@ ENDIF()
...
-ENDIF(UNIX)
-INSTALL(FILES ${CLBLAS_LIBS} DESTINATION lib COMPONENT libraries)
-
-FIND_PACKAGE(clFFT REQUIRED)
+
+OPTION(USE_SYSTEM_CLFFT "Use system clFFT" OFF)
+IF(USE_SYSTEM_CLFFT)
+  FIND_PACKAGE(clFFT REQUIRED)
+ELSE()
+  INCLUDE("${CMAKE_MODULE_PATH}/build_clFFT.cmake")
```
shehzan10: @glehmann can you add `ADD_DEFINITIONS( -DBUILD_STATIC )` here (also any check required around it). This is needed to disable the dll exports on Windows. Otherwise when clfft.h is included in ArrayFire source, the dll definitions will be there and thats what is causing issues.

glehmann: :-(

We really shouldn't have to do this - I'll think a bit more about https://github.com/clMathLibraries/clFFT/pull/75 to see if it can be done directly there.

pavanky: clBLAS is not using the `__dlexport` and `__dlimport` routines. I am not entirely sure if clFFT needs them.

shehzan10: I tested this quickly (by removing the macro conditions and just keeping #define CLFFTAPI (to keep the compiler happy) and it worked fine. all the examples ran ok.

I can go ahead and push this is all parties agree.

shehzan10: Ah well, may have spoken too soon. Shared libs aren't playing too well with that. I guess for now we can just add the definitions in ArrayFire until we figure out a solution.

shehzan10: Ok. after a little bit of research it looks like clBLAS does export it, except not in the standard way.
It uses a .def file (https://github.com/arrayfire/clBLAS/blob/develop/src/clBLAS.def) to export the symbols in case of shared libs. Because such a file does not exist for clFFT, the declspecs are required.

And according to this http://www.cmake.org/cmake/help/v3.0/command/add_library.html, if add_library does not specify SHARED or STATIC, then the value of BUILD_SHARED_LIBS is used to determine shared. And this is exactly the case in clBLAS.

CMakeModules/build_clFFT.cmake
```diff
@@ -0,0 +1,39 @@
...
+  UPDATE_COMMAND ""
+  CONFIGURE_COMMAND ${CMAKE_COMMAND} -Wno-dev "-G${CMAKE_GENERATOR}" <SOURCE_DIR>/src
+    -DCMAKE_CXX_COMPILER:FILEPATH=${CMAKE_CXX_COMPILER}
+    "-DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS} -w"
+    -DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}
+    "-DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS} -w"
+    -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
+    -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+    -DBUILD_CLIENT:BOOL=OFF
+    -DBUILD_SHARED_LIBRARY:BOOL=OFF
```
shehzan10: Given the comments below, I renamed BUILD_SHARED_LIBRARY to BUILD_SHARED_LIBS in the https://github.com/clMathLibraries/clFFT/pull/75

shehzan10:
Merged in #550

PR #455, opened 2015-03-05
Fix pedantic compiler warning.

bkloppenborg:
Several macros in the ArrayFire headers are followed by semicolons, generating pedantic compiler warnings.


9prady9:
is pedantic enabled by default ?

bkloppenborg:
It is on gcc (GCC) 4.8.2 20140120 (Red Hat 4.8.2-15)

PR #460, opened 2015-03-06
Use user supplied or first non-zero dimension in several functions

bkloppenborg:
Use either the user supplied or first non-zero dimension in reductions, median, and mean.


9prady9:
:+1:

PR #468, opened 2015-03-08
fix gtest build with ninja and simplify gtest external project

glehmann:
build with ninja fixed with cmake 3.2's BUILD_BYPRODUCTS external project option

gtest external project code simplified by using imported libraries and
${CMAKE_CFG_INTDIR} to find the right lib path with MSVC

also pass the build flags to the C++ compiler

most of the code comes from https://github.com/glehmann/gtest-external


pavanky:
:+1:

PR #470, opened 2015-03-08
install arrayfire cmake configuration and version file

glehmann:
so arrayfire can be used in a cmake project with the usual find_package command

  find_package(arrayfire REQUIRED)
  include_directories(${ARRAYFIRE_INCLUDE_DIRS})
  tarkget_link_libraries(mytarget ${ARRAYFIRE_LIBRARIES})


glehmann:
@pavanky OK, I'll add a license. In my opinion you should require the use of the BSD 3 clause license - dealing with several licenses in a project is a nightmare!

glehmann:
and of course the example in the commit is quite wrong - the user should choose the backend

    tarkget_link_libraries(mytarget afopencl)

bkloppenborg:
@glehmann could you please clarify the goal of these changes? I think this is what you wish to accomplish:

1. Enable CMake find functionality. This is already done via. the `FindArrayFire.cmake` script
2. Enable 32/64-bit detection, set PACKAGE_VERSION_UNSUITABLE, effectively disabling 32-bit support on 64-bit systems.
3. Set `EXPORT Library` on libraries. This is the default behavior for library targets in CMake and is unnecessary.
4. Set LINK_INTERFACE_LIBRARIES to an empty string. This setting is suppose to move libraries out of the link and into the interface link property. With an empty string as input, does the code actually do anything?
5. Add version information. This can be found in the `version.h` header file. Unfortunately this file is not included in the install step (this should be corrected).

glehmann:
1. I must admit I completely missed `FindArrayFire.cmake`. That being said, the config file has some advantages over the find file, like being automatically discovered by `find_package`.
1. The 32/64 detection and everything in that file is just the standard code implemented in the CMake macro `WRITE_BASIC_CONFIG_VERSION_FILE`. I've not used it because it appeared in CMake 2.8.6, and arrayfire seems to work with older cmake versions
1. The `EXPORT Library` is used with the `install(EXPORT` macro to generate a list of targets that can be imported from another project
1. Yes, it removes all the libraries in `LINK_INTERFACE_LIBRARIES`. This list is mainly useful when the headers of a library are including the headers of another library. Here the problem comes from the use of external project to build clBLAS and clFFT in #454. The libraries are not installed, and thus shouldn't appear in this list.
1. The version is meant to be used from a CMake project, for example with `find_package`.

Many useful informations on config and version files can be found at http://www.cmake.org/cmake/help/v3.0/command/find_package.html

bkloppenborg:
Very nice. I was not aware of this functionality in CMake. It is indeed far superior to the present `FindArrayFire.cmake` script! I fixed a few things in our `CMakeLists.txt` yesterday which introduced a merge conflict with your branch. Could you please resolve the conflicts and we'll merge in your changes. Thank you for your contribution!

PR #471, opened 2015-03-08
Added ORB documentation

pentschev:


9prady9:
:+1: @pentschev if possible, please add code samples as well. I can show you how we do them in the current framework tomorrow.

PR #472, opened 2015-03-09
Remove C++11 features for test. Make CUDA backend using libstdc++

umar456:
We couldn't easily build the CUDA backend using the clang++(default) compiler. This was because the CUDA toolkit requires that clang++ can only use the libstdc++ library when compiling. This commits makes this configuration possible.

* Build Gtest with libstdc++ on OSX
* Removes C++11 features from tests
* When building tests. Use the libstdc++ library if linking to the CUDA backend. Use the libc++ otherwise
* Added an af_make_seq helper function to create af_seq structs in C++03 portably.


pavanky:
Does this not include the `float2` and `double2` fixes ?

PR #475, opened 2015-03-09
Fixes for CMake installation

bkloppenborg:
* `version.h` missing from install step (does not exist when GLOB is called, must be manually specified)
* Fix documentation source directory for CMake install step.
* Doxygen is mandatory if documentation is to be built: make it `REQUIRED` and remove conditional


pavanky:
@bkloppenborg Can you merge from devel and also add the "make package" target ?

PR #494, opened 2015-03-12
Create .tar.gz package for ArrayFire

bkloppenborg:
This pull request creates a .tar.gz package for ArrayFire using CPack. This includes packaging of

* the CPU, CUDA, and OpenCL backends
* documentation
* examples

The packages are set  as components so future Windows and Mac package generators will let users select which components they wish to install.

Additionally, I rewrote the example/CMakeLists.txt file so that it will be usable from both within the ArrayFire source tree and as a standalone script.


pavanky:
:+1:

PR #543, opened 2015-03-27
Remove delete calls

umar456:
* Replace C arrays with std::vector
* Replace raw pointers with scoped/unique_ptr

### DIFF
--- a/src/api/c/print.cpp
+++ b/src/api/c/print.cpp
@@ -9,6 +9,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <vector>
 #include <af/array.h>
 #include <copy.hpp>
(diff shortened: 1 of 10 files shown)


pavanky:
Can we use `scoped_ptr` / `unique_ptr` are probably faster than `std::vector<T>` inside src ? I say this because `std::vector` sets the values to zero by default causing an overhead.  I may even produce false positives in some cases.

umar456:
Hmm. good point. I will make that change

umar456:
I have made the changes for everything but print. That function will be difficult because we want to use C++03 for CUDA and C++11 for everything else. There is probably a clean way to support scoped/unique_ptr there but probably not worth the effort.

PR #546, opened 2015-03-28
fix 1000+ warnings about unused function with clang

glehmann:


umar456:
Awesome! Thanks for fixing this. Could you replace the tab before the change you made in include/af/defines.h?

umar456:
There seem to be a bunch of tabs in that file. If you could replace all of them with spaces while you are at it that would be great!

PR #547, opened 2015-03-30
Changes based on cppcheck static analysis

umar456:
Ran the cppcheck static analysis tool on the code base. These changes are based
on the recommendation of that tool.

* Pass variables by reference
* Prefix iterators for non basic types
* Variable scope reduction
* Remove exceptions from destructors


pavanky:
:+1:

Issue #525, opened 2015-03-20
no submodules in source archive


glehmann:
The current source archive https://github.com/arrayfire/arrayfire/archive/v3.0beta.tar.gz is lacking the git submodules (gtest and assets).

This unfortunately makes hard to build the tests and make the install fail.

bkloppenborg:
@glehmann this package is automatically generated by GitHub. From numerous posts (e.g.
[Adobe brackets](https://github.com/adobe/brackets/issues/9068
[StackOverflow](http://stackoverflow.com/questions/12936014/include-a-github-submodule-automatically-in-download) ) , it appears that GitHub source zip files do not include submodules. This is because the archive is created by git-archive (i.e. part of git) which does not include submodules by default (see the comment by JaZahn [Academic Technology Group Developers Blog](https://blogs.law.harvard.edu/acts/2012/10/18/github-zip-doesnt-include-submodules/). This issue has been around since 2012 so it is unlikely that it will be addressed by the git or GitHub developers anytime soon.

I think the best alternative is to do a git clone then checkout the v3.0beta tag and updating submodules accordingly.

@pavanky what do you think about 
1. doing a `git clone ... --recursive` on the repository, compressing it, and posting it on the ArrayFire website or
2. Creating special instructions for the GitHub zip file?

glehmann:
Yes git archive is not working well with submodule. I had to deal with a similar case in one of my projects, and I ended using cmake source package generator. You may find it useful in arrayfire too: https://github.com/verateam/vera/blob/master/packages/CMakeLists.txt#L27

The generated source archive can be included in the github release, as would be a binary distribution.

shehzan10:
Archives can be downloaded from our CI server ci.arrayfire.org. Click on the "ArrayFire-<OS> >> <Branch>" job and you will be able to get the archive of the latest build.
Static links to the latest build are http://ci.arrayfire.org/job/arrayfire-linux/branch/devel/lastSuccessfulBuild/artifact/artifacts/

Replace linux with windows or tegra or osx.

glehmann:
@shehzan10 The archive at that url contains the sources, but also the .git repository and the build dir. This is not a clean source archive that can be easily used to build arrayfire.

Do I miss something?

bkloppenborg:
@glehmann, @shehzan10 and I discussed your specific comment in detail offline. We're all on the same page now. I'll put together a source package soon.

bkloppenborg:
Closed via #557.

PR #557, opened 2015-04-01
Add source packaging, remove FindArrayFire.cmake script

bkloppenborg:
* Add CPack source packaging. To create the package, execute `make package_source`
* Remove FindArrayFire.cmake (the equivalent script is automatically generated and installed with `make install`)


bkloppenborg:
Fixes #525

PR #568, opened 2015-04-06
Fixes for memory issues in cuFFT and CLFFT

### LINKED ISSUE #567: Make cuFFT use ArrayFire's memory manager
If it is not possible, atleast call garbageCollect() when cuFFT fails before trying again

pavanky:
- Fixes #567


src/backend/cuda/fft.cpp
```diff
@@ -107,27 +109,30 @@ void find_cufft_plan(cufftHandle &plan, int rank, int *n,
...
-    } else
-        AF_ERROR("cuFFTDestroy call failed", AF_ERR_INTERNAL);
+    cufftDestroy(planner.mHandles[slot_index]); // We ignore both return values
+
+    cufftHandle temp;
+    cufftResult res = cufftPlanMany(&temp, rank, n, inembed, istride, idist, onembed, ostride, odist, type, batch);
+
+    // If plan creation fails, clean up the memory we hold on to and try again
+    if (res != CUFFT_SUCCESS) {
+        garbageCollect();
```
9prady9: @umar456 is `garbageCollect` from our memory manager? If so, how will it remove the memory allocated by the `cuFFTPlanMany` call ? I might be missing something, can you please explain.

PR #575, opened 2015-04-13
Remove print messages on errors

umar456:
This will remove the printing of error messages to the standard output by default. You can still view the messages by defining the AF_PRINT_ERRORS environment variable. Example:

```sh
# No errors printed
./your_bin
AF_PRINT_ERRORS=0 ./your_bin 

# Errors printed to cerr
AF_PRINT_ERRORS=1 ./your_bin
```

This PR also removes the iostreams from the src files. This is done to reduce the compile time and startup costs.


pavanky:
I don't want this removed right away. This is our hackish way of printing the stack trace. If we do this, the default behavior only prints the last function call that catches the error.

PR #593, opened 2015-04-22
FEAT Added 3 and 4 array join functions

shehzan10:
* Changed join kernels to be appending kernels
* Overhead is comparable to previous

### DIFF
--- a/include/af/index.h
+++ b/include/af/index.h
@@ -225,9 +225,7 @@ extern "C" {
     */
     AFAPI af_err af_join(af_array *out, const int dim, const af_array first, const af_array second);
 
-    AFAPI af_err af_join3(af_array *out, const int dim, const af_array first, const af_array second, const af_array third);
-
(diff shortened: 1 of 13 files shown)


pavanky:
@shehzan10 Can you change `af_join3` and `af_join4` to `af_join_many` ?

The API should be like this:

```
af_join_many(af_array *out, const unsigned n_arrays, const af_array *inputs);
```

You can error out if `n_arrays` is > 4 for now.

shehzan10:
Will there be a limit to how many arrays that can be given to that?

pavanky:
Limit to 4 for now.

shehzan10:
Ok. It will be easy to expand on that because of the way I rewrote the kernels.

shehzan10:
@pavanky added it to allow up to 10 arrays using the C API (only limited by a switch case).
The C++ API is limited to 4 independent arrays.

PR #609, opened 2015-04-27
Added u64 support to join()

pentschev:


umar456:
why you no add intl

pentschev:
Well pointed @umar456, added it now.

PR #610, opened 2015-04-27
Bug fixes and proper dimension error checks

### LINKED ISSUE #607: Arrays created with the default constructor have numdims of 4
The numdims function returns 4 for af::arrays which are created using the default constructor. It should return 0 because these arrays do not contain values.

```c++
array a;
cout << "# dims: " << a.numdims() << endl; //Prints 4
```

pavanky:
- Fixes #607


src/api/cpp/array.cpp
```diff
@@ -252,11 +253,7 @@ namespace af
     array::~array()
     {
         af_array tmp = get();
-        if (tmp != 0){
-            if(AF_SUCCESS != af_destroy_array(tmp)) {
-                fprintf(stderr, "Error: Couldn't destroy af::array %p", this);
-            }
-        }
+        if (tmp) AF_THROW(af_destroy_array(tmp));
```
umar456: Destructor MUST never throw

PR #619, opened 2015-04-28
Minor indexing refactor

umar456:


pavanky:
This need not be a member function. make it a static function that exists only in array.cpp

umar456:
Tried that didn't work. Need to call private data. Should go away after my proxy object refactor.

PR #620, opened 2015-04-29
Removed execution permissions of source files

pentschev:


pavanky:
@umar456 Can we add githooks for these kinds of things ? I think this is happening because one of the developers has the default permissions to be 755.

umar456:
@pavanky Yeah, I will look into it

pentschev:
I am not entirely sure about this, but it may be that the files were added to the repo from an NTFS/FAT filesystem.

9prady9:
@pentschev But addition of new features mostly starts with Linux platform, never on windows.

PR #622, opened 2015-04-29
Add STGZ, DEB, and RPM installers

bkloppenborg:
Installers tested against Ubuntu 15.04, Debian 8, and Fedora 21.


umar456:
Can you extract the installation out to another file? The main CMakeLists looks intimidating. You can probably put it in `CMakeModule/Installers.cmake` or something.

bkloppenborg:
@umar456 , this can be done for the packaging step, but the individual `INSTALL` commands need to remain where they are. Would this be acceptable?

umar456:
I was thinking more along the lines of using the [include](http://www.cmake.org/cmake/help/v3.0/command/include.html) command. Creating a packages dir in the root folder seems heavy handed.

bkloppenborg:
Humm... I just realized the OpenCL backend isn't packaging clFFT and similar packages. Is their license permissive enough we can ship the binaries?

pavanky:
@bkloppenborg For reference clBLAS and clFFT are unde Apache license. They can be packaged with ArrayFire if needed.

Issue #625, opened 2015-04-30
Race Condition in reduce_first_kernel


nathanjackson:
I believe I've found a race condition in the reduce_first_kernel for the CUDA backend.  On a GTX 750 Ti the following code produces a race condition that is detected by cuda-memcheck:

    #include <arrayfire.h>

    int main()
    {
        af::array data = af::randu( 3000, 1000, f32 );
        float result = af::mean<float>(data);
        return 0;
    }

I'm not sure if its a problem in the mean function or reduce_first_kernel.

Here is the result from cuda-memcheck

```
$ cuda-memcheck --tool racecheck ./test
========= CUDA-MEMCHECK
========= Race reported between Write access at 0x000009f0 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int)
=========     and Read access at 0x00000a30 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int) [2936 hazards]
=========
========= Race reported between Write access at 0x00000998 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int)
=========     and Read access at 0x000009d8 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int) [5872 hazards]
=========
========= Race reported between Write access at 0x00000948 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int)
=========     and Read access at 0x00000988 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int) [11744 hazards]
=========
========= Race reported between Write access at 0x00000a48 in void cuda::kernel::reduce_first_kernel<float, float, af_op_t=0, unsigned int=256>(cuda::Param<float>, cuda::CParam<float>, unsigned int, unsigned int, unsigned int)
[...]

PR #633, opened 2015-05-03
Fix race condition in reduce_first_kernel.

nathanjackson:
I believe this fix should resolve issue #625.  It seems that a `__syncthreads()` was missing.


pavanky:
@nathanjackson I am aware that this will fix the issue. I am working on an alternative that will not require this for newer architectures ( >3.0). I am also removing the need for volatile memory and clean up the code. Thanks for the contribution though!

pavanky:
I am going to merge your commit into the branch I am working on and add the other stuff around it.

nathanjackson:
Should I close this request then?

pavanky:
@nathanjackson I will close it when I merge the other pull request in.

pavanky:
Actually this is dangerous CUDA code. You can not have `__syncthreads()` inside an if condition. It needs to be on the outside.

PR #639, opened 2015-05-04
Fixed OpenCL scan() for OS X

pentschev:
Also fixes where() failures due to its scan() dependency.


pentschev:
Histogram tests were also fixed by this commit.

9prady9:
:+1:

PR #640, opened 2015-05-04
Proxy Indexing

### LINKED ISSUE #282: Try to implement indexing and assignment via proxy classes in C++

umar456:
This PR implements the proxy class to perform indexing. A proxy class is basically a class which delays the evaluation of an expression until a later time. We can use this to determine if the indexing performed for assignment or reading.

The advantages to this approach is that we do not need to store the indexing information in all arrays. Current approach forces us to store the isRef, parent, and indices variable in the af::array class.

Using the proxy class, these variable do not need to be stored for every object. We will only need the reference to the af_array handle to reference the data. The other variables are stored in the proxy class and are only allocated during the indexing. This reduces the size of the af::array from 152 bytes to 8;

- Fixes #282


pavanky:
:+1:

umar456:
Fixed the issue I found earlier in the HelloWorld program. Should we be returning proxy objects for transpose functions?

pavanky:
@umar456 Let us keep the transpose for later ?

umar456:
Ok, but we can't do it until the next major release.

PR #650, opened 2015-05-09
Added tests for gaussiankernel()

pentschev:


test/testHelpers.hpp
```diff
@@ -71,6 +71,46 @@ void readTests(const std::string &FileName, std::vector<af::dim4> &inputDims,
     }
 }
 
+template<typename outType>
+void readTestsOutputOnly(const std::string &FileName, std::vector<af::dim4> &inputDims,
```
umar456: Couldn't you just put
```
0

2
10

1 2 3 4 5 6 7 8 9 10
```

in the test data for the same effect?

pentschev: Not if I want to have the input dimensions, for gaussiankernel() I can still use the test input dimensions as the arguments passed to the method.

umar456: I had to do something like this for assign. what I ended up doing was make a linear input (X, 1,1,1) and use the values without reading them into the array.

In your case I would make (4, 1, 1, 1) input data. and use the values to create a dim4. but i guess this would work as well.

so gauss1_7 would look like:

```
1
4 1 1 1
1
7

7 1 1 1

0.0548     0.1239     0.2022     0.2381     0.2022     0.1239     0.0548
```

PR #657, opened 2015-05-12
Linear algebra for all backends (Under Review)

### LINKED ISSUE #7: Linear Algebra functions
Possible libraries for CPU:
- netlib's LAPACK
- netlib's PLASMA
- Intel's MKL

Possible libraries for CUDA:
- MAGMA
- CULA

Possible libraries for OpenCL:
- clMAGMA
- ViennaCL

pavanky:
- Fixes #7


include/af/lapack.h
```diff
@@ -0,0 +1,75 @@
...
+    AFAPI af_err af_qr(af_array *q, af_array *r, af_array *tau, const af_array in);
+
+    AFAPI af_err af_qr_inplace(af_array *tau, af_array in);
+
+    AFAPI af_err af_cholesky(af_array *out, int *info, const af_array in, const bool is_upper);
+
+    AFAPI af_err af_cholesky_inplace(int *info, af_array in, const bool is_upper);
+
+    AFAPI af_err af_solve(af_array *out, const af_array a, const af_array b,
+                          const af_solve_t options = AF_SOLVE_NONE);
```
umar456: C does not have default arguments.

src/api/c/cholesky.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+af_err af_cholesky(af_array *out, int *info, const af_array in, const bool is_upper)
+{
+    try {
+        ArrayInfo i_info = getInfo(in);
+
+        af_dtype type = i_info.getType();
+
+        ARG_ASSERT(2, i_info.isFloating());                       // Only floating and complex types
+        ARG_ASSERT(2, i_info.isFloating());                       // Only floating and complex types
```
umar456: Repeat

pavanky:
````
         70 - Test_lu_dense_opencl (Failed)   
         76 - Test_inverse_dense_opencl (Failed)
        105 - Test_transpose_inplace_opencl (Failed)
        106 - Test_blas_opencl (OTHER_FAULT)  
        125 - Test_solve_dense_opencl (Failed)
```

pavanky:
The `blas_opencl` test is failing because of AMD's OpenCL implementation when running mutliGPU blas.

src/backend/cpu/triangle.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <af/dim4.hpp>
+#include <af/defines.h>
+#include <Array.hpp>
+#include <triangle.hpp>
+#include <math.hpp>
+
+namespace cpu
+{
+
+template<typename T, bool is_upper>
```
umar-arshad: The is_upper can be a function parameter. I don't think it will give the compiler much of a hint since it is used in conjunction with a runtime variable.

pavanky: Yes, but everything down stream from this function will be compiled as two functions. is_upper needs to be a template inside the kernel for the best performance.

include/af/lapack.h
9prady9: @pavanky @umar-arshad We need to have a CI-Build for just the C API to avoid such mistakes in future. This would also be useful for other language wrapper projects.

pavanky: Might be tricky on windows. May be possible on Linux.

src/backend/opencl/program.cpp
```diff
@@ -6,7 +6,7 @@
  * The complete license agreement can be obtained at:
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
-
+#include <iostream>
```
umar456: what is this for?

pavanky: Not necessary, can be taken out.

src/backend/opencl/qr.cpp
```diff
@@ -0,0 +1,130 @@
...
+        Array<T> in = padArray<T, T>(orig, pDims, scalar<T>(0));  //copyArray<T>(orig);
+        in.resetDims(iDims);
+
+        int MN = std::min(M, N);
+        int NB = magma_get_geqrf_nb<T>(M);
+
+        int NUM = (2*MN + ((N+31)/32)*32)*NB;
+        Array<T> tmp = createEmptyArray<T>(dim4(NUM));
+
+        T *h_tau = new T[MN];
```
umar456: std::vector

src/backend/opencl/qr.cpp
```diff
@@ -0,0 +1,130 @@
...
+    } catch(cl::Error &err) {
+        CL_TO_AF_ERROR(err);
+    }
+}
+
+template<typename T>
+Array<T> qr_inplace(Array<T> &in)
+{
+    try {
+        AF_ERROR("Linear Algebra is disabled on OpenCL", AF_ERR_NOT_CONFIGURED);
```
umar456: is this correct? Its in the WITH_OPENCL_LINEAR_ALGEBRA section

pavanky: Yeah. Magma does non standard stuff to support `qrInplace` consistently. This is the *one* function that will not be supported in OpenCL.

src/backend/cuda/cublasManager.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+        cublasHandle_t get() const
+        {
+            return handle;
+        }
+    };
+
+    cublasHandle_t getHandle()
+    {
+        using boost::scoped_ptr;
```
9prady9: We can avoid having the class `cublasHandle` here. 
```
static std::once_flag blasFlags[DeviceManager::MAX_DEVICES];
static scoped_ptr<cublasHandle> blasHandles[cuda::DeviceManager::MAX_DEVICES];

int device = getActiveDeviceId();

std::call_once( blasFlags[device],  [device] () {
	handle[id].reset(new cublasHandle());
};

return blasHandles[device]->get();
```
We don't need to create classes when it is not required.

src/backend/opencl/solve.cpp
```diff
@@ -0,0 +1,266 @@
...
+                 (*dB)(), B.getOffset(), B.strides()[1],
+                 1, &queue, 0, nullptr, &event);
+
+        magmablas_swapdblk<T>(MN - 1, NB,
+                              (*dT)(), tmp.getOffset() + MN * NB, NB, 0,
+                              (*dA)(), A.getOffset(), A.strides()[1], 1, queue);
+
+#if UNMQR
+        int lwork = (B.dims()[0]-A.dims()[0]+NB)*(B.dims()[1]+2*NB);
+        T *h_work = new T[lwork];
```
umar456: std::vector

test/cholesky_dense.cpp
```diff
@@ -0,0 +1,114 @@
...
+using std::vector;
+using std::string;
+using std::cout;
+using std::endl;
+using af::cfloat;
+using af::cdouble;
+
+///////////////////////////////// CPP ////////////////////////////////////
+//
+// TEST(Cholesky, CPP)
```
9prady9: is this commented out unit test done on purpose ? Can a `TO DO` or `FIX ME` be added so that it will be un-commented or removed later.

src/backend/opencl/qr.cpp
9prady9: hmm, that is sad. May be can do a custom kernel for this one case ourselves. What do you think ?

pavanky: Not easy. I wasted multiple days trying to fix it. I think we should just disable qrInplace for all backends right now instead.

include/af/lapack.h
pavanky: Fixed.

src/api/c/cholesky.cpp
pavanky: Fixed.

src/backend/opencl/qr.cpp
pavanky: fixed.

src/backend/opencl/solve.cpp
pavanky: fixed.

test/cholesky_dense.cpp
pavanky: Done.

PR #669, opened 2015-05-14
Ensure functions in err_common.cpp are not missing

FilipeMaia:
Ensure af_get_last_error and af_err_to_string are exported by including the public header.
The header that declares this function as AFAPI (af/exception.h) was not included in err_common.cpp so the symbols were not visible in the compiled library.

### DIFF
--- a/src/api/c/err_common.cpp
+++ b/src/api/c/err_common.cpp
@@ -7,6 +7,7 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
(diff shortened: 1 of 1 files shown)


pavanky:
@FilipeMaia I am going to merge this PR, but next time can you try to keep the commit message short. You can provide a detailed description after the commit message.

FilipeMaia:
Sorry about that. I'll keep it in mind.

pavanky:
Forgot to add, thanks for fixing these issues!

PR #675, opened 2015-05-15
Add support for s64 and u64 arrays

FilipeMaia:
Currently calling af::where() using an s64 or u64 array as first argument causes:
```Error(11): Function does not support this data type```
This pull request fixes that.

Also the return array has type u32, which is going to be a problem with large arrays. That should be looked into at some point, maybe sooner than later.


pavanky:
@FilipeMaia Can you create a new issue for the return types ? It needs more discussion and I need to get feedback from other engineers.

PR #684, opened 2015-05-18
Api changes for dimensions

pavanky:
- Renamed dim_type to dim_t
- Changed `af_array` from `size_t` to `void *`
- Changed `dim_t` from `int` to `size_t`
- Made sure all kernels in cuda and opencl are using `int` for performance

### DIFF
--- a/src/api/c/shift.cpp
+++ b/src/api/c/shift.cpp
@@ -18,12 +18,12 @@ using af::dim4;
 using namespace detail;
 
 template<typename T>
-static inline af_array shift(const af_array in, const af::dim4 &sdims)
+static inline af_array shift(const af_array in, const int sdims[4])
 {
     return getHandle(shift<T>(getArray<T>(in), sdims));
 }
 
-af_err af_shift(af_array *out, const af_array in, const af::dim4 &sdims)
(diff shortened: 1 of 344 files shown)


src/api/c/shift.cpp
```diff
@@ -18,12 +18,12 @@ using af::dim4;
...
 
 template<typename T>
-static inline af_array shift(const af_array in, const af::dim4 &sdims)
+static inline af_array shift(const af_array in, const int sdims[4])
 {
     return getHandle(shift<T>(getArray<T>(in), sdims));
 }
 
-af_err af_shift(af_array *out, const af_array in, const af::dim4 &sdims)
+af_err af_shift(af_array *out, const af_array in, const int sdims[4])
```
umar456: sdims should be a pointer. This is equivalent but it implies that it is passing all of the values.

pavanky: It is passing all four values.

umar456:
This should have been 2 pull requests

src/api/c/shift.cpp
umar456: No its still a pointer. It just looks like it is passing by value. The [4] is irrelevant.

PR #689, opened 2015-05-20
Fixed memory leak in cpu/where.cpp

munnybearz:


pavanky:
@munnybearz Can you configure your name and email in your gitconfig before future commits ? Right now it is picking up some default values and not listing you as a contributor.

munnybearz:
Hi, ok I've set my client up accordingly :)

PR #698, opened 2015-05-22
statistics features and OSX fixes

pavanky:
- Fixes var test failing on OSX
- cleans up enums
- Adds support for 64 bit ints in statistics.

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -65,7 +65,7 @@ typedef unsigned long long uintl;
 typedef enum {
     AF_SUCCESS            =   0,
     AF_ERR_INTERNAL       =  -1,
-    AF_ERR_NOMEM          = 101,
+    AF_ERR_NO_MEM         = 101,
     AF_ERR_DRIVER         = 102,
     AF_ERR_RUNTIME        = 103,
     AF_ERR_INVALID_ARRAY  = 201,
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -63,20 +63,20 @@ typedef long long intl;
 typedef unsigned long long uintl;
 
 typedef enum {
-    AF_SUCCESS=0,
-    AF_ERR_INTERNAL,
-    AF_ERR_NOMEM,
-    AF_ERR_DRIVER,
-    AF_ERR_RUNTIME,
-    AF_ERR_INVALID_ARRAY,
-    AF_ERR_ARG,
-    AF_ERR_SIZE,
-    AF_ERR_DIFF_TYPE,
-    AF_ERR_NOT_SUPPORTED,
-    AF_ERR_NOT_CONFIGURED,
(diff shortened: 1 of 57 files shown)


include/af/defines.h
```diff
@@ -63,20 +63,20 @@ typedef long long intl;
...
-    AF_ERR_SIZE,
-    AF_ERR_DIFF_TYPE,
-    AF_ERR_NOT_SUPPORTED,
-    AF_ERR_NOT_CONFIGURED,
-    AF_ERR_INVALID_TYPE,
-    AF_ERR_INVALID_ARG,
-    AF_ERR_UNKNOWN
+    AF_SUCCESS            =   0,
+    AF_ERR_INTERNAL       =  -1,
+    AF_ERR_NOMEM          = 101,
```
umar456: AF_ERR_NO_MEM

src/api/c/var.cpp
```diff
@@ -127,8 +127,8 @@ af_err af_var(af_array *out, const af_array in, const bool isbiased, const dim_t
             case f32: output = var<float ,  float >(in, isbiased, dim); break;
             case s32: output = var<int   ,  float >(in, isbiased, dim); break;
             case u32: output = var<uint  ,  float >(in, isbiased, dim); break;
-            case s64: output = var<intl  ,  double>(in, isbiased, dim); break;
-            case u64: output = var<uintl ,  double>(in, isbiased, dim); break;
+            case s64: output = var<intl  ,  double>(in,  isbiased, dim); break;
+            case u64: output = var<uintl ,  double>(in,  isbiased, dim); break;
```
umar456: boo

src/backend/opencl/program.cpp
```diff
@@ -56,7 +56,7 @@ namespace opencl
             prog.build(targetDevices, (defaults + options).c_str());
 
         } catch (...) {
-            SHOW_BUILD_INFO(prog);
+            SHOW_DEBUG_BUILD_INFO(prog);
```
umar456: Is this a mistake?

umar456:
I saw you removed a lot of tests for doubles. Is it moved higher up?

include/af/defines.h
pavanky: I was going to do that but left it in place because of consistency with the old errors. But I guess this would be the right time to change them.

src/api/c/var.cpp
pavanky: I don't know how that happened..

src/backend/opencl/program.cpp
pavanky: Yep. I will fix this.

pavanky:
@umar456 The checks should ideally be in the Array creation functions. This way we are not even allocating memory if they are not supported. It also means the number of checks is fairly small

pavanky:
Fixed all the things

PR #710, opened 2015-05-27
Added Hamming matcher

pentschev:
Moved computer vision functionality from image.h to vision.h

### DIFF
--- a/include/af/vision.h
+++ b/include/af/vision.h
@@ -69,25 +69,27 @@ AFAPI void orb(features& feat, array& desc, const array& image, const float fast
 /**
    C++ Interface wrapper for Hamming matcher
 
-   \param[out]  idx is an array of MxN size, where M is equal to the number of query
-                features and N is equal to n_dist. The value at position IxJ indicates
-                the index of the Jth smallest distance to the Ith query value in the
-                train data array.
-                the index of the Ith smallest distance of the Mth query.
-   \param[out]  idx is an array of MxN size, where M is equal to the number of query
-                features and N is equal to n_dist. The value at position IxJ indicates
-                the Hamming distance of the Jth smallest distance to the Ith query
-                value in the train data array.
-   \param[in]   query is the array containing the data to be queried
-   \param[in]   train is the array containing the data stored as training data
-   \param[in]   dist_dim indicates the dimension to analyze for distance (the dimension
-                indicated here must be of equal length for both query and train arrays)
-   \param[in]   n_dist is the number of smallest distances to return (currently, only 1
-                is supported)
+   \param[out] idx is an array of MxN size, where M is equal to the number of query
+               features and N is equal to n_dist. The value at position IxJ indicates
+               the index of the Jth smallest distance to the Ith query value in the
+               train data array.
+               the index of the Ith smallest distance of the Mth query.
+   \param[out] idx is an array of MxN size, where M is equal to the number of query
+               features and N is equal to n_dist. The value at position IxJ indicates
+               the Hamming distance of the Jth smallest distance to the Ith query
+               value in the train data array.
+   \param[in]  query is the array containing the data to be queried
+   \param[in]  train is the array containing the data stored as training data
+   \param[in]  dist_dim indicates the dimension to analyze for distance (the dimension
+               indicated here must be of equal length for both query and train arrays)
+   \param[in]  n_dist is the number of smallest distances to return (currently, only 1
+               is supported)
+
+   \ingroup cv_func_hamming_matcher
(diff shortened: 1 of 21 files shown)


include/af/vision.h
```diff
@@ -0,0 +1,175 @@
...
+                the Hamming distance of the Jth smallest distance to the Ith query
+                value in the train data array.
+   \param[in]   query is the array containing the data to be queried
+   \param[in]   train is the array containing the data stored as training data
+   \param[in]   dist_dim indicates the dimension to analyze for distance (the dimension
+                indicated here must be of equal length for both query and train arrays)
+   \param[in]   n_dist is the number of smallest distances to return (currently, only 1
+                is supported)
+ */
+AFAPI void hamming_matcher(array& idx, array& dist,
```
pavanky: Can you make this camel case?

src/backend/cpu/hamming.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <intrin.h>
+#define __builtin_popcount __popcnt
+
+#endif
+
+template<typename T>
+inline uint hamming_distance(T v1, T v2)
+{
+    return __builtin_popcount(v1 ^ v2);
+}
```
pavanky: Is this supported on windows and OSX as well ?

pavanky: OK, I see the __MSC_VER up above.

src/backend/cuda/hamming.cu
```diff
@@ -0,0 +1,62 @@
...
+{
+    uint sample_dim = (dist_dim == 0) ? 1 : 0;
+    const dim4 qDims = query.dims();
+    const dim4 tDims = train.dims();
+
+    const dim4 queryTDims = dim4(qDims[1], qDims[0], qDims[2], qDims[3]);
+    const dim4 trainTDims = dim4(tDims[1], tDims[0], tDims[2], tDims[3]);
+    Array<T> queryT = createEmptyArray<T>(queryTDims);
+    Array<T> trainT = createEmptyArray<T>(trainTDims);
+    if (dist_dim == 0)
```
pavanky: This allocates memory on the device. You don't need it if dist_dim is not 0. Re-arrange the code to reduce the memory overhead.

src/backend/cuda/kernel/hamming.hpp
```diff
@@ -0,0 +1,450 @@
...
+                s_dist[tid] = s_dist[tid + 64];
+                s_idx[tid]  = s_idx[tid + 64];
+            }
+        }
+        __syncthreads();
+        if (tid < 32) {
+            if (s_dist[tid + 32] < s_dist[tid]) {
+                s_dist[tid] = s_dist[tid + 32];
+                s_idx[tid]  = s_idx[tid + 32];
+            }
```
pavanky: For whatever reason this is causing race conditions in newer versions of cuda. Just use __syncthreads() for every block. They are not as costly on newer hardware.

src/backend/cuda/kernel/hamming.hpp
```diff
@@ -0,0 +1,450 @@
...
+            if (s_dist[tid + 2] < s_dist[tid]) {
+                s_dist[tid] = s_dist[tid + 2];
+                s_idx[tid]  = s_idx[tid + 2];
+            }
+            if (s_dist[tid + 1] < s_dist[tid]) {
+                s_dist[tid] = s_dist[tid + 1];
+                s_idx[tid]  = s_idx[tid + 1];
+            }
+        }
+        __syncthreads();
```
pavanky: Same goes here. Don't rely on intra-warp synchronization.

src/backend/opencl/kernel/hamming.cl
```diff
@@ -0,0 +1,327 @@
...
+            }
+            if (l_dist[tid + 2] < l_dist[tid]) {
+                l_dist[tid] = l_dist[tid + 2];
+                l_idx[tid]  = l_idx[tid + 2];
+            }
+            if (l_dist[tid + 1] < l_dist[tid]) {
+                l_dist[tid] = l_dist[tid + 1];
+                l_idx[tid]  = l_idx[tid + 1];
+            }
+        }
```
pavanky: You can't rely on intra warp synchronization on OpenCL. Different devices have different criterion. Always use barriers.

src/backend/opencl/kernel/hamming.cl
```diff
@@ -0,0 +1,327 @@
...
```
pavanky: Same as above.

Issue #713, opened 2015-05-27
Inconsistent camel case


pavanky:
- [x] alltrue --> allTrue
- [x] anytrue --> anyTrue
- [x] setunion --> setUnion
- [x] setunique --> setUnique
- [x] setintersect --> setIntersect
- [x] fftconvolve --> fftConvolve
- [ ] medfilt --> medianFilter (medFilt) ?
- [ ] corrcoef --> corrCoef ? (this is similar to matmul, should it even change?)
- [x] meanshift --> meanShift
- [x] colorspace --> colorSpace

pavanky:
@umar456 We need to figure this out today.

pavanky:
@pentschev add stuff as you find here.

pavanky:
While I am here, shouldn't these also change:

- erode3d --> erode3
- dilate3d --> dilate3 ?

pentschev:
Also those:

* deviceprop --> deviceProp
* colorspace --> colorSpace
* histequal --> histEqual

pavanky:
- rgb2gray --> rgbToGray
- rgb2hsv --> rgbToHsv
- hsv2rgb --> hsvToRgb
- gray2rgb --> grayToRgb

pavanky:
I think deviceInfo feels more natural than deviceProp.

PR #718, opened 2015-05-27
API consistency 

umar456:
Updated the function names to be consistent across the library.

Fixes: #713 

_Possible additions_:
Add the deprecate flags for the older versions? This will throw a warning if the older versions are used guiding the user to use the newer versions.

### DIFF
--- a/docs/details/algorithm.dox
+++ b/docs/details/algorithm.dox
@@ -49,7 +49,7 @@ Find the maximum values and their locations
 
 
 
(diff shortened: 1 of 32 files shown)


pavanky:
Stuff that needs to be added to compatible.h

- setUnion, setUnique, setIntersect
- anyTrue, allTrue
- histEqual
- colorSpace
- deviceInfo (deviceprop)

PR #722, opened 2015-05-28
Enh/cmake install paths

### LINKED ISSUE #721: Centralized location for path management
Dear ArrayFire developers,

This is more of a suggestion than a requirement, but for downstream packaging (again) we often need to relocate part of the installation tree to more suitable locations. As it stands now, that means patching individual CMakeLists.txt files in the ArrayFire source tree, which is not ideal and could be made easier by centralizing path management.

Could you guys provide such centralized place where all of the installation paths are set (for instance in the root CMakeLists.txt or under a separate CMake/AFInstallDirs.cmake module) and, for additional packaging niceness, let the individual paths be overridden when calling CMake ?

This would reduce the amount of patching downstream and should be fairly straightforward to implement. Many CMake projects do implement this sort of thing, or use GNUInstallDirs to let CMake handle it.

Many thanks,
Ghis

ghisvail:
- Fixes #721 .

Installation paths are centralized in CMakeModules/AFInstallDirs.cmake and imported in the main CMakeLists.txt. All hardcoded installation paths have been replaced by their corresponding  AF_INSTALL_* variable.

It has been moderately tested on my Ubuntu Linux machine, additional testing may be required for Windows and OSX (particularly the latter, since it uses RPATH).

Let me know if I forgot anything.


pavanky:
@umar456 Did you get a chance to test it on OSX ?

umar456:
Not yet. doing it now.

pavanky:
@ghisvail Is this a partial fix or a complete fix to #721 ?

ghisvail:
It is a full fix.

pavanky:
Thanks, I asked to update the description (changed from addresses to fixes).

umar456:
Look like everything is working as expected.

pavanky:
:+1:

PR #733, opened 2015-05-29
Added arrayfire pro tips page

shehzan10:


shehzan10:
Merge along with some other PR.

pavanky:
I wouldnt call this page pro tips. The title makes it sound like it is helping users with writing more efficient code. This is not the case. Perhaps configuring arrayfire or even just calling it "tips" would be better.

shehzan10:
Updated

PR #734, opened 2015-05-29
Fix of OpenCL backend's ugly memory leak

### LINKED ISSUE #735: Memory leak in OpenCL backend
Continued from this: https://github.com/arrayfire/arrayfire/pull/734

The actual issue is there: https://github.com/unbornchikken/arrayfire/blob/devel/src/backend/opencl/Array.cpp#L283

Look, if there is an OpenCL buffer, and there is a CLBuffer reference to this in the heap, and this CLBuffer never gets deleted, then the reference wil hold forever, and OpenCL buffer won't get released, because **its ref. count will never reach zero**. So there is a memory leak in the heap of size_t and on the GPU of buffer size!

When was the last time you see a C++ code that behaved correctly when there was a `new` without a corresponding `delete` anywhere? :)

My PR was wrong because I though that it is assumed that (cl_mem)(data) will outlive the asynchronous copy operation for sure. Despite it was wrong that leak is still there.

I mentioned if that I thought at first is not the case, you have to create and event that fires when the copy gets finished, and you gotta delete that CLBuffer in its callback, eg:

```C++
void
writeDeviceDataArray(Array<T> &arr, const void * const data, const size_t bytes)
{
    if (!arr.isOwner()) {
        arr = createEmptyArray<T>(arr.dims());
    }

    cl::Buffer& buf = *arr.get();

    const cl::Buffer* data_buf = new cl::Buffer((cl_mem)(data));

    cl::Event e;
    
    getQueue().enqueueCopyBuffer(data_buf, buf,
                                 0, (size_t)arr.getOffset(),
                                 bytes,
                                 nullptr,
                                 e);
                                 
    e.setCallback(
    CL_COMPLETE,
    [](cl_event event, cl_int status, void* userData)
    {
        delete (cl::Buffer*)userData;
    },
    data_buf);


    return;
}
```

unbornchikken:
This fixes that serious memory leak in OpenCL backend, and gets rid of unnecessary CLBuffer referencing.

- Fixes #735

### DIFF
--- a/src/backend/opencl/Array.cpp
+++ b/src/backend/opencl/Array.cpp
@@ -280,6 +280,7 @@ namespace opencl
 
         cl::Buffer& buf = *arr.get();
 
+        clRetainMemObject((cl_mem)(data));
         cl::Buffer data_buf = cl::Buffer((cl_mem)(data));
 
         getQueue().enqueueCopyBuffer(data_buf, buf,
--- a/src/backend/opencl/Array.cpp
+++ b/src/backend/opencl/Array.cpp
@@ -278,11 +278,11 @@ namespace opencl
             arr = createEmptyArray<T>(arr.dims());
         }
 
-        cl::Buffer buf = *arr.get();
+        cl::Buffer& buf = *arr.get();
 
-        const cl::Buffer *data_buf = new cl::Buffer((cl_mem)(data));
+        cl::Buffer data_buf = cl::Buffer((cl_mem)(data));
 
-        getQueue().enqueueCopyBuffer(*data_buf, buf,
+        getQueue().enqueueCopyBuffer(data_buf, buf,
                                      0, (size_t)arr.getOffset(),
                                      bytes);
 


pavanky:
Did you catch this during a code audit or is there a piece of code that was able to reproduce the memory leak ? I am asking because `data_buf` is a pointer for a reason.

`cl::Buffer` does not retain a buffer, but it does release it when it goes out of scope. If `data_buf` is not a pointer, it will destroy `data` when it goes out of scope. Considering that we do not control `data`, this is a really bad consequence of this change.

unbornchikken:
I've caught this with my eyes. This is a memory leak for sure. 

You already have a reference this OpenCL memory, that's (cl_mem)(data) which is a handle to this memory. Till this handle is not released, the memory won't released. So there is no need to hold another reference for this in the heap.

But if there is a chance for (cl_mem)(data) will get released before copy finished, then it is still not a good idea to pile up references to this OCL memory into the heap, because heap usage will increase infinitely and the actual OpenCL memory won't get released ever. So this is a double leak actually.

If the latter is the case, then create an event that signals after the copy gets finished and release this newly created CLBuffer instance in it's callback.

pavanky:
May be I didn't make this explicit. This function is used when users are trying to copy data (created outside of arrayfire) into an existing arrayfire buffer. We are *not supposed* to delete it.

The original code was not leaking GPU memory for sure. It may have been leaking `sizeof(cl::Buffer)` bytes of host memory. 

The original code wasn't leaking OpenCL memory. This PR breaks functionality and tests. I am going to close this issue.

pavanky:
@unbornchikken I may have been hasty in closing the issue. @umar456 privately pointed out that your PR can be merged with one minor addition.

Can you please call `clRetainMemObject` on `data` before creating `data_buf`. This should make the code cleaner and also fix the memory leak on the heap.

unbornchikken:
It's done.

unbornchikken:
It's shortened but I dunno how to add details.

pavanky:
@unbornchikken This looks fine. To add details, you just need to leave an empty space after the commit message followed by the detailed description.

unbornchikken:
I'm using SmartGit. Dunno mutch about the commandline but I'll figure it out if you point me the right direction.

unbornchikken:
Done. Look how pretty it's become!

pavanky:
Ah much better. Thanks for doing this!

Issue #738, opened 2015-05-29
CMakeLists.txt in examples directory does not let you disable backends


pavanky:
- If it finds the necessary config files, it tries to build the examples for all the backends.

PR #740, opened 2015-05-30
Standalone examples

bkloppenborg:
Fixes for #738 , automatically disable backends if CUDA or OpenCL is not detected.


examples/FindOpenCL.cmake
```diff
@@ -0,0 +1,168 @@
+#.rst:
```
shehzan10: Why 2 FindOpenCL.cmake file? one in examples/ and one in examples/CMakeModules

bkloppenborg: Oh, there should only be one. Good catch. I'll fix this and push.

PR #741, opened 2015-05-30
Fix examples, improve flow of getting started guide

### LINKED ISSUE #738: CMakeLists.txt in examples directory does not let you disable backends

bkloppenborg:
- fixes #738


pavanky:
Can you add the remove FindOpenCL commit to this one? I am going to close #740  because it adds the same commits.

PR #747, opened 2015-06-01
Adding capability to build multiple compute versions

shehzan10:
* PTX still builds with 2.0 always


shehzan10:
Need to test on tegra. Don't merge just yet

PR #760, opened 2015-06-02
Fixed 'snprintf' definition conflict in Visual Studio 2015

xumbu:
Since Visual Studio 14 snprintf is already defined by default.

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -25,7 +25,9 @@
         #define true  1
     #endif
     #define __PRETTY_FUNCTION__ __FUNCSIG__
-    #define snprintf sprintf_s
+    #if _MSC_VER < 1900
+	    #define snprintf sprintf_s
+    #endif
     #define STATIC_ static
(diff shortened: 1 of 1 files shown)


shehzan10:
Hi. Thanks for sending this in.

I tried building on VS 2015RC, but it failed with other errors. Can you confirm whether you were able to build any backend on VS2015?

xumbu:
I haven't tried to build arrayfire itself with VS 2015 yet.

shehzan10:
I did a little bit of testing yesterday. The fixes in arrayfire are fairly straightforward and have to do with scope resolution in our tests. However, the fixes in upstream projects also need to be done. It looks like upstream projects like clBLAS, clFFT etc also use the same convention.
FreeImage built with 2013 cannot be statically with 2015. Dynamic works though. It faces the same problem too.
So a lot of software is set to fail, not limited to ArrayFire. For once I guess maybe Microsoft not implementing something would have been better.

I'll accept this PR when we merge other PRs too.

shehzan10:
Merged into #767

PR #762, opened 2015-06-03
Bring back FindclFFT.cmake

marbre:
I am building an arrayfire ebuild to install version 3.0.0 on Gentoo Linux. In general we try to unbundle every lib, as a result we set `USE_SYSTEM_CLFFT` to `ON`. In 3.0_beta a `FindclFFT.cmake` module was shipped, but somehow it was dropped. Instead of providing it only with the Gentoo ebuild, I would prefer to just readd this module to the arrayfire CMakeModules.

### DIFF
--- a/CMakeModules/FindclFFT.cmake
+++ b/CMakeModules/FindclFFT.cmake
@@ -1,0 +1,78 @@
+# - Find clFFT, AMD's OpenCL FFT library
+
+# This script defines the following variables:
+# CLFFT_INCLUDE_DIRS    - Location of clFFT's include directory.
+# CLFFT_LIBRARIES       - Location of clFFT's libraries
+# CLFFT_FOUND           - True if clFFT has been located
+#
+# If your clFFT installation is not in a standard installation directory, you
+# may provide a hint to where it may be found. Simply set the value CLFFT_ROOT
+# to the directory containing 'include/clFFT.h" prior to calling this script.
+#
+# By default this script will attempt to find the 32-bit version of clFFT.
+# If you desire to use the 64-bit version instead, set
+#   set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS ON)
+# prior to calling this script.
+#
+#=============================================================================
+# Copyright 2014 Brian Kloppenborg
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+
+IF(CLFFT_INCLUDE_DIRS)
+    # Already in cache, be silent
+    set (CLFFT_FIND_QUIETLY TRUE)
+ENDIF (CLFFT_INCLUDE_DIRS)
+
+FIND_PATH(CLFFT_SOURCE_DIR
+    NAMES src/include/clFFT.h
+    HINTS   "${CMAKE_SOURCE_DIR}/clFFT"
+            "${CMAKE_SOURCE_DIR}/../clFFT"
+            "${CMAKE_SOURCE_DIR}/../../clFFT"
+    DOC "clFFT source directory.")
+
+FIND_PATH(CLFFT_BUILD_DIR
+    NAMES ./install_manifest.txt
+    HINTS   "${CLFFT_SOURCE_DIR}/build"
+            "${CLFFT_SOURCE_DIR}/src/build"
+    DOC "clFFT build directory.")
+
+FIND_PATH(CLFFT_ROOT_DIR
+    NAMES include/clFFT.h
+    HINTS   "$ENV{CLFFT_ROOT}"
+            "${CMAKE_SOURCE_DIR}/.."
+            "${CMAKE_SOURCE_DIR}/../.."
+            "${CLFFT_BUILD_DIR}/package"
+            "${CMAKE_INSTALL_PREFIX}"
+    DOC "clFFT root directory.")
+
+FIND_PATH(_CLFFT_INCLUDE_DIRS
+    NAMES clFFT.h
+    HINTS "${CLFFT_ROOT_DIR}/include"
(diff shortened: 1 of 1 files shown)


shehzan10:
Can you also re-add the FindclBLAS.cmake file to this PR?

marbre:
Sure, but somehow cmake didn't had a problem with finding clBLAS on my system.

pavanky:
marbre, ideally FindclFFT.cmake / FindclBLAS.cmake should be included from upstream packages. It is not ideal for downstream packages to include them one at a time.

pavanky:
@ghisvail Can you comment on how you are trying to package on Debian ? 

We are thinking about writing a document that will help distros downstream create package arrayfire for their distros. We would appreciate any feedback from you two.

ghisvail:
For now, I have only taken care of the CPU backend, since neither clBLAS, clFFT nor Boost.Compute are available in Debian yet.

With regards to how things are done in Debian,
* all libs go to /usr/lib/<multiarch-triplet>, which is in the ldconfig discovery path
* all includes to /usr/include
* all cmake stuff in /usr/lib/<multiarch-triplet>/cmake/ArrayFire, whereby /usr/lib/<multiarch-triplet>/cmake is a directory normally searched for by find_package according to [1].

[1] http://www.cmake.org/cmake/help/v3.0/command/find_package.html

Since clFFT normally ships with a CMake project config file, it should not require an additional find module to work. So for @marbre , it is a matter of finding which path is appropriate in Gentoo to store the CMake config file of clBLAS / clFFT, so that the CMake find machinery works as expected.

marbre:
@ghisvail Yes, clFFT ships a `FindclFFT.cmake`. Anyway, after a short look into clFFT's `CMakeLists.txt` this file isn't installed to /usr/lib/cmake. Thus, this is an clFFT issue.


@pavanky You're right. clFFT provides it on Find module, but clBLAS doesn't. I should be okay, to just push the `FindCLBLAS.cmake` from this repo to upstream or not?


Some feedback regarding the creation of ebuilds: On Gentoo we try to unbundle libs in general. Therefore we add dependencies to additional packages like clFFT, clBLAS and Boot.Compute for which I have already written ebuilds. arrayfire-forge will follow soon as an own package, as it is prohibited to fetch source in the compile stage (what cmake currently does). At the moment I have to pass `-DBUILD_GRAPHICS=OFF`.
The [wiki](https://github.com/arrayfire/arrayfire/wiki/Build-Instructions-for-Linux) is a real good starting point, but the it would be easier if all CMake Options are documented. Right now, a look into the `CMakeLists.txt` is necessary. Some of the undocumented options are quite self-explanatory (like `BUILD_GRAPHICS`) while others aren't (`INSTALLER_MODE`) and some are even confusing (`BUILD_TEST` vs. `BUILD_GTEST`). Thus an updated build from source guide would be the appreciated.

marbre:
BTW the CUDA backend has absolutely no problems on Gentoo (also all tests are passed), the CPU backend compiles nearly without problems (I will take care for LAPACKE) and only minor tests fail (probably due to missing LAPACKE support), while OpenCL still has some troubles but that are mainly issues with clBLAS and clFFT, which don't provide really clean cmake configurations.

pavanky:
@marbre clBLAS and clFFT have been cleaned up to include some of the changes required for packaging already. If you are going to contribute a FindclBLAS.cmake, I would suggest you base it off the FindclFFT.cmake. You can contribute it back upstream. They are usually more than willing to accept changes.

pavanky:
On a related note, can you tell me what tests are failing in the CPU backend ? For us it is the backend that passes tests consistently.

ghisvail:
@marbre , @pavanky 
Since clBLAS / clFFT are CMake projects, they do *not* require an explicit Find* module.

@marbre should figure out where to put the clBLASConfig.cmake and clFFTConfig.cmake in the Gentoo system tree so that their CMake pick it up.

ghisvail:
Where he is right however is that clFFT 2.4 does not ship clFFTConfig.cmake, but this has been addressed in the develop branch I reckon. Just need to cherry-pick the patch.

ghisvail:
> Just need to cherry-pick the patch.

Actually @marbre , let me know if you are interested in this patch. Now both clBLAS and clFFT are picked up by ArrayFire, without any need for a Find module.

marbre:
@ghisvail Sure I am interested, but shouldn't it be enough to just pick the file from the clBLAS/clFFT develop branch an just install them to the right place?

marbre:
@pavanky I let you know what fails on the CPU after it digged into it deeper. I assume this is a drawback of the unresolved LAPACK on my system. I already managed it to use `pkg_config` to detect LAPACKE and also LAPACK can be found this way, but I didn't yet set the correct variables thus that the CMakeLists.txt of the CPU backend could resolve it.

ghisvail:
I don't know how things work in Gentoo, but in Debian you need to start from whatever upstream tarball / upstream tag to build your binary packages. Hence me having to cherry-pick a patch to be applied onto the current stable version (v3.0.0).

marbre:
@ghisvail On Gentoo we can either provide patches or even additional files and copy them into the source location before starting the compile phase.
So, you apply your patches to arrayfire and not to clBLAS/clFFT? My idea was to patch/add additional files to the clBLAS/clFFT packages.

ghisvail:
@marbre , yes the patches are applied onto clFFT, not AF.
My current status is that I have both clBLAS and clFFT packaged and working and AF does pick them up with the USE_SYSTEM_CLBLAS and USE_SYSTEM_CLFFT options.

marbre:
@ghisvail I just picked the clFFTConfig*.cmake.in files from the current clFFT develop branch. As there is still some effort necessary to get the CMakeLists.txt patched, I would like to come back to your offer and take a look into your patch. Maybe, you would like to contact me by email (marbre <at> linux.sungazer.de).

ghisvail:
@marbre done

pavanky:
@ghisvail if it is going to help the larger community, can you send in the patch to clFFT ? The stuff that I was trying to change (https://github.com/clMathLibraries/clFFT/pull/76)  needs to be cleaned up because of merge conflicts anyway.

ghisvail:
@pavanky 
The patch is just a cherry-pick of Kent's commit in ```develop```, which added the missing cmake config files, onto the stable release of v2.4.

pavanky:
Alright. Thanks for clarifying.

ghisvail:
If @marbre successfully applies this patch onto his packaging, I see no reason to spend efforts towards providing a FindclFFT module for ArrayFire.

pavanky:
We aren't doing this. This issue has been closed for a while now. 

Thanks for helping out!

Issue #761, opened 2015-06-02
FindCBLAS.cmake should use pkg_config to find CBLAS


marbre:
I am currently preparing an ebuild to package arrayfire-3.0.0 for the [Gentoo science overlay](https://github.com/gentoo-science/sci). Ebuilds to build 3.0_beta and from git are already available.
I noticed that a patch I wrote for the previous ebuilds is still necessary to find CBLAS. On Gentoo CBLAS can be found via `pkg_config`, as done with applying this [patch](https://github.com/marbre/sci/blob/arrayfire/sci-libs/arrayfire/files/arrayfire-9999-FindCBLAS.patch). Providing this patch is absolutely okay for the release and our ebuild, but it would be nice to see the usage of `pkg_config` in one of the next arrayfire releases. IMHO this is already implemented to find LAPACKE.
Unfortunately, I won't find enough time within the next weeks to modify the `FindCBLAS.txt` by myself and send a pull request, therefore this report.

pavanky:
This is fixed in hotfixes-3.1.1

PR #763, opened 2015-06-03
cpu/CMakeLists.txt: Only warn about missing LAPACK on APPLE

marbre:
WINDOWS and LINUX depend on LAPACK**E**. Thus the warning

```
-- Could NOT find LAPACK (missing:  LAPACK_LIBRARIES) 
CMake Warning at src/backend/cpu/CMakeLists.txt:43 (MESSAGE):
  LAPACK not found.  Functionality will be disabled
```
is quite irritating and should moved.


shehzan10:
The LAPACKE file is one that we wrote. Given how complicated the FindLAPACK.cmake file is (one that ships with CMake), we felt the need for a more simple and straight forward version on Linux and Windows. The FindLAPACKE script finds 3 things: LAPACKE include file, LAPACK library and LAPACKE library. The FindLAPACK script is still necessary on Apple because we use the Framework on OSX.

That said, the reason the IF(LAPACK_FOUND) is outside the OS IF block is because the FindLAPACKE file sets LAPACK_* variables. I did this in order to be consistent whether it is on Apple or the other OSes. Yes, I agree it defies CMake standards and conventions, and maybe we should take a consensus on whether we should fix the LAPACKE file instead.

As of this moment, I would instead prefer changing the warning comment to be of the order "LAPACK and LAPACKE not found (for Linux and Windows)".

I'll let the other comment on this and we'll take a final call then.

pavanky:
LAPACKE wraps LAPACK. I don't understand how you can find LAPACKE but not LAPACK.

marbre:
@pavanky You're right. I *assumed* that LAPACKE is found, but taking a look into the `FindLAPACKE.cmake` this seems not to be the case. I had similar problems with the provided `FindCBLAS.cmake`, which is no found via `pkg_config` (see https://github.com/arrayfire/arrayfire/issues/761).
I therefore close this pull request and will work on adding the `pkg_config` stuff to yours `FindLAPACKE.cmake` (by a patch for Gentoo). Should I send a new pull request if I am finished, which only adds the `pkg_config` lines?

shehzan10:
Sure, you can do that.

PR #774, opened 2015-06-04
Fix copy/paste typo for INSTALLER_MODE in CUDA backend

bkloppenborg:


pavanky:
@bkloppenborg 

Do you think we should add `${CUDA_LIBRARIES}` to this line ?

https://github.com/arrayfire/arrayfire/blob/hotfixes-3.0.1/examples/CMakeLists.txt#L76

bkloppenborg:
@pavanky , no, I don't think so. That line is used only when the examples are being built as part of ArrayFire, so af* is defined along with its link targets.

pavanky:
Too late. merged already. If something breaks, I am still blaming you :-p

PR #789, opened 2015-06-08
Requires some knowledge of array_proxy_impl. Fixes #787

### LINKED ISSUE #787: af::array::array_proxy::array_proxy(const array_proxy &other) broken
The [current implementation](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/array.cpp#L562) dereferences ``impl`` before it being allocated.

FilipeMaia:
This fixes the constructor problems. The current solution requires some knowledge of the internals of array_proxy_impl. A possible alternative solution could be:
```
impl = new array_proxy_impl;
*impl = *other.impl;
```
which might separate a bit more the internals, but would require adding an empty constructor for array_proxy_impl.

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -558,8 +558,9 @@ namespace af
     {
     }
 
-    af::array::array_proxy::array_proxy(const array_proxy &other) {
-        *impl = *(other.impl);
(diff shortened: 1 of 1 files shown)


pavanky:
ok to test

pavanky:
looks good :+1:

PR #793, opened 2015-06-09
fix missing include

glehmann:
for std::min and std::max


shehzan10:
Can you close this PR and open it to the hotfixes branch instead of devel?

shehzan10:
Merged in PR https://github.com/arrayfire/arrayfire/pull/797

PR #801, opened 2015-06-09
Make all link interfaces private

umar456:
* Libs/exes linking to ArrayFire do not need to be linked to
  libforge


shehzan10:
Can you remove the "INSTALLER_MODE" CMake variables? They are in the main cmake list and the backend ones.

In opencl cmake list, there is an else part to it, you can keep the else part as is.

PR #803, opened 2015-06-10
Fix path for CMake config files.

bkloppenborg:
[skip ci]


shehzan10:
Merged from command line.

PR #805, opened 2015-06-10
Fix for displaying images of type uchar, int and uint

9prady9:


9prady9:
I have tested this on Windows and Linux.

PR #817, opened 2015-06-13
Complex dot product

FilipeMaia:
Provides dot product for complex arrays.


shehzan10:
Thanks for this PR.
I believe we should we using dotc (cong(x) . y) instead of dotu.
@pavanky Thoughts?

Also, you missed instantiating the OpenCL templates for complex types.

FilipeMaia:
I think we should have both, but I was planning to have dot call dotu and dotc call dotc.

pavanky:
I agree the default should be dotu 

Let me review it once I'm back in town.

pavanky:
ok to test

PR #818, opened 2015-06-14
Cuda fixes

shehzan10:
2d5c961 Fix a compilation error when compiling on windows without graphics enabled.

Recommend merging into both master and devel and update the 3.0.1 tag.


pavanky:
Send this PR to devel.

pavanky:
Actually send in the PR to 3.0.2

PR #835, opened 2015-06-18
FEAT: Difference of Gaussians

9prady9:
* CPU, CUDA & OpenCL backends
* Unit tests for non-batched, batched & invalid inputs
* Documentation

### DIFF
--- a/docs/details/image.dox
+++ b/docs/details/image.dox
@@ -570,5 +570,18 @@ Interpolation types of \ref AF_INTERP_NEAREST and \ref AF_INTERP_BILINEAR are al
 Affine transforms can be used for various purposes. \ref af::translate, \ref af::scale and \ref af::skew
 are specializations of the transform function.
 
+=======================================================================
+
+\defgroup image_func_dog DoG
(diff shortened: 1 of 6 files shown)


pavanky:
1. API needs to change. The size of both the gaussian kernels needs to be the same for dog. 
2. Instead of doing `convolve(A, B) - convolve(A, C)`, you can just do `convolve(A, (B - C))`.

9prady9:
@pavanky Gaussian kernel size doesn't need to be same. Different radius kernels can be used to smooth the the input image.

pavanky:
@9prady9 Yes, but for `dog` force using same size.

9prady9:
@pavanky I have made the changes.

9prady9:
[skip ci]

9prady9:
@pavanky Ready for merge.

pavanky:
Why is resize_cpu failing ??

9prady9:
Well, i didn't touch that function so i don't think it is related to this change. Must have been failing from earlier i guess.

PR #841, opened 2015-06-20
Gradient fix

### LINKED ISSUE #839: Bug in grad
CUDA and OpenCL have a bug that causes the second last column and row to get wrong values.
Occurs for sizes > 32.

shehzan10:
Fixes #839


pavanky:
@shehzan10 Is this not affecting the CPU backend ?

shehzan10:
No. Its only a problem for CUDA/OpenCL and only when there is exactly one row/column in the last block.

Issue #816, opened 2015-06-13
Illegal casting


FilipeMaia:
Hi,

The code for [`getInfo()`](https://github.com/arrayfire/arrayfire/blob/master/src/backend/ArrayInfo.cpp#L30) seems to lead to undefined behaviour:
```
const ArrayInfo *info = static_cast<ArrayInfo*>(reinterpret_cast<void *>(arr));
```
as arr is of type af_array which is a pointer to a type derived from ArrayInfo.
Check http://stackoverflow.com/a/5445220.

pavanky:
I think the problem only when multiple inheritance is in play. The SO answer deals with such a case. This should not be a problem in our case. 

@umar456 any thoughts?

umar456:
I can see this being a problem with multiple inheritance but since we only have one base type it is a standard upcast which is safe in C++. I don't think we even need the `reinterpret_cast` in this case. The `static_cast` should be able to handle the conversion. This has been tested on multiple compilers(msvc, clang++, g++, icc) and we haven't run into any issues.

PR #842, opened 2015-06-20
POD ArrayInfo and dim4; Standard Layout Array<T>

umar456:
After #816 I did some extra research on the data layout guarantees that C++ provides and
to make sure that we are performing all casts safely. This PR attempts to conform the Array,
ArrayInfo and dim4 objects into the POD and standard layout guarantees.

ArrayInfo and dim4 are now [POD](http://stackoverflow.com/a/4178176) objects.

Array<T> class has a [standard layout](http://en.cppreference.com/w/cpp/language/data_members#Standard_layout)

The standard layout guarantees that the object will not be padded so reinterpret_casts to the first member of the object is safe. This change allows us to make the ArrayInfo class a member object rather than a base class.

Other changes:
* The default constructor of the dim4 object will no longer initialize the member variables. use dim4(0) to initialize to 0
* Dim4 member variables are now private
* ArrayInfo is not a base class of Array<T>

### DIFF
--- a/include/af/dim4.hpp
+++ b/include/af/dim4.hpp
@@ -14,9 +14,6 @@
 #include <ostream>
 #include <istream>
 #include <vector>
-#if __cplusplus > 199711L // Necessary for NVCC
-//#include <initializer_list>
-#endif
 #include <af/defines.h>
 #include <af/seq.h>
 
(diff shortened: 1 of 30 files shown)


pavanky:
retest this please

PR #854, opened 2015-06-25
PERF: min/max operations on CPU

umar456:
Replace std::min/std::max for integral types. This could probably work on floats with some work.

~20% increase on min/max


umar456:
test this please

pavanky:
Slower by 8% on Linux + gcc.

hax denied.

PR #865, opened 2015-06-30
Allow building against 64bit index OpenBlas

Keno:
OpenBLAS can be built with signatures that take a 64bit value as the index type.
When built in this mode blasint is defined to be int64_t, otherwise it is just a
regular int. Since this causes a signature mismatch for those versions of OpenBLAS,
use blasint throughout and provide a fallback typedef for other BLASes.
Also use the OpenBLAS detection to make it use the correct gemm/gemv signature pointer
type. Note that this reverts part of 9be64f05f7236b64949430d4c18ee0421f18bd56, which
I believe was incorrect as it made `BT` pointless. The correct fix would have been
to detect the specific BLAS that was used in the definition below and set the BT to
be `void`.

### DIFF
--- a/src/backend/cpu/blas.cpp
+++ b/src/backend/cpu/blas.cpp
@@ -26,29 +26,30 @@ namespace cpu
 
 template<typename T, typename BT>
 using cptr_type     =   typename conditional<   is_complex<T>::value,
-                                                const void *,
+                                                const BT *,
                                                 const T*>::type;
 template<typename T, typename BT>
 using ptr_type     =    typename conditional<   is_complex<T>::value,
-                                                void *,
+                                                BT *,
                                                 T*>::type;
 template<typename T, typename BT>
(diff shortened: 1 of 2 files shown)


pavanky:
ready to test

umar456:
[build ci]

pavanky:
test this please

pavanky:
@Keno Ignore the comments, we are trying to kick off the ci using comments.

shehzan10:
This build failed on windows. Please refer to http://ci.arrayfire.org/job/arrayfire-windows-pull/90/consoleFull

Keno:
Which BLAS is used for the windows build?

pavanky:
@Keno MKL

Keno:
Updated to also use `void *` with MKL on windows.

pavanky:
build arrayfire ci

pavanky:
test this please

PR #866, opened 2015-06-30
Link against the binary_dir version of forge.

Keno:
CMake will strip the RPATH from the forge shared library when installing it,
causing link failures later down the line. Use the non-installed version instead.

### DIFF
--- a/CMakeModules/build_forge.cmake
+++ b/CMakeModules/build_forge.cmake
@@ -1,8 +1,14 @@
 INCLUDE(ExternalProject)
 
 SET(prefix ${CMAKE_BINARY_DIR}/third_party/forge)
-SET(forge_location "${prefix}/lib/${CMAKE_SHARED_LIBRARY_PREFIX}forge${CMAKE_SHARED_LIBRARY_SUFFIX}")
 
+IF(WIN32)
+    SET(forge_lib_prefix "${prefix}/lib")
+ELSE(WIN32)
+    SET(forge_lib_prefix "${prefix}/src/forge-ext-build/src")
+ENDIF(WIN32)
+
+SET(forge_location "${forge_lib_prefix}/${CMAKE_SHARED_LIBRARY_PREFIX}forge${CMAKE_SHARED_LIBRARY_SUFFIX}")
 IF(CMAKE_VERSION VERSION_LESS 3.2)
     IF(CMAKE_GENERATOR MATCHES "Ninja")
         MESSAGE(WARNING "Building forge with Ninja has known issues with CMake older than 3.2")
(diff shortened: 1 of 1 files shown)


pavanky:
@Keno We can't use this. We need the existing behavior when we are shipping ArrayFire.

Keno:
I have a better version of this coming up. What exactly is the concern? This only affects the link stage.

pavanky:
From what I understand, this will leave a hard dependency on the location of libforge.so. While this might be OK for development, I think this will be problematic when you create installers and use the libraries on a different machine.

I could be wrong though.

Keno:
Yes, that it inaccurate, the RPATH will still get cleared as usual when the arrayfire libraries are installed. This is only such that during linking the linker can find forge's dependencies to avoid throwing undefined symbol errors. Alternatively you could duplicate the FindGLFW logic in arrayfire, but that seems unnecessarily complication. Please wait a couple of minutes while I prepare my improved patch and then you can see if you run into any problems with it.

pavanky:
Alright, I understand what is happening now. Thanks for clearing it up. The reason we are not seeing this internally is because we have glfw in system path in all of our machines. Even when we did not, I think we were just exporting `LD_LIBRARY_PATH` because that was an exception rather than the norm.

Keno:
Yup, this is for when glfw is not in the system path.

Keno:
Alright, updated to what I think is a slightly better way. Annoyingly one cannot use generator variables in the step targets for ExternalProject_Add, so this just hardcodes the value that it is and then afterwards asserts that that is still correct. Not ideal, but nothing one can do until this is fixed in CMake.

shehzan10:
build arrayfire ci

CMakeModules/build_forge.cmake
```diff
@@ -33,12 +34,14 @@ ExternalProject_Add(
...
 ExternalProject_Get_Property(forge-ext install_dir)
+SET(forge_bindir_location ${binary_dir}/src/${CMAKE_SHARED_LIBRARY_PREFIX}forge${CMAKE_SHARED_LIBRARY_SUFFIX})
+IF(NOT (${forge_bindir_location} STREQUAL ${forge_bin_location}))
+    MESSAGE(WARNING "Did the forge binary location move? (Have ${forge_bindir_location} vs ${forge_bin_location}")
+ENDIF()
 ADD_LIBRARY(forge SHARED IMPORTED)
-SET_TARGET_PROPERTIES(forge PROPERTIES IMPORTED_LOCATION ${forge_location})
-IF(WIN32)
-    SET_TARGET_PROPERTIES(forge PROPERTIES IMPORTED_IMPLIB ${prefix}/lib/forge.lib)
-ENDIF(WIN32)
```
shehzan10: We need the IMPORTED_IMPLIB for windows. You can use the bin directory, but it is needed for windows builds.

Keno: Right, my apologies, will add this back.

Keno:
Updated.

pavanky:
build arrayfire ci

CMakeModules/build_forge.cmake
```diff
@@ -33,9 +34,14 @@ ExternalProject_Add(
...
 ExternalProject_Get_Property(forge-ext install_dir)
+SET(forge_bindir_location ${binary_dir}/src/${CMAKE_SHARED_LIBRARY_PREFIX}forge${CMAKE_SHARED_LIBRARY_SUFFIX})
+IF(NOT (${forge_bindir_location} STREQUAL ${forge_bin_location}))
+    MESSAGE(WARNING "Did the forge binary location move? (Have ${forge_bindir_location} vs ${forge_bin_location}")
+ENDIF()
 ADD_LIBRARY(forge SHARED IMPORTED)
-SET_TARGET_PROPERTIES(forge PROPERTIES IMPORTED_LOCATION ${forge_location})
+SET_TARGET_PROPERTIES(forge PROPERTIES IMPORTED_LOCATION ${forge_bin_location})
 IF(WIN32)
     SET_TARGET_PROPERTIES(forge PROPERTIES IMPORTED_IMPLIB ${prefix}/lib/forge.lib)
```
shehzan10: To make it consistent can you use `${prefix}/src/forge-ext-build/src/forge.lib`

Keno: Sure.

shehzan10:
test this please

shehzan10:
Well, it looks like this wont work on windows.
The problem is that cmake on windows build the lib/dll in the src/<build config>/ directory.
So we would have to then query for the build type and all. On top of that, the concept of rpath doesn't really exist on Windows.

@pavanky should we close this or should we have a huge block of if(unix) else windows? That way windows will use the installed location and linux/osx can use the binary directory ones.

pavanky:
Use a new variable that points to the following:

- `${prefix}/lib/` on Windows
- `${forge_bin_prefix}` on non windows systems.

Keno:
Is this what you had in mind?

pavanky:
@Keno Yes. This looks good.  :+1:

 Thanks for taking care of it!

PR #867, opened 2015-06-30
Add support for BLAS symbol renaming

Keno:
OpenBLAS (and potentially other BLASes), generate a symbol rename file to
move symbols out of a namespace where they might conflict with other BLASes
(if the right options are set). Teach ArrayFire to make use of this file
to be able to still link against that library.

### DIFF
--- a/src/backend/cpu/CMakeLists.txt
+++ b/src/backend/cpu/CMakeLists.txt
@@ -102,18 +102,55 @@ ELSE(${UNIX}) #Windows
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 ENDIF()
 
-ADD_LIBRARY(afcpu SHARED
-           ${cpu_headers}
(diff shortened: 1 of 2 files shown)


Keno:
I have an improved patch here as well - please hold on.

Keno:
Alright, here's an improved version with proper dependency declaration (to make sure the ninja build works) as well as OS X support.

pavanky:
@Keno I see that afcpu_static and afopencl_static have been added as well. Any reason why only these two backends have been updated to have static libraries ?

Keno:
I need to do a bulk rename of the symbols and this is the only way I could get cmake to do it. the _static versions are an implementation detail of the build process and not exported.

pavanky:
ok to test

pavanky:
build arrayfire ci

Keno:
Thanks!

PR #870, opened 2015-06-30
Use reference counts to reduce unnecessary copies

### LINKED ISSUE #791: Reduce the memory overhead for copy-on-write
Assignment operation performs a copy-on-write. However this is not necessary if there are no additional references to the array being assigned. We should check for this and reduce the memory overheads accordingly.

There are two ways to do this:

- Keep the list of child arrays referencing the parent. Make the copies of either the parent or the children based on the memory considerations. This will need to be done for each backend.

- Keep track of the number of references and only perform a copy if there are additional references left. By it's nature, I think this should be done in the C++ API.

pavanky:
Fixes #791 

Added and updated tests to reflect this change.

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -1270,6 +1270,15 @@ extern "C" {
     AFAPI af_err af_retain_array(af_array *out, const af_array in);
 
     /**
+       \ingroup method_mat
+       @{
+
+       Get the use count of `af_array`
+    */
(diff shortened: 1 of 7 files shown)


pavanky:
@umar456 This was surprisingly easy. Perhaps because of the way `array_proxy` was implemented. Thanks for doing that!

PR #883, opened 2015-07-06
Improve mean computation tests

alcinos:


pavanky:
Considering that these tests break devel, I think we should move these tests into a different branch where we can fix them cleanly without other stuff interfering with the code.

This does not update the commit hash of test/data. Can you fix that and send in a PR to stats_fixes?

PR #895, opened 2015-07-08
Fix double free issue. Closes #894

### LINKED ISSUE #894: af::index copy constructor and assignment lead to double free
The copy constructor and assignment operator for af::index are not explicitly defined so their are created by the compiler, who just copies the impl private member. If af::index contains an af_array that array handle is copied without increasing the reference count. This leads to a double free when both the original and the new af::index are destroyed.

The following code exemplifies the problem:

```
#include <arrayfire.h>

int main(int argc, char ** argv){
  af::array A = af::constant(0,1, s32);
  af::index s1;
  // This will just copy the members of af::index,
  // using an implicit assignment operator,
  // without updating the reference counting
  s1 = af::index(A);
  if(argc > 1){
    // without this A will be released twice, once
    // during the destructor of A and once during
    // the destructor of s1.
    af_array tmp;
    af_retain_array(&tmp, s1.get().idx.arr);
  }  
  return 0;
}
```

When run with no arguments on my machine it outputs:
```
index_copy(8076,0x7fff7dac3300) malloc: *** error for object 0x7fdd850e6470: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```
 
the result of the double free.

FilipeMaia:
Closes #894

### DIFF
--- a/include/af/index.h
+++ b/include/af/index.h
@@ -104,6 +104,15 @@ class AFAPI index {
     index(const af::array& idx0);
 
     ///
+    /// \brief Copy constructor
+    ///
+    /// \param[in] idx0 is index to copy.
+    ///
+    /// \sa indexing
+    ///
+    index(const index& idx0);
+
+    ///
     /// \brief Returns true if the \ref af::index represents a af::span object
     ///
     /// \returns true if the af::index is an af::span
@@ -116,6 +125,17 @@ class AFAPI index {
     /// \returns the af_index_t represented by this object
     ///
     const af_index_t& get() const;
+
+    ///
+    /// \brief Assigns idx0 to this index
+    ///
+    /// \param[in] idx0 is the index to be assigned to the /ref af::index
+    /// \returns the reference to this
+    /// 
+    ///
+    index & operator=(const index& idx0);
+
+
 };
 
 ///
--- a/include/af/index.h
+++ b/include/af/index.h
@@ -135,7 +135,21 @@ class AFAPI index {
     ///
     index & operator=(const index& idx0);
 
-
+#if __cplusplus > 199711L
+    ///
+    /// \brief Move constructor
+    ///
+    /// \param[in] idx0 is index to copy.
+    ///
+    index(index &&idx0);
+    ///
+    /// \brief Move assignment operator
+    ///
+    /// \param[in] idx0 is the index to be assigned to the /ref af::index
+    /// \returns a reference to this
+    ///
+    index& operator=(index &&idx0);
+#endif
 };
 
 ///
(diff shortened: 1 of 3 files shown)


pavanky:
@FilipeMaia Can you remove the "Closes #894" from the commit message and put it as a description in the PR?

FilipeMaia:
Can i just force push the amended commit message or do i need a new pull request?

umar456:
Hi @FilipeMaia, Thanks for finding this issue.

The problem is that `af::index` is only used to simplify the `operator()` overloads. It allows us to support any arbitrary combination of `int`, `af::seq` or `af::array` objects to index into an array. It is never meant to be created. I think a better solution would be to make the `af::index` default constructor private so that it cannot be created without referencing one of the other constructors.

pavanky:
Force push is fine.

FilipeMaia:
@umar456, I think that having things in the public API which are not supposed to be manipulated by the users, such as af::index and af::array::array_proxy is a sure way to lead to problems in the future.

Just making the default constructor private is not enough. You also need to avoid the creation of ``af::index::operator=``, and there might be other things we haven't thought about.

pavanky:
@umar456 For better or worse it is out in the public. We can't change the API now.

pavanky:
@FilipeMaia Can you make the code from the failure case as a test ? You can make it a new test in test/index.cpp.

FilipeMaia:
I'm not sure if I succeeded making a test. I'm don't know how to test them...

pavanky:
That should be enough :+1: Any code snippet that was failing makes a valid test.

umar456:
@FilipeMaia Yeah, I forgot about the rule of 3/5. Could you add operator=(index&&) and index(index&&) for the index class with c++11 guards (see:https://github.com/arrayfire/arrayfire/blob/devel/include/af/array.h#L51). These functions will not need to increment the reference count for the array object. 

I still think we should make the default constructor private. The index class was not designed to be used this way. @pavanky This change will not break ABI but might cause issues when recompiling for those who used af::index incorrectly.

FilipeMaia:
@umar456 such a change would break the ABI because the copy constructor, which is currently public, would become private.

umar456:
ABI would not break because the Linker will still create the symbol. The API on the other hand will be different. This will cause issues during recompilation of code which makes use of the af::index default constructor. I would like to make this change but I understand it can cause issues with some users.

FilipeMaia:
Ok, I didn't realise the private/public part would not impact the symbol.

umar456:
build arrayfire ci

umar456:
[build ci]

umar456:
ok to test

umar456:
build arrayfire ci

PR #897, opened 2015-07-10
Fixes to error handling and adding sigmoid

### LINKED ISSUE #896: CUDA backend repeats last error forever
@ debug_cuda.hpp line 23

You're using `cudaPeekAtLastError`, but it never reset error state to cudaSuccess, so any subsequent operation call will reports that last error forever. I believe `cudaGetLastError` should be used here instead.

pavanky:
- Fixes #896


pavanky:
@umar456 @shehzan10 I added a test that will take up all the memory available on the system before recovering. I am not sure if we want to be running this test on developer machines.

PR #921, opened 2015-07-24
SUSAN Detector

### LINKED ISSUE #911: susan corner dectector
[Susan Corner Detector](https://en.wikipedia.org/wiki/Corner_detection#The_SUSAN_corner_detector)

9prady9:
Fixes #911

### DIFF
--- a/examples/computer_vision/susan.cpp
+++ b/examples/computer_vision/susan.cpp
@@ -1,0 +1,89 @@
+/*******************************************************
+ * Copyright (c) 2015, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#include <cstdio>
+#include <arrayfire.h>
+#include <cstdlib>
+
+using namespace af;
+
+static void susan_demo(bool console)
+{
+    // Load image
+    array img_color;
+    if (console)
+        img_color = loadImage(ASSETS_DIR "/examples/images/square.png", true);
+    else
+        img_color = loadImage(ASSETS_DIR "/examples/images/man.jpg", true);
+    // Convert the image from RGB to gray-scale
+    array img = colorSpace(img_color, AF_GRAY, AF_RGB);
+    // For visualization in ArrayFire, color images must be in the [0.0f-1.0f] interval
+    img_color /= 255.f;
+
+    features feat = susan(img, 3, 32.0f, 10, 0.05f, 3);
+
+    if (feat.getNumFeatures() > 0) {
+        printf("Found features, proceeding ahead\n");
+    } else {
+        printf("No features found, exiting\n");
+        return;
+    }
+
+    float* h_x = feat.getX().host<float>();
+    float* h_y = feat.getY().host<float>();
+
+    // Draw draw_len x draw_len crosshairs where the corners are
+    const int draw_len = 3;
+    for (size_t f = 0; f < feat.getNumFeatures(); f++) {
+        int x = h_x[f];
+        int y = h_y[f];
+        img_color(y, seq(x-draw_len, x+draw_len), 0) = 0.f;
+        img_color(y, seq(x-draw_len, x+draw_len), 1) = 1.f;
+        img_color(y, seq(x-draw_len, x+draw_len), 2) = 0.f;
+
+        // Draw vertical line of (draw_len * 2 + 1) pixels centered on  the corner
+        // Set only the first channel to 1 (green lines)
+        img_color(seq(y-draw_len, y+draw_len), x, 0) = 0.f;
+        img_color(seq(y-draw_len, y+draw_len), x, 1) = 1.f;
+        img_color(seq(y-draw_len, y+draw_len), x, 2) = 0.f;
+    }
+
+    printf("Features found: %lu\n", feat.getNumFeatures());
+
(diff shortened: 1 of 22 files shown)


9prady9:
[skip ci]

examples/computer_vision/susan.cpp
```diff
@@ -0,0 +1,89 @@
...
+        img_color = loadImage(ASSETS_DIR "/examples/images/man.jpg", true);
+    // Convert the image from RGB to gray-scale
+    array img = colorSpace(img_color, AF_GRAY, AF_RGB);
+    // For visualization in ArrayFire, color images must be in the [0.0f-1.0f] interval
+    img_color /= 255.f;
+
+    features feat = susan(img, 3, 32.0f, 10, 0.05f, 3);
+
+    if (feat.getNumFeatures() > 0) {
+        printf("Found features, proceeding ahead\n");
```
pentschev: I think this printf is unnecessary.

9prady9: @pentschev :smile: sure, i will remove it. I just found it rather too silent when it doesn't say anything when no corners are found.

include/af/vision.h
```diff
@@ -166,6 +166,29 @@ AFAPI void nearestNeighbour(array& idx, array& dist,
  */
 AFAPI array matchTemplate(const array &searchImg, const array &templateImg, const matchType mType=AF_SAD);
 
+
+/**
+   C++ Interface for SUSAN corner detector
+
+   \param[in]  in is input grayscale/intensity image
+   \param[in]  radius Nuclei radius for each pixel neighborhood
```
pentschev: Nuclei is plural, passes the idea that there would be multiple values, I think the singular form nucleus is a better choice.

src/backend/cpu/harris.cpp
```diff
@@ -166,7 +160,7 @@ unsigned harris(Array<float> &x_out, Array<float> &y_out, Array<float> &resp_out
     else {
         gaussian1D<convAccT>(h_filter, (int)filter_len, sigma);
     }
-    Array<convAccT> filter = createHostDataArray(filter_len, h_filter);
+    Array<convAccT> filter = createDeviceDataArray<convAccT>(dim4(filter_len), (const void*)h_filter);
```
pavanky: I was worried you wouldn't catch something like this. Good job :+1:

include/af/vision.h
```diff
@@ -325,6 +348,27 @@ extern "C" {
     */
     AFAPI af_err af_match_template(af_array *out, const af_array search_img, const af_array template_img, const af_match_type m_type);
 
+    /**
+       C Interface for SUSAN corner detector
+
+       \param[out] out is af_features struct composed of arrays for x and y
+                   coordinates, score, orientation and size of selected features
+       \param[in]  in is input grayscale/intensity image
+       \param[in]  radius Nuclei radius for each pixel neighborhood
```
pentschev: Same thing about Nuclei in C++ method above.

include/af/vision.h
9prady9: This is corrected.

src/backend/cpu/harris.cpp
9prady9: :smile:

include/af/vision.h
9prady9: Yup, corrected in both locations.

9prady9:
@pentschev @pavanky Can one of you run susan example in cuda backend and let me know if it is crashing. It is finding correct # of corners, but failing in graphics interestingly.

src/backend/cpu/harris.cpp
pavanky: Can you make the changes in other files as well ? grep for `createHostDataArray` and make changes to avoid unnecessary copies. The two files that I could find are `where.cpp` and `orb.cpp` in cpu backend.

src/api/c/susan.cpp
```diff
@@ -0,0 +1,78 @@
...
+    try {
+        ArrayInfo info = getInfo(in);
+        af::dim4 dims  = info.dims();
+
+        ARG_ASSERT(1, dims.ndims()==2);
+        ARG_ASSERT(2, radius < 10);
+        ARG_ASSERT(3, diff_thr > 0.0f);
+        ARG_ASSERT(4, geom_thr > 0.0f);
+        ARG_ASSERT(5, (feature_ratio > 0.0f && feature_ratio <= 1.0f));
+        ARG_ASSERT(6, (dims[0] >= (dim_t)(2*edge+1) || dims[1] >= (dim_t)(2*edge+1)));
```
pentschev: I think this check must be changed to max(radius, (2*edge+1)) to ensure not trying to read out of image bounds.

pavanky:
I am unsubscribing from this issue until Monday :p sorry guys, I need to leave.

src/backend/cpu/harris.cpp
9prady9: Sure, but this is only in cpu backend right ?

9prady9:
After the recent changes to kernels, tests passed on my machine and tegra as well. Hopefully ci builds will all report green! :smile:

pentschev:
@9prady9 looks like Tegra is still failing, can you check that when you have time?

9prady9:
@pentschev Yup, i was fooled by RelWithDebInfo mode, apparently the problem still persists in release mode. Will try to fix it today and let you know. :checkered_flag:

9prady9:
@pavanky Ready for merge.

PR #928, opened 2015-08-04
Minor changes

pavanky:


src/backend/opencl/magma/laset.cpp
```diff
@@ -7,6 +7,50 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
+/***********************************************************************
+ * Based on MAGMA library http://icl.cs.utk.edu/magma/
+ * Below is the original copyright.
```
shehzan10: Why don't we add the Magma license to the LICENSES directory instead of pasting the full licenses in each of the files? For the example/random123, its only a single file, so it doesn't make that much of a difference.

shehzan10: It looks like BSD 3-clause. So we can give the copyright, but the detailed text can be pointed to the existing BSD license file

pavanky: I was making it consistent with the remaining magma files (they have the license included in the file). I did not want to remove license for existing files, so just added them everywhere.

PR #929, opened 2015-08-04
Changed default cuda stream to be non-zero

### LINKED ISSUE #926: Change default stream of CUDA backend
The default stream used to launch all CUDA kernels should not be stream 0. Launching a kernel on stream0 synchronizes all other streams. 

We should create a new stream and use that instead.

Changes to make:
- Launch all kernels with a stream parameter

9prady9:
Fixes #926

### DIFF
--- a/include/af/cuda.h
+++ b/include/af/cuda.h
@@ -1,0 +1,51 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#include <af/defines.h>
+#include <af/exception.h>
+#include <cuda.h>
+#include <cuda_runtime.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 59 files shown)


shehzan10:
What about calls to thrust etc? I tried to do a quick search (check for sort), but didn't find any stream changes for that.

9prady9:
there is something called as `thrust::cuda::par ` , not sure how to use it with other functions, but that's the one i think.

shehzan10:
looks like its similar to a state variable http://stackoverflow.com/questions/24368197/getting-cuda-thrust-to-use-a-cuda-stream-of-your-choice

9prady9:
It's just another parameter to the functions, thrust functions are overloaded. One of those versions take policy as first argument which is what this thrust::cuda::par thing is.

9prady9:
This failed on tegra because cuda toolkit version is 6.5 which comes with thrust where `thrust::cuda::par.on` is not available. It is working on linux machine because it was using 7.0 toolkit on that.

9prady9:
@pavanky this is a block, cuda-6.5 ships with thrust 1.7.2 and this `cuda::par::on` is only available from 1.8.0 version of thrust that comes from cuda-7.0.

9prady9:
Finally, all :green_apple: @pavanky Please merge it.

PR #936, opened 2015-08-07
New FFT functions

### LINKED ISSUE #613: Add fft r2c and c2r support
Currently fft converts all inputs to complex and performs an in place operation on them.

Add functions where users can call r2c and c2r transforms explicitly. The output of these functions from all fft libraries only stores non-redundant data. We'll need to figure out how this plays with the rest of the library.

This may be useful in some cases for a few users.

### LINKED ISSUE #790: Support in place FFTs
Performing FFTs on large data sets will result in the GPUs running out of memory. Having inplace support would be beneficial in such cases.

pavanky:
- Added in place FFTs
- Added real to complex FFTs
- Added complex to real FFTs
- Cleaned up complex to complex FFTs
- Fixes #790 
- Fixes #613

### DIFF
--- a/include/af/signal.h
+++ b/include/af/signal.h
@@ -371,7 +371,7 @@ AFAPI array idft(const array& in, const dim4 outDims);
 AFAPI array idft(const array& in);
 
 /**
-   C Interface for real to complex fast fourier transform for one dimensional signals
+   C++ Interface for real to complex fast fourier transform for one dimensional signals
 
    \param[in]  in is a real array
    \param[in]  dims is the requested padded dimensions before the transform is applied
@@ -388,7 +388,7 @@ array fftR2C(const array &in,
              const double norm_factor = 0);
 
 /**
-   C Interface for real to complex fast fourier transform for one dimensional signals
+   C++ Interface for real to complex fast fourier transform for one dimensional signals
 
    \param[in]  in is a real array
    \param[in]  norm_factor is the normalization factor with which the input is scaled before the transformation is applied
--- a/include/af/signal.h
+++ b/include/af/signal.h
@@ -370,6 +370,20 @@ AFAPI array idft(const array& in, const dim4 outDims);
  */
 AFAPI array idft(const array& in);
 
+template<int rank>
+array fftR2C(const array &in,
+             const dim4& dims,
+             const double norm_factor = 0);
+
+
+template<int rank>
+array fftR2C(const array &in,
+             const double norm_factor = 0);
+
+template<int rank>
+array fftC2R(const array &in, bool is_odd = false,
+                 const double norm_factor = 0);
+
 /**
    C++ Interface for convolution any(one through three) dimensional data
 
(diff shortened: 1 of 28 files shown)


include/af/signal.h
```diff
@@ -295,7 +371,55 @@ AFAPI array idft(const array& in, const dim4 outDims);
 AFAPI array idft(const array& in);
 
 /**
-   C++ Interface for convolution any(one through three) dimensional data
+   C Interface for real to complex fast fourier transform for one dimensional signals
```
9prady9: this is not c interface, seems to be a typo.

include/af/signal.h
```diff
@@ -295,7 +371,55 @@ AFAPI array idft(const array& in, const dim4 outDims);
...
+
+   \ingroup signal_func_fft_r2c
+*/
+template<int rank>
+array fftR2C(const array &in,
+             const dim4& dims,
+             const double norm_factor = 0);
+
+/**
+   C Interface for real to complex fast fourier transform for one dimensional signals
```
9prady9: same typo as above.

src/backend/cpu/copy.cpp
```diff
@@ -63,6 +63,30 @@ namespace cpu
         return out;
     }
 
+    template<typename T>
+    void multiply_inplace(Array<T> &in, double val)
```
9prady9: Can't we do this using `detail::copyArray<inType, outType>` if we modify it to have scale parameter ? like the following
```
template<typename inType, typename outType>
void copyArray(Array<outType> &out, Array<inType> const &in, double scale_factor);
//....
Array<T> input = ....;
copyArray(input, input, scale_value);
```
Just wondering if that is possible.

9prady9: tested it in standard c++ and it is working fine. Not sure if this will cause any issues in CUDA/OpenCL kernels.

src/backend/cuda/copy.cu
```diff
@@ -70,9 +70,16 @@ namespace cuda
         ARG_ASSERT(1, (in.ndims() == dims.ndims()));
         Array<outType> ret = createEmptyArray<outType>(dims);
         kernel::copy<inType, outType>(ret, in, in.ndims(), default_value, factor);
+        CUDA_CHECK(cudaDeviceSynchronize());
         return ret;
     }
 
+    template<typename T>
```
9prady9: Oh, you are using kernel::copy for cuda then.

src/backend/cpu/copy.cpp
pavanky: CUDA and OpenCL just wrap other functions. I wrote CPU backend first and did not think about using existing functions. I will change this.

9prady9: Ya, i noticed that below.

9prady9:
:+1:

9prady9:
But why ci builds failing in cuda_mem_lock unit test ?

pavanky:
I disabled the memory manager accidentally in cuda backend. I will need to renable it.

9prady9:
Ah, ok.

9prady9:
Let me know once you do it, will merge it.

pavanky:
@9prady9 made all the changes. Wait until tests pass before merging.

9prady9:
Sure

PR #943, opened 2015-08-10
Stream Ops (Review only until string function is added)

### LINKED ISSUE #231: Add new functions to stream / read / write af::array
Function ideas

- For C API
```
af_readarray(af_array *arr, const char *file_name);
af_savearray(const char *file_name, const af_array arr);
``` 

- For C++ API
```
array readArray(std::string file_name);
void saveArray(std::string file_name, const array in);
ostream array::operator<<() 
```

shehzan10:
* Fixes #231

### DIFF
--- a/docs/details/util.dox
+++ b/docs/details/util.dox
@@ -1,0 +1,137 @@
+/**
+\addtogroup arrayfire_func
+@{
+\defgroup print_func_print print
+
+\brief Print the array to screen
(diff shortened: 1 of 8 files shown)


pavanky:
Can you also add `af_array_to_string` that has the same API as `af_print_array_gen` but takes in a char buffer as input ?

PR #952, opened 2015-08-18
modified cpu::getInfo to display CPU information

### LINKED ISSUE #903: Possibly add device properties for CPU backend
Right now it just provides a warning and does not return anything.

One bit of feedback we received over email says the following:

> Would it be possible to return some suitable information, e.g. platform=CPU or x86_64?

@umar456 @shehzan10 @9prady9 thoughts ?

9prady9:
Fixes #903


9prady9:
Samples outputs on Tegra and x86 architectures are given below, if any changes are needed please give feedback here.
Tegra
```
ArrayFire v3.1.0 (CPU, 32-bit Linux, build 30dd47a)
[0] Unknown: Unknown Max threads(1) 
```
x86
```
ArrayFire v3.1.0 (CPU, 64-bit Linux, build dc6b11c)
[0] Intel: Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz Max threads(4) GNU Compiler Collection(GCC/G++) 5.2.0
```
For debug mode, the toolkit(compiler used to build arrayfire) is also displayed.

shehzan10:
Is the compiler showing where the library was compiled or where the executable was compiled. It's very likely those two will be different.

9prady9:
This string is created using CMAKE_<LANG>_COMPILER_* variables. It now indicates with which compiler the library is compiled. Now that you said, CUDA/OpenCL backends show the current available toolkit/OpenCL version in the info. May be this field is irrelevant for CPU backend: showing compiler with which it was build doesn't make sense. May be showing supported SIMD features is better.

9prady9:
@pavanky What do you want to do about what @shehzan10 pointed out ?

pavanky:
@9prady9 What you did is alright. The compiler version in version.h is the one that is generated when compiling the library. This is the one that gets hard coded into the library. This is the one we should be reporting.

PR #954, opened 2015-08-18
FEAT Added image IO using memory functions and tests

### LINKED ISSUE #930: ImageIO in Memory
The plan is to add API to allow images to be loaded from/written to host side memory. FreeImage has the necessary API to do this.

Possible ArrayFire side API is:
```
array loadImageFromMemory(const void *ptr);  // Decodes and deep copies the image binary pointer to an array. Note the missing is_color argument.

void* saveImageToMemory(const char* file_ext, const array &in); // Deep copies and encodes the input array into a stream created by FI. 
```

In save to memory, the stream needs to be freed using FI API call at the end of its use. 
We can either copy it into a user defined buffer and return this to the user (immediately deleting the stream) and ask the user to free that memory similar to the host function.
The other option is to ask the user to call an AF api that in turn calls the FI api to free this memory.

shehzan10:
* Load and save image formats to memory
* Fixes #930

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -326,6 +326,7 @@ namespace af
     typedef af_colormap ColorMap;
     typedef af_norm_type normType;
     typedef af_ycc_std YCCStd;
+    typedef af_image_format imageFormat;
 }
 
 #endif
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -285,6 +285,22 @@ typedef enum {
     AF_COLORMAP_BLUE    = 6     ///< Blue hue map
 } af_colormap;
 
+typedef enum {
+    AF_FIF_BMP          = 0,    ///< FreeImage Enum for Bitmap File
+    AF_FIF_ICO          = 1,    ///< FreeImage Enum for Windows Icon File
+    AF_FIF_JPEG         = 2,    ///< FreeImage Enum for JPEG File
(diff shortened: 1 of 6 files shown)


include/af/defines.h
```diff
@@ -285,6 +285,22 @@ typedef enum {
...
+    AF_FIF_PNG          = 13,   ///< FreeImage Enum for Portable Network Graphics File
+    AF_FIF_PPM          = 14,   ///< FreeImage Enum for Portable Pixelmap (ASCII) File
+    AF_FIF_PPMRAW       = 15,   ///< FreeImage Enum for Portable Pixelmap (Binary) File
+    AF_FIF_TIFF         = 18,   ///< FreeImage Enum for Tagged Image File Format File
+    AF_FIF_PSD          = 20,   ///< FreeImage Enum for Adobe Photoshop File
+    AF_FIF_HDR          = 26,   ///< FreeImage Enum for High Dynamic Range File
+    AF_FIF_EXR          = 29,   ///< FreeImage Enum for ILM OpenEXR File
+    AF_FIF_JP2          = 31,   ///< FreeImage Enum for JPEG-2000 File
+    AF_FIF_RAW          = 34    ///< FreeImage Enum for RAW Camera Image File
+} af_image_format;
```
umar456: Create a typedef for af::imageFormat for C++

PR #960, opened 2015-08-20
Updating the cl.hpp version

ghisvail:
This PR allows to build ArrayFire on both `i386` and `amd64` architectures using the latest Khronos OpenCL headers.

### DIFF
--- a/src/backend/opencl/cl.hpp
+++ b/src/backend/opencl/cl.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008-2013 The Khronos Group Inc.
+ * Copyright (c) 2008-2015 The Khronos Group Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and/or associated documentation files (the
@@ -33,14 +33,14 @@
  *       Bruce Merry, February 2013.
  *       Tom Deakin and Simon McIntosh-Smith, July 2013
  *   
- *   \version 1.2.6
- *   \date August 2013
+ *   \version 1.2.7
+ *   \date January 2015
  *
  *   Optional extension support
  *
  *         cl
  *         cl_ext_device_fission
- *                #define USE_CL_DEVICE_FISSION
+ *				#define USE_CL_DEVICE_FISSION
  */
 
 /*! \mainpage
@@ -147,22 +147,17 @@
 
 #ifdef _WIN32
 
-#include <windows.h>
 #include <malloc.h>
-#include <iterator>
-#include <intrin.h>
 
-#if defined(__CL_ENABLE_EXCEPTIONS)
-#include <exception>
-#endif // #if defined(__CL_ENABLE_EXCEPTIONS)
-
-#pragma push_macro("max")
-#undef max
 #if defined(USE_DX_INTEROP)
 #include <CL/cl_d3d10.h>
 #include <CL/cl_dx9_media_sharing.h>
 #endif
 #endif // _WIN32
+
+#if defined(_MSC_VER)
+#include <intrin.h>
+#endif // _MSC_VER
 
 // 
 #if defined(USE_CL_DEVICE_FISSION)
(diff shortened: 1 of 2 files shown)


ghisvail:
added another commit which fix a cl::Platform construct inside the `WITH_GRAPHICS` if branch, which I did not test initially but was caught in the Jenkins build logs. Hopefully, this should build fine on osx and linux with `-DBUILD_GRAPHICS=ON` now.

9prady9:
@ghisvail Thank you, i started the ci builds for linux and osx. Will start windows once they are green.

9prady9:
@ghisvail Something went wrong in osx build, it is complaining about the following now
```
[ 35%] Building CXX object src/backend/opencl/CMakeFiles/afopencl.dir/platform.cpp.o
/var/lib/jenkins-slave/workspace/arrayfire-osx-pull/src/backend/opencl/platform.cpp:410:45: error: use of undeclared identifier 'CGLGetCurrentContext'
            CGLContextObj cgl_current_ctx = CGLGetCurrentContext();
```
Weird error!

ghisvail:
Seems that a similar problem happened in [Boost.Compute](https://github.com/boostorg/compute/issues/328) and is apparently due to OpenCL - OpenGL interoperability being different on OSX.

Since this is a completely separate issue, I'd suggest to solve it in a different topic branch and merge this PR if all the other platforms are ok.

pavanky:
@ghisvail We can not merge with a branch failing to build

ghisvail:
@pavanky I can't help you further on that. AFAIC this issue is fixed, and I could not care less about OSX specific issues.

pavanky:
@ghisvail I wasn't asking you to fix it. I was simply stating our policy. 

@9prady9 

The old `cl.hpp` file was including GL headers. The new one does not seem to be including any GL headers.

Can you merge this from command line on our osx machine and add the following to `platform.cpp` and check if it is compiling:

```
#if defined(__APPLE__) || defined(__MACOSX)
#include <OpenGL/OpenGL.h>
#include <libkern/OSAtomic.h>
#else
#include <GL/gl.h>
#endif // !__APPLE__
```

pavanky:
@umar456 Well this brings up another question, should we even have `cl.hpp` as a part of our repo ? It looks like cl.hpp is exists everywhere and can be found fairly easily.

9prady9:
@pavanky I am testing this change now on OSX.

shehzan10:
@pavanky The issue is that (as was previous pointed out) NVIDIA doesn't ship cl.hpp. And the FindOpenCL.cmake does look at the CUDA installation for cl.h (not for cl.hpp). Hence if the CUDA SDK is used for OpenCL headers, it will fail.

There are 2 ways to move forward:
1. Keep the status quo.
2. Remove cl.hpp from the repo and add a cl.hpp requirement to the FindOpenCL.cmake file. This will automatically disqualify all CUDA-based OpenCL header directories. It should not be so much of a problem for our build systems because we generally have an alternate (Intel/AMD) OpenCL SDK installed as well. However, this may become a potential problem for users compiling on windows with only a CUDA SDK. Users may be able to simply download the cl.hpp file and copy it to CUDA/include/CL to make it work.

pavanky:
@umar456 @shehzan10 

I think it would be nice to add option to find `cl.hpp` and if it is not found, automatically download it from Khronos as an external dependency.

umar456:
It looks like the cl.hpp does not have a stable API considering we had to change our source for a minor/patch update. We can do what @pavanky suggested but I am worried how we would handle minor changes to the headers on different systems.

9prady9:
@pavanky Including GL headers did the trick.

pavanky:
@umar456 Valid point about API breakage. 

@9prady9 Can you merge this PR from command line after changing `platform.cpp` to include GL headers ?

ghisvail:
Please also consider that Khronos no longer ships `cl.hpp` as part of OpenCL 2.0. A new header has been recently added though (`cl2.hpp`). So relying on the availability of a system-installed `cl.hpp` could prove equally as fragile in the future. FYI, Debian unstable now ships the Khronos OpenCL 2.0 headers only.

9prady9:
Handling this with #961

PR #965, opened 2015-08-24
Fixes to building nonfree algorithms

pavanky:
- `BUILD_NONFREE` is now `OFF` by default
- `BUILD_SIFT` is an additional option
- `BUILD_NONFREE` sets `BUILD_SIFT` to be `ON`
- A message about the patent is now displayed
- All non free source code files have "_nonfree" added to their file name.


pavanky:
@ghisvail @glehmann @marbre 

You guys should check out this pull request. We have SIFT in devel branch which is nonfree. It is disabled by default, but the source repo will still contain the algorithm. 

We will be using "_nonfree" to all source files that have non free algorithms. You can delete the non free parts using the following command.

```
find arrayfire -name "*_nonfree*" -delete
```

ghisvail:
I am happy with that, thanks for providing the filter :+1:

shehzan10:
BUILD_NONFREE is still on https://github.com/arrayfire/arrayfire/blob/devel/CMakeLists.txt#L33

ghisvail:
So we should be able to strip the non-free stuff from the release tarball / tagged release using:
```
find arrayfire -name "*_nonfree*" -delete
```
and be able to build the project and run the test suite with `-DBUILD_NONFREE=OFF` right ?

pavanky:
@shehzan10 You are looking at devel. Look at the file from this PR.

shehzan10:
Oops. I thought this was merged. My bad.

pavanky:
@ghisvail `BUILD_NONFREE` is `OFF` by default.  afopencl will build fine if you delete *_nonfree* with the default options. You can explicitly call 
`-DBUILD_NONFREE=OFF` if you choose to.

The test is also called "sift_nonfree", so if the deletion is done prior to building the test is also never built.

pavanky:
@umar456 wait for the builds to go through.

PR #975, opened 2015-08-28
Add C index_t creation helpers

jramapuram:
Adding helper functions to initialize `af_index_t` because this structure cannot be mangled to a rust type (because it has a union type). These helpers merely initialize an `af_index_t` internally (thus the ** pointer). It also checks to see if there was already a data item contained within the pointer (to prevent dangling pointers). 

Test:
```c++
#include <arrayfire.h>
#include <cstdio>
#include <cstdlib>

using namespace af;

extern "C" {
    af_err af_make_array_index(af_index_t** result, const af_array in);
    af_err af_make_seq_index(af_index_t** result, const af_seq in);
    af_err af_randu(af_array *out, const unsigned ndims, const dim_t * const dims, const af_dtype type);
    af_err af_print_array(af_array arr);
    af_err af_create_array(af_array *arr, const void * const data, const unsigned ndims, const dim_t * const dims, const af_dtype type);
}

{
    try {


        // Select a device and display arrayfire info
        int device = argc > 1 ? atoi(argv[1]) : 0;
        af::setDevice(device);
        af::info();

        printf("Test the index creation routine\n");
        long long n = 100;
        af_dtype type = af_dtype::f32;
        af_array F;
        af_randu(&F, 1, &n, type);
        af_index_t* test = NULL;
        printf("%d", af_make_array_index(&test, F));
        af_print_array(test->idx.arr);


    } catch (af::exception& e) {
        throw;
    }

    #ifdef WIN32 // pause in Windows
    if (!(argc == 2 && argv[1][0] == '-')) {
        printf("hit [enter]...");
        fflush(stdout);
        getchar();
    }
    #endif
    return 0;
}
```

Result:
```bash
Test the index creation routine
0[100 1 1 1]
    0.4385
    0.4604
    0.2502
    0.4947
    0.0530
    0.3377
    0.3968
    0.8744
    0.4822
    0.0428
    0.5084
    0.6545
    0.5126
    0.2643
    0.0520
    0.2920
    0.3194
    0.8109
    0.1541
    0.4452
    0.2080
    0.6110
    0.3073
    0.4156
    0.2343
    0.8793
    0.6462
    0.9264
    0.5786
    0.5538
    0.3557
    0.7229
    0.2783
    0.6192
    0.5876
    0.3750
    0.2405
    0.4148
    0.0937
    0.6326
    0.2849
    0.7793
    0.2133
    0.4131
    0.0328
    0.5360
    0.7214
    0.3546
    0.2527
    0.9847
    0.9200
    0.1872
    0.1820
    0.5984
    0.7082
    0.0421
    0.7305
    0.9400
    0.8432
    0.6116
    0.2674
    0.0208
    0.8331
    0.7218
    0.6087
    0.6301
    0.5298
    0.8127
    0.0593
    0.4797
    0.9910
    0.8793
    0.9996
    0.2068
    0.4372
    0.3727
    0.4468
    0.2376
    0.0345
    0.7308
    0.4942
    0.3031
    0.8094
    0.1293
    0.7832
    0.0732
    0.1236
    0.2229
    0.7420
    0.6273
    0.1528
    0.0119
    0.1726
    0.4784
    0.8051
    0.6872
    0.3022
    0.8077
    0.4071
    0.7513
```


src/api/c/index.cpp
```diff
@@ -36,6 +36,38 @@ static void indexArray(af_array &dest, const af_array &src, const unsigned ndims
...
+af_err af_make_array_index(af_index_t** result, const af_array in)
+{
+    try {
+        if(*result) {
+            // avoid a dangling pointer
+            return AF_ERR_NOT_SUPPORTED;
+        }
+        *result = new af_index_t;
+        (*result)->idx.arr = in;
+        (*result)->isBatch = true;
```
9prady9: @jramapuram Shouldn't batch be false like seq ?

9prady9:
Also, for the sake of completeness, i think we should have the following as well. @pavanky & @jramapuram What do you think ?

```
AFAPI af_free_index(af_index_t* indexer);
```

jramapuram:
Is there any benefit to a free call that merely is one line, i.e:
```c++
delete(indexer);
```

src/api/c/index.cpp
jramapuram: Hmm, I was under the impression that array_index has `isBatch = true;` & seq_index has `isSeq = true;' , i.e. this is mainly here to tell what is in the union struct.

9prady9:
:smile: hence my stmt `for the sake of completeness`

src/api/c/index.cpp
jramapuram: Oh I get it now. isSeq has sufficient information to tell if it is an array or seq.

Should we have an additional parameter then in the functions? i.e. 
```c++
af_err af_make_array_index(af_index_t** result, const af_array in, bool isBatch)
```

9prady9: I just checked the documentation for the struct. `isBatch` is used only if the indexer is an seq. So, it  will be false for af_array type indexer always.

jramapuram: Ok perfect. I will add the extra field to `af_make_seq_index` then.

jramapuram:
Actually.... we might need it from Rust now that I think about it... still need to free the C struct that we are holding a ptr to.

9prady9:
There you go, i missed the actual reason :smile:

pavanky:
Hmm, there is no reason for the af_index_t to be a double pointer over here. double pointers are only used when using vectors of some sort. The API is also in consistent  with af_make_seq (which returns a af_seq).

jramapuram:
So the entire issue is that we can't mangle the union over to rust.. af_seq is fine as it's a trivial struct.
You can't create (malloc/new) a new struct by using a single pointer.

These helpers allow for us to never need to hold the actual af_index_t, but a pointer to it.

pavanky:
Fair enough. That is a valid argument.

Can you change af_make_* to af_create_* to be consistent with other pointers to structs / classes we have ? This would also mean changing af_free_* to af_release_*

jramapuram:
Done. I already had af_free as af_release_* since I noticed that in the style of other functions.

pavanky:
Would it be possible to also create a new function "af_creae_seq_param_index" that takes begin, end and step instead of af_seq ? This way you do not have to call two functions when calling creating a index from a sequence.

jramapuram:
@pavanky Done

pavanky:
build arrayfire ci

jramapuram:
@pavanky : Is the OSX issue expected?

pavanky:
OSX is finicky. Issue not related to PR. I'll need to treat again.

pavanky:
@jramapuram 

I think we've overlooked something. How do you plan to index 2D matrices ? If you are creating one index at a time and you can not use unions in rust, it seems like you will not be able to create an array of af_index_t required for af_index_gen as well.

jramapuram:
Why can't we create an array of `af_index_t`?
It's simply a loop with:
```rust
let mySeqs = ...;
let mut indices = Vec::new();
for seq_array in mySeqs{
  indices.push(Index::new(seq));
}
```

where internally in `Index::new()` we call `af_create_seq_index` & holds a handle to it. 
Thus each `Index` type will be 1 `index_t`

pavanky:
@jramapuram Can you send `indices` to the `af_index_gen` in the C API ?

jramapuram:
I believe we can via a get() call that returns the raw handle:

```rust 
    pub fn index_gen(&self, input: Array, ndims: Dim4, indices: &[Index]) -> Result<Array, AfError> {
        unsafe{
            let mut temp: i64 = 0;
            let err_val = af_index_gen(&mut temp as MutAfArray
                                      , input.get() as AfArray
                                      , ndims.get().as_ptr() as DimT
                                      , indices.as_ptr() as *const IndexT);
            match err_val {
                0 => Ok(Array::from(temp)),
                _ => Err(AfError::from(err_val)),
            }
        }
    }
``` 

This is similar to how we pass in Dim4 while creating a new array:
```rust
    pub fn new<T>(dims: Dim4, slice: &[T], aftype: Aftype) -> Result<Array, AfError> {
        unsafe {
            let mut temp: i64 = 0;
            let err_val = af_create_array(&mut temp as MutAfArray,
                                          slice.as_ptr() as *const c_void,
                                          dims.ndims() as c_uint,
                                          dims.get().as_ptr() as * const c_longlong,
                                          aftype as uint8_t);
            match err_val {
                0 => Ok(Array {handle: temp}),
                _ => Err(AfError::from(err_val)),
            }
        }
```

pavanky:
Hmm, does this return `af_index_t *` or `af_index_t **` ?  I am asking because you are storing a vector of `af_index_t *`

jramapuram:
Each `Index::new()` call returns 1 `af_index_t*` (you can think of this as a malloc of 1)

However, with the above function you can pass in an array of `Index`:
```rust
indices: &[Index]
```

Now, when do:
```rust
indices.as_ptr() as *const IndexT);
```
this is converting the array to a pointer and simply typecasting it to a const pointer of type `af_index_t` (aka af_index_t*)

Sidenote: the above `index_gen` function doesnt need to be a struct member [ie no &self is needed] and it can just be part of the module

9prady9:
@pavanky  I just noticed it is branched off from master, shouldn't this PR branch be based off devel rather than master ?

pavanky:
@9prady9 I think it is fine.

9prady9:
Ok.

9prady9:
:+1:

jramapuram:
Hmm you are right @9prady9 . Still, all new functions though.

9prady9:
@jramapuram @pavanky I am going to push the changes we discussed over gitter regarding these utility functions to #977 PR. Closing this PR.

PR #977, opened 2015-08-28
Heterogeneous API - backend independent C-API for ArrayFire

9prady9:


pavanky:
@9prady9 For now manually include `src/api/c/seq.cpp` while compiling hapi. This solves the issue.

CMakeLists.txt
```diff
@@ -168,6 +170,11 @@ IF(${BUILD_OPENCL})
     ADD_SUBDIRECTORY(src/backend/opencl)
 ENDIF()
 
+IF(${BUILD_HETEROGENOUS_API})
+    ADD_SUBDIRECTORY(src/api/hapi)
+    ADD_SUBDIRECTORY(hapi_examples)
```
pavanky: I don't think the examples need to exist in different directories. We can add dummy variations of all hapi specific functions in each backend.

hapi_examples/test.cpp
```diff
@@ -0,0 +1,67 @@
...
+    af_set_backend(AF_BACKEND_OPENCL);
+    testBackend();
+
+    #ifdef WIN32 // pause in Windows
+    if (!(argc == 2 && argv[1][0] == '-')) {
+        printf("hit [enter]...");
+        fflush(stdout);
+        getchar();
+    }
+    #endif
```
pavanky: Why not use the C++ API ? We should have `setBackend()` in C++ as well.

include/af/defines.h
```diff
@@ -316,6 +321,12 @@ typedef enum {
     AF_FIF_RAW          = 34    ///< FreeImage Enum for RAW Camera Image File
 } af_image_format;
 
+typedef enum {
+    AF_BACKEND_CPU,
+    AF_BACKEND_CUDA,
+    AF_BACKEND_OPENCL
```
pavanky: Add values.

include/af/hapi.h
```diff
@@ -0,0 +1,21 @@
...
+ ********************************************************/
+
+#pragma once
+#include <af/defines.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+AFAPI af_err af_set_backend(const af_backend bknd);
```
pavanky: Document this

src/api/hapi/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
+
+ADD_LIBRARY(af SHARED
+            ${hapi_headers}
+            ${hapi_sources})
+
+IF(${BUILD_CPU})
+    ADD_DEPENDENCIES(af afcpu)
+ENDIF()
+
```
pavanky: Technically hapi does not need to wait for any backend to be built. There is no reason for this.

src/api/hapi/symbol_manager.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (cudaHandle) {
+        isCUDALoaded = true;
+        activeHandle = cudaHandle;
+    }
+    oclHandle = openDynLibrary(LIB_AF_OCL_NAME);
+    if (oclHandle) {
+        isOCLLoaded = true;
+        activeHandle = oclHandle;
+    }
+}
```
pavanky: We need to figure out the preferred order or atleast provide the user with a choice. Right now it defaults to OpenCL > CUDA > CPU.

src/api/hapi/symbol_manager.cpp
```diff
@@ -0,0 +1,102 @@
...
+        isOCLLoaded = false;
+    }
+}
+
+af_err AFSymbolManager::setBackend(af::Backend bknd)
+{
+    af_err retCode = AF_SUCCESS;
+    activeBknd = bknd;
+    switch (activeBknd) {
+        case af::Backend::AF_BACKEND_CPU:
```
pavanky: If you make `af_backend` an enum with pre defined numbers you can have a lookup table like `isBackendLoaded[kind]`

src/api/hapi/symbol_manager.hpp
```diff
@@ -0,0 +1,80 @@
...
+            funcHandle = (af_func)GetProcAddress(activeHandle, symbolName);
+#else
+            funcHandle = (af_func)dlsym(activeHandle, symbolName);
+#endif
+            if (!funcHandle) {
+                return AF_ERR_SYM_LOAD;
+            }
+            return funcHandle(args...);
+        }
+
```
pavanky: This function has to be the best thing about this entire process :D

src/api/hapi/symbol_manager.hpp
```diff
@@ -0,0 +1,80 @@
...
+    private:
+        bool isCPULoaded;
+        bool isCUDALoaded;
+        bool isOCLLoaded;
+
+        LibHandle cpuHandle;
+        LibHandle cudaHandle;
+        LibHandle oclHandle;
+
```
pavanky: Read what I said earlier about lookup tables

pavanky:
@9prady9 good work! This went faster than I thought. Are the tests passing on the dev machines?

CMakeLists.txt
```diff
@@ -168,6 +170,11 @@ IF(${BUILD_OPENCL})
     ADD_SUBDIRECTORY(src/backend/opencl)
 ENDIF()
 
+IF(${BUILD_HETEROGENOUS_API})
```
pavanky: Change this be "BUILD_AF"

include/af/defines.h
umar456: Also add AF_BACKEND_DEFAULT which will pick CUDA->OpenCL -> CPU

hapi_examples/test.cpp
9prady9: Sure, it's just that i just worked on only C-API right now.

include/af/defines.h
9prady9: That's the behavior anyway right except that there is no specific enum for that.

src/api/hapi/CMakeLists.txt
9prady9: Sure, hapi is just an c++ library that will use af*.so's. We can remove the dependency.

src/api/hapi/symbol_manager.cpp
9prady9: I didn't quite understand what is the choice here ? isn't that the reason af_set_backend function is there. In the even of such function is not called upon, activeHandle points to the backend in the order you just mentioned.

src/api/hapi/symbol_manager.hpp
9prady9: true.

src/api/hapi/symbol_manager.cpp
9prady9: agree.

9prady9:
@pavanky All builds passed. I think we can merge this if there are no more additions or missing functionality.

pavanky:
Wait until #975 is merged into devel, merge from devel and add those functions as well.

pavanky:
@9prady9 
Can you add the list of new functions added in the description of this issue: https://github.com/arrayfire/arrayfire/issues/984

include/af/hapi.h
```diff
@@ -0,0 +1,56 @@
+/*******************************************************
```
pavanky: I think this file should now be called `af/backend.h`

pavanky:
Tests are going to fail if all tests are being run instead of specifying each backend separately using `ctest -R backend`. This is because `libaf.so` does not currently know the location of each individual file at the time of the test.

pavanky:
There is also the following error that occurs at the exit of each program. I am not sure if this is limited to `belle`.

```
 $ ./build/test/info_unified 
Running main() from gtest_main.cc
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Info/0, where TypeParam = float
[ RUN      ] Info/0.All
ERROR: GLFW wasn't able to initalize
ArrayFire v3.1.0 (OpenCL, 64-bit Linux, build 4cb64cb)
[0] NVIDIA  : Tesla K20c 
-1- NVIDIA  : Quadro 6000 
-2- NVIDIA  : Tesla K20c 
ArrayFire v3.1.0 (OpenCL, 64-bit Linux, build 4cb64cb)
-0- NVIDIA  : Tesla K20c 
[1] NVIDIA  : Quadro 6000 
-2- NVIDIA  : Tesla K20c 
ArrayFire v3.1.0 (OpenCL, 64-bit Linux, build 4cb64cb)
-0- NVIDIA  : Tesla K20c 
-1- NVIDIA  : Quadro 6000 
[2] NVIDIA  : Tesla K20c 
[       OK ] Info/0.All (699 ms)
[----------] 1 test from Info/0 (699 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (699 ms total)
[  PASSED  ] 1 test.

[ FATAL ] /home/pavan/Work/arrayfire/test/gtest/include/gtest/internal/gtest-port.h:2019:: pthread_key_delete(key_)failed with error 22
Aborted (core dumped)
```

pavanky:
@shehzan10 @umar456 @9prady9 

One way to deal with tests would be to use a priority list of paths for loading the libraries instead of looking for absolute paths. This includes something like the following.

1) `libaf$BACKEND`
2) `$AF_PATH/lib/libaf$BACKEND`
3) `$AF_BUILD_PATH/src/build/lib$BACKEND`

SO when running tests we could run using something like

`AF_BUILD_PATH=/full/path/to/af/build ctest -R unified`

9prady9:
@pavanky @umar456  @shehzan10 If anyone of you also tried the latest changes, most of the functions passed tests on `unified` category as well. Exceptions are given below
* sort_by_index_unified
* sort_index_unified
* blas_unified
* imageio_unified

Both sort ones failed with value errors where as the other two were seg faults. Are you seeing anything similar on your machine too ?

9prady9:
@pavanky I have just pushed the changes related to search path order you have suggested earlier. Please take a look at them.

pavanky:
@9prady9 I am not sure how the `*_unified` tests were even running. The dynamic loader was perhaps picking up some other files?

9prady9:
Looks like there are some issues on windows.
Didn't quite understand your last comment, we are not overriding any system path library ? PATH or LD_LIBRARY_PATH is the first place the library is searched for.

9prady9:
@pavanky What should we do about the failing tests in unified category ?

9prady9:
Figured out the fix for imageio test. However, i am still perplexed why sort tests are failing.

9prady9:
build arrayfire ci

PR #988, opened 2015-09-08
Try PkgConf first to find LAPACKE

marbre:
I adjusted `FindLAPACKE.cmake` to use PkgConf to find lapacke. The problem with the current implementation on Gentoo is that the libraries are not named liblapacke and liblapack. On Gentoo the name of the library depends on the installed package, allowing to install multiple lapack packages side-by-side. To show the differences I quote from the *.pc files:

lapacke-reference
~~~
Libs: -L/usr/lib64 -lreflapacke
~~~

lapack-reference
~~~
Libs: -L/usr/lib64 -lrefblas
~~~

atlas
~~~
Libs: -L${libdir} -lptlapack 
Libs.private: -L${libdir} -latlas -lm -lpthread
~~~

Please let me know if there is anything to improve this patch.

### DIFF
--- a/CMakeModules/FindLAPACKE.cmake
+++ b/CMakeModules/FindLAPACKE.cmake
@@ -18,70 +18,23 @@ FIND_PACKAGE(PkgConfig)
 
 #Determine from PKG
 IF(PKG_CONFIG_FOUND AND NOT LAPACKE_ROOT)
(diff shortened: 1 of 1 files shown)


marbre:
Redone against hotfixes branch as PT #991.

PR #989, opened 2015-09-08
Try PkgConf first to find CBLAS

marbre:
Quite similar to PR #988. This fixes issue #761. Any feedback is appreciated.


pavanky:
We can't have PkgConf as REQUIRED on windows. It needs to be optional and only be enabled on Linux systems.

marbre:
Thanks for the comment. Removed REQUIRED and made PKG_CHECK_MODULES optional like in FindLAPACKE.cmake.

marbre:
@pavanky Removed the MESSAGE here and also in PR #988.

pavanky:
@marbre Sorry to do this again, I think the commits need to go to hotfixes-3.1.1 instead of devel.

marbre:
@pavanky No problem. Done so for the FindLAPACKE.cmake PR too.

PR #994, opened 2015-09-11
fix freeimage bitmap resource leaks

vakopian:
Currently, in `af_load_image()`, `af_load_image_memory()`, `af_save_image()` and `af_save_image_memory()` functions, if an error happens after the FreeImage bitmap is loaded, these functions are exited (e.g. by throwing an exception) without unloading the bitmap. This results in resource leaks.
The proposed patch fixes this using the RAII wrapper class called `FI_BitmapResource`. The same could be done using the C++11's `unique_ptr`, but I'm not sure if ArrayFire has made the move to the new standard yet.

### DIFF
--- a/src/api/c/imageio.cpp
+++ b/src/api/c/imageio.cpp
@@ -58,6 +58,24 @@ static void FI_Init()
     static FI_Manager manager = FI_Manager();
 }
 
+class FI_BitmapResource
+{
+public:
+    explicit FI_BitmapResource(FIBITMAP * p) :
+        pBitmap(p)
+    {
+    }
+
+    ~FI_BitmapResource()
+    {
+        FreeImage_Unload(pBitmap);
+    }
+private:
+    FIBITMAP * pBitmap;
+};
+
+
+
 // Helpers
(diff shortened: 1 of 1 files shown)


pavanky:
Why not use `unique_ptr` and `FreeImage_Unload` as the deleter ?

vakopian:
Right, that would be the preferred way to do it - does ArrayFire already require C++11 compliant compiler ? If yes I'll make another pull request using unique_ptr

pavanky:
Good point. It is required for CPU and OpenCL backends. However the CUDA backend does not require it because older versions of CUDA did not support it.

I am going to accept it and change it to use unique pointers later if it is compliant.

pavanky:
build arrayfire ci

pavanky:
@vakopian These are problems with our build system. We are looking into it.

9prady9:
build arrayfire ci

9prady9:
@pavanky I just noticed this, this PR is a bug fix, hence shouldn't it go into hot-fixes for 3.1.1 ?

pavanky:
This isn't critical. It can stay in devel.

9prady9:
Ok, merging then.

PR #1007, opened 2015-09-18
Unified Documentation

shehzan10:
* Change unified backend default priority to CUDA -> OpenCL -> CPU
* Updated unified/basic.cpp to use C++ API
* Added documentation for unified backend


pavanky:
@shehzan10 approx is breaking on OSX

shehzan10:
Looks like its running out of memory.
I'll check tegra too.

PR #1031, opened 2015-09-28
Unified fixes

shehzan10:


shehzan10:
build arrayfire ci

shehzan10:
@pavanky @umar456 @9prady9 This is something I discovered from our OSX build machine failing the PR.

When the machine has CUDA Toolkit installed, but does not have a CUDA device, it can build CUDA (along with CPU and OpenCL).
Then, when running the unified backend, the CUDA backend and it's libs will get loaded fine and unified backend run the device setup for CUDA. And this is where ArrayFire will throw an exception as no devices will be found.

This is not an issue when the CUDA SDK is unavailable. In that case CUDA will fail to load and unified will fallback to OpenCL.

To fix this I can think of 2 solutions:
1. Simply move the afcuda lib to a back up location. In this case afcuda lib will not be found. This is essentially do nothing from our end.
2. Check for environment variables like AF_UNIFIED_DISABLE_BACKEND where BACKEND is cpu/cuda/opencl.

What do you guys think? Any other solutions?

9prady9:
IMHO, i don't think this is something that we should fix in code. If we do such a fix for `unified` API, then to be uniform, we should do a similar fix in `c` API too - where a built backend will not crash if compatible device doesn't exist (I am under the impression that this is the current behavior - cuda lib can be built if cuda sdk is present but any cuda application instance will crash since the machine has is no cuda device, let me know if i am wrong).

shehzan10:
build arrayfire ci

PR #1033, opened 2015-09-29
Unit tests for statistics functions

### LINKED ISSUE #780: Add unit tests for statistics wrapper functions
* [x] mean
* [x] var
* [x] stdev
* [x] cov
* [x] median
* [x] corrcoef

9prady9:
Fixes #780 

Unit tests for the following statistics functions
* median
* corrcoef
* mean
* var
* cov
* stdev


shehzan10:
I think we are losing a lot of performance here by doing a cast array. In my opinion, the casting should happen at the reduction or at the multiplication of the weights (which is also being casted to output type). https://github.com/arrayfire/arrayfire/blob/devel/src/api/c/stats.h#L46

9prady9:
I am not entirely sure that is the case, castArray uses JIT i think.

shehzan10:
Just because its JIT does not mean it does not take time. JIT ops when combined with other JIT ops are helpful. Here you are combining one JIT op (castArray) with a reduce_all. Hence the JIT op has to be completed synchronously hence providing no benefit. This essentially turns the castArray into a synchronous memcpy.

9prady9:
I never said, JIT won't take any time :smile: but agree about what you said regarding castArray turning into synchronous op.

Will try to separate out-type from in-type for common mean function.

shehzan10:
build arrayfire windows ci

9prady9:
build arrayfire windows ci

9prady9:
Deferring it through #1034

PR #1037, opened 2015-10-01
FindCBLAS.cmake: s/cblas/blas in pkg-config mode

ghisvail:
BLAS detection does not work on Debian / Ubuntu due to the `libblas-dev` package shipping a `blas.pc` alongside headers and libraries and not a `cblas.pc`. This patch solves the detection.


pavanky:
@ghisvail This was contributed by @marbre to work on Gentoo. Replacing cblas with blas may break builds on his package.  

@marbre Can you check if simply using `blas` instead of `cblas` works on gentoo ? If so we could just accept the change. If not we should perhaps check for `blas` when `cblas` fails ?

ghisvail:
It would not harm to check for both.

marbre:
On Gentoo cblas and blas can be provided by different packages. For example the user can install blas-reference and cblas-reference. Another possibility is to install Atlas, which provides both interfaces. The user finally can choose which library to use with `eselect {blas,cblas}`.

Of cause, cblas requires a blas package installed, but not the other way around. Thus, checking for blas, it can not be guaranteed that cblas is available. Further `cblas.pc` and `blas.pc` pass back different Libs:

***cblas.pc***
~~~
prefix=/usr
libdir=${prefix}/lib64
includedir=${prefix}/include
Name: refcblas
Description: C wrapper interface to the F77 reference BLAS implementation
Version: 20110218
URL: http://www.netlib.org/blas/
Libs: -L${libdir} -lrefcblas
Private: -lm
Cflags: -I${includedir}/cblas
Requires: blas
~~~

***blas.pc***
~~~
prefix=/usr
libdir=/usr/lib64

Name: refblas
Description: Basic Linear Algebra Subprograms F77 reference implementations
Version: 
URL: http://www.netlib.org/refblas/
Libs: -L/usr/lib64 -lrefblas
Libs.private: -lm
~~~

I assume the cleanest way could be that Debian provides a blas.pc **and** a cblas.pc, with `libblas-dev` and `libatlas-base-dev`. But thats somehow a matter of taste.
If not, and if `PKG_CHECK_MODULES(PC_CBLAS cblas)` fails, the cblas can still be found with the other code in the `FindCBLAS.cmake` module, which I didn't touched. Thus, detecting cblas should also work on Debian without the patch, since I assume it worked before commit 644d074867ff35821a87b3b6794efe6e7cd4006a.

pavanky:
@marbre thanks for the feedback.

@ghisvail I am on vacation and have limited time for testing. Do you mind maintaining this patch in your fork until we have a more permanent fix ?

ghisvail:
@pavanky I am happy to hold on that patch. Afterall, it is distro specific.

> On Gentoo cblas and blas can be provided by different packages.
Also the case on `Arch` apparently. Not the case on `Debian` and `Fedora` derivatives.

> I assume the cleanest way could be that Debian provides a blas.pc and a cblas.pc.
Fair point, worth asking the Debian maintainers about this.

Since I don't want to break `Gentoo`'s setup, I am now convinced that this patch should be kept downstream on Debian.

Feel free to close if you agree.

PR #1048, opened 2015-10-05
Fix memory leak in median & Windows fixes

### LINKED ISSUE #1045: memory leak in af::median
Using af::average to repeatedly calculate the average of an image works fine, and takes **less than a second**:

    try {
      for (int i=0; i<1000; i++) {
        std::cout << "I: " << i << std::endl;
        af::array a = af::randn(1000, 1000);
        af_print(af::mean(af::flat(a)));
      }
    } catch (af::exception &exception) {
      std::cout << "Exception: " << exception.what();
    }

However, as soon as I switch from mean to median:

    try {
      for (int i=0; i<1000; i++) {
        std::cout << "I: " << i << std::endl;
        af::array a = af::randn(1000, 1000);
        af_print(af::median(af::flat(a)));
      }
    } catch (af::exception &exception) {
      std::cout << "Exception: " << exception.what();
    }

It takes around **4 seconds** and then aborts at iteration ~750 / 1000 with the following error:

    Exception: ArrayFire Exception(999): Unknown error
    In C:\workspace\af_inst_source\src\api\cpp\median.cpp:3

Using it in a real project, I also got a CUDA out of memory error in another function after some time, which suggests to me that there must be a memory leak in af::median.

### LINKED ISSUE #1047: Fix windows build with blas excluding MKL
Currently the builds on windows fail when not using MKL.
This is probably because of the cmake options/defines.

shehzan10:
Fixes #1045 
Fixes #1047

### DIFF
--- a/src/api/c/median.cpp
+++ b/src/api/c/median.cpp
@@ -72,7 +72,8 @@ static af_array median(const af_array& in, const dim_t dim)
     af_seq slices[4] = {af_span, af_span, af_span, af_span};
     slices[dim] = af_make_seq(mid-1.0, mid-1.0, 1.0);
 
-    AF_CHECK(af_index(&left, getHandle<T>(sortedIn), input.ndims(), slices));
+    af_array sortedIn_handle = getHandle<T>(sortedIn);
+    AF_CHECK(af_index(&left, sortedIn_handle, input.ndims(), slices));
 
     if (nElems % 2 == 1) {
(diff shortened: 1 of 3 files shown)


pavanky:
@shehzan10 Can you run through all `src/api/c/` files and check if other places are doing something like this ?

shehzan10:
This does not seem to be a concern in any other use of getHandle.

PR #1053, opened 2015-10-08
CPU Fallback for CUDA LAPACK API for CUDA <= 6.5

shehzan10:
Adds fallback CPU capability for CUDA lapack functions who's implementations were disabled for versions of CUDA prior to 7.0.


shehzan10:
build arrayfire tegra ci

shehzan10:
build arrayfire tegra ci

pavanky:
@shehzan10 Instead of using CPU fallback directly, make it an option so that the user has to enable it explicitly.

PR #1064, opened 2015-10-19
Misc Fixes

### LINKED ISSUE #541: Document supported and the return types of all the functions
Some functions return outputs that are different types than the inputs. Document the behavior for such functions.

Functions to document:

- [x] sum
- [x] product
- [x] anyTrue
- [x] allTrue
- [x] count
- [x] histogram
- [x] logical operations

@9prady9 @shehzan10 Anything else we can add here?

### LINKED ISSUE #1067: Unified API: Add function to return WHICH backends are available
Currently `af_get_backend_count` returns the number of available backends, however it would be useful to expose **which** backends are available (or separate functions for `af_is_cuda_available`, etc). This calculation is already being done internally during the function call.

shehzan10:
[skip ci]

Fixes #773 
Fixes #541
Fixes #1067


shehzan10:
build arrayfire windows ci

PR #1076, opened 2015-10-29
Replaced deviceSychronize calls with async versions

9prady9:
[skip ci]

All checks have passed, last commit is just a style fix.


pavanky:
What about OpenCL?

9prady9:
Will check on that as well. It crossed my mind late after sending the PR for cuda backend.

9prady9:
@pavanky I have checked OpenCL backend, all the device to host (& vice versa) copy calls are using `CL_TRUE` which is equivalent of CUDA's `async copy followed by stream sync`. So, OpenCL backend doesn't need any modifications.

The PR is ready for merge.

PR #1091, opened 2015-11-06
Version guards, dims check

shehzan10:
[skip ci]

### DIFF
--- a/src/api/c/device.cpp
+++ b/src/api/c/device.cpp
@@ -144,9 +144,12 @@ af_err af_device_array(af_array *arr, const void *data,
         AF_CHECK(af_init());
 
         af_array res;
-        af::dim4 d((size_t)dims[0]);
-        for(unsigned i = 1; i < ndims; i++) {
+
+        DIM_ASSERT(1, ndims >= 1);
+        dim4 d(1, 1, 1, 1);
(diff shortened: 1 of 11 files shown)


pavanky:
@shehzan10 Can you make sure there are no other places like this?

src/api/c/device.cpp
```diff
@@ -144,9 +144,12 @@ af_err af_device_array(af_array *arr, const void *data,
...
         af_array res;
-        af::dim4 d((size_t)dims[0]);
-        for(unsigned i = 1; i < ndims; i++) {
+
+        DIM_ASSERT(1, ndims >= 1);
+        dim4 d(1, 1, 1, 1);
+        for(unsigned i = 0; i < ndims; i++) {
             d[i] = dims[i];
+            DIM_ASSERT(3, dims[i] >= 1);
         }
```
shehzan10: @pavanky while the older implementation in this would still be accurate, I think it is important to ensure that ndims >= 1 before running the loop. If you think this is not necessary, I'll revert the commit.

shehzan10:
I found the af_device_array to be the only dubious case.

shehzan10:
Merge via command line. Sending with PR #1097

PR #1109, opened 2015-11-18
Build and install documentation in a separate output folder.

ghisvail:
This PR provides more conventional settings for building and installing the Doxygen generated HTML documentation. The HTML documentation is now generated under a separate `html` directory instead of raw `${CMAKE_CURRENT_BINARY_DIR}`.

Besides this approach being standard practice, it has these additional benefits:
1) No tempering by `CMake` specific cruft. Hence the removal of the `CMakeFiles` exclusion pattern,
2) No conflicts with other potential docs (manpage, latex, pdf...). Hence the removal of the `man` exclusion pattern.
3) The HTML documentation is directly installed under `${AF_INSTALL_DOC_DIR}/html` which is the expected install path for HTML related content.

In addition, the `COMMAND cmake -E remove ${CMAKE_CURRENT_BINARY_DIR}/.git` is removed due to the fact that the `.git` folder in `assets` may have already been stripped from the release tarball, as it is the policy in Linux distributions. Instead an exclusion pattern ".git" is used on the corresponding installation target.


pavanky:
@ghisvail Erm, any reason this was closed ? Our build infra structure was incomplete last week. I was waiting until it was back to normal before merging. Did you just happen to do a clean up and forget the PR ?

ghisvail:
Oops I must have deleted my branch by mistake. I'll check if i kept a copy
locally, otherwise I'll resubmit.

Sorry about that.

pavanky:
@ghisvail Doesn't this PR show you a revert button ?

ghisvail:
Unfortunately, it does not. I'll resubmit. Apologize for the mess.

ghisvail:
Actually it does, my bad. Re-opened.

shehzan10:
This PR is opened for devel. However, it can go into hotfixes. I'll be merging it into hotfixes via command line.

shehzan10:
@pavanky @ghisvail the documentation is getting built at build/doc/html and getting installed to package/share/ArrayFire/doc/html (where share/ArrayFire/doc is AF_INSTALL_DOC_DIR). The ArrayFire/doc directory has no files/dirs other than the html directory.

I understand that this is how the PR was designed to be, but is this the way we want to ship installers?

pavanky:
@shehzan10 I think the idea is that there will be other sub directories in the future. I think it is fine to ship it this way.

@marbre Will this affect the gentoo packaging in anyway?

ghisvail:
> I think the idea is that there will be other sub directories in the future.

That's the idea. It can be manpages, example source code, latex files, pdf builds...

You might also have both an API (via Doxygen) and a user (via Sphinx for instance) HTML documentation.

shehzan10:
Merged by #1115

marbre:
@pavanky Thanks for keeping Gentoo in mind. Shouldn't be a problem with our ebuild, otherwise I'll adjust the ebuild as soon as I bump to 3.2.1.

PR #1110, opened 2015-11-21
Add missing linkage with libdl

ghisvail:
As discussed previously on gitter, this commit adds the missing linkage of the unified backend to the system `libdl`.


shehzan10:
Merged via command line through #1115

PR #1111, opened 2015-11-21
Minor docs fixes

shehzan10:
[skip ci]

The current documentation for indexing operator uses dummy groups (http://www.arrayfire.com/docs/group__array__mem__operator__paren__many.htm and http://www.arrayfire.com/docs/group__array__mem__operator__paren__one.htm).

Removed these 2 groups and using a single group for operator.

This even more adversely affects older versions of doxygen as the dummy groups are listed at the top level. This change fixes that as well.

### DIFF
--- a/docs/details/array.dox
+++ b/docs/details/array.dox
@@ -16,27 +16,6 @@ to \ref af::array objects.
 
 ===============================================================================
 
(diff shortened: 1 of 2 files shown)


shehzan10:
Merged via command line through #1115

PR #1119, opened 2015-11-24
Tests as standalone project

### LINKED ISSUE #1108: Make the ArrayFire testsuite a separate project
As for the [examples](https://github.com/arrayfire/arrayfire/blob/devel/examples/CMakeLists.txt), it would be really nice to make the testsuite a separate project, and to support building the tests either with the ArrayFire targets or existing binaries.

The specific use case I have in mind concerns continuous integration inside Linux distributions [1][2], whereby binary packages gets automatically tested on changes to their corresponding dependencies.

[1] https://ci.debian.net
[2] https://fedoraproject.org/wiki/Koschei

shehzan10:
Fixes #1108 

Test using
```
cmake arrayfire/test -DArrayFire_DIR=/path/to/af_install/share/ArrayFire/cmake
```

Allows BUILD_NONFREE and USE_RELATIVE_TEST_DIR options as with the previous build action.

Still requires submodules. If not found, will try to init and update them. The expectation is that the source test directory will always be within arrayfire, ie. arrayfire/test.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,5 +1,5 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
-PROJECT(arrayfire-examples)
+PROJECT(ArrayFire-Examples)
 
 # Find CUDA and OpenCL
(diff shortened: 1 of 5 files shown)


shehzan10:
Closing this and opening PR on hotfixes.

PR #1120, opened 2015-11-24
Tests standalone

### LINKED ISSUE #1108: Make the ArrayFire testsuite a separate project

shehzan10:
Fixes #1108

Test using
```
cmake arrayfire/test -DArrayFire_DIR=/path/to/af_install/share/ArrayFire/cmake
```

Allows BUILD_NONFREE and USE_RELATIVE_TEST_DIR options as with the previous build action.

Still requires submodules. If not found, will try to init and update them. The expectation is that the source test directory will always be within arrayfire, ie. arrayfire/test.

If the test directory has the submodules, then there is no need for it to be in arrayfire. It can be referred to directly as `cmake /path/to/test -DArrayFire_DIR=/path/to/af_install/share/ArrayFire/cmake`


shehzan10:
@ghisvail Can you test this out at your end. I've tried to cover most possibilities. Let me know if I missed anything.

pavanky:
@shehzan10 This is not necessary. The test directory should mimic the examples directory. There is no need to introduce the missing cmake modules.

shehzan10:
It is mimicking the examples. The examples directory also has its own CMakeModules directory.

pavanky:
Hmm. Perhaps we need to clean that up as well. Let me review it.

shehzan10:
For the examples, when you ship it with the Installers, the FindOpenCL file is going to be required. Hence the modules directory. Same case with tests.

Issue #1029, opened 2015-09-28
ArrayFire to OpenGL


rggjan:
Is it possible to go from ArrayFire to OpenGL and back using an AMD graphics card?

There is an article on how to do it using CUDA (http://arrayfire.com/remote-off-screen-rendering-with-opengl/). Any hints on using a similar method for OpenCL, so it can be used cross-platform?

What I would like to do in the end is take an af::array, convert it to an OpenGL texture and render it with my custom shader to the screen or to another texture (assuming a given OpenGL context, for example from Qt).

9prady9:
@rggjan 
As per the current state of the ArrayFire framework, it is not possible to do this over OpenCL backend. Reason being, currently we do not expose the API necessary to allow the user to set an OpenCL context or pass on an OpenGL context for sharing. The issue #1002 is related to the same topic. Once we have the functionality being discussed in #1002 implemented in ArrayFire, you would be able to do what you explained.

Having said that, can you please explain in detail (if that is possible) what this custom shader is doing? 

ArrayFire uses OpenGL-Interoperability for it's graphics. If your custom shader is doing something that others might also use, we are glad to add it as an option for [`af::image`](http://www.arrayfire.com/docs/classaf_1_1Window.htm#aed174450be6db4dfa5ef65b7baccbe57) function in ArrayFire. As of now, we can do the following while displaying an array as image in ArrayFire.
* Set a [ColorMap](http://www.arrayfire.com/docs/defines_8h.htm#a553ceda8a1d8946efac3b08e642574ae) for image display
* Request the ArrayFire to maintain aspect ratio of the image while displaying.

However, if you are planning on use ArrayFire as compute backend for some Qt application, you would have to some data management (copying etc.) like [this](https://github.com/9prady9/wildfire) application (This app is still a work in progress, so it might not be doing everything in an ideal way).

rggjan:
Thanks for the explanations.

My use case is actual 3D rendering: I would like to process an image (out of a video stream) with arrayfire functions, and then use that as a texture in an interactive 3D OpenGL scene.

I don't think rendering triangles and polygons with arrayfire makes a lot of sense, so I would like to go with direct OpenGL for that.

I can currently do this by copying the data from an af::array to the CPU and uploading it again to the GPU as a texture, but that seems like wasted time to me, since it was on the GPU memory anyway. 

[...]

PR #1140, opened 2015-12-04
Functionality to support integration with external OpenCL contexts

### LINKED ISSUE #1002: Using arrayfire with external OpenCL context
To use arrayfire together with external libraries it would be helpful if external contexts could be used.

1) Minimum solution:
Allow the user to call something like afcl_init(device_id, context); before using the library.

2) Great solution:
Allow the user to switch between different contexts, or even different queues calling something like: afcl_set_context(context), afcl_set_device(device_id), afcl_set_queue(queue)

9prady9:
[skip ci]

Fixes #1002 
Effects #1029


include/af/opencl.h
```diff
@@ -147,6 +194,61 @@ namespace afcl
...
+    if (err!=AF_SUCCESS) throw af::exception("Failed to push user provided device/context to ArrayFire pool");
+}
+
+/**
+   Set active device using cl_context and cl_device_id
+
+   \param[in] dev is the OpenCL device id that is to be set as Active device inside ArrayFire
+   \param[in] ctx is the OpenCL cl_context being used by ArrayFire
+*/
+static inline void setDevice(cl_device_id dev, cl_context ctx)
```
shehzan10: Use #if AF_API_VERSION for each function separately. That's the style we are following.

include/af/opencl.h
```diff
@@ -63,6 +63,53 @@ AFAPI af_err afcl_get_device_id(cl_device_id *id);
...
+
+   \param[in] dev is the OpenCL device id that has to be popped
+   \param[in] ctx is the cl_context object to be removed from ArrayFire pool
+
+   \note Any reference counts incremented for cl_* objects by ArrayFire internally are decremented
+   by this func call and you won't be able to call `afcl_set_device_context` on these objects after
+   this function has been called.
+*/
+AFAPI af_err afcl_pop_device_context(cl_device_id dev, cl_context ctx);
+#endif
```
pavanky: Is insert / remove better than push / pop ?

9prady9: push/pop !? i guess :) I like the way they sound in comparison to insert/delete.

shehzan10: PUSH/POP should be using only for FILO (stack-type, or by extension queue-like). If its like a vector, then use insert remove.

9prady9: If we stick to by-book definitions, then `push` is correct for adding the device, and `remove` is correct deleting the device. because adding is doing push_back. remove is index based delete.

pavanky: Just do add and delete then. No need for semantics.

9prady9:
@pavanky Ready for merge. OSX failed while running tests with some java exception, but i think it is a not related to this.

Issue #1144, opened 2015-12-07
Create gravity simulation example in ArrayFire


pavanky:
Similar to what is happening here: http://jsfiddle.net/unkking88/30c69sa9/

PR #1152, opened 2015-12-09
initial gravity example

syurkevi:
Addresses issue #1144
Needs fixes to collision code. Problems with indexing...
Should simulation speed be based on timestep instead of framerate?

### DIFF
--- a/examples/graphics/gravity_sim.cpp
+++ b/examples/graphics/gravity_sim.cpp
@@ -17,43 +17,36 @@ using namespace std;
 static const int width = 512, height = 512;
 static const int pixels_per_unit = 20;
 
-af::array p_x;
-af::array p_y;
-af::array vels_x;
-af::array vels_y;
-af::array forces_x;
-af::array forces_y;
-
-void simulate(float dt){
-    p_x += vels_x * pixels_per_unit * dt;
-    p_y += vels_y * pixels_per_unit * dt;
+void simulate(af::array *pos, af::array *vels, af::array *forces, float dt){
+    pos[0] += vels[0] * pixels_per_unit * dt;
+    pos[1] += vels[1] * pixels_per_unit * dt;
 
     //calculate distance to center
-    af::array diff_x = p_x - width/2;
-    af::array diff_y = p_y - height/2;
+    af::array diff_x = pos[0] - width/2;
+    af::array diff_y = pos[1] - height/2;
     af::array dist = sqrt( diff_x*diff_x + diff_y*diff_y );
 
     //calculate normalised force vectors
-    forces_x = -1 * diff_x / dist;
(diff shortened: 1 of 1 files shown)


9prady9:
@umar456 Yes, convolution with Gaussian is separable, we just need to split the 2d kernel into two separate 1d kernels.

pavanky:
@syurkevi Can you also remove the `#ifdef` for the windows at the end? We realized that we no longer need it and fixed in the hotfixes branch. We fixed for the remaining examples and since this is a new one it is best not to include it at all.

unknowazrael:
@pavanky Can I ask you some other questions? I want to replace the graphics into openGL. But I meet problems. 
1.How can I get the elements from the array? I currently use like this 
image
but I could not free memory (if I add "delete[] host_a", then) 
image
2.What is the right way to do arrayfire+openGL? If I do this every frame  
image
will it spend much time to copy an array on the device to the host?

9prady9:
@unknowazrael What do you mean by arrayfire+opengl ?

They are following possible ways to do opengl graphics with arrayfire:

- Use arrayfire graphics functionality, which is not currently extensive but good for simple debugging.
- Use arrayfire with your own opengl 
    - via CPU - you don't need anything special but it would be slower than interop option.
    - using opengl interop
        - cpu/cuda backend, you can use current release 3.2.1
        - opencl backend, you would need current devel(beta) build of arrayfire

unknowazrael:
@9prady9 Thanks!  I download v3.2.1 and test success.

unknowazrael:
@9prady9  
Use arrayfire with your own opengl 
  using opengl interop
      cpu/cuda backend, you can use current release 3.2.1

Now I use 3.2.1. May I ask how to using opengl interop? Is there an example?

9prady9:
@unknowazrael 
- We have [CUDA-Interop example](https://github.com/arrayfire/arrayfire_opengl_interop) code in our github, but i believe this has been written when ArrayFire was v2.*, which is not API compatible with current stable release, v3.*.  However, the set of operations you would need to do for OpenGL interoperation with CUDA backend can be understood from this code.
- As far as CPU backend goes, you just need to use the functions you pointed out earlier: `host<T>` and then copy it into OpenGL buffer for rendering.

pavanky:
@unknowazrael @9prady9 This is getting off topic, can you either create a new issue or discuss this on our gitter chat room?

unknowazrael:
@pavanky Sorry, I have a problem.

How to design collision between particle and particle like this
http://stackoverflow.com/questions/2481479/2d-colliding-n-body-simulation-fast-collision-detection-for-large-number-of-bal

pavanky:
@unknowazrael We can't help you write the code. The example should provide you as a starting point to do what you are looking for.

PR #1156, opened 2015-12-11
Visual Studio Multiprocess Compilation options

shehzan10:
[skip ci]

On CI, we run multi process builds using msbuild.exe options.
This PR is to test whether adding the MP option makes the builds any faster.

### DIFF
--- a/src/backend/cuda/kernel/identity.hpp
+++ b/src/backend/cuda/kernel/identity.hpp
@@ -23,14 +23,14 @@ namespace kernel
     __global__
     static void identity_kernel(Param<T> out, int blocks_x, int blocks_y)
     {
-        unsigned idz = blockIdx.x / blocks_x;
-        unsigned idw = blockIdx.y / blocks_y;
+        const dim_t idz = blockIdx.x / blocks_x;
+        const dim_t idw = blockIdx.y / blocks_y;
 
-        unsigned blockIdx_x = blockIdx.x - idz * blocks_x;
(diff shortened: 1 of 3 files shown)


shehzan10:
build arrayfire windows ci

src/backend/cuda/kernel/identity.hpp
```diff
@@ -23,14 +23,14 @@ namespace kernel
...
 
-        unsigned blockIdx_x = blockIdx.x - idz * blocks_x;
-        unsigned blockIdx_y = blockIdx.y - idw * blocks_y;
+        const dim_t blockIdx_x = blockIdx.x - idz * blocks_x;
+        const dim_t blockIdx_y = blockIdx.y - idw * blocks_y;
 
-        unsigned idx = threadIdx.x + blockIdx_x * blockDim.x;
-        unsigned idy = threadIdx.y + blockIdx_y * blockDim.y;
+        const dim_t idx = threadIdx.x + blockIdx_x * blockDim.x;
+        const dim_t idy = threadIdx.y + blockIdx_y * blockDim.y;
```
9prady9: Are these creating issues on any other CUDA kernels ? using `long long` (i know it is not long long always, but for 64 bit systems it is) had caused perf issues (dim_t made it slow) in convolve kernel.

shehzan10: Yeah you are right. My bad. I'll fix this.

PR #1174, opened 2015-12-17
Changes to make CPU backend functions asynchronous calls

### LINKED ISSUE #832: Make CPU backend Async
The OpenCL and CUDA backend functions in ArrayFire are asynchronous whereas the CPU backend is not. We need to unify the behavior of all of the backends so that they are consistent.

Function coverage w.r.t each file in `<REPOSITORY_ROOT>/src/backend/cpu/*.cpp`:
- [x] approx.cpp
- [x] Array.cpp
- [x] assign.cpp
- [x] bilateral.cpp
- [x] blas.cpp
- [x] cholesky.cpp
- [x] convolve.cpp
- [x] copy.cpp
- [x] diagonal.cpp
- [x] diff.cpp
- [x] fast.cpp
- [x] fftconvolve.cpp
- [x] fft.cpp
- [x] gradient.cpp
- [x] harris.cpp
- [x] hist_graphics.cpp
- [x] histogram.cpp
- [x] homography.cpp
- [x] hsv_rgb.cpp
- [x] identity.cpp
- [x] iir.cpp
- [x] image.cpp
- [x] index.cpp
- [x] inverse.cpp
- [x] iota.cpp
- [x] ireduce.cpp
- [x] join.cpp
- [x] lookup.cpp
- [x] lu.cpp
- [x] match_template.cpp
- [x] math.cpp
- [x] meanshift.cpp
- [x] medfilt.cpp
- [x] morph.cpp
- [x] nearest_neighbour.cpp
- [ ] orb.cpp
- [x] plot3.cpp
- [x] plot.cpp
- [x] qr.cpp
- [x] random.cpp
- [x] range.cpp
- [x] reduce.cpp
- [x] regions.cpp
- [x] reorder.cpp
- [x] resize.cpp
- [x] rotate.cpp
- [x] scan.cpp
- [x] select.cpp
- [x] set.cpp
- [x] shift.cpp
- [x] sift.cpp
- [x] sobel.cpp
- [x] solve.cpp
- [x] sort_by_key.cpp
- [x] sort.cpp
- [x] sort_index.cpp
- [x] surface.cpp
- [x] susan.cpp
- [x] svd.cpp
- [x] tile.cpp
- [x] transform.cpp
- [x] transpose.cpp
- [x] triangle.cpp
- [x] unwrap.cpp
- [x] where.cpp
- [x] wrap.cpp

9prady9:
Fixes #832


src/backend/cpu/CMakeLists.txt
```diff
@@ -47,11 +47,24 @@ IF(NOT UNIX)
     ADD_DEFINITIONS(-DAFDLL)
 ENDIF()
 
+INCLUDE(ExternalProject)
```
shehzan10: Create a build_threads.cmake file in CMakeModules for this external project.
Follow clBLAS.

src/backend/cpu/CMakeLists.txt
```diff
@@ -47,11 +47,24 @@ IF(NOT UNIX)
...
+ExternalProject_Add(
+    threads
+    PREFIX ${CMAKE_BINARY_DIR}/third_party/threads
+    GIT_REPOSITORY https://github.com/alltheflops/threads.git
```
shehzan10: Add a GIT_TAG. It can be a branch for the review process, but needs to be added.

src/backend/cpu/CMakeLists.txt
```diff
@@ -151,6 +164,7 @@ TARGET_LINK_LIBRARIES(afcpu
                             PRIVATE ${CBLAS_LIBRARIES}
                             PRIVATE ${FFTW_LIBRARIES})
 
+ADD_DEPENDENCIES(afcpu threads)
```
shehzan10: Is this going to be static or dynamic?

src/backend/cpu/CMakeLists.txt
9prady9: That is header only library, do we need it at the root level ?

src/backend/cpu/CMakeLists.txt
9prady9: It is a header only stuff.

src/backend/cpu/CMakeLists.txt
shehzan10: The external project build commands have their own files which are in CMakeModules. Although it will be called only from the CPU backend, the cmake file should be in CMakeModules. That's how we are going about with clBLAS/clFFT as well.

9prady9:
build arrayfire ci

shehzan10:
build arrayfire windows ci

pavanky:
@9prady9 Can you make sure `cmake` detects if the threads submodule is initialized or not (similar to what we are doing with assets).

Right now it fails only when trying to compile

pavanky:
Code looks good. Things to change require the following:

- [ ] CMake checks if `threads` submodule has been initialized
- [ ] Make sure `memAlloc` and `memFree` are not being called from the queued thread
- [ ] Add `getQueue().sync()` after each `getQueue.enqueue()` in debug mode.
- [ ] Add the ability to make the algorithm synchronous at runtime based on environment variable.
- [ ] Perhaps consider making the previous one a compile time option as well

pavanky:
@umar456 What do you think about the last point ?

shehzan10:
It would be nice to have a compile time option

9prady9:
build arrayfire ci

umar456:
What would be the benefit of having both? Couldn't you just export the variable and make it persistent?

pavanky:
@umar456 try setting and unsetting the variable on windows. It is not pretty.

pavanky:
@9prady9 Can you pull from devel and resolve merge conflicts? You will need to fix `transform.cpp` in CPU backend.

9prady9:
@pavanky I will start on this after 3.2.2 -> devel PR. I believe that is the last hot fixes before v3.3.0 release, is that right ?

pavanky:
@9prady9 Sure.

shehzan10:
build arrayfire ci

PR #1180, opened 2015-12-21
(OpenCL) Change path of clBLAS/FFT builds, (CUDA) Create stream when device is active

shehzan10:
* Shorten clBLAS/FFT-external to clBLAS/FFT-ext. This is because windows has a 246/260 characted limit on the paths it can build. This is causing issues.

* Creating streams for all devices at the start is not useful and increased the init time (~800ms/device). So create the stream only when the device is set to active the first time.

### DIFF
--- a/src/backend/cuda/platform.cpp
+++ b/src/backend/cuda/platform.cpp
@@ -332,10 +332,10 @@ DeviceManager::DeviceManager()
 
     sortDevices();
 
-    for(int i = 0; i < nDevices; i++) {
-        setActiveDevice(i, cuDevices[i].nativeId);
-        CUDA_CHECK(cudaStreamCreate(&streams[i]));
-    }
+    // Initialize all streams to 0.
+    // Streams will be created in setActiveDevice()
+    for(int i = 0; i < (int)MAX_DEVICES; i++)
+        streams[i] = (cudaStream_t)0;
 
     const char* deviceENV = getenv("AF_CUDA_DEFAULT_DEVICE");
     if(!deviceENV) {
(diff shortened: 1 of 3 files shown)


src/backend/cuda/platform.cpp
```diff
@@ -332,10 +342,10 @@ DeviceManager::DeviceManager()
...
     sortDevices();
 
-    for(int i = 0; i < nDevices; i++) {
-        setActiveDevice(i, cuDevices[i].nativeId);
-        CUDA_CHECK(cudaStreamCreate(&streams[i]));
-    }
+    // Initialize all streams to 0.
+    // Streams will be created in setActiveDevice()
+    for(int i = 0; i < (int)MAX_DEVICES; i++)
+        streams[i] = (cudaStream_t)0;
```
9prady9: Good catch, but can you change this to the following:
```
for(int i = 0; i < nDevices; i++) {
    // first call to the following function will automatically
    // create a stream for that particular device
    setActiveDevice(i, cuDevices[i].nativeId);
}
```
so that `cudaStreamCreate` is not called upon during actual code execution.

shehzan10: Creating cuda streams is taking about 800ms/device at initialization. So doing it for all available devices is making it slow on systems with multiple GPUs. Doing what you suggested does that same thing as before, simply in a different manner.

This change creates the stream when the device is switched, ie. set device and not when the code is executed.

9prady9: I understand what it is doing, which is exactly what i avoided it in first place, but it had a bug i didn't notice that got fixed now.  

Init time is not a concern in many cases, since it is going to be a little higher whether this particular call is there in the initialization or not. So, doing it in the beginning is better than doing later which makes `setDevice` a very quick switch.

pavanky: It is taking 500ms for each device. I don't think we should initialize it if it is not necessary.

PR #1194, opened 2015-12-30
Fast - Build Type with O0 compilation for speed up

shehzan10:
[skip ci]

I'll test this with PR-Build using Fast config overnight.


shehzan10:
Closing this as it was opened with devel as target

PR #1211, opened 2016-01-03
Build fix for CUDA backend when using boost 1.60

### LINKED ISSUE #1210: CUDA backend does not build with Boost 1.60
Apparently something in this combination requires setting `-std=c++11`. This needs fixing.

https://svn.boost.org/trac/boost/ticket/11852

pavanky:
- Fixes #1210

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -57,6 +57,8 @@ FOREACH(VER 20 30 32 35 50 52 53)
 ENDFOREACH()
 
 IF(UNIX)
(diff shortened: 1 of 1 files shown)


pavanky:
@arrayfire/core-devel I am force adding `__STIRCT_ANSI__` for CUDA backend. I grepped through CUDA and gcc header files. The only places it seems to affect is that it removes custom data types. 

I did this because boost 1.60 added support for `__float128` data type from gcc. This is not defined when nvcc is being used and causes problems. Look at #1120 for more information.

If anyone feels like this is a hack instead of a proper fix let me know how else to resolve it :-/

pavanky:
build arrayfire ci

PR #1221, opened 2016-01-09
CPU Offload for OpenCL

### LINKED ISSUE #1196: Use CPU libraries when possible for CPU OpenCL devices
For example it might be faster to use MKL / OpenBLAS when performing linear algebra functions or FFT operations for CPU devices.

There are two ways it can be achieved.

1) Copy data to "host" memory, do the operation, copy back. This might be useful only after a particular size.

2) Consider using clEnqueueMapBuffer. Check if the performance of this function is faster than dual copies mentioned in (1).

### LINKED ISSUE #1219: expose getActiveBackend()
It would be useful to be able to query which actual backend is currently being used when arrayfire is in the unified backend mode. As far as I can see this information is available internally via the AFSymbolManager class and its getActiveBackend() method, but is not exposed as an API call. Does anything speak against making this call available in the next release?

shehzan10:
* CPU Offload for OpenCL function when device has unified memory (CPU/ Integrated graphics)
  * Use the AF_OPENCL_CPU_OFFLOAD=1 environment variable
  * Add for matmul and all lapack functions
* Fix NONFREE CMake Options
* Better handling of CUDA Lapack on Tegra X1
* Add getActiveBackend function

Fixes #1219 
Fixes #1196


umar456:
I don't like how we have so much copied code in each of the backends. We need to figure out an alternate way to do this.

umar456:
Squash

@shehzan10	Add OpenCL-CPU fallback for LU	45abbc3
 @shehzan10	Add OpenCL-CPU fallback for Cholesky	88e910d
 @shehzan10	Add OpenCL-CPU fallback for QR	872acfb
 @shehzan10	Add OpenCL-CPU fallback for SVD	59a9df0
 @shehzan10	Add OpenCL-CPU fallback for solve	ffb191c
 @shehzan10	Add OpenCL-CPU fallback for inverse	4e2d46c

And

@shehzan10	Fix bug in OpenCL JIT when calling functions that return same value …	d5077ec
 @shehzan10	Fix bug in CUDA JIT when calling functions that return same value …	ac25f5b

pavanky:
I'd rather have CPU offload for different functions to be separate. Easier to debug / revert if something borks. Don't make a huge commit out of those.

src/backend/opencl/blas.cpp
```diff
@@ -113,6 +118,10 @@ template<typename T>
 Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs,
                 af_mat_prop optLhs, af_mat_prop optRhs)
 {
+    if(OpenCLCPUOffload()) {
```
umar456: Wouldn't this change the behavior of the function. The CPU version does not seem to be asynchronous.

pavanky: The behavior is also off by default. Users will enable it using a flag. We can document the flag saying that it can cause synchronization. The "mapping" operation itself is synchronous by the way. It would require too much overhead with very little payback to make this operation asynchronous.

shehzan10:
@pavanky can confirm this, but I think most OpenCL magma based lapack functions are synchronous anyway as they do part of the work on host. So matmul would probably be the only one with the change in behavior per the users choice.

pavanky:
@shehzan10 Well sort of. Magma uses events and queues to overlap CPU and GPU computation, but it does have synchronization points. I agree that adding matmul to the list (for a non default option) is not a terrible thing.

src/backend/cuda/blas.cpp
```diff
@@ -197,40 +200,37 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs,
...
+//                 N,
+//                 lhs.get(), lhs.strides()[0],
+//                 rhs.get(), rhs.strides()[0],
+//                 &out)));
+//
+//    if(both_conjugate)
+//        return createValueArray(af::dim4(1), conj(out));
+//    else
+//        return createValueArray(af::dim4(1), out);
+//}
```
pavanky: Remove this.

src/backend/opencl/Array.hpp
```diff
@@ -209,6 +210,48 @@ namespace opencl
...
+
+            T *ptr = nullptr;
+            try {
+                if(ptr == nullptr) {
+                    if(is_const()) {
+                        ptr = (T*)getQueue().enqueueMapBuffer(*const_cast<cl::Buffer*>(get()), true, CL_MAP_READ,
+                                getOffset(), getDataDims().elements() * sizeof(T));
+                    } else {
+                        ptr = (T*)getQueue().enqueueMapBuffer(*(get()), true, CL_MAP_READ|CL_MAP_WRITE,
+                                getOffset(), getDataDims().elements() * sizeof(T));
```
pavanky: Remove is_const and make this a function instead.

src/backend/opencl/blas.cpp
```diff
@@ -168,45 +177,42 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs,
...
+//            (*rhs.get())(),  rhs.getOffset(), rhs.strides()[0],
+//            scratch(),
+//            1, &getQueue()(), 0, nullptr, &event())
+//        );
+//
+//    if(both_conjugate)
+//        transpose_inplace<T>(out, true);
+//
+//    return out;
+//}
```
pavanky: Remove this as well

src/backend/opencl/inverse.cpp
```diff
@@ -19,6 +21,10 @@ namespace opencl
 template<typename T>
 Array<T> inverse(const Array<T> &in)
 {
+    if(OpenCLCPUOffload()) {
+        if (in.dims()[0] == in.dims()[1])
+            return cpu::inverse(in);
```
pavanky: There is no else part here. Does it not do CPU offload if it is not square ?

shehzan10: When square, the same code as below will need to run on the CPU side. Given CPU is enabled, solve will also run on CPU. So the effect is the same.

test/backend.cpp
```diff
@@ -21,11 +21,26 @@
...
+    }
+}
+
 template<typename T>
 void testFunction()
 {
     af_info();
 
+    printf("Active Backend Enum = %s\n", getActiveBackendString());
+
```
pavanky: How is this a test?

test/blas.cpp
```diff
@@ -36,6 +36,7 @@ template<typename T, bool isBVector>
 void MatMulCheck(string TestFile)
 {
     if (noDoubleTests<T>()) return;
+    af::info();
```
pavanky: I am assuming this got added accidentally ?

shehzan10:
build arrayfire tegrak1 ci
build arrayfire tegrax1 ci

shehzan10:
Tested OpenCL-CPU Offload on the PR Builds. This PR is ready to merge if no further changes are required.

PR #1222, opened 2016-01-09
Improvements to OpenCL DeviceManager 

pavanky:
Sorting for OpenCL devices:
 
1. GPUs > Accelerators > CPUs.

2. IN GPUs:
   a. Discreet preferred to integrated
   b. AMD > NVIDIA > APPLE > Intel / BEIGNET

3. IN CPUs Intel > AMD > POCL > other

4. While everything above is the same:
   a. Higher OpenCL compute version preferred
   b. Higher amount of memory preferred

-------------------------------

Environment flags for choosing the proper device type.

- `AF_OPENCL_DEFAULT_DEVICE_TYPE`: - Setting default device to the first instance of a particular type.

- `AF_OPENCL_DEVICE_TYPE`:  Restrict OpenCL to a particular device type.

[skip ci]

### DIFF
--- a/src/api/c/imageio2.cpp
+++ b/src/api/c/imageio2.cpp
@@ -58,14 +58,15 @@ static af_err readImage_t(af_array *rImage, const uchar* pSrcLine, const int nSr
                     pDst0[indx] = (T) *(src + (x * step + FI_RGBA_RED));
                     pDst1[indx] = (T) *(src + (x * step + FI_RGBA_GREEN));
                     pDst2[indx] = (T) *(src + (x * step + FI_RGBA_BLUE));
+                    if (fi_color == 4) pDst3[indx] = (T) *(src + (x * step + FI_RGBA_ALPHA));
                 } else {
                     // Non 8-bit types do not use ordering
                     // See Pixel Access Functions Chapter in FreeImage Doc
-                    pDst0[indx] = (T) *(src + (x * step + FI_RGBA_RED));
-                    pDst1[indx] = (T) *(src + (x * step + FI_RGBA_GREEN));
-                    pDst2[indx] = (T) *(src + (x * step + FI_RGBA_BLUE));
+                    pDst0[indx] = (T) *(src + (x * step + 0));
+                    pDst1[indx] = (T) *(src + (x * step + 1));
+                    pDst2[indx] = (T) *(src + (x * step + 2));
+                    if (fi_color == 4) pDst3[indx] = (T) *(src + (x * step + 3));
                 }
-                if (fi_color == 4) pDst3[indx] = (T) *(src + (x * step + FI_RGBA_ALPHA));
             }
             indx++;
         }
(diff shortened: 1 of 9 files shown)


pavanky:
@9prady9 Any problems you foresee with this ?

@shehzan10 Will this affect jenkins in anyway ?

pavanky:
Don't merge yet, testing the order on various machines. Right now this is putting a Quadro 600 ahead of K20c because of memory. We should perhaps check for OpenCL version.

shehzan10:
WRT ci, i might have to change a thing here or there but nothing significant.

As you said, device sorting for GPUs should first be version and then memory.

pavanky:
Alright fixed. Review please.

shehzan10:
Here is the output from Windows-01 on CI:
```
[0] NVIDIA  : Quadro K5000
-1- INTEL   : Intel(R) HD Graphics 4600
-2- INTEL   : Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
-3- Experimental OpenCL 2.0 CPU Only Platform: Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
```

In an ideal case, the `Experimental OpenCL 2.0 CPU` should be device 2 and the regular CPU should be device 3.

Is this because "Experimental OpenCL..." is not one of our "recognized" platforms, and that is why it shows the full device name and platform?

9prady9:
@pavanky Using version->memory seems like a reasonable sort order criteria, but I think it won't be close to ideal unless we sort them based on flops - don't we do this in CUDA ? - i think so, may be wrong.

`Experimental OpenCL 2.0` is experimental as it says, therefore we should keep it as low priority. So, i think what we have now is fine.

pavanky:
@shehzan10 That is correct. I don't think we should prioritize experimental platforms anyway. 

@9prady9 NVIDIA GPUs have more or less the same architecture so it is easy to calculate flops. This is not possible with OpenCL devices so we have to use the next best thing.

shehzan10:
Thoughts on adding a `AF_OPENCL_DEFAULT_DEVICE_TYPE` environment variable? This can take `GPU`, `CPU`, `ACC`. It would set the default device to the first device of the type?

There won't really be corresponding CUDA Or CPU variable for this as they are only of a single type.

pavanky:
@shehzan10 Any reason why we should have "default_device_type"  instead of having "AF_OPENCL_DEVICE_TYPE=CPU". or something such.

This would use CL_DEVICE_TYPE_CPU instead of CL_DEVICE_TYPE_ALL when querying for devices and not query for any GPUs whatsoever. 

This would be easier to implement than "default type" which would require querying all devices and then sort them in a different manner using different rules. I feel like this would bloat up the code.

shehzan10:
I'm not talking about changing the device sorting order. I'm talking about just setting the active device using an environment variable.

For example (from the above list), if I wanted to run code on CPU, I would have to run af::info, get the device ID and then set AF_OPENCL_DEFAULT_DEVICE to that. With the DEFAULT_DEVICE_TYPE (or whatever we decide to call it), we can just use DEFAULT_DEVICE_TYPE=CPU and it would set the initial default device to 2 (or 0 for GPU, 1 if we have a IGPU value for integrated GPU).

pavanky:
Well in that case, it is fairly easy to implement. But what use case does it cover that the other option does not ? (i.e. the ability to use the CPU device by default by disabling other devices).

shehzan10:
I think the advantage of this way is that we are not "disabling" other device types? I'm fine with either way though. Just something to pick the type would be nice in my opinion.

pavanky:
@shehzan10 implemented both versions.

src/api/c/imageio2.cpp
```diff
@@ -61,9 +61,9 @@ static af_err readImage_t(af_array *rImage, const uchar* pSrcLine, const int nSr
                 } else {
                     // Non 8-bit types do not use ordering
                     // See Pixel Access Functions Chapter in FreeImage Doc
-                    pDst0[indx] = (T) *(src + (x * step + 0));
-                    pDst1[indx] = (T) *(src + (x * step + 1));
-                    pDst2[indx] = (T) *(src + (x * step + 2));
+                    pDst0[indx] = (T) *(src + (x * step + FI_RGBA_RED));
+                    pDst1[indx] = (T) *(src + (x * step + FI_RGBA_GREEN));
+                    pDst2[indx] = (T) *(src + (x * step + FI_RGBA_BLUE));
```
shehzan10: These enums should not be used here. I put the comment in there for that specific purpose.
Adding tests for the native functions is on my todo list.

src/api/c/imageio2.cpp
```diff
@@ -239,15 +239,15 @@ static void save_t(T* pDstLine, const af_array in, const dim4 dims, uint nDstPit
...
+                    *(pDstLine + x * step + FI_RGBA_BLUE ) = (T) pSrc2[indx]; // b -> 2
                 } else {
                     // Non 8-bit types do not use ordering
                     // See Pixel Access Functions Chapter in FreeImage Doc
-                    *(pDstLine + x * step + 0) = (T) pSrc0[indx]; // r -> 0
-                    *(pDstLine + x * step + 1) = (T) pSrc1[indx]; // g -> 1
-                    *(pDstLine + x * step + 2) = (T) pSrc2[indx]; // b -> 2
+                    *(pDstLine + x * step + FI_RGBA_RED  ) = (T) pSrc0[indx]; // r -> 0
+                    *(pDstLine + x * step + FI_RGBA_GREEN) = (T) pSrc1[indx]; // g -> 1
+                    *(pDstLine + x * step + FI_RGBA_BLUE ) = (T) pSrc2[indx]; // b -> 2
```
shehzan10: Same as above comment.

src/api/c/imageio2.cpp
pavanky: In that case I need to change this imageio as well.

shehzan10: No imageio.cpp is fine. I checked your PR when you did it. I check this file as well. Somehow the u8 code skipped my eye.
This only affects 16-bit and 32-bit images. That's why u8 is using the enums in the if(u8) case above. That's why imageio.cpp should be fine.

Essentially the 4xu8 are 32-bits, so they keep that as one unit. But for 16 and 32-bit it doesn't matter, so they don't use enums.

pavanky: Yes, but I changed everything to use macros in imageio. I am removing the macros for non 8 bit types.

shehzan10: Ah ok. I'll add tests for this asap so that we know everything is working fine.

pavanky:
Alright somebody merge this thing.

Issue #1239, opened 2016-01-19
Improve heuristics for Memory Management


pavanky:
Right now the heuristics are fairly simplistic. 

Possible improvements:

- Check for maximum available memory
- Check for alternatives to `std::map` to improve garbage collection performance
- Increase the size of max_buffers
- Allow users to tweak the max_buffers, threshold of max memory

mchandra:
I tried out this merge on an Intel CPU and a NVIDIA GPU. On both platforms, `memsize` returns 0 and so the GC starts after 1 GB. Our application uses much more memory than that. I tweaked the threshold to 4 GB and got much better performance. Would be good if there's an option to set the `max_bytes` limit by hand.

PR #1241, opened 2016-01-21
Updates to Memory Manager and Garbage Collection Heuristics

### LINKED ISSUE #1243: Tests failing in Debug Mode on Windows
* Meanshift
    * Test copied F32 array into double-pointer array
* ORB
    * In fast_pyramid, vector indices are access after reserve rather than resize
* OCL_EXT
    * New devices are not being pushed to mPlatforms in OpenCL Device Manager

### LINKED ISSUE #1245: Add proper documentation for deviceInfo
af::deviceInfo(char *d_name, char *d_platform, char *d_toolkit, char *d_compute);

This is a dangerous function as we do not know the size of the different char* needed. I check (see below) the devprop implementation and it is :  64, 10, 64,10.

It may differ for a different platform ! ...

It should be a define or it should be indicated in the documentation / header of the function

void devprop(char* d_name, char* d_platform, char *d_toolkit, char* d_compute)
{
    static CPUInfo cinfo;
    snprintf(d_name, 64, "%s", cinfo.vendor().c_str());
    snprintf(d_platform, 10, "CPU");
    // report the compiler for toolkit
    snprintf(d_toolkit, 64, "%s", AF_COMPILER_STR);
    snprintf(d_compute, 10, "%s", "0.0");
}

shehzan10:
Ref #1239 

* Add getHostMemorySize and getDeviceMemorySize functions
  * Print memory size in CPU and OpenCL info
  * Change opencl::getDevice to accept id
  * Fix multi-line error strings
* Updates to Memory Manager and Garbage Collection Heuristics
  * Add getMaxMemorySize virtual function to MemoryManager
    * Used to compute the max_bytes for each memory manager instance
  * max_bytes
    * This is now a part of memory_info
    * Gets it's size from the device memory rather than a generic size
    * For CPU and CUDA Pinned Memory, use getHostDeviceMemroy
  * MAX_BUFFERS is now 1000
* Fix debug builds from source on Windows
* Fix failing tests on debug builds on Windows (Fixes #1243)

Fixes #1245


src/backend/host_memory.cpp
```diff
@@ -0,0 +1,113 @@
...
+#endif
+    unsigned int size = 0;      /* 32-bit */
+    size_t len = sizeof( size );
+    if ( sysctl( mib, 2, &size, &len, NULL, 0 ) == 0 )
+        return (size_t)size;
+    return 0L;          /* Failed? */
+#endif /* sysctl and sysconf variants */
+
+#else
+    return 0L;          /* Unknown OS. */
```
pavanky: We should probably change this to a non zero value because we don't want to fail on platforms we haven't considered.

shehzan10:
build arrayfire osx ci

shehzan10:
build arrayfire windows ci

The cfd60f1 commit only affects Windows debug builds. So running only windows.

shehzan10:
build arrayfire windows ci

PR #1263, opened 2016-02-04
Graphics fixes

### LINKED ISSUE #1244: Correct af::Window default constructors behaviour
Currently, the following code stub, doesn't do anything which needs to resolved appropriately.

```c++
af::Window foo;
af::Window bar(foo);
```

9prady9:
- Fixes #1244

### DIFF
--- a/include/af/graphics.h
+++ b/include/af/graphics.h
@@ -30,6 +30,8 @@ namespace af
 
    \brief Window object to render af::arrays
 
+   Windows are not CopyConstructible or CopyAssignable.
+
    \ingroup graphics_func
  */
 class AFAPI Window {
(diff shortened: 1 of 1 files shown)


include/af/graphics.h
```diff
@@ -39,10 +64,15 @@ class AFAPI Window {
          * cell in the grid is being rendered currently */
         int _r;
         int _c;
-        ColorMap _cmap;
+        ColorMap        _cmap;
+        af_plot_limits  _range;
+        af_color        _color;
```
pavanky: Now this will definitely break API.

include/af/graphics.h
```diff
@@ -43,6 +45,9 @@ class AFAPI Window {
 
         void initWindow(const int width, const int height, const char* const title);
 
+        Window(const Window&);                 // Prevent copy-construction
```
pavanky: Pradeep this may cause missing symbol errors I think. Don't we need to have " {} "  as well?

@umar456 I am not entirely sure about this can you take a look ?

9prady9: Worked on my Arch but not sure if it will on Windows.

9prady9:
build arrayfire ci

9prady9:
@pavanky All jenkins builds passed without any issues.

PR #1278, opened 2016-02-12
BUGFIX: Fixed bug in CUDA and OpenCL when re-using same JIT nodes.

### LINKED ISSUE #1279: BUG in CUDA and OpenCL JIT
`f` should be equal to `-g` should they come out to be identical.
```cpp
#include <arrayfire.h>
#include <cstdio>
#include <cstdlib>

using namespace af;

const int NDIM = 4;

{
    try {

        af::array a = af::constant(2, 5,5);
        af::array b = af::constant(1, 5,5);
        af::array c = a + b;
        af::array d = a - b;
        af::array e = c * d;
        e.eval();   
        af_print(e);
        af::array f = c - d;
        f.eval();   
        af_print(f);
        af::array g = d - c;
        g.eval();   
        af_print(g);

    } catch (af::exception& e) {

    }

    return 0;
}
```

pavanky:
- Fixes #1279

### DIFF
--- a/src/backend/cuda/JIT/BinaryNode.hpp
+++ b/src/backend/cuda/JIT/BinaryNode.hpp
@@ -126,11 +126,7 @@ namespace JIT
 
         void resetFlags()
         {
-            m_set_id = false;
-            m_gen_func = false;
-            m_gen_param = false;
-            m_gen_offset = false;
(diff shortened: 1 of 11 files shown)


pavanky:
build arrayfire osx ci

PR #1287, opened 2016-02-16
Added loadImage support for 32-bit integer images

yshady:
Reading INT32 Tiff files using af::loadImage() resulted in empty (zeros) f32 arrays. I tracked down the reason to the following line in src/api/c/imageio.cpp:97
```
const T *src = (T*)(pSrcLine - y * nSrcPitch);
```
If the read pixel data are actually in int/uint format, dereferencing this src pointer results in incorrect values. I added support just for 32-bit case, but the same can be done for other pixel depths.

### DIFF
--- a/src/api/c/imageio.cpp
+++ b/src/api/c/imageio.cpp
@@ -186,7 +186,10 @@ af_err af_load_image(af_array *out, const char* filename, const bool isColor)
         if(fi_bpc != 8 && fi_bpc != 16 && fi_bpc != 32) {
             AF_ERROR("FreeImage Error: Bits per channel not supported", AF_ERR_NOT_SUPPORTED);
         }
-
+        
+        // data type
+        FREE_IMAGE_TYPE image_type = FreeImage_GetImageType(pBitmap);
+        
         // sizes
         uint fi_w = FreeImage_GetWidth(pBitmap);
         uint fi_h = FreeImage_GetHeight(pBitmap);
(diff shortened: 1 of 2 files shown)


pavanky:
@yshady can you make sure your commit has your proper email id ?

pavanky:
Would it be possible to add support for 32 bit int images in saveImage / loadImageNative / saveImageNative ?

yshady:
Sure, I'll do this right away.

yshady:
I added the support in af_load_image_native. saveImage functions always output a float image. The returned af::array from loadImage functions is always f32, so I think it is consistent.

pavanky:
@yshady Can you merge these two commits into a single commit ? You can use `git rebase` and force push to your branch.

Also the commit message is a bit long. What you can do is have the main message be "Added support for loading 32 bit integer images" and provide any details in the next line.

pavanky:
Thanks for the contribution BTW!

shehzan10:
Please make the following style changes:
* Block tabs -> 4 spaces
* `switch(cond) {` in one line. Fix the alignment of the closing bracket as well.
* DWORD -> uint/unsigned int, WORD -> int

yshady:
@shehzan10 will do for the tabs.
I'm using `DWORD` and `LONG` as suggested by the FreeImage manual (v3.17.0 section 28). `int` and `unsigned int` can have different sizes on different architectures/compilers etc... Just trying to keep it consistent with FreeImage.

shehzan10:
We don't use DWORD-like types anywhere in ArrayFire which means using it here would require definitions. Long is not safe since it is highly OS dependent (thats why ArrayFire uses long long and unsigned long long for 64-bit integer types).

I think it is safe enough to use int and unsigned int.

yshady:
I disagree. DWORD and LONG are type-defined in FreeImage.h and are the same type used for the pointers inside FreeImage structure. They are safer than using int and unsigned int which can be 64-bit.

pavanky:
@shehzan10 I am assuming `LONG` from freeimage is typedef'd to a 32 bit signed integer.

@yshady While what you are doing is consistent with freeimage, the code is not consistent with the rest of arrayfire. We currently do not support (nor are we planning to support) any architecture where `int` and `unsigned int` are not 32 bits.  In the grand scheme of things your code is OK, but it is a bit hard to read for not familiar with free image data types.

yshady:
Done

pavanky:
:+1:

pavanky:
build arrayfire ci

src/api/c/imageio.cpp
```diff
@@ -204,21 +207,36 @@ af_err af_load_image(af_array *out, const char* filename, const bool isColor)
                 else if(fi_bpc == 16)
                     AF_CHECK((readImage<ushort, AFFI_RGBA, AFFI_RGBA>)(&rImage, pSrcLine, nSrcPitch, fi_w, fi_h));
                 else if(fi_bpc == 32)
-                    AF_CHECK((readImage<float,  AFFI_RGBA, AFFI_RGBA>)(&rImage, pSrcLine, nSrcPitch, fi_w, fi_h));
+                    switch(image_type) {
+						case FIT_UINT32: AF_CHECK((readImage<uint,  AFFI_RGBA, AFFI_RGBA>)(&rImage, pSrcLine, nSrcPitch, fi_w, fi_h)); break;
```
shehzan10: This spacing is still tabbed

shehzan10:
build arrayfire linux ci
build arrayfire osx ci

PR #1288, opened 2016-02-16
Fixes for OSX OpenCL LAPACK and CMake Build Options

### LINKED ISSUE #662: Linear algebra functions not working on OSX intel GPU

### LINKED ISSUE #1285: Minimum CMake version
The `CMakeLists.txt` file says `CMAKE_MINIMUM_REQUIRED(VERSION 2.8)` but certain files (e.g. `src/backend/cuda/CMakeLists.txt`) use the `PRIVATE` keyword in `TARGET_LINK_LIBRARIES` call which was only introduced in from CMake 3.0.

This causes errors like:
```
/usr/bin/ld: cannot find -lPRIVATE
```

### LINKED ISSUE #1286: Add BUILD_* Options to Examples and Tests when building standalone
Currently, the backends can only be turned off by moving the CMake config files.
Add BUILD_* Options only when building out of source.

shehzan10:
* Fixes #662 
* FIxes #1285 
* Fixes #1286


shehzan10:
Tested and confirmed that LAPACK OpenCL tests are running correctly with this "forced" fallback.

PR #1293, opened 2016-02-22
Fixes to reduce memory usage for CPU backend in async mode. 

pavanky:


src/backend/MemoryManager.cpp
```diff
@@ -258,7 +257,7 @@ void MemoryManager::printInfo(const char *msg, const int device)
             unit = "MB";
         }
 
-        std::cout << " |  " << std::right << std::setw(14) << kv.first << " "
+        std::cout << "|  " << std::right << std::setw(14) << kv.first << " "
```
umar456: I know this is not part of this PR but you should look into changing this into printfs. `cout` should only be used in templates.

pavanky: I didn't do it!

src/backend/cpu/kernel/sift_nonfree.hpp
```diff
@@ -969,6 +969,7 @@ unsigned sift_impl(Array<float>& x, Array<float>& y, Array<float>& score,
                    const bool compute_GLOH)
 {
     in.eval();
+    getQueue().sync();
```
umar456: Why is this required here?

src/backend/cpu/memory.cpp
```diff
@@ -112,13 +112,30 @@ void printMemInfo(const char *msg, const int device)
 template<typename T>
 T* memAlloc(const size_t &elements)
 {
-    return (T *)getMemoryManager().alloc(elements * sizeof(T), false);
+    T *ptr = nullptr;
+
+    try {
```
umar456: Can't we return a null pointer in case of a failure here?.

src/backend/cpu/queue.hpp
```diff
@@ -48,32 +49,45 @@ typedef async_queue queue_impl;
...
+    void enqueue(const F func, Args... args)
+    {
+        count++;
+        if(sync_calls) { func( args... ); }
+        else           { aQueue.enqueue( func, args... ); }
 #ifndef NDEBUG
-    sync();
+        sync();
+#else
+        if (checkMemoryLimit() || count >= 25) {
```
umar456: we should probably figure out a way to sync to a certain number of elements in the queue. Going all they way down to zero seems sub optimal

src/backend/cpu/scan.cpp
```diff
@@ -27,8 +27,7 @@ template<af_op_t op, typename Ti, typename To>
 Array<To> scan(const Array<Ti>& in, const int dim)
 {
     dim4 dims     = in.dims();
-    Array<To> out = createValueArray<To>(dims, 0);
-    out.eval();
+    Array<To> out = createEmptyArray<To>(dims);
```
umar456: are you sure this does not expect 0 values.

umar456:
couple of questions. Other then that it looks good.

src/backend/cpu/kernel/sift_nonfree.hpp
pavanky: The rest of the algorithm is synchronous. This function is not being called in a queue.

src/backend/cpu/memory.cpp
pavanky: We return nullptr for 0 elements elsewhere. It'll get confusing.

src/backend/cpu/queue.hpp
pavanky: This is a work around for now. We should probably remove it entirely by having a proper solution in the future.

src/backend/cpu/scan.cpp
pavanky: I checked. I also ran the tests.

pavanky:
Don't merge this yet. I am trying to see if I can fix this properly without using the work around.

PR #1298, opened 2016-02-23
Fixes to and reorganization of OSX Installer

### LINKED ISSUE #1223: OSX Installer Fixes
Currently, the CMake config files are part of the "Extra" package in the OSX installer.

When one or more of the libs are not selected to be installed and the extra package is, then the ArrayFireBackend.cmake files for the backend are installed anyway.

This will cause errors when building with cmake, which will complain about missing libs.

My suggestion is as follows:
* The ArrayFireBackend.cmake files should be part of the lib packages.
* The ArrayFireConfig.cmake files (the main one), should be installed if even one of the backends is installed.
* The "Extra" package should be renamed "examples" package
* Docs should be part of the ArrayFire headers package, or a separate package.

Alternatively, installing everything as one package can work too. If the system does not have CUDA/OpenCL SDKs installed, that backend will be turned off anyway.

Any other ways to handle this?

shehzan10:
* ArrayFire<Backend>.config and forge are now installed with the library component
* ArrayFireConfig.cmake files are installed by all repos
* Examples and Documentation and now independent components
* No changes to install paths
* When make osx_installer is called, it creates a new directory called
  osx_install_files which are then used to generate the installers
* osx_installer target depends on make install being called first

Fixes #1223 
[skip ci]

### DIFF
--- a/CMakeModules/osx_install/OSXInstaller.cmake
+++ b/CMakeModules/osx_install/OSXInstaller.cmake
@@ -13,15 +13,28 @@ SET(OSX_INSTALL_DIR ${CMAKE_MODULE_PATH}/osx_install)
 ################################################################################
 SET(OSX_TEMP "${CMAKE_BINARY_DIR}/osx_install_files")
 
+# Common files - libforge, ArrayFireConfig*.cmake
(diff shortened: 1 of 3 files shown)


shehzan10:
I want to make another small update to this. Will reopen when done.

PR #1299, opened 2016-02-23
Changes to get device on which an array was created.

### LINKED ISSUE #1181: Add backend and device id information to af_array
3.2 implemented a function called `getBackendId`, but that does not provide us information about the device the array was created on. We will need to return information about the device the array was created on so the user is aware if they can be inter-op'd on or not.

~~Considering that the unified backend is more orless thread safe~~(It is not thread safe). The arrays returning the same backend and the same device may have been created in different threads. We will also need to provide information about "contexts" that is unique to each device per thread.

pavanky:
- Fixes #1181 

[skip ci]

### DIFF
--- a/test/array.cpp
+++ b/test/array.cpp
@@ -454,3 +454,34 @@ TEST(Device, unequal)
         ASSERT_EQ(ptr, b.device<float>());
     }
 }
+
+TEST(DeviceId, Same)
+{
+    array a = randu(5,5);
+    ASSERT_EQ(getDevice(), getDeviceId(a));
+}
+
+TEST(DeviceId, Different)
+{
+    int ndevices = getDeviceCount();
+    if (ndevices < 2) return;
+
+    int id0 = getDevice();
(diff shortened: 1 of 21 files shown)


shehzan10:
build arrayfire windows ci

test/array.cpp
```diff
@@ -454,3 +454,28 @@ TEST(Device, unequal)
...
+{
+    array a = randu(5,5);
+    ASSERT_EQ(getDevice(), getDeviceId(a));
+}
+
+TEST(DeviceId, Different)
+{
+    if (getDeviceCount() < 2) return;
+    array a = randu(5,5);
+    ASSERT_EQ(getDeviceId(a), 0);
```
shehzan10: This should not be asserted without doing an explicit `setDevice`. It is possible to run tests on other devices.

pavanky: Ok, I'll fix this.

PR #1305, opened 2016-02-25
Support to set visibility of windows programmatically

9prady9:
[skip ci]

### DIFF
--- a/src/api/c/image.cpp
+++ b/src/api/c/image.cpp
@@ -264,6 +264,28 @@ af_err af_is_window_closed(bool *out, const af_window wind)
 #endif
 }
 
+af_err af_set_visibility(const af_window wind, const bool is_visible)
+{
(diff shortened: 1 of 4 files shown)


src/api/c/image.cpp
```diff
@@ -264,6 +264,28 @@ af_err af_is_window_closed(bool *out, const af_window wind)
 #endif
 }
 
+af_err af_set_visibility(const af_window wind, const bool is_visible)
+{
+#if defined(WITH_GRAPHICS)
+    if(wind==0) {
```
jnicholls: space between paren and `if`

pavanky: He's asleep now and we have to release soon :( looks like we'll have to fix it at a later point.

9prady9: Are you talking about space between parenthesis and if in line 277 ?

PR #1309, opened 2016-02-29
README.md changes

mlloreda:
* Reorganized structure of the document. Most notably, the example code has been pulled near the top and the contact section has been pushed to the bottom of the document.
* Citation information has been moved to a new file, `CITATION.md`. ArrayFire is probably not cited enough to warrant a spot in the README.
* Removed `techical@arrayfire.com` email address. Users should use the issue tracker or the Google Groups mailing list, instead.
* Example code now using two-space indentation.


umar456:
Squash

pavanky:
@mlloreda I agree with @umar456 here. Here are a list of things that need to be done.

- Squash the commits to one or two.
- Move `CITATION.md` to a `.github` directory
- I think we should also link to the citation page from `README.md` page.

pavanky:
I like the over all structure. My only concern is about the contact / support being at the bottom of the page. However I am not too worried about it if we can reduce the overall length of the README page.

I think `Installing` and `Building` should be two alternatives (as subsections) under `Getting ArrayFire` similar to this: https://github.com/arrayfire/arrayfire-python#getting-started

pavanky:
When this is settled, I think we should change the code sample to be something simpler.

@arrayfire/core-devel suggestions for the code sample ?

PR #1315, opened 2016-03-03
histogram - doing more work per thread block

anshumang:
@pavanky Here is the PR. 
Taking the global memory bin atomic update loop outside the loop did not change performance.

1 test fails now but the grid size is unchanged. With a grid size of occupancy*multiProcessorCount, all tests fail. Performance is same (and better than upstream) even if grid size is unchanged. 

Did not use memory allocator for allocating global counter to track number of completed blocks.
Using the null stream to do the memcpy that initializes above counter to zero.

### DIFF
--- a/src/backend/cuda/kernel/histogram.hpp
+++ b/src/backend/cuda/kernel/histogram.hpp
@@ -30,37 +30,53 @@ __forceinline__ __device__ int minimum(int a, int b)
 
 template<typename inType, typename outType, bool isLinear>
 static __global__
-void histogramKernel(Param<outType> out, CParam<inType> in,
-                     int len, int nbins, float minval, float maxval, int nBBS)
+void histogramKernel(Param<outType> out, CParam<inType> in, unsigned long long int *d_nb_blocks_completed, 
+                     unsigned long long int tot_nb_blocks, int len, int nbins, float minval, float maxval, int nBBS)
 {
+    __shared__ unsigned long long int s_nb_blocks_completed;
+
     SharedMemory<outType> shared;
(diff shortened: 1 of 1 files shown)


9prady9:
@anshumang Is it the GFOR test that is failing for you as well ?

I think the performance doesn't degrade with grid size same as upstream because the blocks that doesn't have any work are just returning.

pavanky:
@anshumang This looks commit looks like you did not change existing behavior except for adding atomic variables. Is that correct ?

anshumang:
@9prady9 Yes it the GFOR test failing for me too.
Yes the blocks with blockIdx.x>=90 don't do any work so the performance impact is negligible but I shouldn't need to run them to get a correct output. That's where I am confused.

@pavanky blockIdx.x should go from 0 to tot_nb_blocks-1, that's why used ">="
Have not changed the grid size because tests were failing.

pavanky:
@anshumang You are incrementing the "block done" count before going ahead with the computation. It should be a simple > not >=.

anshumang:
@pavanky Isn't atomicAdd returning the old value? That goes into s_nb_blocks_completed and I compare that to tot_nb_blocks.

pavanky:
Oops, I missed that. OK now I am unsure why the tests are failing. The kernel code looks identical..

pavanky:
@anshumang Are you still working on this ? If so can you just send a new PR to devel instead. We want to release 3.3.1 tomorrow if possible.

anshumang:
@pavanky I haven't worked on this since the PR. If I get a chance to debug the failing test, I will send a new PR to devel. Thanks for the review.

PR #1349, opened 2016-03-18
Corrected checksum for Boost Compute

Brainiarc7:
Hello,

The currently specified checksum for Boost Compute is invalid.

I've updated it to the correct value.

Regards,

Brainiarc7


pavanky:
@Brainiarc7 I don't know what you are seeing, but I still get the old md5sum..

shehzan10:
As I said in the issue you created, that checksum is for an empty file. Cmake will create an empty file when it fails. This may make cmake pass, but have you actually tried to build it? Because when the archive is extracted, it will be empty and the compilation will give missing includes. This PR will surely break compilation.

PR #1368, opened 2016-03-28
Prevent inclusion of <sys/sysctl.h> on gnu hurd.

ghisvail:
According to the current [build log](https://buildd.debian.org/status/fetch.php?pkg=arrayfire&arch=hurd-i386&ver=3.3.1%2Bdfsg1-1&stamp=1459100721), `<sys/sysctl.h>` get included when compiled on the gnu-hurd architecture although it should not be. This commit adds the necessary guard to prevent this.

### DIFF
--- a/src/backend/host_memory.cpp
+++ b/src/backend/host_memory.cpp
@@ -16,7 +16,7 @@
 #include <sys/types.h>
 #include <sys/param.h>
 
-#if defined(BSD)
+#if defined(BSD) && !defined(__gnu_hurd__)
 #include <sys/sysctl.h>
 #endif
 


pavanky:
I wonder where `BSD` is coming from..

ghisvail:
According to the list of compiler macros maintained [here](https://sourceforge.net/p/predef/wiki/OperatingSystems/), it should probably be:

```
#if defined(__FreeBSD_kernel__)
```

instead of `BSD` (which is not considered an architecture but a version identifier), and would include both the FreeBSD kernel and GNU/kFreeBSD kernels. Perhaps this should be enough to exclude GNU/Hurd actually.

PR #1380, opened 2016-04-07
Disable CPU Async if GCC Version is less than 4.8.4

### LINKED ISSUE #1378: CPU async causing crashes
See arrayfire/arrayfire-python#75.

shehzan10:
[skip ci]

Fixes #1378


shehzan10:
In the current code, if the compiler is older than 4.8.4, then there is no way to compile it for Async also (cant force it either since the option itself is entirely removed). This would cause a lot of problems.
There are 2 options to fix it:

1. Instead of deleting the variable itself, we can set it to OFF when the compiler is older.
2. Have a separate variable, say `FORCE_CPU_ASYNC_COMPILE` that would bypass the compiler version check.

I think the first option is more clean.

shehzan10:
Fixed the last comment. CMake will turn off the async variable if an older compiler is detected, but user can set it to on.

PR #1387, opened 2016-04-18
Bugfix for rgb2gray

### LINKED ISSUE #1386: rgb2gray misbehaves for non-float inputs
The following code:
`auto rgb = af::array(10, 10, 3, af_dtype::u8);`
`auto gray = af::rgb2gray(rgb);`
gives wrong values and segfaults.
I think that the implementation of `rgb2gray` wrongly uses array `in`, where it should use `input` (which is `in` cast to float) (lines 45-47 in arrayfire/src/api/c/rgb_gray.cpp).

pavanky:
- Fixes #1386


pavanky:
@shehzan10 merge after review

PR #1373, opened 2016-03-31
Improvements to sort functions

shehzan10:
This pull request adds:
* PERF: Improve the performance of batched sort along 0th dimension
* FEAT: Add sorting along higher dimensions
* Enable larger tests for sorting functions
* FEAT: Add multi dimension support for media
* Add multi dimensions tests for median

Fixes #291 #394

PR #1397, opened 2016-04-26
Improves sort_by_key instantiations

shehzan10:
CPU and OpenCL use single source file to compile to separate object files.
CUDA requires generating files using CMake based on a template file which is then added as source.


umar456:
There HAS to be a better way to do this.

shehzan10:
I'm open to ANY other way you may have for this. I don't like it either. But right now, this is the only way its going to compile well on all architectures.

umar456:
You can make multiple .o targets and pass a different -D flag to each one using the same source file. It will require some changes to the CMake files but it will be hell of a lot better then adding 20 files.

shehzan10:
Ok, that might work. I'll look into it.

shehzan10:
build arrayfire tegrak1 ci
build arrayfire tegrax1 ci

src/backend/cpu/kernel/sort_by_key/sort_by_key_impl.cpp
```diff
@@ -9,11 +9,13 @@
 
 #include <kernel/sort_by_key_impl.hpp>
 
+// SBK_TYPES:float double int uint intl uintl short ushort char uchar
```
umar456: Very nice

FilipeMaia:
I don't know if there's anything that can be done, but it became extremely slow to compile arrayfire after this merge.

shehzan10:
Well, actually it became slowly after #1373 because its like adding a bunch of new kernels as sort_by_key requires All-to-All mapping for types. With #1373 the number of instantiations per file were so many that NVCC was crashing even on the best machines. This fixed the crashing part.

I have been thinking about this and it might make a difference if we remove the dim template from the batched function. The dim template is not actually used in any of the actual sort calls. I'm looking into the time saved - its about a minute and a half reduction (mostly on the CUDA backend).

pavanky:
@FilipeMaia Can you verify if it is after this PR ? We had other PRs with similar names.

FilipeMaia:
@shehzan10 is probably correct. I haven't tried just after #1373 but it's the new sort kernels that cause the slowness, so the reason must be #1373. 
I have no solution, just pointing out that the slowness is at a level where it becomes a practical concern.

pavanky:
@FilipeMaia that is a problem we have to deal with because of using a template based library like thrust.. :-/ Too many instantiations.

Need to figure out if there is any easier way to deal with this.

PR #1404, opened 2016-04-29
Cleanup included header files, bug fixes

### LINKED ISSUE #1398: Bug in af::seq when using floating point numbers
`af::seq(0, 2.9, 0.1)` returns only 20 elements.

A possible work around for now:

```
af::array sequence(double start, double end, double step)
{
    double diff = end - begin;
    double step_fixed = fabs(step) * (signbit(diff) ? -1 : 1);
    int len = (int)(diff + step_fixed)/step;
   return range(len) * step + begin;
}
```

### LINKED ISSUE #1406: size_of is missing a case for u64
The internal function `size_of(const dtype)` does not have a case for `u64`. Moreover, the default returns `sizeof(float)` instead of failing.

In consequence, bad things happen when the function is used with this data type. For example, calling `pinned(count, u64)` does *not* fail, but allocates only *half* of the memory requested -- an error which might be really hard to debug!

As a side note, I would recommend that *all* functions switch()ing over dtypes fail with an assert when an unhandled type is encountered.

shehzan10:
Fixes #1398 
Fixes #1406 

[skip ci]

### DIFF
--- a/test/memory.cpp
+++ b/test/memory.cpp
@@ -20,6 +20,8 @@ using std::vector;
 using std::string;
 using std::cout;
 using std::endl;
+using af::cfloat;
+using af::cdouble;
 
 const size_t step_bytes = 1024;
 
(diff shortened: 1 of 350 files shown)


test/memory.cpp
```diff
@@ -74,6 +76,159 @@ TEST(Memory, Scope)
...
+TYPED_TEST(MemAlloc, PtrScope2048)
+{
+    memAllocPtrScopeTest<TypeParam>(2048);
+}
+
+TYPED_TEST(MemAlloc, PtrScope2293)
+{
+    memAllocPtrScopeTest<TypeParam>(2293);
+}
+
```
pavanky: :+1: 

@umar456 would approve!

PR #1408, opened 2016-05-05
OpenCL graphics fixes, Update dependency build tags, CMake clean up

shehzan10:
* Fix OpenCL Graphics failures with error -30
* Update clBLAS, clFFT, Forge to staged branches
* Clean up CMake verbosity and use package handlers
* Fix NVVM and cusolver for when CUDA toolkit is changed


pavanky:
@shehzan10 let's try and not mix bug fixes and build fixes into the same PR in the future.

PR #1409, opened 2016-05-08
Fixes to build arrayfire with gcc 6.1.1

pavanky:
- CUDA backend being forced to use c++98


pavanky:
This removes annoying warnings from CPU and OpenCL backends, but introduces a ton in CUDA backend like this.

```
cc1: warning: command line option ‘-std=c++98’ is valid for C++/ObjC++ but not for C
```
I can't figure out if this can be resolved.

PR #1411, opened 2016-05-09
CUDA Kernels: Use ints instead of dim_t in index calculations

### LINKED ISSUE #1410: unwrap much slower on CUDA compared to OpenCL on same GPU
Code to reproduce:

```
include <arrayfire.h>
#include <cstdio>
#include <cstdlib>

using namespace af;

af::array A;
void bench_unwrap()
{
    array B = unwrap(A, 512, 1, 256, 1);
}

{
    af::info();
    A = randu(16384, 1000);
    printf("Time: %lfms\n", 1000 * timeit(bench_unwrap));
    return 0;
}
```

-------

**Results**

OpenCL:
```
ArrayFire v3.4.0 (OpenCL, 64-bit Linux, build 4d625e7)
[0] NVIDIA  : GeForce GTX 950, 2047 MB
-1- AMD     : Intel(R) Core(TM) i7-3770K CPU @ 3.50GHz, 14997 MB
Time: 7.991200ms
```

CUDA:
```
ArrayFire v3.4.0 (CUDA, 64-bit Linux, build 4d625e7)
Platform: CUDA Toolkit 7.5, Driver: 364.19
[0] GeForce GTX 950, 2048 MB, CUDA Compute 5.2
Time: 27.438000ms
```

shehzan10:
[skip ci]

Fixes #1410

### DIFF
--- a/src/backend/cuda/kernel/approx.hpp
+++ b/src/backend/cuda/kernel/approx.hpp
@@ -27,13 +27,13 @@ namespace cuda
         ///////////////////////////////////////////////////////////////////////////
         template<typename Ty, typename Tp>
         __device__ inline static
-        void core_nearest1(const dim_t idx, const dim_t idy, const dim_t idz, const dim_t idw,
+        void core_nearest1(const int idx, const int idy, const int idz, const int idw,
                            Param<Ty> out, CParam<Ty> in, CParam<Tp> pos,
                            const float offGrid, const bool pBatch)
         {
-            const dim_t omId = idw * out.strides[3] + idz * out.strides[2]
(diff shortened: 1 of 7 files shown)


pavanky:
@shehzan10 OpenCL is using `dim_t` as well. This does not explain #1410

shehzan10:
The performance is comparable to OpenCL after that fix.

pavanky:
Nevermind, that is it :-/ Looks like the OpenCL is using 32 bit ints by default. Can you change OpenCL kernels to ints as well  to be consistent

shehzan10:
Done

PR #1412, opened 2016-05-10
Transform batching

### LINKED ISSUE #1346: Support One-to-One batch mode in transform
Allow batch-mode processing for translate() to allow

1. The same image to be translated according to specified vectors of row and column shifts.
2. Multiple images to be translated by the same scalar row and column shifts.

This is useful in image processing, for instance, to register and add multiple images to reduce noise. An implementation of this capability in the generalized transform() function would be useful as well.

shehzan10:
Fixes #1346 

This PR adds higher forms of batching for transform. There is no change in API. However it breaks 1 specific case of batching in the previous version (read [note below](#behavior-change)).

* Rename ConvolveBatchKind to AF_BATCH_KIND 
* Add support for batching on 4th dimension
* Add support for one-to-one batching in transform
* Batching now follows similar batch parameters as convolve
* On dim 2 and dim 3, the image and kernels can:
      * All ones
      * If image has dim[d] = N > 1, then, the kernel dim[d] can either be 1 or N
      * Similarly, if kernel has dim[d] = K > 1, then, the image dim[d] can either be 1 or K
      * Where d is 2 or 3
* If transforms are batched, then output dimensions has to be specified

### <a name="behavior-change"></a>Behavior Change
This PR changes the existing behavior of transform in 1 specific case
Previously, for inputs `input.dims = [X, Y, N, 1]` and `transform = [3, 2, K, 1]`, the output was `[X, Y, N * K, 1]` where **both N and K > 1**.
This has now been changed as follows:
* N != K is **NOT** allowed if both N and K > 1.
* When N=K, the batching is one-to-one
* To achieve the old behavior, use `moddims` to change the batch dimension of transform from `[3, 2, K, 1]` to `[3, 2, 1, K]`. This will give the output `[X, Y, N, K]` (can be changed back to `[X, Y, N * K]` with `moddims`). `Moddims` operation is a no-op so there is not performance hit.

**There is NO CHANGE when either N or K are 1.**
[skip ci]


src/backend/defines.hpp
```diff
@@ -41,3 +41,11 @@ clipFilePath(std::string path, std::string str)
     #define __AF_FILENAME__ (clipFilePath(__FILE__, "src/").c_str())
 #endif
 
+typedef enum {
+    AF_BATCH_UNSUPPORTED = -1, /* invalid inputs */
+    AF_BATCH_NONE,             /* one signal, one filter   */
+    AF_BATCH_SIGNAL,           /* many signal, one filter  */
+    AF_BATCH_KERNEL,           /* one signal, many filter  */
+    AF_BATCH_SAME,             /* signal and filter have same batch size */
+    AF_BATCH_DIFF,             /* signal and filter have different batch size */
```
pavanky: Can you change the name to `AF_BATCH_LHS` and `AF_BATCH_RHS` instead of signal and kernel ?

shehzan10: Resolved

PR #1414, opened 2016-05-12
BUGFIX fix incorrect results from timeit

### LINKED ISSUE #1413: Bug in timeit
`timeit` is giving incorrect results. There are the following issues:
* Its taking the minimum time
* It might not sun a significant number of iterations
* This can result in weird results.

Sample code to reproduce
```
/*******************************************************
 * Copyright (c) 2014, ArrayFire
 * All rights reserved.
 *
 * This file is distributed under 3-clause BSD license.
 * The complete license agreement can be obtained at:

#include <arrayfire.h>
#include <stdio.h>
#include <math.h>
#include <cstdlib>

using namespace af;

// create a small wrapper to benchmark
static array A; // populated before each timing
static void fn()
{
#if 1   // IN PLACE
    array B = A.copy(), tau;
    qrInPlace(tau, B);
    tau.eval();
    B.eval();
#else
    #if 0   // Decompose Q and R
        array q, r, tau;
        qr(q, r, tau, A);
        q.eval();
        r.eval();
        tau.eval();
    #else
        array out, tau;
        qr(out, tau, A);
        out.eval();
        tau.eval();
    #endif
#endif
}

double loopFn()
{
    timer begin =  timer::start();
    for(int i = 0; i < 100; i++) {
        fn();
    }
    af::sync();
    double elapsed = timer::stop(begin);
    return elapsed / 100;
}


int main(int argc, char ** argv)
{
    double peak = 0;
    try {
        int device = argc > 1 ? atoi(argv[1]) : 0;
        setDevice(device);
        info();

        printf("Benchmark N-by-N matrix multiply\n");
        for (int n = 128; n <= 2048; n += 128) {

            printf("%4d x %4d: ", n, n);
            A = randu(n,n);
            double time = timeit(fn); // time in seconds
            double timeLoop = loopFn(); // time in seconds
            printf(" %4.4f ms | %4.4f ms\n", time * 1000, timeLoop * 1000);
            fflush(stdout);
        }
    } catch (af::exception& e) {
        throw;
    }

    return 0;
}
```

### LINKED ISSUE #1418: gray2rgb() should work for images of size Nx1 (and 1x1) too
currently, gray2rgb() requires numdims() == 2, which appears to be an unnecessary restriction.

There's a workaround for the Nx1 case (out = gray2rgb(in.T()).T(), but that fails if in is 1x1.

shehzan10:
* Timeit now computes the median time from a select number of trial runs that
  are synced after each iteration
* The median time is then used to compute how many batches of the functions to
  run based on some minimum time
* It now runs batch of trial runs and then averages the times from each batch

Fixes #1413
Fixes #1418

### DIFF
--- a/src/api/c/rgb_gray.cpp
+++ b/src/api/c/rgb_gray.cpp
@@ -109,8 +109,13 @@ af_err convert(af_array* out, const af_array in, const float r, const float g, c
         af_dtype iType     = info.getType();
         af::dim4 inputDims = info.dims();
 
-        ARG_ASSERT(1, (inputDims.ndims()>=2));
+        // 2D is not required.
+        ARG_ASSERT(1, info.elements() > 0);
+
+        // If RGB is input, then assert 3 channels
+        // else 1 channel
         if (isRGB2GRAY) ARG_ASSERT(1, (inputDims[2]==3));
+        else            ARG_ASSERT(1, (inputDims[2]==1));
 
         af_array output = 0;
         switch(iType) {
--- a/src/api/cpp/timing.cpp
+++ b/src/api/cpp/timing.cpp
@@ -10,6 +10,7 @@
 #include <af/timing.h>
 #include <af/device.h>
 #include <math.h>
+#include <vector>
 #include <algorithm>
 
 using namespace af;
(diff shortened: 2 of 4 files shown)


pavanky:
@shehzan10 while we are at it, let's use std::chrono and get rid of the unnecessary include files.

pavanky:
build arrayfire ci

shehzan10:
build arrayfire linux ci
build arrayfire osx ci

shehzan10:
build arrayfire linux ci
build arrayfire osx ci

pavanky:
@shehzan10 what about this PR?

shehzan10:
cusolver QR functions are not playing well with the stream addition. LU, Cholesky etc are fine.
Failing tests are:
inverse
qr
rank
solve

I haven't gotten to the bottom of the issue yet.

[jenkins/Linux-x64 Build Log](http://ci.arrayfire.org/job/arrayfire-linux/job/pr-build/467/consoleFull)

shehzan10:
Change cusolver back to default stream
* Synchronize the ArrayFire stream when cusolver getHandle is done
* The stream form of cusolver require using cudaDeviceSynchronize and
cudaStreamSynchronize all over the place
* Even then, getrs in solve_lu failed on any stream other than 0
* patch available at: https://gist.github.com/shehzan10/414c3d04a40e7c4a03ed3c2e1b9072e7

PR #1415, opened 2016-05-12
FEAT Add fallback to CUDA OpenGL Interop

### LINKED ISSUE #1402: Improved error messages and fallback options graphics functions
Right now trying to run graphics example on CUDA GPUs without display attached fail with [errors such as these](https://github.com/arrayfire/arrayfire-python/issues/78).

We should look into providing a warning message about GL sharing being unavailable and that it is using fallback instead.

We can sort of do that in OpenCL backend by setting a flag, but the CUDA backend has no such option.

shehzan10:
This comes into play when devices are not graphics capable but OpenGL is
available.

On Windows, when all devices are in TCC mode or not connected to display, CUDA
throws an error `CUDA Error (63): OS call failed or operation not supported on this OS`.

On Linux, if all devices are in TCC mode, the graphics part will still run as
long as OpenGL is available.

The checking function uses the [cudaGLGetDevices](http://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__OPENGL.html#group__CUDART__OPENGL_1g3471ecaa5b827c94f2c55ab51fde1751).
This function returns
* Linux: `cudaErrorNoDevices (38)` if all the devices are in TCC mode
* OSX: This function is not supported and returns `cudaErrorNotSupported (71)`.
There is no need for fallback in these cases.

Fixes #1402

### DIFF
--- a/src/backend/cuda/hist_graphics.cu
+++ b/src/backend/cuda/hist_graphics.cu
@@ -21,23 +21,35 @@ namespace cuda
 template<typename T>
 void copy_histogram(const Array<T> &data, const fg::Histogram* hist)
 {
-    const T *d_P = data.get();
+    if(InteropManager::checkGraphicsInteropCapability()) {
+        const T *d_P = data.get();
 
-    InteropManager& intrpMngr = InteropManager::getInstance();
+        InteropManager& intrpMngr = InteropManager::getInstance();
 
-    cudaGraphicsResource *cudaVBOResource = intrpMngr.getBufferResource(hist);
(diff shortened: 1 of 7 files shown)


pavanky:
build arrayfire ci

shehzan10:
build arrayfire ci

shehzan10:
This pull request has passed the build tests. The statuses were not set due to a [Jenkins security fix](https://wiki.jenkins-ci.org/display/JENKINS/Plugins+affected+by+fix+for+SECURITY-170) that disabled the relevant plugin from being able to set them (this has been fixed now).

Issue #1429, opened 2016-05-26
[OpenCL] Template specialization in af/opencl.h causing multiple definitions at link time.


bkloppenborg:
The presence of the template specialization for `af::array::device` in `af/opencl.h` precludes the ability to include the header from multiple files as it results in conflicting definitions at link time.

pavanky:
Fixed in devel

PR #1430, opened 2016-05-26
Move template specialization to the OpenCL backend.

bkloppenborg:
Fixes issue #1429

### DIFF
--- a/include/af/opencl.h
+++ b/include/af/opencl.h
@@ -434,27 +434,5 @@ static inline platform getPlatform()
 */
 }
 
-namespace af
-{
- /**
-     \addtogroup opencl_mat
-     @{
- */
-
-#if !defined(AF_OPENCL)
-template<> AFAPI cl_mem *array::device() const
-{
-    cl_mem *mem_ptr = new cl_mem;
-    af_err err = af_get_device_ptr((void **)mem_ptr, get());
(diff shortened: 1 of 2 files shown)


pavanky:
I really don't like having af::array inside src/backend

umar456:
Its either that or `#ifdefs` in src/api

pavanky:
Or remove AFAPI from template instantiation ?

pavanky:
Maybe you need to make it static as well.

umar456:
Then its a different function. The idea is to have the same interface as the `device<T>()` function which is a member function.

shehzan10:
Wouldn't this be an API breaking change or was it a mistake that the declaration was removed too?

umar456:
The API is the same but the definition shouldn't have been in the header. The definition is already in include/af/array.h

shehzan10:
Might be a bad idea, but cast `cl_mem` as `void*` and document it. Users can cast it on their own. CPU/CUDA can return NULL. This way the standard c++ -> c -> backend track can be taken.

umar456:
That is not a good design. It is confusing, not portable, and error prone. I don't understand why this approach is a problem.

pavanky:
@umar456 It shouldn't matter if it is a different function. It exists only within the header. We are not breaking ABI for the library.

umar456:
The ABI is not being broken. A function is being added not removed.

PR #1433, opened 2016-05-30
correct cmake args for local example compile

gpryor:
The suggested cmake invocation for a local arrayfire installation only worked when utilizing `ArrayFire_DIR` as opposed to `ArrayFire_ROOT` and also by specifying the explicit directory containing the referred cmake file. cmake was version 3.5.2 -- perhaps a versioning issue?


pavanky:
@gpryor Thanks for catching this. It's been `ArrayFire_DIR` for a while. It looks like the documentation hasn't been fixed

PR #1434, opened 2016-05-30
update ocl cpp headers

### LINKED ISSUE #1344: Moving from cl.hpp to cl2.hpp in OpenCL backend.
`cl2.hpp` seems to be the way forward. it is backwards compatible with older versions of OpenCL as well as OpenCL 2.0. It also seems to have a more terse way of dealing with some commonly used functionality.

Things to do for this to happen:

- [x] Remove `cl.hpp` from the source repo.
- [x] Add `cl2.hpp` as an [external dependency](https://github.com/KhronosGroup/OpenCL-CLHPP) similar to boost compute.
- [x] Change the internal functions wherever they are being used to the new API.

9prady9:
Fixes #1344 

- [x] Removed old cl.hpp header file
- [x] Added `OpenCL-CLHPP` header only repository as external project
- [x] Change API to use latest header functions
- [x] Provide mechanism to automatically choose among installed `cl2.hpp` and the one generated by ArrayFire's external project.


src/backend/opencl/platform.cpp
```diff
@@ -22,8 +22,6 @@
 
 #endif
 
-#include <cl.hpp>
```
shehzan10: Don't you need to include cl2.hpp?

CMakeModules/build_cl2hpp.cmake
```diff
@@ -0,0 +1,28 @@
...
+    CONFIGURE_COMMAND ${CMAKE_COMMAND} -Wno-dev "-G${CMAKE_GENERATOR}" <SOURCE_DIR>
+    -DCMAKE_SOURCE_DIR:PATH=<SOURCE_DIR>
+    -DCMAKE_CXX_COMPILER:FILEPATH=${CMAKE_CXX_COMPILER}
+    -DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}
+    -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
+    -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+    -DBUILD_DOCS:BOOL=OFF
+    -DBUILD_EXAMPLES:BOOL=ON
+    -DBUILD_TESTS:BOOL=OFF
+    )
```
shehzan10: Does this really require CMake to run? Why can't we just git-clone it like boost compute?

src/backend/opencl/platform.cpp
9prady9: :) That is the third task, i just made sure it is working as it is with old header using the external project setup. What @pavanky suggested on the issue is that functions/functors have become more terse. I have no idea what are the new things, so still have to see what this new header does. Then only i can change it to cl2.hpp.

CMakeModules/build_cl2hpp.cmake
pavanky: Looks like it is auto generated

9prady9: It is running a python script to generate the header. So, just git clone won't do the job i think. unless there is some CMAKE trick to do such things.

pavanky: Hmm this might be a problem for people who don't have python on their machines. 

Can we host an auto-generated cl2.hpp someplace and download that instead?

shehzan10: Python is required for clBLAS from source (not for external clBLAS)

pavanky: Good point. So we just recommend people have python installed. We could also put the generated `cl2.hpp` some place if needed at a later point.

@9prady9 looks like what you did here is ok :+1:

shehzan10: What is the fallback for when there is no network access? Package maintainers would require this.

pavanky: @shehzan10 `cl2.hpp` is already shipped by ubuntu, debian and archlinux.

9prady9: Wouldn't that be the case for all cmake external projects ?

shehzan10: @9prady9 that is the case. clBLAS/clFFT/compute all have options to use a system one.
@pavanky the latest ones may be, but we still have to provide for it.

Also, would there be an issue with ArrayFire-External OpenCL interop if someone is using cl.hpp in their code? I know this is essentially a wrapper for OpenCL-C, but need to verify.

9prady9: Just changing from `cl.hpp` to cl2.hpp` won't work i, because i am getting compilation errors by doing that.

pavanky: @shehzan10 what do you mean latest ones ? Latest version of the distros ? 

Also, No need to change externally. We only interface with OpenCL C externally.

9prady9:
build arrayfire ci

9prady9:
I was able to reproduce the issue that happened on build machines, weird that i didn't run into this when i was implementing it. I am looking into it.

9prady9:
Cause of windows ci build has been addressed in [this](https://github.com/arrayfire/arrayfire/pull/1436) PR.

9prady9:
build arrayfire osx opencl ci
build arrayfire linux-03 opencl ci

9prady9:
build arrayfire osx opencl ci

9prady9:
build arrayfire osx opencl ci
build arrayfire linux-03 ci

9prady9:
I am still waiting for https://github.com/KhronosGroup/OpenCL-CLHPP/pull/14 to merge. Once that is done, i will update the external project git url and branch.

9prady9:
I have moved the CLHPP repository ownership to ArrayFire and update the URL.

9prady9:
build arrayfire linux opencl ci
build arrayfire osx opencl ci
build arrayfire linux-03 ci

9prady9:
@shehzan10 / @pavanky : any reason why builds aren't starting ?

9prady9:
@pavanky All builds passed, this is ready for merge now.

pavanky:
@9prady9 since we are targeting OpenCL 1.2 and above, I think we should add the following flag to avoid annoying messages

```
$ git di src/backend/opencl/CMakeLists.txt
diff --git a/src/backend/opencl/CMakeLists.txt b/src/backend/opencl/CMakeLists.txt
index 29713da..403e990 100644
--- a/src/backend/opencl/CMakeLists.txt
+++ b/src/backend/opencl/CMakeLists.txt
@@ -3,6 +3,8 @@ PROJECT(ARRAYFIRE)
 
 FIND_PACKAGE(OpenCL REQUIRED)
 
+ADD_DEFINITIONS(-DCL_USE_DEPRECATED_OPENCL_1_2_APIS)
+
 IF(NOT USE_SYSTEM_CL2HPP)
     INCLUDE("${CMAKE_MODULE_PATH}/build_cl2hpp.cmake")
 ENDIF(NOT USE_SYSTEM_CL2HPP)
```

This also has the benefit of providing a work around for the buggy system headers provided for archlinux.

PR #1435, opened 2016-05-31
[WIP] Evaluating multiple outputs inside a single kernel

### LINKED ISSUE #1247: Give users the ability to compile custom JIT kernels.
Consider the following code:

```
result1 = a1*x1 + a2*x2 + a3*x3 + a4*x4;
result2 = a1*y1 + a2*y2 + a3*y3 + a4*y4;
result3 = a1*z1 + a2*z2 + a3*z3 + a4*z4;

result1.eval();
result2.eval();
result3.eval();
```

The above code results in the firing of 3 separate kernels and hence 3 round trips to the global memory to fetch `a1, a2, a3, a4`. Instead, if there were a command like `eval(result1, result2, result3)` which fires a single kernel to compute `result1, result2, result3`, the memory would be reused inside the kernel and thus save on global reads.

pavanky:
This is a work in progress. So far implemented only in OpenCL as a POC.

TODO:

- [ ] CUDA implementation
- [ ] OpenCL implementation
- [ ] CPU implementation

Issues that need to be resolved:

- [ ] Intel CPU + OpenCL slowed down for code shown in #1247 
- [ ] output types need to be of same type

[skip ci]

Will fix #1247

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -1243,11 +1243,12 @@ namespace af
        @{
     */
     inline array &eval(array &a) { a.eval(); return a; }
-    inline void eval(array &a, array &b) { eval(a); b.eval(); }
-    inline void eval(array &a, array &b, array &c) { eval(a, b); c.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d) { eval(a, b, c); d.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e) { eval(a, b, c, d); e.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e, array &f) { eval(a, b, c, d, e); f.eval(); }
+    AFAPI void eval(array &a, array &b);
+    AFAPI void eval(array &a, array &b, array &c);
+    AFAPI void eval(array &a, array &b, array &c, array &d);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e, array &f);
+    AFAPI void eval(int num, array *arrays);
     /**
        @}
     */
@@ -1344,6 +1345,16 @@ extern "C" {
     /**
       @}
     */
+
+
+    /**
+       Evaluate multiple arrays together
+    */
+    AFAPI af_err af_eval_multiple(const int num, af_array *arrays);
+    /**
+      @}
+    */
+
 
     /**
         \ingroup method_mat
--- a/src/backend/opencl/jit.cpp
+++ b/src/backend/opencl/jit.cpp
@@ -10,6 +10,7 @@
 #include <af/dim4.hpp>
 #include <Array.hpp>
 #include <map>
+#include <vector>
 #include <stdexcept>
 #include <copy.hpp>
 #include <JIT/Node.hpp>
@@ -35,10 +36,8 @@ using cl::NDRange;
 using std::string;
 using std::stringstream;
 
-static string getFuncName(Node *node, bool is_linear, bool *is_double)
+static string getFuncName(std::vector<Node *> nodes, bool is_linear, bool *is_double)
 {
-    node->setId(0);
-
     stringstream hashName;
     stringstream funcName;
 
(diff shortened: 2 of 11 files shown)


include/af/array.h
```diff
@@ -1243,11 +1243,12 @@ namespace af
...
-    inline void eval(array &a, array &b, array &c) { eval(a, b); c.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d) { eval(a, b, c); d.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e) { eval(a, b, c, d); e.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e, array &f) { eval(a, b, c, d, e); f.eval(); }
+    AFAPI void eval(array &a, array &b);
+    AFAPI void eval(array &a, array &b, array &c);
+    AFAPI void eval(array &a, array &b, array &c, array &d);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e, array &f);
+    AFAPI void eval(int num, array *arrays);
```
9prady9: Would it be bad to have `std::vector<array>` here instead of c-style array since it is a C++ API function ?

src/backend/opencl/jit.cpp
```diff
@@ -48,108 +47,128 @@ static string getFuncName(Node *node, bool is_linear, bool *is_double)
...
+    static const char *blockStart = "{\n\n";
+    static const char *blockEnd = "\n\n}";
+
+    static const char *linearIndex = "\n"
+        "uint groupId  = get_group_id(1) * get_num_groups(0) + get_group_id(0);\n"
+        "uint threadId = get_local_id(0);\n"
+        "int idx = groupId * get_local_size(0) * get_local_size(1) + threadId;\n"
+        "if (idx >= oInfo.dims[3] * oInfo.strides[3]) return;\n";
+
+    static const char *generalIndex = "\n"
```
9prady9: We can probably use raw string literals here that would make modifying this set of strings much easier in future, also would be easy to read.

include/af/array.h
umar456: We avoid using std lib classes in the ArrayFire interface to avoid compiler incompatibilities.

include/af/array.h
```diff
@@ -1243,11 +1243,12 @@ namespace af
...
-    inline void eval(array &a, array &b) { eval(a); b.eval(); }
-    inline void eval(array &a, array &b, array &c) { eval(a, b); c.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d) { eval(a, b, c); d.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e) { eval(a, b, c, d); e.eval(); }
-    inline void eval(array &a, array &b, array &c, array &d, array &e, array &f) { eval(a, b, c, d, e); f.eval(); }
+    AFAPI void eval(array &a, array &b);
+    AFAPI void eval(array &a, array &b, array &c);
+    AFAPI void eval(array &a, array &b, array &c, array &d);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e);
+    AFAPI void eval(array &a, array &b, array &c, array &d, array &e, array &f);
```
umar456: Variadic template for c++11?

umar456:
```cpp
template<typename... ARGS>
AFAPI void eval(ARGS... args) {
    af_eval_multiple(sizeof...(args), {args...});
}
```
I think.

pavanky: I think we should add that as an additional API, but only expose it if user is compiling with c++11 support.

src/backend/opencl/jit.cpp
pavanky: Yeah, I wanted to do that as well. But my emacs config did not have C++11 support :p I did not want to screw up the indentation.

include/af/array.h
pavanky: We can have a inline header only implementation for std::vector<array> as well.

shehzan10:
Reminder to add the API to unified.

pavanky:
closing this issue so it will not build tests

PR #1444, opened 2016-06-03
BUGFIX: Correct input arrays validation in convolve fns

### LINKED ISSUE #1442: Convolve3 Produces an Error is the Z-Dimension is 1
When using ```af::convolve3```, if the 3rd dimension of either the signal or filter (dim 2), is 1, it will produce an error.

Easy repro:

```
auto x = af::constant(1,3,3,1,1);
auto y = af::constant(2,2,2,1,1);
auto z = af::convolve3(x,y);
```

9prady9:
Fixes #1442 

Earlier to this change, if an n-1 dimensional array and filter were
provided as inputs to af_convolve(n) function, it used to error out
as it expected the nth dimension of arrays to be >1.

After this change, there is no such restriction. For example,
2d arrays passed onto af_convolve3 should give same output as
passing them to af_convolve2. Have added a test checking the same.

### DIFF
--- a/src/api/c/convolve.cpp
+++ b/src/api/c/convolve.cpp
@@ -40,7 +40,7 @@ AF_BATCH_KIND identifyBatchKind(const dim4 &sDims, const dim4 &fDims)
     dim_t sn = sDims.ndims();
     dim_t fn = fDims.ndims();
 
-    if (sn==baseDim && fn==baseDim)
+    if ((sn==baseDim && fn==baseDim) || (sn==baseDim-1 && fn==baseDim-1))
         return AF_BATCH_NONE;
     else if (sn==baseDim && (fn>baseDim && fn<=4))
         return AF_BATCH_RHS;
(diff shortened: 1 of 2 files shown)


src/api/c/convolve.cpp
```diff
@@ -40,7 +40,7 @@ AF_BATCH_KIND identifyBatchKind(const dim4 &sDims, const dim4 &fDims)
     dim_t sn = sDims.ndims();
     dim_t fn = fDims.ndims();
 
-    if (sn==baseDim && fn==baseDim)
+    if ((sn==baseDim && fn==baseDim) || (sn==baseDim-1 && fn==baseDim-1))
```
pavanky: This will still be a problem if you pass column arrays to convolve3.

src/api/c/convolve.cpp
```diff
@@ -74,6 +74,8 @@ af_err convolve(af_array *out, const af_array signal, const af_array filter)
 
         AF_BATCH_KIND convBT = identifyBatchKind<baseDim>(sdims, fdims);
 
+        ARG_ASSERT(1, (sdims.ndims()>=1));
+        ARG_ASSERT(2, (fdims.ndims()>=1));
```
pavanky: We should just check `std::min(sdims.ndims(), fdims.ndims())` and put to the right fft.

9prady9: If we do just that condition, how can we identify right batch operation from the [list of 4](http://arrayfire.org/docs/group__signal__func__convolve3.htm) types of batches we handle ?

pavanky: passing a 2d matrix to convolve3 and convolve2 is generating the same outputs..

shehzan10:
Add a test for this case

9prady9:
Incorrect behavior, will open a new PR with correct changes.

PR #1448, opened 2016-06-04
Function to let user control the number of fft plans cached by fft plan cacher

### LINKED ISSUE #1443: Increase the maximum number of fft caches
100 may be?

9prady9:
Fixes #1443

### DIFF
--- a/include/af/signal.h
+++ b/include/af/signal.h
@@ -1055,7 +1055,7 @@ AFAPI af_err af_fft_convolve2(af_array *out, const af_array signal, const af_arr
 AFAPI af_err af_fft_convolve3(af_array *out, const af_array signal, const af_array filter, const af_conv_mode mode);
 
 /**
-   C++ Interface for finite impulse response  filter
+   C Interface for finite impulse response  filter
 
    \param[out] y is the output signal from the filter
    \param[in] b is the array containing the coefficients of the filter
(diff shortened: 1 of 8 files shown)


include/af/signal.h
```diff
@@ -1078,6 +1078,19 @@ AFAPI af_err af_fir(af_array *y, const af_array b, const af_array x);
    \ingroup signal_func_iir
 */
 AFAPI af_err af_iir(af_array *y, const af_array b, const af_array a, const af_array x);
+
+#if AF_API_VERSION >= 34
+/**
+   C Interface for setting plan cache size
+
+   This function is doesn't do anything if called when CPU backend is active. The plans associated with
```
pavanky: Unnecessary "is"

src/api/c/fft.cpp
```diff
@@ -247,3 +247,14 @@ af_err af_fft3_c2r(af_array *out, const af_array in, const double norm_factor, c
 {
     return fft_c2r<3>(out, in, norm_factor, is_odd);
 }
+
+af_err af_set_fft_plan_cache_size(size_t cache_size)
+{
+#ifndef AF_CPU
```
pavanky: Don't do this. Just have a dummy inline function in CPU backend's fft.hpp. The compiler will optimize it out.

src/backend/cuda/fft.cpp
```diff
@@ -25,11 +27,15 @@ using std::string;
 namespace cuda
 {
 
+typedef std::pair<string, cufftHandle> FFTPlanPair;
+typedef std::deque<FFTPlanPair> FFTPlanCache;
```
pavanky: If we used a map we can get by using only one data type. Considering that the size is going to be pretty small,I don't think it will affect the performance.

pavanky: Hmm, but then it might get hard to figure out which plan top pop. Ok, let's leave this in place.

9prady9:
build arrayfire linux ci

PR #1456, opened 2016-06-10
Fixes to OpenCL-OpenGL interop with different platforms

pavanky:


src/backend/opencl/platform.cpp
```diff
@@ -673,22 +672,26 @@ void DeviceManager::markDeviceForInterop(const int device, const fg::Window* wHa
 #if defined(_WIN32) || defined(_MSC_VER)
                 CL_WGL_HDC_KHR, (cl_context_properties)wHandle->display(),
 #else
-                CL_GLX_DISPLAY_KHR, (cl_context_properties)wHandle->display(),
+                //CL_GLX_DISPLAY_KHR, (cl_context_properties)wHandle->display(),
 #endif
                 CL_CONTEXT_PLATFORM, (cl_context_properties)plat(),
```
pavanky: This is a terrible way to fix this, I know. But this seems to work without a hitch on my machine. Without removing it, it segfaults on the platform that supports opencl-opengl interop, but is not connected to the display.

@shehzan10 @syurkevi @9prady9 can you look at it as well and tell me why this may break on other machines.

pavanky: @bkloppenborg can you have a look at this as well?

9prady9: When only one of the GPUs is connected to display and others are not, may be the OpenCL context we create from the OpenGL context should have both the devices for the interop to function properly. I don't have a multi-GPU setup locally, @pavanky Which machine did you test this on ?

pavanky: You can't dot hat when the devices are from different platforms. I am testing this on my personal machine which has an AMD GPU as well as an NVIDIA GPU.

9prady9: In those cases (GPUs are of different vendors), we can't possibly do anything from user side code, can we ?

pavanky: Well the patch I sent seems to make it work. I was wondering if it is a good idea.

pavanky:
@9prady9 this seems to make it work.

9prady9:
:+1:

pavanky:
build arrayfire linux ci
build arrayfire osx ci

pavanky:
build arrayfire windows ci

pavanky:

PR #1462, opened 2016-06-17
Support for evaluating multiple outputs simultaneously

### LINKED ISSUE #1247: Give users the ability to compile custom JIT kernels.

pavanky:
- Fixes #1247

### DIFF
--- a/src/backend/cuda/platform.cpp
+++ b/src/backend/cuda/platform.cpp
@@ -488,6 +488,12 @@ bool synchronize_calls() {
     return sync;
 }
 
+bool& evalFlag()
+{
+    static bool flag = true;
+    return flag;
+}
+
 }
 
 af_err afcu_get_stream(cudaStream_t* stream, int id)
--- a/src/backend/opencl/Array.cpp
+++ b/src/backend/opencl/Array.cpp
@@ -31,10 +31,8 @@ namespace opencl
     template<typename T>
     void Array<T>::genBufferNode() const
     {
-        bool is_linear = isLinear();
         BufferNode *buf_node = new BufferNode(dtype_traits<T>::getName(),
-                                              shortname<T>(true),
-                                              is_linear);
+                                              shortname<T>(true));
         const_cast<Array<T> *>(this)->node = Node_ptr(reinterpret_cast<Node *>(buf_node));
     }
 
@@ -213,7 +211,7 @@ namespace opencl
             KParam kinfo = *this;
             BufferNode *bufNode = reinterpret_cast<BufferNode *>(node.get());
             unsigned bytes = this->getDataDims().elements() * sizeof(T);
-            bufNode->setData(kinfo, data, bytes);
+            bufNode->setData(kinfo, data, bytes, isLinear());
         }
         return node;
     }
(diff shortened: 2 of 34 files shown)


src/backend/cuda/platform.cpp
```diff
@@ -488,6 +488,12 @@ bool synchronize_calls() {
     return sync;
 }
 
+bool& evalFlag()
```
umar456: This doesn't need to be a reference does it?

src/backend/opencl/Array.cpp
```diff
@@ -149,23 +156,65 @@ namespace opencl
...
-                                                  bytes,
-                                                  is_linear);
-            const_cast<Array<T> *>(this)->node = Node_ptr(reinterpret_cast<Node *>(buf_node));
+            bufNode->setData(kinfo, data, bytes, isLinear());
         }
+        return node;
+    }
 
+    template<typename T>
+    Node_ptr Array<T>::getNode() const
```
umar456: should probably return a const pointer

src/backend/cuda/platform.cpp
pavanky: This function contains a static variable. I am using the function to set it.

src/backend/opencl/jit.cpp
```diff
@@ -66,90 +66,106 @@ static string getFuncName(Node *node, bool is_linear, bool *is_double)
...
+        outParamStream << "__global " << node->getTypeStr() << " *out" << id << ", \n";
+        outWriteStream << "out" << id << "[idx] = " << "val" << id << ";\n";
+        node->genOffsets(offsetsStream, is_linear);
+        node->genFuncs(opsStream);
+        opsStream << "//" << ++count << std::endl << std::endl;
+    }
 
+    // Put various blocks into a single stream
+    stringstream kerStream;
+    kerStream << kernelVoid;
```
umar456: I feel like many of these stream operations could be converted into string concatenations, but I don't know if it will be better for performance

test/jit.cpp
```diff
@@ -115,3 +115,128 @@ TEST(JIT, CPP_JIT_Reset_Unary)
...
+{
+    using af::array;
+
+    const int num = 1 << 16;
+    af::array a = af::randu(num, s32);
+    af::array b = af::randu(num, s32);
+    af::array x = a + b;
+    af::array y = a - b;
+
+    af::eval(x);
```
umar456: These aren't very good tests although I don't know how we could test something like this

PR #1465, opened 2016-06-20
Fixes when building large kernels on AMD devices

### LINKED ISSUE #1457: Make AMD platform to be consistent with the rest
Currently it is set to 16 for AMD, but 20 for everything else. This is because the program seems to be hanging when calling `clBuildProgram` for larger kernels.

pavanky:
- Fixes #1457


pavanky:
build arrayfire linux ci
build arrayfire osx ci

PR #1478, opened 2016-06-28
Add shared(cuda)/local(opencl) memory limit check for bilateral

### LINKED ISSUE #1475: bilateral slower on CUDA compared to OpenCL
``` cpp
#include <arrayfire.h>

void bench()
{
    const int iter = 100;
    for (int n = 512; n <= 2048; n += 512) {
        af::array in = af::randu(n, n);

        af::array out;
        af::timer::start();
        for (int i = 0; i < iter; i++) {
            out = af::bilateral(in, 2, 2);
        }
        af::sync();
        double t = af::timer::stop() / iter;
        printf("Size: %d, Time: %lf ms\n", n, 1000 * t);
    }
}

{
    try {
        af::setBackend(AF_BACKEND_OPENCL);
        af::setDevice(1);
        af::info();
        af::sync();

        bench();
    } catch (af::exception &e) {
        std::cout << e.what() << std::endl;
    }

    try {
        af::setDevice(0);
        af::info();
        af::sync();

        bench();
    } catch (af::exception &e) {
        std::cout << e.what() << std::endl;
    }
    return 0;
}
```

### LINKED ISSUE #1536: Working example kernel
It would be very convenient if there were a working example kernel, say to compute `c = a + b`, where `a, b, c` are each an `array`, instead of the empty shell function (`src/backend/opencl/kernel/example.cl`) that exists now. Would be much easier for users to modify that, and implement custom algos as kernels.

Thanks!

9prady9:
[skip arrayfire ci]
- [ ] Fixes #1475 
- [x] Fixes #1536
- [x] Add shared memory check for bilateral function in cuda/opencl backends

### DIFF
--- a/include/af/util.h
+++ b/include/af/util.h
@@ -128,7 +128,7 @@ namespace af
 #endif
 
     // Purpose of Addition: "How to add Function" documentation
-    AFAPI array exampleFunction(const array& a, const array& b, const af_someenum_t param);
+    AFAPI array exampleFunction(const array& in, const af_someenum_t param);
 
 #if AF_API_VERSION >= 34
     ///
@@ -253,7 +253,7 @@ extern "C" {
 
     // Purpose of Addition: "How to add Function" documentation
     AFAPI af_err af_example_function(af_array* out,
-                                     const af_array a, const af_array b,
+                                     const af_array in,
                                      const af_someenum_t param);
 
     ///
--- a/include/af/util.h
+++ b/include/af/util.h
@@ -128,7 +128,7 @@ namespace af
 #endif
 
     // Purpose of Addition: "How to add Function" documentation
-    AFAPI array exampleFunction(const array& in, const af_someenum_t param);
+    AFAPI array exampleFunction(const array& a, const array& b, const af_someenum_t param);
 
 #if AF_API_VERSION >= 34
     ///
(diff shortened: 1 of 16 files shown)


pavanky:
@9prady9 are you doing any more work on this ?

9prady9:
@pavanky Yes, i was last week but didn't have any break through so far as to why OpenCL kernel is out performing CUDA kernel even though both are identical in algorithm.

pavanky:
@9prady9 change opencl to use `native_exp` for consistency

pavanky:
@9prady9 what is the performance comparison now ?

include/af/util.h
```diff
@@ -128,7 +128,7 @@ namespace af
 #endif
 
     // Purpose of Addition: "How to add Function" documentation
-    AFAPI array exampleFunction(const array& in, const af_someenum_t param);
+    AFAPI array exampleFunction(const array& a, const array& b, const af_someenum_t param);
```
pavanky: This breaks API! Never do this.

9prady9: This function doesn't serve any other purpose than tutorial help. Shall i revert it back and just use single input array to do `a+a` style operation ?

9prady9:
@pavanky Well, after changing opencl to native_exp, it is back to square one.

include/af/util.h
pavanky: Even if it servers no purpose, we should not be breaking API

pavanky:
@9prady9 The problem may be because of the compiler. I re-wrote the code to use fewer registers on both CUDA and OpenCL. Now both of them seem to take the same amount of time. I am merging your changes into my branch and sending a new PR.

PR #1501, opened 2016-07-14
FIX: find_backend(Unified) in build tree ArrayFireConfig.cmake file

schuhschuh:
The `ArrayFireUnified.cmake` file with the exported Unified backend targets to be imported by `ArrayFireConfig.cmake` file are not found when using the build tree directly because it is located in a different path then the other backend target export files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,7 @@ ENDIF(FORGE_FOUND AND NOT USE_SYSTEM_FORGE)
 ## configuration to be used from the binary directory directly
 SET(INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 SET(BACKEND_DIR "src/backend/\${lowerbackend}")
+SET(UNIFIED_DIR "src/api/unified")
 CONFIGURE_FILE(
     ${CMAKE_MODULE_PATH}/ArrayFireConfig.cmake.in
     ${CMAKE_CURRENT_BINARY_DIR}/ArrayFireConfig.cmake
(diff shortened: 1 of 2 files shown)


schuhschuh:
I mean, in my own project, I specify `ArrayFire_DIR` to be the build directory of the ArrayFire project configured with CMake. No `make install` step.

schuhschuh:
BTW: Another solution would have been to change the following line in `src/api/unified/CMakeLists.txt`:

```
EXPORT(TARGETS af FILE ArrayFireUnified.cmake)
```
to
```
EXPORT(TARGETS af FILE ${ARRAYFIRE_BINARY_DIR}/src/backend/unified/ArrayFireUnified.cmake)
```

Not sure if that's what you prefer.

pavanky:
I don't know if the second solution will break the "installed" versions.


Can you check both of your solutions to see if either or both work when you "install" ?

schuhschuh:
Neither has an effect on the installed version because the above EXPORT command only affects the file that is written to the build tree. It is different from `install(EXPORT)` used to create the file for the installation. The change of ArrayFireConfig.cmake.in also does not affect the installation because in this case `BACKEND_DIR` and `UNIFIED_DIR` are both empty strings and thus the `targetFile` path is identical in if and else branch in `find_backend`.

schuhschuh:
I'll test the installation with this PR change. Unfortunately, ArrayFire is always completely rebuild whenever CMake is run again even if nothing changed... guess a side effect of using `file(GLOB)` although CMake advocates [against it](http://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil).

pavanky:
I don't think the complete rebuild is because of globbing, it is possible because of few generated files.

pavanky:
rebased into https://github.com/arrayfire/arrayfire/pull/1526

Issue #1505, opened 2016-07-15
Interop with thrust::device_vector


pavanky:
Implement this in `af/cuda.h`

pavanky:
being handled here: https://github.com/arrayfire/arrayfire-extras

pavanky:
https://github.com/arrayfire/arrayfire-extras/issues/3

Issue #1504, opened 2016-07-15
Add support for STL containers


pavanky:
The following type of functionality needs to be supported:

Getting data from an af::array:
```cpp
af::array a = af::randu(5,5);
 // Performed by having an "operator std::vector<T>" method in af::array class
std::vector<float> ha1 = a;
std::valarray<float> ha2 = a;
```

Creating af::array:
```cpp
std::vector<float> ha1{1,2,3,4,5};
std::vallarray<float> ha2{2,4,6};
af::array a1 = ha1; // overload = operator
af::array a2 = ha2; // overload = operator
```

pavanky:
Being handled here: 
https://github.com/arrayfire/arrayfire-extras

pavanky:
https://github.com/arrayfire/arrayfire-extras/issues/2

Issue #1503, opened 2016-07-14
Explicit instantiation for array::device<void>()


schuhschuh:
Minor request following up on the Gitter chat with @pavanky regarding the template argument needed to get a device memory pointer, I wonder if it would be reasonable to add an explicit instantiation of the function for `T=void`?

Would be more explicit than `void *ptr = a.device<char>()`, suggesting the template type would actually match the unknown data type. Also, when calling a function overload specific to `void *` where another function specialization may handle the `char *` differently, one cannot just pass `a.device<char>()` as argument without an explicit cast to avoid the wrong overload to be called. None of this is critical and can obviously dealt with easily on the client code side. But I was just wondering if it wouldn't be easy to allow `a.device<void>()`, document it, and provide the needed template specialization for the linker?

Issue #529, opened 2015-03-23
Support OpenCV interoperability


pavanky:
The functions will go under `af/opencv.h`. This file will NOT be included by default (in arrayfire.h). The users will have to include this file when they are trying to interop with OpenCV. The definitions and implementations will reside in the header file as static functions.

The functions from here can be used for reference:
https://github.com/arrayfire-community/arrayfire_opencv

@mcclanahoochie If you have the time, I will assign this to you. Please note the following requirements.

Only the following functions need to be implemented:

```
// conversion for gpu
void mat_to_array(const cv::Mat& input, af::array& output, bool transpose = true) ;
void mat_to_array(const vector<Mat>& input, af::array& output, bool transpose = true);
af::array mat_to_array(const cv::Mat& input, bool transpose = true) ;


// conversion for cpu
void array_to_mat(const af::array& input, cv::Mat& output, int type = CV_32F, bool transpose = true) ;
Mat array_to_mat(const af::array& input, int type = CV_32F, bool transpose = true) ;
```

pavanky:
CUDA and OpenCL arrays from OpenCV will need to be supported before this issue is closed.

pavanky:
Being handled here: https://github.com/arrayfire/arrayfire-extras/issues/4

PR #1512, opened 2016-07-21
adds stl, thrust, opencv interop functions

syurkevi:
Addresses interop issues: #1504 , #1505, #529 
Rider issue #1503 .

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -12,10 +12,57 @@
 #include <af/seq.h>
 #include <af/util.h>
 #include <af/index.h>
+#include <af/exception.h>
 
 #ifdef __cplusplus
 #include <af/traits.hpp>
 #include <vector>
+#ifdef AF_ENABLE_STL_CAST
+#include <valarray>
+#endif
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    /**
+        \ingroup construct_mat
+        @{
+    */
+    /**
+       Create an \ref af_array handle initialized with user defined data
+
+       This function will create an \ref af_array handle from the memory provided in \p data
+
+       \param[out]  arr The pointer to the returned object.
+       \param[in]   data The data which will be loaded into the array
+       \param[in]   ndims The number of dimensions read from the \p dims parameter
+       \param[in]   dims A C pointer with \p ndims elements. Each value represents the size of that dimension
+       \param[in]   type The type of the \ref af_array object
+
+       \returns \ref AF_SUCCESS if the operation was a success
+    */
+    AFAPI af_err af_create_array(af_array *arr, const void * const data, const unsigned ndims, const dim_t * const dims, const af_dtype type);
+    /**
+      @}
+    */
+
+    /**
+       \ingroup method_mat
+       @{
+    */
+    /**
+       \brief Reduce the reference count of the \ref af_array
+    */
+    AFAPI af_err af_release_array(af_array arr);
+    /**
+      @}
+    */
+
+#ifdef __cplusplus
+}
+#endif
+
 namespace af
 {
 
@@ -682,6 +729,64 @@ namespace af
            @}
         */
 
+#ifdef AF_ENABLE_STL_CAST
+        template<typename T> array& operator=(const std::vector<T> &other)
+        {
+            //TODO: Unsafe. loses data if af_weak_copy fails
+            if(this->arr != 0) {
+                af_err err = af_release_array(this->arr);
+                if(err != AF_SUCCESS) { throw af::exception(err); }
+            }
+
+            af_array temp = 0;
+            dim_t sz = other.size();
+            af_dtype type = (af_dtype)dtype_traits<T>::af_type;
+            af_err err = af_create_array(&temp, (const void*)&other[0], 1, &sz, type);
+            if(err != AF_SUCCESS) { throw af::exception(err); }
+            this->arr = temp;
+            return *this;
+        }
+
+        template<typename T> array& operator=(std::valarray<T> &other)
+        {
+            //TODO: Unsafe. loses data if af_weak_copy fails
+            if(this->arr != 0) {
+                af_err err = af_release_array(this->arr);
+                if(err != AF_SUCCESS) { throw af::exception(err); }
+            }
+
+            af_array temp = 0;
+            dim_t sz = other.size();
+            af_dtype type = (af_dtype)dtype_traits<T>::af_type;
+            af_err err = af_create_array(&temp, &other[0], 1, &sz, type);
+            if(err != AF_SUCCESS) { throw af::exception(err); }
+            this->arr = temp;
+            return *this;
+        }
+
+        template<typename T> operator std::vector<T>() {
+            std::vector<T> out(this->elements());
+            af::dtype oType = (af::dtype)dtype_traits<T>::af_type;
+            if(type() != oType) {
+                this->as(oType).host(&out[0]);
+            } else {
+                this->host(&out[0]);
+            }
+            return out;
+        }
+
+        template<typename T> operator std::valarray<T>() {
+            std::valarray<T> out(this->elements());
+            af::dtype oType = (af::dtype)dtype_traits<T>::af_type;
+            if(type() != oType) {
+                this->as(oType).host(&out[0]);
+            } else {
+                this->host(&out[0]);
+            }
+            return out;
+        }
+#endif
+
         // INDEXING
         // Single arguments
 
@@ -1275,21 +1380,6 @@ extern "C" {
     */
 
     /**
-       Create an \ref af_array handle initialized with user defined data
-
-       This function will create an \ref af_array handle from the memory provided in \p data
-
-       \param[out]  arr The pointer to the returned object.
-       \param[in]   data The data which will be loaded into the array
-       \param[in]   ndims The number of dimensions read from the \p dims parameter
-       \param[in]   dims A C pointer with \p ndims elements. Each value represents the size of that dimension
-       \param[in]   type The type of the \ref af_array object
-
-       \returns \ref AF_SUCCESS if the operation was a success
-    */
-    AFAPI af_err af_create_array(af_array *arr, const void * const data, const unsigned ndims, const dim_t * const dims, const af_dtype type);
-
-    /**
        Create af_array handle
 
        \param[out]  arr The pointer to the retured object.
@@ -1324,11 +1414,6 @@ extern "C" {
        Needs to used in conjunction with the two functions above
     */
     AFAPI af_err af_get_data_ptr(void *data, const af_array arr);
-
-    /**
-       \brief Reduce the reference count of the \ref af_array
-    */
-    AFAPI af_err af_release_array(af_array arr);
 
     /**
        Increments an \ref af_array reference count
--- a/include/af/cuda.h
+++ b/include/af/cuda.h
@@ -12,6 +12,11 @@
 #include <cuda.h>
 #include <cuda_runtime.h>
 
+#include <thrust/host_vector.h>
+#include <thrust/device_vector.h>
+#include <arrayfire.h>
+#include <af/traits.hpp>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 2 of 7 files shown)


include/af/array.h
```diff
@@ -12,10 +12,57 @@
...
+    */
+
+    /**
+       \ingroup method_mat
+       @{
+    */
+    /**
+       \brief Reduce the reference count of the \ref af_array
+    */
+    AFAPI af_err af_release_array(af_array arr);
```
umar456: why are these moved?

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+/*
+ * arrayfire -> opencv
+ */
+// af::array conversion to cv::Mat
+cv::Mat toMat(const af::array& input, int type = CV_32F, bool transpose = true);
+
+
+//helper conversion function
+void copy_data(Mat& input, af::array& output, bool transpose) {
```
umar456: I think this will throw an error if included in multiple files in the same binary

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
+#pragma once
```
umar456: The file should be `opencv.h`

include/af/opencv3.hpp
```diff
@@ -0,0 +1,160 @@
+#pragma once
```
umar456: rename to `opencv3.h`

include/af/opencv2.hpp
syurkevi: Only one should be included at a time. Easy workaround would be to use:
#include "opencv2/core/version.hpp"
#if CV_MAJOR_VERSION == 2
#include opencv2.hpp
#elif CV_MAJOR_VERSION == 3
#include opencv3.hpp
#endif

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
...
+#include <vector>
+#include <opencv2/core/core.hpp>
+#include <arrayfire.h>
+
+namespace afcv
```
umar456: i think we should have a subnamespace instead of this.
i.e.
namespace af {
namespace opencv {
}
}

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
...
+{
+
+/*
+ * opencv -> arrayfire
```
umar456: Document

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
...
+ */
+// conversion for cv::Mat
+af::array array(const std::vector<cv::Mat>& input, bool transpose = true);
+af::array array(const cv::Mat& input, bool transpose = true);
+
+/*
+ * arrayfire -> opencv
```
umar456: document

include/af/array.h
syurkevi: They are used in the new assignment operators and need to be defined before since the new operators are templated.

include/af/opencv2.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * opencv -> arrayfire
+ */
+
+// conversion for cv::Mat
+af::array array(const cv::Mat& input, bool transpose)
+{
+    af::array output;
+    if (input.empty()) { return output; }
+    cv::Mat tmp;
+    if (input.channels() == 1)
```
umar456: will this only work for float types?

include/af/opencv3.hpp
```diff
@@ -0,0 +1,160 @@
...
+        int* data = output.ptr<int>(0);
+        input_.as(s32).host((void*)data);
+    } else if (type == CV_64F) {
+        double* data = output.ptr<double>(0);
+        input_.as(f64).host((void*)data);
+    } else if (type == CV_8U) {
+        uchar* data = output.ptr<uchar>(0);
+        input_.as(b8).host((void*)data);
+    } else {
+        throw std::runtime_error(string("array to mat error "));
```
umar456: Use af::exception instead

src/api/cpp/array.cpp
```diff
@@ -1001,6 +1002,17 @@ af::dtype implicit_dtype(af::dtype scalar_type, af::dtype array_type)
...
 
+#define INSTANTIATE(T)                                              \
+    template<> AFAPI T* array::device() const                       \
+    {                                                               \
+        void *ptr = NULL;                                           \
+        AF_THROW(af_get_device_ptr(&ptr, get()));                   \
+        return (T *)ptr;                                            \
+    }
+
+INSTANTIATE(void)
```
umar456: seems unnecessary to make a macro for one type

umar456:
I don't see tests for these functions

include/af/array.h
```diff
@@ -682,6 +729,64 @@ namespace af
...
+            af_array temp = 0;
+            dim_t sz = other.size();
+            af_dtype type = (af_dtype)dtype_traits<T>::af_type;
+            af_err err = af_create_array(&temp, &other[0], 1, &sz, type);
+            if(err != AF_SUCCESS) { throw af::exception(err); }
+            this->arr = temp;
+            return *this;
+        }
+
+        template<typename T> operator std::vector<T>() {
```
umar456: I really really really don't want this function in arrayfire. It will allow the following code to work:

```
vector<float> my_funciton(vector<float>& foo) {
  for(auto &b : foo) {
    b++;
  }
  return foo;
}
int main () {
  af::array bar = randu(10, 10);
  bar = my_function(bar);
  return 0;
}
```

include/af/opencv2.hpp
umar456: I mean if you included this specific header into two or more cpp source files then you will encounter a multiple definition error because this is not a template function

include/af/opencv2.hpp
pavanky: This is not consistent with what we are doing with other interop namespaces.

src/api/cpp/array.cpp
pavanky: We have the macro defined already some other place. Use that instead.

include/af/opencv2.hpp
pavanky: @syurkevi Umar is right, make the functions static.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+/**
+   Interop function that converts a thrust::device_vector to an af::array
+
+   \param[in] tvec thrust::device_vector to be converted
+   \return af::array with converted contents of the thrust vector
+
+   \ingroup cuda_mat
+ */
+template<typename T>
+static af::array array(thrust::device_vector<T> &tvec)
```
pavanky: Templated functions need not be static.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+{
+    af::array output;
+    if(tvec.empty()) return output;
+    T* ptr = thrust::raw_pointer_cast(tvec.data());
+    af::array tmp = af::array(tvec.size(), ptr, afDevice);
+    output = tmp.copy();
+    tmp.lock();
```
pavanky: Don't perform a copy by default. Have a flag that is set to false and copy when the user requests it.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+   \return af::array with converted contents of the thrust vector
+
+   \ingroup cuda_mat
+ */
+template<typename T>
+static af::array array(thrust::host_vector<T> &tvec)
+{
+    af::array output;
+    if(tvec.empty()) return output;
+    //af_dtype type = (af_dtype)dtype_traits<T>::af_type;
```
pavanky: Remove this.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+   \param[in] arr af::array to be converted to thrust::host_vector
+   \return thrust::host_vector with converted contents of af::array
+
+   \ingroup cuda_mat
+ */
+template<typename T>
+static thrust::host_vector<T> toHostVector(af::array &arr)
+{
+    af_dtype vec_type = (af_dtype)af::dtype_traits<T>::af_type;
+    if(arr.type() != vec_type) { throw af::exception("Thrust vector data type and array data type are mismatching"); }
```
pavanky: throw in new line.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+    //todo check continuity
+    thrust::host_vector<T> hvec(arr.elements());
+    T* h_ptr = thrust::raw_pointer_cast(hvec.data());
+    arr.host(h_ptr);
```
pavanky: `.host()` does the proper transformations to make it linear. You don't need to check for continuity.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+ */
+template<typename T>
+static thrust::device_vector<T> toDeviceVector(af::array &arr)
+{
+    af_dtype vec_type = (af_dtype)af::dtype_traits<T>::af_type;
+    if(arr.type() != vec_type) { throw af::exception("Thrust vector data type and array data type are mismatching"); }
+    arr.lock();
+    //todo check continuity
+    thrust::device_ptr<T> af_ptr = thrust::device_pointer_cast(arr.device<T>());
+    thrust::device_vector<T> dvec(af_ptr, af_ptr + arr.elements());
```
pavanky: Same with `.device`

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+   \param[in] arr af::array to be converted to thrust::device_vector
+   \return thrust::device_vector with converted contents of af::array
+
+   \ingroup cuda_mat
+ */
+template<typename T>
+static thrust::device_vector<T> toDeviceVector(af::array &arr)
+{
+    af_dtype vec_type = (af_dtype)af::dtype_traits<T>::af_type;
+    if(arr.type() != vec_type) { throw af::exception("Thrust vector data type and array data type are mismatching"); }
```
pavanky: throw in new line.

include/af/cuda.h
```diff
@@ -117,5 +122,93 @@ static inline void setNativeId(int nativeId)
...
+    arr.lock();
+    //todo check continuity
+    thrust::device_ptr<T> af_ptr = thrust::device_pointer_cast(arr.device<T>());
+    thrust::device_vector<T> dvec(af_ptr, af_ptr + arr.elements());
+    arr.unlock();
```
pavanky: `unlock` needs to be called when `device_vector` goes out of scope. We can do this but it would require a little bit more work.

include/af/array.h
```diff
@@ -682,6 +729,64 @@ namespace af
            @}
         */
 
+#ifdef AF_ENABLE_STL_CAST
+        template<typename T> array& operator=(const std::vector<T> &other)
+        {
+            //TODO: Unsafe. loses data if af_weak_copy fails
+            if(this->arr != 0) {
+                af_err err = af_release_array(this->arr);
+                if(err != AF_SUCCESS) { throw af::exception(err); }
```
pavanky: Have a message in the exception.

include/af/array.h
```diff
@@ -682,6 +729,64 @@ namespace af
...
+            }
+
+            af_array temp = 0;
+            dim_t sz = other.size();
+            af_dtype type = (af_dtype)dtype_traits<T>::af_type;
+            af_err err = af_create_array(&temp, &other[0], 1, &sz, type);
+            if(err != AF_SUCCESS) { throw af::exception(err); }
```
pavanky: Have a message in the exception.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+   \returns an cv::Mat with the contents of the input af::array
+*/
+static cv::Mat toMat(const af::array& input, int type = CV_32F, bool transpose = true);
+/**
+  @}
+*/
+
+
+//helper conversion function
+static void copy_data(cv::Mat& input, af::array& output, bool transpose) {
```
pavanky: just `copy` is fine. Set `transpose=false` as default value.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+/**
+   This function will copy data from an \ref af::array to a cv::Mat
+
+   \param[in]  input The af::array from which data which will be loaded into cv::Mat
+   \param[in]  type The resulting datatype of the output cv::Mat
+   \param[in]  transpose flag specifying if the input should be transposed during copy
+
+   \returns an cv::Mat with the contents of the input af::array
+*/
+static cv::Mat toMat(const af::array& input, int type = CV_32F, bool transpose = true);
```
pavanky: No reason to separate definition and declaration.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+  @}
+*/
+
+
+//helper conversion function
+static void copy_data(cv::Mat& input, af::array& output, bool transpose) {
+    const unsigned w = input.cols;
+    const unsigned h = input.rows;
+    const unsigned nchannels = input.channels();
+    if (nchannels > 4) { throw af::exception("mat to array error "); }
```
pavanky: Explain why it is failing. The message does not say much.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+    if (input.channels() == 1)
+        { input.convertTo(tmp, CV_32F); }
+    else if (input.channels() == 2)
+        { input.convertTo(tmp, CV_32FC2); }
+    else if (input.channels() == 3)
+        { input.convertTo(tmp, CV_32FC3); }
+    else if (input.channels() == 4)
+        { input.convertTo(tmp, CV_32FC4); }
+
+    copy_data(tmp, output, transpose);
```
pavanky: There are several things wrong here.

- copy_data is only used once in this file.
- It should also not be used by the user because it performs no checks.
- The function is not actually performing a copy. It is creating a new output array.

Just move all the relevant code here.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+    const int channels = input.dims(2);
+    int ndims = input.numdims();
+    af::array input_;
+    if (transpose) {
+        if(channels == 1) {
+            input_ = input;
+        } else {
+            input_ = af::constant(0, channels, input.dims(1), input.dims(0));
+            for(int ii=0; ii<channels; ++ii) {
+                input_(ii, af::span, af::span) = af::moddims(input(af::span, af::span, ii), 1, input.dims(1), input.dims(0));
```
pavanky: Use multiple lines for RHS.

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+        output = cv::Mat(input.dims(1), input.dims(0), CV_MAKETYPE(type, channels));
+    } else {
+        if(channels == 1) {
+            input_ = input.T();
+        } else {
+            input_ = af::constant(0, channels, input.dims(0), input.dims(1));
+            for(int ii=0; ii<channels; ++ii) {
+                input_(ii, af::span, af::span) = af::moddims(input(af::span, af::span, ii).T(), 1, input.dims(0), input.dims(1));
```
pavanky: Use multiple lines for RHS

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+    int ndims = input.numdims();
+    af::array input_;
+    if (transpose) {
+        if(channels == 1) {
+            input_ = input;
+        } else {
+            input_ = af::constant(0, channels, input.dims(1), input.dims(0));
+            for(int ii=0; ii<channels; ++ii) {
+                input_(ii, af::span, af::span) = af::moddims(input(af::span, af::span, ii), 1, input.dims(1), input.dims(0));
+            }
```
pavanky: This is just a reorder operation. Remove for loop. `array input_ = af::reorder(input, 2, 1, 0);`

include/af/opencv3.h
```diff
@@ -0,0 +1,191 @@
...
+        }
+        output = cv::Mat(input.dims(1), input.dims(0), CV_MAKETYPE(type, channels));
+    } else {
+        if(channels == 1) {
+            input_ = input.T();
+        } else {
+            input_ = af::constant(0, channels, input.dims(0), input.dims(1));
+            for(int ii=0; ii<channels; ++ii) {
+                input_(ii, af::span, af::span) = af::moddims(input(af::span, af::span, ii).T(), 1, input.dims(0), input.dims(1));
+            }
```
pavanky: This is just a reorder operation. Remove for loop. `array input_ = af::reorder(input, 2, 0, 1);`

include/af/opencv.h
```diff
@@ -0,0 +1,190 @@
+#pragma once
+#include <vector>
+#include <opencv2/core/core.hpp>
+#include <arrayfire.h>
+
```
pavanky: You include opencv2 both times.

pavanky:
Moving to https://github.com/arrayfire/arrayfire-extras

PR #1464, opened 2016-06-19
Build time improvements for Scan By Key

shehzan10:
* Fix cl2hpp dependency for scan compilation
* Split scan*.cu files
* Remove dim template from scan_dim

PR #1524, opened 2016-07-29
Minor fixes

shehzan10:
Selected commits from #1464 (Removed scan by key commits)

[skip arrayfire ci]


shehzan10:
@pavanky you can pull these into #1526 as well.

PR #1525, opened 2016-07-29
Avoid instantiating templates for sort direction

umar456:
This PR moves the isAscending template parameter into the function parameter. This should reduce the compile times and binary sizes. It should also not impact the performance of the operation.


pavanky:
Using the commit in this PR: https://github.com/arrayfire/arrayfire/pull/1526

PR #1534, opened 2016-08-05
Performance improvements to CUDA JIT

pavanky:
- Call NVVM / libdevice functions when possible
- Updates to JIT kernel generation 
- Performance within 2-8 percent of OpenCL

[skip arrayfire ci]


shehzan10:
build arrayfire windows ci

src/backend/cuda/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ FIND_PACKAGE(Boost REQUIRED)
 INCLUDE("${CMAKE_MODULE_PATH}/CLKernelToH.cmake")
 INCLUDE("${CMAKE_MODULE_PATH}/FindNVVM.cmake")
 
+OPTION(USE_LIBDEVICE "Use libdevice for CUDA JIT" ON)
```
shehzan10: I don't think `USE_LIBDEVICE` says anything about the variable without using the description. I would suggest either `BUILD_CUDA_WITH_LIBDEVICE` or `USE_CUDA_LIBDEVICE` or similar. Something that suggests it is associated with JIT in CUDA.

src/backend/cuda/CMakeLists.txt
```diff
@@ -314,6 +316,47 @@ CL_KERNEL_TO_H(
     EOF "1"
     )
 
+SET(libdevice_bc "")
+IF (USE_LIBDEVICE)
+  SET(libdevice_computes "")
+  LIST(APPEND libdevice_computes "20" "30" "35" "50")
```
shehzan10: Use CMake LIST `COMPUTES_DETECTED_LIST` here. This includes both detected computes as well as manually set computes.

src/backend/cuda/CMakeLists.txt
```diff
@@ -314,6 +316,47 @@ CL_KERNEL_TO_H(
...
+    SOURCES ${libdevice_bc}
+    VARNAME libdevice_files
+    EXTENSION "hpp"
+    OUTPUT_DIR ${libdevice_headers}
+    TARGETS libdevice_targets
+    NAMESPACE "cuda"
+    EOF "1"
+    )
+
+  MESSAGE(STATUS "LIBDEVICE found.")
```
shehzan10: As suggested before, prefix CUDA in these variables/messages.

src/backend/cuda/CMakeLists.txt
```diff
@@ -314,6 +316,47 @@ CL_KERNEL_TO_H(
...
+  ADD_DEFINITIONS(-DUSE_LIBDEVICE)
+ELSE()
+  MESSAGE(STATUS "LIBDEVICE not found on system. CUDA JIT may be slower")
```
shehzan10: I don't think "CUDA JIT may be slower" succinctly talks about the performance. I think it may cause confusion over what "slower" is, which in fact is not actually slow, ie. there is no performance degradation compared to previous versions of ArrayFire (as I understand it at least).

shehzan10:
Will these changes mean drop in support for the tegra devices? I'm assuming it's still going to be supported.

src/backend/cuda/CMakeLists.txt
pavanky: libjit doesn't have all computes.

src/backend/cuda/CMakeLists.txt
pavanky: It is slower than using libdevice. Sometimes by as much as 2-3x

pavanky:
@shehzan10 This should still work with both tegras.

pavanky:
@shehzan10 can you kick off windows tests after the machine is up ?

shehzan10:
build arrayfire windows ci

shehzan10:
[Windows](http://ci.arrayfire.org/job/arrayfire-windows/job/pr-build/492/console) fails with the same errors as linux was.

shehzan10:
Can one of @arrayfire/core-devel run the CUDA tests on an OSX machine?

shehzan10:
build arrayfire windows ci

umar456:
Everything is building and compiling on OSX.

PR #1542, opened 2016-08-12
Use compute library from Boost 1.61.

ghisvail:
With the availability of Boost 1.61, usage of a separate Boost-Compute dependency is no longer required. Here is the patch I applied to make this happen for Debian, whilst keeping the old behavior for earlier versions of Boost.


shehzan10:
We have had an internal discussion about this, and decided to no do this for version 1.61.
The reason being that the components of compute::sort functions that are being used by ArrayFire either had bugs or are feature incomplete at v1.61. Hence we decided to stick to our own builds/tags to avoid the failures.

In case you have applied this patch and seen sort/sort_index/sort_by_key OpenCL tests pass, then we'll look into it once again.

pavanky:
@ghisvail as shehzan mentioned, this patch will make sure arrayfire compiles but it does not guarantee all tests will pass. 

@shehzan10 This patch only optionally adds support for using System boost compute. I think we should accept the PR but add a notice saying that certain functions might be broken.

shehzan10:
@pavanky the problem with adding a message is that although the compiler person/system might see this, it wont be seen by the end users. Unless we add some message there (using #if or something), I'm afraid it won't be enough.

ghisvail:
> @pavanky the problem with adding a message is that although the compiler person/system might see this, it wont be seen by the end users.

Same thing happens to other libraries in the system. Take the packaged BLAS libraries of your distro of choice, they are likely to be sub-optimally configured for *your* specific machine. So, the end-user who seeks optimality should know he or she has to build AF and its dependencies from source anyway or perhaps use your installers.

> add a notice saying that certain functions might be broken.

Consequently, if such disclaimer is required because a patched version of `clBLAS` / `clFFT` / `clSPARSE` / `Boost Compute` is recommended, then it should be displayed during the configuration.

pavanky:
> Consequently, if such disclaimer is required because a patched version of clBLAS / clFFT / clSPARSE / Boost Compute is recommended, then it should be displayed during the configuration.

That's my point really. We have been doing this already with clBLAs and clFFT. We should just warn the maintainer and let them decide what to do.

pavanky:
@ghisvail Sorry there was some confusion on our end. We thought we were using a non release version as a dependency in the devel branch. 

It looks like 1.61 release should work just fine with our current code base :+1:

pavanky:
@ghisvail 

That said I am worried about disabling `build_boost_compute.cmake` all together if 1.61 or greater is found.

There may be cases where we would want to build with a particular version of boost compute and this patch disables that ability. 

We need to figure out a way to have the option to build with custom versions of boost compute as well.

ghisvail:
I see what you mean.

Another solution, now that Boost 1.61 is out, the `USE_SYSTEM_BOOST_COMPUTE` should attempt to use the `compute` library from Boost if `Boost_VERSION >= 1.61`, otherwise find a system installed `Boost.Compute` as it did before.

This way the default path is still to use your embedded version regardless whether Boost 1.61 is available or not. I can update my PR is you are happy with this scenario.

pavanky:
@ghisvail Just so we are on the same page, is this what you mean:

1) `USE_SYSTEM_BOOST_COMPUTE` is `OFF` -- uses `build_boost_compute.cmake`
2) `USE_SYSTEM_BOOST_COMPUTE` is `ON` 
  a) If Boost_VERSION >= 1.61 use the system Boost install.
  b) else use system installed Boost_Compute.

If yes, go ahead and update your patch to reflect this :+1:

ghisvail:
That is correct.

ghisvail:
This should do it.

PR #1545, opened 2016-08-16
dilate, erode wrong result when the filter center is not set

vakopian:
The current implementation of `morph` (which is used for `dilate` and `erode` functions) seems to assume that the center of the filter is always set. This happens in [backend/cpu/kernel/morph.hpp:40](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cpu/kernel/morph.hpp#L40) and [backend/cuda/kernel/morph.hpp:105](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cuda/kernel/morph.hpp#L105).
In contrast, matlab's `imdilate` and `imerode` makes no such assumption and (for floating types) initializes the value to `Inf` for `erode` and to `-Inf` for `dilate`.
For example in matlab ` imdilate(ones(3, 3), zeros(3, 3))` would return 3x3 all `-Inf`, whereas arrayfire will return 3x3 all 1s.

### DIFF
--- a/src/backend/cpu/kernel/morph.hpp
+++ b/src/backend/cpu/kernel/morph.hpp
@@ -9,6 +9,7 @@
 
 #pragma once
 #include <af/defines.h>
+#include <limits>
 #include <Array.hpp>
 #include <utility.hpp>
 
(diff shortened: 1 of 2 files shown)


vakopian:
pls see https://github.com/arrayfire/arrayfire/pull/1546

PR #1546, opened 2016-08-16
dilate, erode wrong result when the filter center is not set

vakopian:
The current implementation of `morph` (which is used for `dilate` and `erode` functions) seems to assume that the center of the filter is always set. This happens in [backend/cpu/kernel/morph.hpp:40](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cpu/kernel/morph.hpp#L40) and [backend/cuda/kernel/morph.hpp:105](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cuda/kernel/morph.hpp#L105).
In contrast, matlab's `imdilate` and `imerode` makes no such assumption and (for floating types) initializes the value to `Inf` for `erode` and to `-Inf` for `dilate`.
For example in matlab ` imdilate(ones(3, 3), zeros(3, 3))` would return 3x3 all `-Inf`, whereas arrayfire will return 3x3 all 1s.

### DIFF
--- a/src/backend/cpu/kernel/morph.hpp
+++ b/src/backend/cpu/kernel/morph.hpp
@@ -38,7 +38,7 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
                 // j steps along 2nd dimension
                 for(dim_t i=0; i<dims[0]; ++i) {
                     // i steps along 1st dimension
-                    T filterResult = isDilation ? (std::is_integral<T>::value ? std::numeric_limits<T>::lowest() : -std::numeric_limits<T>::infinity())
+                    T filterResult = IsDilation ? (std::is_integral<T>::value ? std::numeric_limits<T>::lowest() : -std::numeric_limits<T>::infinity())
 			: (std::is_integral<T>::value ? std::numeric_limits<T>::max() : std::numeric_limits<T>::infinity());
 
                     // wj,wi steps along 2nd & 1st dimensions of filter window respectively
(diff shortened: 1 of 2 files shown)


pavanky:
There is no reason for us to reproduce Matlab's behavior.

If there is an algorithm reasoning for returning infinities, we can consider adding that as an option.

vakopian:
this is not about the infinity case, that was just an illustration of the problem. If you use an off-centered filter like this
```
0 0 0
1 0 0
0 0 0
```
the result is just wrong (irrespective of infinity)

vakopian:
here is a full example:
```c++
using namespace af;
array in = constant(0, 4, 4);
in(1, 1, 0, 0) = 1;
print("in", in);
array filter = af::constant(0, 3, 3);
filter(1, 0, 0, 0) = 1;
print("filter", filter);
array out = dilate(in, filter);
print("dilate", out);
```
this produces
```
in
[4 4 1 1]
    0.0000     0.0000     0.0000     0.0000 
    0.0000     1.0000     0.0000     0.0000 
    0.0000     0.0000     0.0000     0.0000 
    0.0000     0.0000     0.0000     0.0000 

filter
[3 3 1 1]
    0.0000     0.0000     0.0000 
    1.0000     0.0000     0.0000 
    0.0000     0.0000     0.0000 

dilate
[4 4 1 1]
    0.0000     0.0000     0.0000     0.0000 
    0.0000     1.0000     1.0000     0.0000 
    0.0000     0.0000     0.0000     0.0000 
    0.0000     0.0000     0.0000     0.0000 
```
which is wrong, as there should only be a single `1` in the result.

pavanky:
@vakopian Thanks for the example.

The fix for CPU looks fine. I am not sure if the CUDA kernels can accept STL values like that. The OpenCL version is not fixed at all.


I can merge your PR into my local branch and fix the issues I raised here.

vakopian:
@pavanky: Please do. I was not sure how to fix OpenCL at all (the code is very different there). For CUDA, I thought the template resolution would happen before the kernel takes over, but again, please fix as necessary. Thanks

pavanky:
Pulled in changes from this PR into #1547

pavanky:
@vakopian I accidentally squashed your commit along with mine, so I reverted the squash and sent in your commit separately. Sorry about that.

PR #1550, opened 2016-08-20
optimize morph by pre-computing the mask linear offsets

vakopian:
As discussed in https://github.com/arrayfire/arrayfire/pull/1547#issuecomment-241145933, this patch is to optimize the algorithm for `erode` and `dilate`. Before this patch, the following code took about 110 seconds, while after the patch it takes about 4 seconds.

```
in = randn(3000, 3000);
filter = constant(1, 21, 21);
dilate(in, filter);
```

This change assumes that the input and output arrays have the same dimensions, which allows to use the linear indexing for both input and output batches. The way `morph` is currently being used for `erode` and `dilate`, this assumption is definitely satisfied.

Only the 2D version of morph is implemented here, mainly to show the idea.

### DIFF
--- a/src/backend/cpu/kernel/morph.hpp
+++ b/src/backend/cpu/kernel/morph.hpp
@@ -12,7 +12,7 @@
 #include <Array.hpp>
 #include <utility.hpp>
 #include <ops.hpp>
-#include <cassert>
+#include <kernel/pad.hpp>
 
 namespace cpu
 {
@@ -31,12 +31,12 @@ void getOffsets(const af::dim4& strides, const Array<T>& mask, std::vector<dim_t
 
     offsets.reserve(mask.elements());
     for (dim_t j = 0; j < dim1; ++j) {
-	for (dim_t i = 0; i < dim0; ++i) {
-	    if (filter[ getIdx(fstrides, i, j) ] > (T)0) {
-		dim_t offset = (j - R1) * strides[1] + (i - R0) * strides[0];
-		offsets.push_back(offset);
-	    }
-	}
+        for (dim_t i = 0; i < dim0; ++i) {
+            if (filter[ getIdx(fstrides, i, j) ] > (T)0) {
+                dim_t offset = (j - R1) * strides[1] + (i - R0) * strides[0];
+                offsets.push_back(offset);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


vakopian:
there is a bug in the implementation, I'll fix and re-open later

pavanky:
@vakopian interesting, I wouldn't have thought that just the offset calculations were that costly.

I wonder if the performance improves when you do the same for convolve as well.

vakopian:
My first attempt for implementing this offset pre-computation, was to simply pre-compute both the vertical and horizontal offsets, and avoid the need for pad and crop. That worked, but it only gave about 30% improvement. So I'm thinking, it must not be just the pre-computation. Most likely accessing the memory linearly helps with cpu cache hits and brings us this 20x improvement.
So, if we can keep this linear access, it's a real possibility that convolve would benefit from this too.

pavanky:
@vakopian The "Kernel" / neighborhood mask was linear in old code also..

Unless I am missing something, the only reduction seemed to be in the number of offset calculations.

vakopian:
In the previous implementation when looping over i and j, the filter masks could shift the positions back-n-forth, so the index `getIdx(istrides, offi, offj)` for accessing the input data could jump between rows and columns. In the new implementation the index `n + offsets[oi]` only moves forward within the linearized 2D batch. Note that, this only works because of padding (and which is why padding is necessary). Without padding, trying to do linear batches would create problems near the boundaries

pavanky:
@vakopian using the function `getIdx(istrides, offi, offj)`  was probably the issue. It ended up generating offsets identical to the pre computed offsets (when not at the edges), but the compiler was probably unable to optimize as well as it is doing in the PR.

Let me do some experiments and get back to you.

For reference on i7 3770k using gcc 6.1, I get 81.5s with the example you provided.

pavanky:
The time is down to 42s with the following patch applied. I just removed `getIdx` function and used the proper offset calculations.

```
diff --git a/src/backend/cpu/kernel/morph.hpp b/src/backend/cpu/kernel/morph.hpp
index d990bb8..44d0e7a 100644
--- a/src/backend/cpu/kernel/morph.hpp
+++ b/src/backend/cpu/kernel/morph.hpp
@@ -37,7 +37,7 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
                 // j steps along 2nd dimension
                 for(dim_t i=0; i<dims[0]; ++i) {
                     // i steps along 1st dimension
-                    T filterResult = inData[ getIdx(istrides, i, j) ];
+                    T filterResult = inData[j * istrides[1] + i];

                     // wj,wi steps along 2nd & 1st dimensions of filter window respectively
                     for(dim_t wj=0; wj<window[1]; wj++) {
@@ -46,11 +46,11 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
                             dim_t offj = j+wj-R1;
                             dim_t offi = i+wi-R0;

-                            T maskValue = filter[ getIdx(fstrides, wi, wj) ];
+                            T maskValue = filter[wj * fstrides[1] + wi];

                             if ((maskValue > (T)0) && offi>=0 && offj>=0 && offi<dims[0] && offj<dims[1]) {

-                                T inValue   = inData[ getIdx(istrides, offi, offj) ];
+                                T inValue   = inData[istrides[1] * offj + offi];

                                 if (IsDilation)
                                     filterResult = std::max(filterResult, inValue);
@@ -61,7 +61,7 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
                         } // window 1st dimension loop ends here
                     } // filter window loop ends here

-                    outData[ getIdx(ostrides, i, j) ] = filterResult;
+                    outData[ostrides[1] * j + i] = filterResult;
                 } //1st dimension loop ends here
             } // 2nd dimension loop ends here
```

pavanky:
Time is down to 18s after the following patch is applied. This patch removes constant bounds check inside the inner most loop.

```cpp
diff --git a/src/backend/cpu/kernel/morph.hpp b/src/backend/cpu/kernel/morph.hpp
index d990bb8..70394ef 100644
--- a/src/backend/cpu/kernel/morph.hpp
+++ b/src/backend/cpu/kernel/morph.hpp
@@ -35,22 +35,22 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
             // either channels or batch is handled by outer most loop
             for(dim_t j=0; j<dims[1]; ++j) {
                 // j steps along 2nd dimension
+                const int jstart = std::max(0, (int)(j - R1)), jend = std::min((int)dims[1], (int)(j - R1 + window[1]));
                 for(dim_t i=0; i<dims[0]; ++i) {
                     // i steps along 1st dimension
-                    T filterResult = inData[ getIdx(istrides, i, j) ];
+                    T filterResult = inData[j * istrides[1] + i];
 
-                    // wj,wi steps along 2nd & 1st dimensions of filter window respectively
-                    for(dim_t wj=0; wj<window[1]; wj++) {
-                        for(dim_t wi=0; wi<window[0]; wi++) {
-
-                            dim_t offj = j+wj-R1;
-                            dim_t offi = i+wi-R0;
+                    const int istart = std::max(0, (int)(i - R0)), iend = std::min((int)dims[0], (int)(i - R0 + window[0]));
 
-                            T maskValue = filter[ getIdx(fstrides, wi, wj) ];
-
-                            if ((maskValue > (T)0) && offi>=0 && offj>=0 && offi<dims[0] && offj<dims[1]) {
+                    // wj,wi steps along 2nd & 1st dimensions of filter window respectively
+                    for(dim_t offj=jstart; offj<jend; offj++) {
+                        dim_t wj = offj - (j - R1);
+                        for(dim_t offi=istart; offi<iend; offi++) {
+                            dim_t wi = offi - (i - R0);
 
-                                T inValue   = inData[ getIdx(istrides, offi, offj) ];
+                            T maskValue = filter[wj * fstrides[1] + wi];
+                            if (maskValue) {
+                                T inValue   = inData[istrides[1] * offj + offi];
 
                                 if (IsDilation)
                                     filterResult = std::max(filterResult, inValue);
@@ -61,7 +61,7 @@ void morph(Array<T> out, Array<T> const in, Array<T> const mask)
                         } // window 1st dimension loop ends here
                     } // filter window loop ends here
 
-                    outData[ getIdx(ostrides, i, j) ] = filterResult;
+                    outData[ostrides[1] * j + i] = filterResult;
                 } //1st dimension loop ends here
             } // 2nd dimension loop ends here
```

pavanky:
For reference, after applying the patch from this PR results in the performance to be about 3.2s.

@vakopian I am going to incorporate your changes eventually. I am trying to document which of the individual changes resulted in the performance improvements so we can use this information for other functions.

vakopian:
very interesting. BTW, since you mentioned `convolve`, after the changes to dilate, profiling my application shows the bottleneck in `one2one_2d`, which is only called from `convolve`...

vakopian:
@pavanky there is one more optimization that would be really good to add to `morph`. In the special case (which occurs pretty often), when the filter contains only a single nonzero element, all the iterations and max/min can be avoided, as the whole operation becomes a simple "shift" operation. However, I'm not sure if it's better to do that optimization at the kernel level, or at a higher level. The benefit of doing it at a higher level, is that the "shift" operation can be easily implemented using the high level ArrayFire functions. On the other hand, doing it at the kernel level has the benefit that we already count the number of non-zeros in the filter, so it's easier to check if we are handling the special case.

9prady9:
@vakopian Any chance you would be able to get to this in the coming week ? Also, please rebase your changes from lastest master.

9prady9:
Commits from this will be merged in https://github.com/arrayfire/arrayfire/pull/2238
The commits are cherry-picked to preserve author info.

PR #1571, opened 2016-09-07
explicitly instantiate void* type for array::device<void>()

syurkevi:
Fixes issue #1503


shehzan10:
build arrayfire osx cpu ci

PR #1581, opened 2016-09-09
Minor features and bug fixes for 3.4

### LINKED ISSUE #1370: Feature Request: clamp
This can be implemented as following in C++:

```
af::array clamp(af::array in, double lower, double upper)
{
    return af::min(upper, af::max(lower, in));
}
```

However it might be better to use builtin functions if possible. This can be added to JIT once a node with 3 inputs is available as discussed here: https://github.com/arrayfire/arrayfire/issues/1345

### LINKED ISSUE #1450: array.device is sometimes returning raw pointers
I am mixing CUDA and arrayfire code extensively. I am using CUDA surface objects, so I need to do a copy from the arrayfire device memory to the cuda array memory. Currently, I'm using cudaMemcpy3D(), and I am assuming that the arrayfire data is in packed form, in H x W x C order (height varies first.) I'm using float32 elements here.

Here's the code I use:

```
// release any previous content, make a cuda array, copy from arrayfire array
void CudaArray::to_surface(const af::array &in) 
{
	if (in.dims()[3] != 1)
		throw "CudaArray: attempt to allocate a 4D CUDA surface";

	// free up old CUDA array and allocate new one
	free();
	dims_ = in.dims();
	allocate();

	void *d_srcptr = in.device<float>();	// does an implicit eval()
	af::sync();		// ensure all arrayfire work is complete before we trigger CUDA code

	cudaMemcpy3DParms params = { 0 };
	params.srcPtr = make_cudaPitchedPtr(d_srcptr, dims_[0] * sizeof(float), dims_[0], dims_[1]);	// dims[0] is what varies fastest
	params.dstArray = cuda_array;
	params.kind = cudaMemcpyDeviceToDevice;
	params.extent = make_cudaExtent(dims_[0], dims_[1], dims_[2]);

	checkCuda(cudaMemcpy3D(&params));
	in.unlock();                        // done with d_srcptr
}
```

I use the following to get the data out of the surface back to an array:

```
// make a new arrayfire array, copy data from surface to the array
af::array CudaArray::to_array() 
{
	// ensure CUDA is done before we start copying
	checkCuda(cudaDeviceSynchronize());
	checkCuda(cudaGetLastError());

	af::array out(dims_);

	void *d_dstptr = out.device<float>();

	cudaMemcpy3DParms params = { 0 };
	params.srcArray = cuda_array;
	params.dstPtr = make_cudaPitchedPtr(d_dstptr, dims_[0] * sizeof(float), dims_[0], dims_[1]);
	params.kind = cudaMemcpyDeviceToDevice;
	params.extent = make_cudaExtent(dims_[0], dims_[1], dims_[2]);

	checkCuda(cudaMemcpy3D(&params));
	out.unlock();                       // done with d_dstptr

	return out;
}
```

Everything works fine as long as to_surface() is invoked only on a fresh array, i.e., an array created by `array foo = loadImage("file.png", true); CudaArray baz(foo);   // this works just fine`

However, if I do any kind of arrayfire processing on the array, e.g.,
```
array subwindow(const array &in)
{
    int h = in.dims()[0], w = in.dims()[1];
    array out = in(seq(h/4, h/4+h/2-1), seq(w/4, w/4+w/2-1), span);
    return out;
}

int main()
{
    af::setDevice(0); .. make a window ..

    array in = loadImage("baz.jpg", true);
    in = subwindow(in);          // remove this line and the code works
    CudaArray baz(in);
    win.image(in.to_array());  // shows what appears to be result of a bad stride 
}
```

So the question is, after the call to subwindow, is the memory layout of the array in still tightly packed with floats with no padding and a pitch in bytes equal to the height * 4?

### LINKED ISSUE #1502: API function to query status of device memory user lock
Functions that access the device memory of an `array` using the `device()` function have to call `unlock()` to unset the user lock when done. However, when the device memory was actually claimed previously by client code that wants to manage the memory itself, this function should indeed not unset the user lock. A function to query the status of the user lock would be needed, such as:

```c++
void foo(af::array &a)
{
  const bool was_locked = a.islocked(); // not yet existing API function
  void *ptr = a.device<char>(); // will set user lock
  // do something with the device memory
  if (!was_locked) a.unlock(); // need to unset user lock if it wasn't set before
}
```

### LINKED ISSUE #1529: Lazy evaluation is too lazy
The following examples do not work anymore due to increased memory usage:

- rbm
- fractal

### LINKED ISSUE #1558: symbols of af::eval with multiple parameters different in 3.3 and devel
They were inline function calls earlier. They are now part of the library. This needs to be fixed.

pavanky:
- Fixes #1558 
- Fixes #1370 
- Fixes #1502 
- Fixes #1450 
- Fixes #1529


shehzan10:
build arrayfire linux ci
build arrayfire osx ci
build arrayfire windows ci

PR #1582, opened 2016-09-11
Cleanup freeimage api calls

### LINKED ISSUE #1356: Clean up the interface with Freeimage
Right now we are loading in raw data and ignoring any changes that might occur because of the system architecture.  We have fixes in places for certain cases, but that is not the proper way to deal with the data.

Free image documentation specifies how to do this. Check out page 28 of the pdf from here:
[FreeImage3154.pdf](https://github.com/arrayfire/arrayfire/files/186314/FreeImage3154.pdf)

9prady9:
Fixes #1356 

[skip arrayfire ci]

### DIFF
--- a/include/af/graphics.h
+++ b/include/af/graphics.h
@@ -1071,9 +1071,9 @@ AFAPI af_err af_set_axes_limits_3d(const af_window wind,
    \ingroup gfx_func_window
 */
 AFAPI af_err af_set_axes_titles(const af_window wind,
-                                const char * const xtitles,
-                                const char * const ytitles,
-                                const char * const ztitles,
+                                const char * const xtitle,
+                                const char * const ytitle,
+                                const char * const ztitle,
                                 const af_cell* const props);
 #endif
 
--- a/src/api/c/imageio.cpp
+++ b/src/api/c/imageio.cpp
@@ -134,14 +134,8 @@ af_err af_load_image(af_array *out, const char* filename, const bool isColor)
         // for statically linked FI
         FI_Init();
 
-        // set your own FreeImage error handler
-        FreeImage_SetOutputMessage(FreeImageErrorHandler);
-
         // try to guess the file format from the file extension
-        FREE_IMAGE_FORMAT fif = FreeImage_GetFileType(filename);
-        if (fif == FIF_UNKNOWN) {
-            fif = FreeImage_GetFIFFromFilename(filename);
-        }
+        FREE_IMAGE_FORMAT fif = identifyFIF(filename);
 
         if(fif == FIF_UNKNOWN) {
             AF_ERROR("FreeImage Error: Unknown File or Filetype", AF_ERR_NOT_SUPPORTED);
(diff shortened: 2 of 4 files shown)


CMakeModules/FindFreeImage.cmake
```diff
@@ -46,23 +58,37 @@ FIND_LIBRARY( FREEIMAGE_STATIC_LIBRARY
     /sw/lib
     /opt/local/lib
     DOC "The FreeImage library")
+FIND_LIBRARY( FREEIMAGE_CPP_STATIC_LIBRARY
```
9prady9: @shehzan10 Can you please check if there is a static version of freeimageplus.

pavanky:
If we are going to use the C++ api, we should use the static version of the library (or the linked library with arrayfire).

The ABI for C++ changes quite frequently and we can not rely on the version we compiled with will be the same one present on the users systems.

shehzan10:
Although it was me who suggested that we should use the C++ API, I do not think that is the best way to go about it for us. My reasons are:
* The FreeImagePlus library is similar to ArrayFire CPP API for ArrayFire C. There is no performance to be gained out of using it.
* The FreeImagePlus may not be ubuquitous as freeimage and unless we double check, I am not sure of how many distributions provide it in package manager.

pavanky:
Can't we clean up using the C api ? In the existing implementation we are assuming a lot of things which we can get away from.

shehzan10:
> Can't we clean up using the C api ? In the existing implementation we are assuming a lot of things which we can get away from.

That's what I think is the best path forward.

9prady9:
I agree that we should ditch freeimageplus, the c++ api of freeimage isn't that helpful anyway. All it has done so far is that it abstracted out the common checks that we were doing earlier, which barely saved any effort.

As far as C API code we had earlier, i am not sure there is anything to clean up other than handling the data correctly - what pavan referred to in the FreeImage documentation's page 28. We may extract common code (such as the checks offered by `fipImage::identifyFIF` function) as helper functions into `imageio_helper.h`.

9prady9:
@pavanky / @shehzan10 I have reverted the changes regarding freeimageplus and cleaned up the freeimage api calls a bit now.

include/af/graphics.h
```diff
@@ -1071,9 +1071,9 @@ AFAPI af_err af_set_axes_limits_3d(const af_window wind,
    \ingroup gfx_func_window
 */
 AFAPI af_err af_set_axes_titles(const af_window wind,
-                                const char * const xtitles,
-                                const char * const ytitles,
-                                const char * const ztitles,
+                                const char * const xtitle,
+                                const char * const ytitle,
+                                const char * const ztitle,
```
shehzan10: This has been fixed in devel. You can undo this commit.

shehzan10:
Close this PR and open it to hotfix-3.4.1

9prady9:
Since we are not going to freeimageplus, i will correct the branch name etc. and reopen a new PR.

PR #1587, opened 2016-09-12
Vector field example and bug fixes

pavanky:
This pulls in changes from #1586 

- Fixes performance issues in CUDA select and replace
- Fixes bug in CPU element wise operations
- Turn on CPUID by default


shehzan10:
build arrayfire linux ci
build arrayfire osx cpu ci

PR #1591, opened 2016-09-15
BUGFIX: Add missing unified call for set_fft_plan_cache_size

9prady9:
[skip arrayfire ci]

I found this problem while debuggin why rust wrapper builds were failing on Windows. Interestingly, windows compiler didn't complain about this while building arrayfire though.


shehzan10:
It probably was not complaining because there was no calls to that function.

9prady9:
Yes, that could be it. But the interesting part is that rust wrapper didn't complain about it on Linux/OSX.

PR #1595, opened 2016-09-19
Bugfixes to image morph and nearest neighbors

### LINKED ISSUE #1594: Change initial values of min and max to be +/- infinity
I think this was missed in the 3.4 release. 
As per  https://github.com/arrayfire/arrayfire/pull/1547#issuecomment-240919098

### LINKED ISSUE #1596: Nearest Neighbour Distance Calculation wrong results on CUDA and OpenCL Backend
There appears to be an issue in the nearest neighbour function in the CUDA and OpenCL Backend (CPU returns the correct results).
Here is a small Test Case Illustrating the issue:

```C++
TEST(NearestNeighbour, CUDA)
{
	double train[1 * 2] = {
		5, 5,
	};

	double query[5 * 2] = {
		0, 0,
		3.5, 4,
		5, 5,
		6, 5,
		8, 6.5
	};
	array t(2, 1, train);
	array q(2, 5, query);
	array indices;
	array distances;
	nearestNeighbour(indices, distances, q, t, 0, 1, AF_SSD);

	double expectedDistances[5] = {
		(5 - 0) * (5 - 0) + (5 - 0) * (5 - 0),
		(5 - 3.5) * (5 - 3.5) + (5 - 4) * (5 - 4),
		(5 - 5) * (5 - 5) + (5 - 5) * (5 - 5),
		(5 - 6) * (5 - 6) + (5 - 5) * (5 - 5),
		(5 - 8) * (5 - 8) + (5 - 6.5) * (5 - 6.5)
	};

	double* actualDistances = distances.host<double>();
	for (int i = 0; i < 5; i++)
	{
		EXPECT_EQ(expectedDistances[i], actualDistances[i]);
	}
}
```
The formula for calculating the sum of square distance between Point p1 and p2 is: `(p1.x - p2.x)^2 + (p1.y - p2.y)^2`
When trying to calculate the distance of every point in query to (5 | 5) the CPU Backend returns the correct results (those in expectedDistances). Those are:
`   50.0000     3.2500     0.0000     1.0000    11.2500 `

However the CUDA and OpenCL backend return the following values:
`   50.0000    27.2500    25.0000    26.0000    34.0000 `

This issue does not only affect the `AF_SSD` match type, but also the `AF_SAD` (sum of absolute difference). Here the correct results would be: 
`10.0000     2.5000     0.0000     1.0000     4.5000 `
but instead are:
`   10.0000     6.5000     5.0000     6.0000     8.0000 `

I am using ArrayFire v3.4.0 installed with the binaries from your website.
```
ArrayFire v3.4.0 (CUDA, 64-bit Windows, build 10d9716)
Platform: CUDA Toolkit 7.5, Driver: CUDA Driver Version: 7050
[0] GeForce GT 730M, 1024 MB, CUDA Compute 3.5
```

pavanky:
- Fixes issues with erode and dilate at corner cases
- Fixes #1594 
- Fixes #1596

### DIFF
--- a/src/backend/cuda/kernel/nearest_neighbour.hpp
+++ b/src/backend/cuda/kernel/nearest_neighbour.hpp
@@ -36,7 +36,16 @@ struct dist_op<T, To, AF_SAD>
 {
     __device__ To operator()(T v1, T v2)
     {
-        return abs((double)v1 - (double)v2);
+        return fabsf((float)v1 - (float)v2);
+    }
+};
+
+template<typename To>
+struct dist_op<double, To, AF_SAD>
+{
+    __device__ To operator()(double v1, double v2)
+    {
+        return fabs((double)v1 - (double)v2);
     }
 };
 
@@ -130,7 +139,7 @@ __global__ void nearest_neighbour_unroll(
 
         // Load one query feature that will be tested against all training
         // features in current block
-        if (tid < feat_len && valid_feat) {
+        if (tid < feat_len) {
             s_query[tid] = query.ptr[tid * nquery + j];
         }
         __syncthreads();
@@ -269,7 +278,7 @@ __global__ void nearest_neighbour(
 
         // Load one query feature that will be tested against all training
         // features in current block
-        if (tid < feat_len && valid_feat) {
+        if (tid < feat_len) {
             s_query[tid] = query.ptr[tid * nquery + j];
         }
         __syncthreads();
@@ -379,10 +388,8 @@ __global__ void select_matches(
     __shared__ To s_dist[THREADS];
     __shared__ unsigned s_idx[THREADS];
 
+    s_dist[sid] = max_dist;
     if (f < nfeat) {
-        s_dist[sid] = max_dist;
-        __syncthreads();
-
         for (unsigned i = threadIdx.y; i < nelem; i += blockDim.y) {
             To dist = in_dist[f * nelem + i];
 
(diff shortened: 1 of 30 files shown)


pavanky:
@shehzan10 I am not making my PRs run tests run automatically. Can you launch and run the tests when possible

shehzan10:
build arrayfire linux ci
build arrayfire osx ci
build arrayfire windows ci

pavanky:
@shehzan10 you can kill old windows build. I missed something earlier by not testing OpenCL backend.

pavanky:
@umar456 @shehzan10 
Ideally the `options` should be a daughter class of `stringstream` with customizations. This will get rid of hacks like the last commit I have here and other work arounds like (`toNum`).

May be the payoff is not that great, but that is something to consider going forward.

src/backend/cuda/math.hpp
```diff
@@ -93,26 +93,30 @@ namespace cuda
...
-    template<> __device__  char   limit_max<char>()   { return 0x7f; }
-    template<> __device__  char   limit_min<char>()   { return 0x80; }
-    template<> __device__  float  limit_max<float>()  { return  CUDART_INF_F; }
-    template<> __device__  float  limit_min<float>()  { return -CUDART_INF_F; }
-    template<> __device__  double limit_max<double>() { return  CUDART_INF; }
-    template<> __device__  double limit_min<double>() { return -CUDART_INF; }
-    template<> __device__  short  limit_max<short>()  { return 0x7fff; }
-    template<> __device__  short  limit_min<short>()  { return 0x8000; }
-    template<> __device__  ushort limit_max<ushort>() { return ((ushort)1) << (8 * sizeof(ushort) - 1); }
+    template <typename T> __device__ T maxval() { return 1u << (8 * sizeof(T) - 1); }
```
umar456: Hmm. I liked the old name

src/backend/cuda/kernel/nearest_neighbour.hpp
```diff
@@ -392,26 +399,26 @@ __global__ void select_matches(
...
-                __syncthreads();
+    // Reduce best matches and find the best of them all
+    for (unsigned i = blockDim.y / 2; i > 0; i >>= 1) {
+        if (threadIdx.y < i) {
+            To dist = s_dist[sid + i];
+            if (dist < s_dist[sid]) {
+                s_dist[sid] = dist;
+                s_idx[sid]  = s_idx[sid + i];
             }
+            __syncthreads();
```
umar456: Isn't this still undefined?

shehzan10:
build arrayfire windows ci

src/backend/cuda/kernel/nearest_neighbour.hpp
pavanky: It was already like that (look at the left). I fixed it in one place but missed this. I'll fix it.

src/backend/cuda/math.hpp
pavanky: I did not want it to be confused with numeric limits. I could change it back if you feel strongly about.

umar456: Isn't it doing the same thing as numeric limits. If anything it should have the same interface.

pavanky: It has different behavior than numeric limits for floats and doubles. It had same behavior before, but this commit changed the behavior so I changed the name as well.

shehzan10:

pavanky:
@shehzan10 I missed the change in __syncthreads like @umar456  mentioned. Can you fix it in a different PR.

BTW, it would also be better if you guys start commenting this as reviews.

PR #1598, opened 2016-09-20
Sparse eval

### LINKED ISSUE #1597: Allow eval() for sparse inputs
Missed adding eval support for Sparse in front end. eval is supported by the SparseArray class internally.
This just requires changing the getInfo function in af_eval().

shehzan10:
Fixes #1597 

[skip arrayfire ci]

### DIFF
--- a/src/api/c/device.cpp
+++ b/src/api/c/device.cpp
@@ -15,6 +15,7 @@
 #include <platform.hpp>
 #include <Array.hpp>
 #include <handle.hpp>
+#include <sparse_handle.hpp>
 #include "err_common.hpp"
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


src/api/c/device.cpp
```diff
@@ -165,25 +166,43 @@ static inline void eval(af_array arr)
     return;
 }
 
+template<typename T>
+static inline void sparseEval(af_array arr)
```
umar456: A better(simpler) approach would be to specialize eval to accept sparse arrays

umar456:
```
template<typename T, bool sparse = false>
static inline void eval(af_array arr)
{
    if(sparse) {
        getSparse<T>(arr).eval();
    } else {
        getArray<T>(arr).eval();
    }
    return;
}
```

PR #1599, opened 2016-09-20
Add Conjugate Gradient example to benchmarks

shehzan10:
Compare the performance and memory usage of sparse vs dense using conjugate
gradient example

[skip arrayfire ci]

### DIFF
--- a/examples/benchmarks/cg.cpp
+++ b/examples/benchmarks/cg.cpp
@@ -1,0 +1,139 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+#include <arrayfire.h>
+#include <iostream>
+
+using namespace af;
+
+static size_t dimension = 4 * 1024;
+static const int maxIter = 10;
+static const int sparsityFactor = 7;
+
+static array A;
+static array spA;  // Sparse A
+static array x0;
+static array b;
+
+void setupInputs()
+{
+    // Generate a random input: A
+    array T = randu(dimension, dimension, f32);
+    // Create 0s in input.
+    // Anything that is no divisible by sparsityFactor will become 0.
+    A = floor(T * 1000);
+    A = A * ((A % sparsityFactor) == 0) / 1000;
+    // Make it positive definite
+    A = transpose(A) + A + A.dims(0)*identity(A.dims(0), A.dims(0), f32);
+
+    // Make A sparse as spA
+    spA = sparse(A);
+
+    // Generate x0: Random guess
+    x0 = randu(A.dims(0), f32);
+
+    //Generate b
+    b = matmul(A, x0);
+
+    std::cout << "Sparsity of A = "
+              << 100.f * (float)sparseGetNNZ(spA) / (float)spA.elements()
+              << "%" << std::endl;
+    std::cout << "Memory Usage of A = "
+              << A.bytes() / (1024.f * 1024.f)
+              << " MB" << std::endl;
+    std::cout << "Memory Usage of spA = "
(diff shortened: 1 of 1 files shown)


examples/benchmarks/cg.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+using namespace af;
+
+static size_t dimension = 4 * 1024;
+static const int maxIter = 10;
+static const int sparsityFactor = 7;
+
+static array A;
+static array spA;  // Sparse A
+static array x0;
```
umar456: this looks too similar to a hex number.

examples/benchmarks/cg.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+void denseConjugateGradient(void)
+{
+    array x = constant(0, b.dims(), f32);
+    array r = b - matmul(A, x);
+    array p = r;
+
+    for (int i = 0; i < maxIter; ++i) {
```
umar456: doesn't the timeit function execute this multiple times anyway?

examples/benchmarks/cg.cpp
```diff
@@ -0,0 +1,139 @@
...
+        array Ap = matmul(in, p);
+        array alpha_num = dot(r, r);
+        array alpha_den = dot(p, Ap);
+        array alpha = alpha_num/alpha_den;
+        r = r - tile(alpha, Ap.dims())*Ap;
+        x = x + tile(alpha, Ap.dims())*p;
+        array beta_num = dot(r, r);
+        array beta = beta_num/alpha_num;
+        p = r + tile(beta, p.dims())*p;
```
umar456: `p *= r + tile(beta, p.dims());`

examples/benchmarks/cg.cpp
```diff
@@ -0,0 +1,139 @@
...
+    array r = b - matmul(in, x);
+    array p = r;
+
+    for (int i = 0; i < maxIter; ++i) {
+        array Ap = matmul(in, p);
+        array alpha_num = dot(r, r);
+        array alpha_den = dot(p, Ap);
+        array alpha = alpha_num/alpha_den;
+        r = r - tile(alpha, Ap.dims())*Ap;
+        x = x + tile(alpha, Ap.dims())*p;
```
umar456: `x += tile(alpha, Ap.dims())*p;`

examples/benchmarks/cg.cpp
```diff
@@ -0,0 +1,139 @@
...
+    array x = constant(0, b.dims(), f32);
+    array r = b - matmul(in, x);
+    array p = r;
+
+    for (int i = 0; i < maxIter; ++i) {
+        array Ap = matmul(in, p);
+        array alpha_num = dot(r, r);
+        array alpha_den = dot(p, Ap);
+        array alpha = alpha_num/alpha_den;
+        r = r - tile(alpha, Ap.dims())*Ap;
```
umar456: r -= tile(alpha, Ap.dims())*Ap;

examples/benchmarks/cg.cpp
shehzan10: x0 is a mathematical notation used in conjugate gradient.

examples/benchmarks/cg.cpp
shehzan10: This is not for timing, conjugate gradient is an iterative solver.

umar456: Right. I don't know what I was thinking.

examples/benchmarks/cg.cpp
kaatish: @umar456 These expressions are not equivalent. We want something like this `p = r + scalar*p`.

umar456: doh! Clearly its time for bed.

examples/benchmarks/cg.cpp
pavanky: @shehzan10 FYI, try to use X0 in such cases.

PR #1604, opened 2016-09-21
fix for surface rendering function

### LINKED ISSUE #1603: bug in Window::surface function - incorrect surface rendering.
This has been initially been reported via [google group forum post](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/arrayfire-users/SMfSfcXVeCM/Qo3R4oc1CAAJ).

Code to reproduce the bug is given below.

```cpp
#include <arrayfire.h>
#include <cstdio>
#include <cstdlib>

using namespace af;

{
    try {
        af::info();

        Window w;

        array A = randu(10, 10);

        w.setAxesLimits(0.0f, 10.0f, 0.0f, 10.0f, 0.0f, 1.0f);

        while(!w.close()) {
            w.surface(A);
        }

    } catch (af::exception& e) {
        throw;
    }

    return 0;
}
```

9prady9:
Fixes #1603

### DIFF
--- a/src/api/c/histeq.cpp
+++ b/src/api/c/histeq.cpp
@@ -53,7 +53,7 @@ static af_array hist_equal(const af_array& in, const af_array& hist)
     Array<float> idxArr  = lookup<float, T>(normCdf, getArray<T>(vInput), 0);
 
     Array<T> result = cast<T>(idxArr);
-    result.modDims(input.dims());
+    result = modDims(result, input.dims());
 
     AF_CHECK(af_release_array(vInput));
 
--- a/src/api/c/plot.cpp
+++ b/src/api/c/plot.cpp
@@ -265,8 +265,8 @@ af_err plotWrapper(const af_window wind, const af_array X, const af_array Y,
 //
 //    // Force the vectors to be row vectors
 //    // This ensures we can use join(0,..) and skip reorder
-//    xIn.modDims(rowDims);
-//    yIn.modDims(rowDims);
+//    xIn = modDims(xIn, rowDims);
+//    yIn = modDims(yIn, rowDims);
 //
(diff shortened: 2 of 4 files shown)


9prady9:
@shehzan10 Apparently, regular `modDim` backend function is also doing evaluate before it is modifying the meta data of Array. 

The problem lies in dims being modified before data is evaluated. Since each JIT node is storing the original dims, when eval is called after moddims is done, the node is still using original dims to compute the array values.

@pavanky Does changing `m_dims` member variable in buffer JIT node whenever moddims is called have any side effects ?

pavanky:
@shehzan10 didn't we fix this ?

pavanky:
Ok, it looks like me and shehzan found the exact same bug for when inputs are **NOT** vectors and fixed that. We forgot to fix cases where input is a vector. It is a simple enough fix.

@9prady9 can you look at the following commit and send in a proper patch.
18a4e5547c729ae9bbb45926a4d9d39fea166b38

9prady9:
@pavanky The fix is using function version of `modDims` instead of `Array::modDims` ?

pavanky:
@9prady9 yes, the function version of modDims is updating the metadata properly. So it is easier to just call that than re-implement changing the metadata in surface.

9prady9:
@shehzan10 This is ready for merge.

pavanky:
@9prady9 can you look to see if any other functions in `src/api/c/` are using `arr.modDims` and change those as well ? (except `modDims` function ofcourse :) )

9prady9:
Sure

9prady9:
There is only file `histeq.cpp` that used that and i changed that now.

Issue #731, opened 2015-05-28
Missing functions


umar456:
- [ ] arg (phase angle in radians)
- [ ] epsilon
- [ ] erfinv
- [ ] erfcinv
- [ ] factorial
- [ ] trunc
- [ ] pow2
- [ ] root
- [ ] sign
- [ ] nonzeros

Either we add these functions or we talk about them in release notes

pavanky:
- mul has been changed to prod
- nonzeros was a C only api. The C++ was called count.
- We can't do anything about Eigen, SVD, matpow, hessenberg (no upstream support)
- filter was a wrapper around convolve. We had it for legacy reasons. I dont see why we should continue supporting it.

pavanky:
Realistically the following functions need to be done be:

- [ ] arg (phase angle in radians)
- [ ] epsilon
- [ ] erfinv
- [ ] erfcinv
- [ ] factorial
- [ ] trunc
- [ ] pow2
- [ ] root
- [ ] sign
- [ ] rank
- [ ] det
- [ ] minfilt (wrap around erode)
- [ ] maxfilt (wrap around dilate)

umar456:
* nonzeros was an af::array member function
* should we create a mul function in compatible and deprecate it?
* I figured. I wanted a comprehensive list
* deprecate?

pavanky:
About non zeros.. yeah it was a member function because it did something special for sparse matrices. We can add it back, but I don't think it does anything useful.

- We can put mul and filter in compatible.h if need be.

pavanky:
Finished the necessary functions for 3.0

shehzan10:
Marking minfilt, maxfilt, det, rank, svd as completed

Issue #101, opened 2014-11-12
af::exception should derive from std::exception


pavanky:
The files that are affected:

```
$ grep -rl exception src/api/cpp/ include/
src/api/cpp/exception.cpp
src/api/cpp/error.hpp
include/arrayfire.h
include/af/exception.h
```

PR #1608, opened 2016-10-03
erfinv and erfcinv

billpage:
Ref:
1. [https://www.accelereyes.com/arrayfire/c/group__array__func__erfinv.htm](https://www.accelereyes.com/arrayfire/c/group__array__func__erfinv.htm)
2. [https://www.accelereyes.com/arrayfire/c/group__array__func__erfcinv.htm](https://www.accelereyes.com/arrayfire/c/group__array__func__erfcinv.htm)
3. [Iissues #731](https://github.com/arrayfire/arrayfire/issues/731)
4. [arrayfire-python PR #101](https://github.com/arrayfire/arrayfire-python/pull/101)

I attempted to add these missing functions to ArrayFire by following the pattern that I saw with erf and erfc. See also related pull request for `arrayfire-python`. But I am having trouble calling this from modified `arrayfire-python` bindings.  Apparently the new routines are not seen by the bindings, e.g.

```
...
RuntimeError: Error in af_err unified::AFSymbolManager::call(const
char*, CalleeArgs ...) [with CalleeArgs = {void**, void*}]
In file src/api/unified/symbol_manager.hpp:68
Failed to load symbol: af_erfinv
```

What might I be missing?


pavanky:
@billpage 

1) You need to add the C api functions in both src/api/c/unary.cpp and src/api/unified/unary.cpp. You missed the former.

2) Simply adding the enums will not add the functionality. It'll default to use a noop function which will error out. You should look at src/backend/{cpu,cuda,opencl}/unary.hpp to see what is happening.

3) Considering that `erfinv` and `erfcinv` are not standard math functions in C++ and OpenCL, adding them to `unary.hpp` will not work either. Even for CUDA, it has only been added in CUDA 8.0.

billpage:
Thanks for the missing details. As I understand it, Boost is a backend
dependency and these functions are available there.

http://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/math_toolkit/sf_erf/error_inv.html

CUDA 8.0 has just been released.

What is the policy for functionality that is missing in earlier CUDA/OpenCL
releases? Should one expect an error when compiling against arraryfire
built with an earlier version or do you allow a fall-back to CPU?  My
strong preference is for the latter case since that would allow me to write
my code in a way that simply performs better when newer libraries are
available.

Is it worth my while to persist in implementing something of this kind?  If
not, it might be desirable to change the arraryfire documentation.

pavanky:
@billpage Boost is not a dependency for the CPU backend, just CUDA (for scoped_ptr) and OpenCL (for boost.compute).

- As for CUDA 8 being released, OSX and Tegra do not have 8.0 yet.

The best way to implement these functions would be to do something similar to how complex math functions. You can see an example here: https://github.com/arrayfire/arrayfire/blob/devel/src/api/c/unary.cpp#L343

Because of JIT, functions implemented this way are combined into a single kernel and are quite efficient.

pavanky:
Having looked at boost's implementation, I don't think it is easy to implement such a function using existing functionality. I'll try to pull your changes and implement fallback versions for CPU and OpenCL.

9prady9:
@billpage @pavanky Closing the PR since it doesn't have a complete implementaion of the functions. Please move the discussion to #2066. Thank you.

PR #1612, opened 2016-10-07
Preparing for release with CUDA 8

shehzan10:
This PR does the following:
* If CUDA 8 is being used, `CUDA_COMPUTE_20` is no longer in the default computes list. This can be enabled using `cmake -DCUDA_COMPUTE_20=ON`
  * For CUDA versions older than 8, 20, 30, 50 is the default compute selection.
* If CUDA 8 is being used and `CUDA_COMPUTE_20` if OFF, then the default compute version of PTX compilation will be 30 in case of multiple compute devices being present. The behavior when a single compute is found remains the same.
* For libdevice, check the compute version of the device and select the appropriate header.

The compute 2x deprecation follows CUDA 8's deprecation of computes 2.x.

[skip arrayfire ci]

### DIFF
--- a/src/backend/cuda/jit.cpp
+++ b/src/backend/cuda/jit.cpp
@@ -327,7 +327,8 @@ void compute_to_libdevice_table(const char **buffer, size_t *bc_buffer_len, int 
         *buffer        = compute_30_bc;
         *bc_buffer_len = compute_30_bc_len;
     } else {
-        AF_ERROR("Invalid Compute for libdevice", AF_ERR_INTERNAL);
+        *buffer        = compute_30_bc;
+        *bc_buffer_len = compute_30_bc_len;
     }
 }
 
--- a/src/backend/cuda/jit.cpp
+++ b/src/backend/cuda/jit.cpp
@@ -302,6 +302,35 @@ static string getKernelString(string funcName, std::vector<Node *> nodes, bool i
                                                 \
     } while(0)
 
+void compute_to_libdevice_table(const char **buffer, size_t *bc_buffer_len, int compute)
+{
+    // Source: http://docs.nvidia.com/cuda/libdevice-users-guide/basic-usage.html#version-selection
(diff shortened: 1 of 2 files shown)


src/backend/cuda/jit.cpp
```diff
@@ -302,15 +302,50 @@ static string getKernelString(string funcName, std::vector<Node *> nodes, bool i
...
+        *bc_buffer_len = compute_35_bc_len;
+    } else if (compute >  37 && compute <  50) {
+        *buffer        = compute_30_bc;
+        *bc_buffer_len = compute_30_bc_len;
+    } else if (compute >= 50 && compute <= 53) {
+        *buffer        = compute_50_bc;
+        *bc_buffer_len = compute_50_bc_len;
+    } else if (compute >  53) {
+        *buffer        = compute_30_bc;
+        *bc_buffer_len = compute_30_bc_len;
```
pavanky: I think you can just default to this instead of erroring out.

shehzan10:
build arrayfire windows ci

src/backend/cuda/jit.cpp
```diff
@@ -327,7 +327,8 @@ void compute_to_libdevice_table(const char **buffer, size_t *bc_buffer_len, int
         *buffer        = compute_30_bc;
         *bc_buffer_len = compute_30_bc_len;
     } else {
-        AF_ERROR("Invalid Compute for libdevice", AF_ERR_INTERNAL);
+        *buffer        = compute_30_bc;
+        *bc_buffer_len = compute_30_bc_len;
     }
```
pavanky: You should also check fir `if defined(__LIBDEVICE_COMPUTE_XY)` otherwise `compute_xy` and `compute_xy_len` variables will not be defined.

shehzan10: I pushed a change directly for this. It only really affected the compute_50 one since 20 and 30 are available in CUDA 6/6.5.

PR #1614, opened 2016-10-11
FEAT: Add CPU offload to Sparse matmul

### LINKED ISSUE #1602: Add OpenCL-CPU Fallback for Sparse Functions
This would be similar to CPU fallback for BLAS/LAPACK when running on a Unified Memory device (CPU/Integrated Graphics).

Should be be available only when compiled with MKL.

shehzan10:
Add CPU offload for Unified devices for Sparse Matmul
* Uses CPU fallback code when not using MKL

Fixes #1602

### DIFF
--- a/src/backend/opencl/cpu/cpu_sparse_blas.cpp
+++ b/src/backend/opencl/cpu/cpu_sparse_blas.cpp
@@ -1,0 +1,535 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#if defined(WITH_OPENCL_LINEAR_ALGEBRA)
+#include <cpu/cpu_sparse_blas.hpp>
+
+#include <stdexcept>
+#include <string>
+#include <cassert>
+
+#include <af/dim4.hpp>
+#include <complex.hpp>
+#include <err_common.hpp>
+#include <math.hpp>
+#include <platform.hpp>
+
+namespace opencl
+{
+namespace cpu
+{
+
+using namespace common;
+
+using std::add_const;
+using std::add_pointer;
+using std::enable_if;
+using std::is_floating_point;
+using std::remove_const;
+using std::conditional;
+using std::is_same;
+
+template<typename T, class Enable = void>
+struct blas_base {
+    using type = T;
+};
+
+template<typename T>
+struct blas_base <T, typename enable_if<is_complex<T>::value>::type> {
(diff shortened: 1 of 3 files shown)


shehzan10:
build arrayfire osx mkl opencl ci

shehzan10:
The failures in the PR are caused by #1617

src/backend/opencl/cpu/cpu_sparse_blas.cpp
```diff
@@ -0,0 +1,530 @@
...
+SPARSE_FUNC(mv , float   , s)
+SPARSE_FUNC(mv , double  , d)
+SPARSE_FUNC(mv , cfloat  , c)
+SPARSE_FUNC(mv , cdouble , z)
+
+SPARSE_FUNC_DEF( mm )
+SPARSE_FUNC(mm , float   , s)
+SPARSE_FUNC(mm , double  , d)
+SPARSE_FUNC(mm , cfloat  , c)
+SPARSE_FUNC(mm , cdouble , z)
```
umar456: undef SPARSE_FUNC after usage

src/backend/opencl/cpu/cpu_sparse_blas.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+template<typename T, int value>
+scale_type<T> getScale()
+{
+    static T val = scalar<T>(value);
+    return getScaleValue<scale_type<T>, T>(val);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+#ifdef USE_MKL // Implementation using MKL
```
umar456: It would be super nice if you could combine the two ifdefs somehow. There is a lot of repeat code here

src/backend/opencl/cpu/cpu_sparse_blas.cpp
```diff
@@ -0,0 +1,530 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#if defined(WITH_OPENCL_LINEAR_ALGEBRA)
```
umar456: remove the file if that flag is set

src/backend/opencl/cpu/cpu_sparse_blas.cpp
umar456: I understand. But you can wrap the two backends into a similar interface and call it from a single function instead of making two versions of the same thing that share the same initial logic.

src/backend/opencl/cpu/cpu_sparse_blas.cpp
pavanky: How would you deal with it in the source code that calls functions from this file ?

PR #1627, opened 2016-10-17
Device specific info query functions

### LINKED ISSUE #119: Get launch parameters for OpenCL based on the device
Right now OpenCL kernels are launched with 256 threads per group. 
Try to get supported sizes from `DeviceManager` and use the values accordingly.

Add the following queryable information to opencl::DeviceManager class
- [x] Type of local memory
- [x] Check for device specific local memory limitations
- [x] Launch configuration (global, local work group sizes) for given kernel
- [x] Vectorization options.
- [ ] Type of device.

9prady9:
Fixes #119

[skip arrayfire ci]

### DIFF
--- a/src/backend/opencl/program.cpp
+++ b/src/backend/opencl/program.cpp
@@ -12,6 +12,8 @@
 #include <kernel_headers/KParam.hpp>
 #include <debug_opencl.hpp>
 #include <iostream>
+#include <algorithm>
+#include <dispatch.hpp>
 
 using cl::Buffer;
 using cl::Program;
@@ -66,5 +68,33 @@ namespace opencl
             SHOW_BUILD_INFO(prog);
             throw;
         }
+    }
+
+    NDRange getKernelLaunchCfg1D(const Kernel& kernel)
+    {
+        int device = getActiveDeviceId();
+
+        size_t dim0ThreadCount = getDeviceWorkGroupSize(kernel, device);
+
+        // make sure you are not launching what the device can't handle
+        dim0ThreadCount = std::min(getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>(), dim0ThreadCount);
+
+        return cl::NDRange(dim0ThreadCount, 1, 1);
+    }
+
+    NDRange getKernelLaunchCfg2D(const Kernel& kernel)
+    {
+        int device = getActiveDeviceId();
+
+        size_t dim0ThreadCount = kernel.getWorkGroupInfo<CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE>(getDevice(device));
+
+        size_t recomWorkGroupSize = getDeviceWorkGroupSize(kernel, device);
+
+        size_t dim1ThreadCount = divup(recomWorkGroupSize, dim0ThreadCount);
+
+        // make sure you are not launching what the device can't handle
+        dim1ThreadCount = std::min(getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>(), dim1ThreadCount);
+
+        return cl::NDRange(dim0ThreadCount, dim1ThreadCount, 1);
     }
 }
--- a/src/backend/opencl/kernel/morph.hpp
+++ b/src/backend/opencl/kernel/morph.hpp
@@ -79,12 +79,12 @@ void morph(Param out, const Param in, const Param mask)
     auto morphOp = KernelFunctor< Buffer, KParam, Buffer, KParam, Buffer, cl::LocalSpaceArg,
                                   int, int >(*entry.ker);
 
-    NDRange local(THREADS_X, THREADS_Y);
+    NDRange local = getKernelLaunchCfg2D(*entry.ker);
 
-    int blk_x = divup(in.info.dims[0], THREADS_X);
-    int blk_y = divup(in.info.dims[1], THREADS_Y);
+    int blk_x = divup(in.info.dims[0], local[0]);
+    int blk_y = divup(in.info.dims[1], local[1]);
     // launch batch * blk_x blocks along x dimension
-    NDRange global(blk_x * THREADS_X * in.info.dims[2], blk_y * THREADS_Y * in.info.dims[3]);
+    NDRange global(blk_x * local[0] * in.info.dims[2], blk_y * local[1] * in.info.dims[3]);
 
     // copy mask/filter to constant memory
     cl_int se_size   = sizeof(T)*windLen*windLen;
@@ -94,8 +94,8 @@ void morph(Param out, const Param in, const Param mask)
     // calculate shared memory size
     const int halo    = windLen/2;
     const int padding = 2*halo;
-    const int locLen  = THREADS_X + padding + 1;
-    const int locSize = locLen * (THREADS_Y+padding);
+    const int locLen  = local[0] + padding + 1;
+    const int locSize = locLen * (local[1]+padding);
 
     morphOp(EnqueueArgs(getQueue(), global, local),
             *out.data, out.info, *in.data, in.info, *mBuff,
--- a/src/backend/opencl/platform.cpp
+++ b/src/backend/opencl/platform.cpp
@@ -1014,6 +1014,57 @@ void DeviceManager::markDeviceForInterop(const int device, const forge::Window* 
     }
 }
 #endif
+
+bool isDedicatedLocalMem(int device)
+{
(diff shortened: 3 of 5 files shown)


src/backend/opencl/program.cpp
```diff
@@ -67,4 +69,32 @@ namespace opencl
...
+
+        size_t recomWorkGroupSize = getDeviceWorkGroupSize(kernel, device);
+
+        size_t dim1ThreadCount = divup(recomWorkGroupSize, dim0ThreadCount);
+
+        // make sure you are not launching what the device can't handle
+        dim1ThreadCount = std::min(getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>(), dim1ThreadCount);
+
+        return cl::NDRange(dim0ThreadCount, dim1ThreadCount, 1);
+    }
```
pavanky: I don't think this should be a function (atleast under this name). This gives the impression that this function can be used to query any 2D. Functions that have static allocation of resources based on the launch configuration will not work.

src/backend/opencl/program.cpp
```diff
@@ -67,4 +69,32 @@ namespace opencl
...
     }
+
+    NDRange getKernelLaunchCfg1D(const Kernel& kernel)
+    {
+        int device = getActiveDeviceId();
+
+        size_t dim0ThreadCount = getDeviceWorkGroupSize(kernel, device);
+
+        // make sure you are not launching what the device can't handle
+        dim0ThreadCount = std::min(getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>(), dim0ThreadCount);
```
pavanky: Isn't it pretty much guaranteed that `kern.getWorkGroupInfo<CL_KERNEL_WORK_GROUP_SIZE>(dev)` is always less than `getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>()` ?

src/backend/opencl/program.cpp
```diff
@@ -67,4 +69,32 @@ namespace opencl
...
+        int device = getActiveDeviceId();
+
+        size_t dim0ThreadCount = kernel.getWorkGroupInfo<CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE>(getDevice(device));
+
+        size_t recomWorkGroupSize = getDeviceWorkGroupSize(kernel, device);
+
+        size_t dim1ThreadCount = divup(recomWorkGroupSize, dim0ThreadCount);
+
+        // make sure you are not launching what the device can't handle
+        dim1ThreadCount = std::min(getDevice(device).getInfo<CL_DEVICE_MAX_WORK_GROUP_SIZE>(), dim1ThreadCount);
```
pavanky: Same as above

src/backend/opencl/program.cpp
9prady9: I agree, this is not a single bullet for all cases. It is just like many other convenience functions we have internally. We can have a better name and cautionary comments before that function explaining where it should be used and what it does.

Don't you think the number returned by `clGetKernelWorkGroupInfo` for the param name `CL_KERNEL_WORK_GROUP_SIZE` takes into account the static `local` allocations declared in the kernel ?

src/backend/opencl/program.cpp
9prady9: I can't give you example case right away, but i added it as a cautionary check to warrant unexpected values.

9prady9: Ok

src/backend/opencl/program.cpp
pavanky: If the constant is defined in the kernel itself, it should be fine. If it is a compile time option (based on the threads) that is also being used for the launch configuration, it is going to be an issue (for example, checkout convolve). 

One solution would be to always have dynamically allocated `__local` memory in OpenCL, but this may not give the best performance.

9prady9: Agree, but having this chuck of code repeated every where we want to compute the 2D local launch configuration is not an ideal solution either. Any suggestions for better name ?

9prady9:
@arrayfire/core-devel Any suggestions about how should we make type of device information available ? (referring the checklist in #119 )

One possible way i can think of is through enums.

pavanky:
@9prady9 are you talking about the info being available internally  or externally ?

9prady9:
Good question, i was asking about internally  earlier, but do we need such information externally as well ?

9prady9:
@arrayfire/core-devel What are the next steps for this PR ?

9prady9:
@arrayfire/core-devel What are the next steps for this PR ?

shehzan10:
build arrayfire linux ci
build arrayfire osx opencl ci

src/backend/opencl/program.cpp
umar456: I agree with @pavanky. I don't think we can use this function in it's current state. We need to have a function LIKE this but not in its current implementation.

Many of the kernels we are using have hard coded values they expect for the NDRange. Of course this is not ideal because each device is different. This function should have a lookup table that takes into account the device and the kernel to return an appropriate value.

An even better solution would return a kernel and its configuration given a generic name for the kernel and its device. If there are no specializations for a specific kernel a default version would be returned.

9prady9:
I don't think lookup table is a good solution because of the following reason - It is hard to keep track of all the devices being released into the market. We may specialise for some known(ones we have access to) cards and default to a particular launch configuration for unknown devices, but it would be better to have a solution that can scale with the new devices being released into the market without the developer's intervention - an algorithm that can use device information & kernel requirements and generate an appropriate launch configuration for that kernel on that specific device.

@umar456 Can you elaborate on your last idea ? What did you mean by generic name ?

umar456:
@9prady9 Yes and if we don't have a known device we can do what you are saying. As you know OpenCL is not performance portable, so we can't rely on the settings to get the correct configuration. This is also true for kernels. Here is what I think should happen:

* Kernels should be vendor/device type specific
* Launch configuration should be vendor/device type specific
* If there is no entry for the device/vendor use a default kernel/configuration

pavanky:
@9prady9 @umar456 Device and Kernel specific lookup tables are not going to scale. I completely disagree with that idea. 

The goal of the original issue was to look at key device specific configurations and categorize them into a known number of categories like the following:

- CPU vs GPU
- Scalar support vs Vector support
- Has local memory or not
- OpenCL 2.0 support (this can be important for a bunch of kernels)

This information can be used to write one kernel that behaves differently on different devices or write multiple kernels.

Automatically figuring out launch groups and threads is the *next* step after this. The performance gains made here are going to be relatively modest compared to the previous set of improvements. Even here, I disagree with having a lookup table for each kernel and device. We need to have a well defined set of parameters that a device can fall in to.

Some of those categories should be:

1. The ideal number of threads per group along dim 0
2. The ideal number of total groups 
3. Amount of shared memory available.
4. Number of queues it can support concurrently (this will be useful in the really long run).

This is not an exhaustive list. But it is a good start. It may also end up us requiring to do *some* hard coding into device / vendor specific lookup tables. But it is not / should not kernel dependent. The launch configurations can be figured out within each kernel based on the parameters for each device.

If we have a known set of parameters, they can also be exposed to the user for fine tuning in case we get it wrong.

9prady9:
Agree on all the points. So, what are the take away goals here ?

Following are the new information that a developer can query in OpenCL backend right now.

- CPU vs GPU -> `getDeviceType()`
- Scalar support vs Vector support -> `preferredVectorWidth()`
- Has local memory or not -> `getLocalMemType()` & `isDedicatedLocalMem()`
- Amount of total shared/local memory available -> `getDeviceLocalMemorySize()`
- `getDeviceWorkGroupSize()` returns the maximum work group size on the device for a specific kernel, note that this is not same as maximum global work group size.
- `getDeviceHardwareSIMDWidth()` returns hardware SIMD width a.k.a warp value, wavefront value for CUDA & NVIDIA hardware respectively.
- OpenCL 2.0 support - there is no function right now that checks this explicitly, we can add it.

Should this PR limit to above information query functions and deal with figuring out launch configuration related changes for later ?

pavanky:
@umar456 @9prady9 Another set of functions that can sit on top of this could use the fact that many kernels use similar access patterns. There could be functions that returned kernel launch configurations (based on parameters mentioned earlier) for the following types of categories:

- Kernels that operate on vectors (scan, sum, etc. along dim0, along dim other.)
- Kernels that operate on windows (1D, 2D & 3D)
- Kernels that transform images / signals (approx2, transform, rotate, resize)

umar456:
Closing this PR for now. We will do this work in a future PR.

PR #1629, opened 2016-10-17
OS X Installer Fixes

shehzan10:
* Fixes Forge `share/Forge/*` being installed both in `/usr/local/*` and /usr/local/share/Forge/*` (the former was wrong).
* Fixes `postinstall` script not working due to not having +x enabled.
* Fixes how user is set in `postinstall` script.
* Fixes installation errors due to brew installation commands for packages that may have already been installed.
  * This is done by checking if packages are already installed and then installing only if needed.
* **Fixes library symlinks becoming real files on their own. The copy from install directory to temp directory for symlinks was copying the files they point to rather than create symlinks. (No wonder the installers were so big in size). The installer size has become a 3rd of what it used to be.**

[skip arrayfire ci]

These fixes have been retroactively been applied to the ArrayFire-v3.4.1_OSX.pkg installer available on the arrayfire.com/download site.


CMakeModules/osx_install/forge_scripts/postinstall
```diff
@@ -14,13 +14,16 @@ if [ ! -f $brew ]; then
     exit 1
 fi
 
-user=$(ps aux | grep console | grep -v 'grep\|root' | cut -d' ' -f1 | head -n1)
+#user=$(ps aux | grep console | grep -v 'grep\|root' | cut -d' ' -f1 | head -n1)
+user=$(stat -f '%Su' $HOME)
```
mlloreda: :+1:

CMakeModules/osx_install/OSXInstaller.cmake
```diff
@@ -261,10 +296,10 @@ IF(BUILD_GRAPHICS)
     PKG_BUILD(  PKG_NAME        ForgeLibrary
                 DEPENDS         OSX_INSTALL_SETUP_FORGE_LIB
                 TARGETS         forge_lib_package
-                INSTALL_LOCATION /usr/local/
+                INSTALL_LOCATION /usr/local/lib
```
umar456: This doesn't look right. Was it really installing the libraries in the /usr/local directory? I don't see a change in the PKG_BUILD function that is associated with this change.

umar456: @shehzan10 clarified that this is indeed correct.

PR #1630, opened 2016-10-17
FEAT Function to return allocated bytes

### LINKED ISSUE #1607: Create a function to return allocated bytes
Currently there is no real way to query the number of bytes being held by the original array after doing something like this:

```
array a = randu(5,5);
array b = a(af::seq(1,3), af::seq(1,3));
//b.allocated() ?
```

shehzan10:
Fixes #1607

### DIFF
--- a/src/api/c/internal.cpp
+++ b/src/api/c/internal.cpp
@@ -177,18 +177,18 @@ af_err af_get_allocated_bytes(size_t *bytes, const af_array arr)
         size_t res = 0;
 
         switch (ty) {
-        case f32: res = getArray<float  >(arr).getDataDims().elements() * sizeof(float);            break;
-        case f64: res = getArray<double >(arr).getDataDims().elements() * sizeof(double);           break;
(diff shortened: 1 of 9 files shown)


src/api/c/internal.cpp
```diff
@@ -168,3 +168,32 @@ af_err af_is_owner(bool *result, const af_array arr)
...
+
+af_err af_get_allocated_bytes(size_t *bytes, const af_array arr)
+{
+    try {
+        af_dtype ty = getInfo(arr).getType();
+
+        size_t res = 0;
+
+        switch (ty) {
+        case f32: res = getArray<float  >(arr).getDataDims().elements() * sizeof(float);            break;
```
umar456: Can you move this into the backend? This type of functionality should not be exposed in the API layer.

pavanky: This file is located misleadingly. This should have ideally been under `src/backend/internal.cpp`. All backends use this file internally so this is technically part of the backend code, not the api code.

shehzan10:
build arrayfire linux ci

PR #1634, opened 2016-10-24
use scoped_array instead of scoped_ptr when managing array resources

vakopian:
There are several places in the code where `scoped_ptr` is used to clean memory allocated as an array. Instead `scoped_array` should be used.

### DIFF
--- a/src/backend/cuda/jit.cpp
+++ b/src/backend/cuda/jit.cpp
@@ -45,10 +45,10 @@
 #include <vector>
 #include <nvvm.h>
 #include <boost/functional/hash.hpp>
-#include <boost/scoped_ptr.hpp>
+#include <boost/scoped_array.hpp>
 
 using std::vector;
-using boost::scoped_ptr;
+using boost::scoped_array;
 
 namespace cuda
 {
@@ -409,7 +409,7 @@ static char *irToPtx(string IR, size_t *ptx_size)
         size_t log_size = 0;
         nvvmGetProgramLogSize(prog, &log_size);
         printf("%ld, %zu\n", IR.size(), log_size);
-        scoped_ptr<char> log(new char[log_size]);
+        scoped_array<char> log(new char[log_size]);
         nvvmGetProgramLog(prog, log.get());
         printf("LOG:\n%s\n%s", log.get(), IR.c_str());
         NVVM_CHECK(comp_res, "Failed to compile program");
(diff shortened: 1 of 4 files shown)


shehzan10:
@vakopian Since this would go into a hotfix, can you base your changes off of hotfix-3.4.2 branch? You can force push to your branch.

pavanky:
`boost:scoped_ptr` was used because CUDA did not support `c++11` back in the day. `c++11` is turned on by default under certain circumstances.

Perhaps it is time to optionally enable `shared_ptr` if `__cplusplus > 199711L` . This can be done by doing something like this in a header file

```
#if __cplusplus > 199711L
#include <memory>
#define cuda_shared_ptr std::shared_ptr
#else
#include <boost/scoped_array.hpp>
#define cuda_shared_ptr boost:scoped_array
#endif
```

And then use `cuda_shared_ptr` instead of `scoped_array`

umar456:
These are great changes. Thank you for your contributions. I agree with @pavanky about moving towards the standard libraries for the newer versions of CUDA. Could you add the macro @pavanky suggested in his comment?

vakopian:
@pavanky Unless I'm missing something, `std::shared_ptr` is still not the correct way to cleanup these arrays, because `std::shared_ptr` doesn't have an array specialization (so it will still call `delete` instead of `delete[]`). 
Perhaps `std::unique_ptr` would work (that one does have a specialization for arrays), but only if these arrays are truly not shared

umar456:
Ahh yes. @vakopian is right. Only `unique_ptr` has the specialization for array objects. The standard only added the array specialization for `shared_ptr` in C++17.

@vakopian we need the arrays to be reference counted because they will be shared amongst other `af_array`s in the form of views. We can either use the `scoped_array` object or create an array deleter which would call `delete[]` during destruction. I am in the favor of using the second approach.

vakopian:
@umar456 the second approach is fine, but the `#define` for the older versions will need to take that into account. So instead of
`#define cuda_shared_ptr boost::scoped_array`
we'll need
`#define cuda_shared_ptr(p, deleter) boost::scoped_array(p)`

Then when using it we'll have to do something like this:
`cuda_shared_ptr<const char> ptx(irToPtx(jit_ker, &ptx_size), array_deleter<const char>());`

Feels too much trouble to just avoid the use of `boost::scoped_array`

umar456:
Thats a good point. I think your current approach is fine. Once we deprecate the older CUDA toolkits we can do a cleanup remove these at that time.

This actually troubles me. I didn't think we were leaking memory. I guess the standard library was handling this case even though it is not standard behavior.

vakopian:
I think It's more an undefined behavior than a leak (which is arguably a more serious issue). I only discovered it by running under valgrind. So it might be a good idea to make some small valgrind tests part of your standard unit testing

shehzan10:
@vakopian this was merged by mistake into devel. However, I have undone the commit.
Can you open a new PR to point to `hotfix-3.4.2`?

vakopian:
here it is for `hotfix-3.4.2`: https://github.com/arrayfire/arrayfire/pull/1637

PR #1636, opened 2016-10-26
Revert "use scoped_array instead of scoped_ptr when managing array resources"

umar456:
Reverts arrayfire/arrayfire#1634


pavanky:
Something broke ?

PR #1639, opened 2016-11-02
Revert to setting chart axes limits automatically

shehzan10:
* The axes limits will always be the minimum/maximum of the current and new limit.
* The user can still set limits from API calls
  * If the user sets a limit from the API call, then the automatic limit setting will be disabled.
* This PR requires https://github.com/arrayfire/forge/pull/120

TODO:
Update the forge build tag to a version once the version tag is created in Forge.

### DIFF
--- a/CMakeModules/build_forge.cmake
+++ b/CMakeModules/build_forge.cmake
@@ -42,13 +42,13 @@ ELSE()
     SET(byproducts BYPRODUCTS ${forge_location})
 ENDIF()
 
-SET(FORGE_VERSION 0.9.0)
+SET(FORGE_VERSION 0.9.1)
 
 # FIXME Tag forge correctly during release
 ExternalProject_Add(
     forge-ext
     GIT_REPOSITORY https://github.com/arrayfire/forge.git
-    GIT_TAG v${FORGE_VERSION}
+    GIT_TAG devel
     PREFIX "${prefix}"
     INSTALL_DIR "${prefix}"
     UPDATE_COMMAND ""
--- a/src/api/c/graphics_common.cpp
+++ b/src/api/c/graphics_common.cpp
@@ -233,15 +233,15 @@ void ForgeManager::setWindowChartGrid(const forge::Window* window,
 {
     ChartMapIter iter = mChartMap.find(window);
 
-        if(iter != mChartMap.end()) {
-
-    }
-
     if(iter != mChartMap.end()) {
         // ChartVec found. Clear it.
         // TODO: Should we clear this even if r = old_r and c = old_c?
-        for(int i = 0; i < (int)(iter->second).size(); i++)
-            if((iter->second)[i] != NULL) delete (iter->second)[i];
+        for(int i = 0; i < (int)(iter->second).size(); i++) {
+            if((iter->second)[i] != NULL) {
+                delete (iter->second)[i];
+                mChartAxesOverrideMap.erase((iter->second)[i]);
+            }
+        }
         (iter->second).clear();
     }
 
@@ -271,6 +271,9 @@ forge::Chart* ForgeManager::getChart(const forge::Window* window, const int r, c
             // Chart has not been created
             chart = new forge::Chart(ctype);
             (iter->second)[c * gRows + r] = chart;
+
+            // Set Axes override to false
+            mChartAxesOverrideMap[chart] = false;
         }
     } else {
         // The chart map for this was never created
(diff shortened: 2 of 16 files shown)


shehzan10:
@arrayfire/core-devel if you believe there could be a better heuristic for setting the computed axes limits, we can integrate that.

CMakeModules/build_forge.cmake
```diff
@@ -42,13 +42,13 @@ ELSE()
...
 
-SET(FORGE_VERSION 0.9.0)
+SET(FORGE_VERSION 0.9.1)
 
 # FIXME Tag forge correctly during release
 ExternalProject_Add(
     forge-ext
     GIT_REPOSITORY https://github.com/arrayfire/forge.git
-    GIT_TAG v${FORGE_VERSION}
+    GIT_TAG devel
```
9prady9: @shehzan10 I have merged forge PR, please change it back to `${FORGE_VERSION}`. I wanted to talk about this for a while now. Having devel here is not good because if any breaking change of forge has been pushed to devel in the mean while. It cause build failures in arrayfire until related changes are incorporated into arrayfire. We should either keep it as tag or complete commit number.

9prady9:
(forge PR)[https://github.com/arrayfire/forge/pull/120] is taken care of. This PR is ready w.r.t that change.

src/api/c/graphics_common.cpp
```diff
@@ -178,15 +233,15 @@ void ForgeManager::setWindowChartGrid(const forge::Window* window,
...
-        if(iter != mChartMap.end()) {
-
-    }
-
     if(iter != mChartMap.end()) {
         // ChartVec found. Clear it.
         // TODO: Should we clear this even if r = old_r and c = old_c?
-        for(int i = 0; i < (int)(iter->second).size(); i++)
-            if((iter->second)[i] != NULL) delete (iter->second)[i];
+        for(int i = 0; i < (int)(iter->second).size(); i++) {
```
9prady9: We can avoid clearing if grid size is same but i believe we should reset the chart properties to default because the new charts might be of same properties.

src/api/c/graphics_common.cpp
```diff
@@ -366,6 +424,24 @@ forge::VectorField* ForgeManager::getVectorField(forge::Chart* chart, int nPoint
     return mVcfMap[keypair];
 }
 
+bool ForgeManager::getChartAxesOverride(forge::Chart* chart)
+{
+    ChartAxesOverrideIter iter = mChartAxesOverrideMap.find(chart);
+    if (iter == mChartAxesOverrideMap.end()) {
+        AF_ERROR("Chart Not Found!", AF_ERR_ARG);
```
9prady9: Isn't this error code reserved for API level arguments ? Charts are not exposed to user and the current callee is also not exposed externally. Therefore, i think it should be `AF_ERR_INTERNAL`. The string message can leave more clues if needed.

src/api/c/graphics_common.cpp
shehzan10: I think clearing is a fair solution for now. There is no API in forge that would "clear" the properties as you have mentioned - which would probably do the same thing as delete and create anyway. That is, delete all the objects associated with the chart and reset axes limits/titles.

src/api/c/graphics_common.cpp
shehzan10: Yeah you are right, I'll change that.
In any case, it should never execute that statement.

src/api/c/graphics_common.cpp
9prady9: The only properties users can change are axes titles and limits, just setting them to default values does the trick. But of course, there is no convenience function to reset all values in one go, we have to call `setAxesLimits` and `setAxesTitles` individually.

shehzan10: It's not just that, you will have to remove the association between the chart and the render-able objects.

9prady9: Ah, right - forgot about those.

src/api/c/plot.cpp
```diff
@@ -64,6 +64,43 @@ forge::Chart* setup_plot(const forge::Window* const window, const af_array in_,
...
+        && cmin[2] == 0 && cmax[2] == 0) {
+            // No previous limits. Set without checking
+            cmin[0] = step_round(dmin[0], false);
+            cmax[0] = step_round(dmax[0], true);
+            cmin[1] = step_round(dmin[1], false);
+            cmax[1] = step_round(dmax[1], true);
+            if(order == 3) cmin[2] = step_round(dmin[2], false);
+            if(order == 3) cmax[2] = step_round(dmax[2], true);
+        } else {
+            if(cmin[0] > dmin[0])       cmin[0] = step_round(dmin[0], false);
```
shehzan10: This is basically to ensure that the limits can only grow and not shrink. So minimum can only become lesser and maximum can only become greater.

Issue #133, opened 2014-11-16
Check for launch configuration limits in CUDA backend.


pavanky:
Verify against [`cudaDeviceProp`](http://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp) instead of a hard coded value like 65535.

pavanky:
@umar456 This is only going to be a limitation for pre-kepler cards. Do we need to worry about the issue or leave it as a known issue ?

umar456:
We should support everything CUDA supports. At least things that are not deprecated in the new versions.

pavanky:
hmm  but it only affects Fermi GPUs in weird configurations (eg: very tall). The architecture is 5 years old at this point and adding support around the limitation requires quite a bit of change in code.

pavanky:
I stand corrected. The limit for y and z dimensions is still 65535 and we hit this limit for certain configurations. We should look into fixing this properly.

pavanky:
@shehzan10 do you want to take a shot at figuring out how to deal with this ?

vakopian:
I started hitting this bug recently too. Other reports have indicated a large second dimension being the culprit, but I think I have an example without a large second dimension:
```
    int sz = 8000;
    array A = randu(1000000, 3); // 1000000x3
    array idx = tile(randu(sz), 1, 3, 300); // 8000x3x300
    array B = A(idx, span); // 7200000x3
```
This produces
```
terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Internal error:998):
In function void cuda::evalNodes(std::vector<cuda::Param<T> >&, std::vector<cuda::JIT::Node*>) [with T = unsigned int]
In file src/backend/cuda/jit.cpp:658
CU Error (1)

In function af::array::array_proxy::operator af::array()
In file src/api/cpp/array.cpp:705
Signal: SIGABRT (Aborted)
```
It does work if I reduce `sz` to `6000`.
af_info:
```
ArrayFire v3.4.2 (CUDA, 64-bit Linux, build 801f40f)
Platform: CUDA Toolkit 8, Driver: 367.48
[0] Quadro K2200, 4042 MB, CUDA Compute 5.0
```

InBetweenNames:
What would it take to get this issue fixed in ArrayFire?  I'm able to work around it by only processing up to 65535 elements at once, but it's a bit of a hassle to do and it affects code clarity.

[...]

PR #1641, opened 2016-11-08
CUDA Launch configuration fixes for memcopy and join

### LINKED ISSUE #1513: memcopy and join have launch configuration issues
Hello,
I'm running the following code that exploits wrap/unwrap to mix/unmix channels of an rgb image.

``` cpp
#include <iostream>

#include <arrayfire.h>

af::array mix_chans(const af::array& a, const af::array& b, const af::array& c) {
    unsigned d0 = a.dims(0);
    unsigned d1 = a.dims(1);
    auto a_ = af::moddims(a, 1, d0*d1);
    auto b_ = af::moddims(b, 1, d0*d1);
    auto c_ = af::moddims(c, 1, d0*d1);
    auto u = af::join(0, a_, b_, c_); 
    return af::wrap(u, d0*3, d1, 3, 1, 3, 1, 0, 0);
}
void unmix_chans(const af::array& abc, af::array& a, af::array& b, af::array& c) {
    unsigned d0 = abc.dims(0)/3;
    unsigned d1 = abc.dims(1);
    auto abc_ = af::flat(abc);
    auto u = af::unwrap(abc_, 3, 1, 3, 1, 0, 0);
    a = af::moddims(u(0, af::span, af::span, af::span), d0, d1);
    b = af::moddims(u(1, af::span, af::span, af::span), d0, d1);
    c = af::moddims(u(2, af::span, af::span, af::span), d0, d1);
}

int main() {
    int w = 1920;
    int h = 1080;
    af::array r = af::randu(w, h);
    af::array g = af::randu(w, h);
    af::array b = af::randu(w, h);
    af::array rgb = mix_chans(r, g, b);
    af::array r_,g_,b_;
    unmix_chans(rgb, r_, g_, b_);
    std::cerr << af::max<double>(af::abs(r - r_)) << " " <<
                 af::max<double>(af::abs(g - g_)) << " " <<
                 af::max<double>(af::abs(b - b_)) << std::endl;
}
```

I get the following CUDA/AF Internal error:

```
In function memcopy
In file src/backend/cuda/kernel/memcopy.hpp:85
CUDA Error (9): invalid configuration argument


In function af::array af::moddims(const af::array&, unsigned int, const dim_t*)
In file src/api/cpp/data.cpp:265
Aborted
```

When dimensions of the image are halved then everything runs fine.
I'm using ArrayFire v3.3.2 (CUDA, 64-bit Linux, build f65dd97).

shehzan10:
This is, of course, a semi-permanent fix. Although it should work in all scenarios for these 2 kernels, it is not complete, as it does not fix other kernels with the same problems.

This should ideally become a function that kernels can call on to fix the launch config. This function wont be applied everywhere as some kernels do things differently. But it can be used across a majority of the kernels.

Fixes #1513 
Ref #133 

[skip arrayfire ci]


shehzan10:
build arrayfire linux ci
build arrayfire windows ci
build arrayfire osx cuda ci

shehzan10:
build arrayfire windows ci

shehzan10:
build arrayfire windows ci

src/backend/cuda/kernel/memcopy.hpp
```diff
@@ -26,34 +26,43 @@ namespace kernel
...
+        T *d_out = out;
+        T const *d_in = in;
+
+        if(iz < idims.dim[2] && iw < idims.dim[3]) {
+            d_out = d_out + iz * ostrides.dim[2] + iw * ostrides.dim[3];
+            d_in  = d_in  + iz * istrides.dim[2] + iw * istrides.dim[3];
+
+            for (int iy = yy; iy < idims.dim[1]; iy += incy) {
+                T const *d_in_ = d_in + iy * istrides.dim[1];
+                T *d_out_ = d_out + iy * ostrides.dim[1];
```
umar456: `d_out_` and `d_out` can be confusing if you are not careful. Rename these variable so they are more differentiable.

src/backend/cuda/kernel/memcopy.hpp
```diff
@@ -63,24 +72,48 @@ namespace kernel
...
-        // FIXME: DO more work per block
-        uint blocks_x = divup(idims[0], threads.x);
-        uint blocks_y = divup(idims[1], threads.y);
+        dim3 blocks(blocksPerMatX * idims[2],
+                    blocksPerMatY * idims[3],
+                    1);
+
+        // This X case is mostly for fermi cards which have a limit of 65 * 1024 (65535)
+        // Kepler and newer have a limit of 2 * 1024 * 1024 * 1024 (2B)
+        // which is unlikely to be every hit
```
umar456: every-> ever

src/backend/cuda/kernel/join.hpp
```diff
@@ -73,6 +73,15 @@ namespace cuda
                         blocksPerMatY * X.dims[3],
                         1);
 
+            int maxBlocksY = cuda::getDeviceProp(cuda::getActiveDeviceId()).maxGridSize[1];
+            // Increase the work per block until the condition is satisfied
+            // Increase threads only upto 512
+            while(blocks.y > maxBlocksY) {
```
shehzan10: I made the PR description to suggest better ways to do this. Right now, this is the simplest way to get rid of this bug. We definitely need a more permanent solution for this. But until we figure out what that is, this will work.

src/backend/cuda/kernel/join.hpp
```diff
@@ -73,6 +73,15 @@ namespace cuda
...
+                if(threads.x * threads.y < 512) threads.y *= 2;
```
pavanky: What happens when this condition becomes false ? Is the kernel iterating a bunch of times ?

src/backend/cuda/kernel/join.hpp
shehzan10: I had a similar version first https://github.com/arrayfire/arrayfire/pull/1641/commits/fda93803120241e8fd923b00cf5826e91e7f199a. It did not take into account that if blocks.y remains greater than maxBlocksY, the error will remain. Hence the while loop.

src/backend/cuda/kernel/join.hpp
shehzan10: This if statement only ensures that the maximum threads per block is 512. When this becomes false, we simply keep increasing the work per thread to get into limits of max blocks.

src/backend/cuda/kernel/join.hpp
pavanky: You were dividing old `blocksPerMatY` by  `2`, this one divides the `maxBlockSize` by `X.dims[3]` which ensures that it falls within the given size.

src/backend/cuda/kernel/join.hpp
shehzan10: Yes, that is the basis of this kind of loop. It assumes that the kernels operate in a specific manner for X and Y.
The X thread/blocks will operate on 0 and 2 dims, Y operates on 1 and 3 with threads doing multiple elements. Only those kind of kernels can use this method of reducing no. of blocks. So once we hit 512 threads, we simply increase the number of iterations per thread.
It's definitely not universal.

pavanky: Just double checking. Anyway you can increase the number of threads to 1024 since 2.0 support is effectively dropped.

src/backend/cuda/kernel/join.hpp
pavanky: @shehzan10 The for loop inside the kernel should work even if `blocksPermatY` was `1`.

pavanky:
@shehzan10 can you rebase the branch and resolve the conflicts ?

mlloreda:
Work on this has moved over to #1807.

Issue #821, opened 2015-06-15
Sparse matrix support for all backends


pavanky:
Figure out the list of alternatives that exist and implement the following functions:

- [x] Sparse matrix Creation
  - [x] From Dense
  - [x] From CSR arrays
  - [x] From COO arrays
- [ ] Format Conversion
  - [x] To Dense
  - [ ] CSR <--> COO
- [x] Matrix multiply 
  - [x] CSRMM (sparse matrix, dense matrix with options for sparse)
  - [x] CSRMV (sparse matrix, dense vector with options for sparse)
- [x]  ~~Sparse Matrix Transpose (Same as CSR --> CSC)~~ Transpose is an available option for matmul.

Use [cuSparse](http://docs.nvidia.com/cuda/cusparse/) for CUDA, [clSparse](https://github.com/clMathLibraries/clSPARSE) for OpenCL.
CPU can use either MKL or fallback. For distribution installers, we will use MKL, with the possibility of a fallback option which will need to be implemented from scratch.

pavanky:
@shehzan10 Can you look at the potential library options for CPU sparse support?

Some of these functions are fairly easy to implement. We could implement our own versions if there are not any standard free libraries out there.

shehzan10:
The enum for the type of storage will be `af_sparse_type` (could also use `af_sparse_matrix_type` but I think that will be too long).

`af_sparse_type` will contain `dense`, `COO`, `CSR`, `CSC`. Do we need powers of 2 ordering for this or will a natural number sequence be enough?
Note: CSC will not be supported on OpenCL unless,
* clSparse starts supporting it
* We use the CPU fallback implementation that we will implement.

The C-struct `af_sparse_array` will be internal and will be exposed as a void*. This will allow more flexibility in the future.
The C++ API will contain a new class `af::sparse` (or) `af::sparse_array`. It's members will be the struct pointed, the type of storage, functions to handle common tasks like get().

`af::matmul` will be overloaded to handle this sparse input, and a new C function `af_matmul_sparse` will be added. I don't think we need a dot operation as well.
Will matmul will have the restriction of running the same storage types?

Will sparse functions also need to support transposing options?

ghisvail:
> Use cuSparse for CUDA, clSparse for OpenCL.

Just as an FYI, `clSPARSE` is on my packaging backlog but it is not compatible with OpenCL 2.0 yet which is Debian's default. You guys might want to push for it upstream?

pavanky:
[...]

PR #1642, opened 2016-11-10
FEAT: Add CSR <--> COO Conversion in all backends

shehzan10:
* Directly convert between `AF_STORAGE_COO` <--> `AF_STORAGE_CSR` using the `sparseConvertTo` function.
  * CPU and OpenCL use native code (no libraries)
  * CUDA uses cusparse
* Fixed a bug in `COO` row/col indices from dense
* Add tests

Fixes CSR <-> COO in #821 

[skip arrayfire ci]

### DIFF
--- a/src/backend/cpu/sparse.cpp
+++ b/src/backend/cpu/sparse.cpp
@@ -339,18 +339,52 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in_)
 ////////////////////////////////////////////////////////////////////////////////
 // Common to MKL and Not MKL
 ////////////////////////////////////////////////////////////////////////////////
-template<typename T, af_storage src, af_storage dest>
+template<typename T, af_storage dest, af_storage src>
 SparseArray<T> sparseConvertStorageToStorage(const SparseArray<T> &in)
 {
-    // Dummy function
-    // TODO finish this function when support is required
-    AF_ERROR("CPU Backend only supports Dense to CSR or COO", AF_ERR_NOT_SUPPORTED);
-
     in.eval();
 
-    SparseArray<T> dense = createEmptySparseArray<T>(in.dims(), (int)in.getNNZ(), dest);
+    SparseArray<T> converted = createEmptySparseArray<T>(in.dims(), (int)in.getNNZ(), dest);
+    converted.eval();
 
-    return dense;
+    if(src == AF_STORAGE_CSR && dest == AF_STORAGE_COO) {
+
+        auto func = [=] (SparseArray<T> out, const SparseArray<T> in_) {
+            Array<T  > ovalues = out.getValues();
+            Array<int> orowIdx = out.getRowIdx();
+            Array<int> ocolIdx = out.getColIdx();
+
+            Array<T  > ivalues = in_.getValues();
+            Array<int> irowIdx = in_.getRowIdx();
+            Array<int> icolIdx = in_.getColIdx();
+
+            kernel::csr_coo<T>()(ovalues, orowIdx, ocolIdx, ivalues, irowIdx, icolIdx);
+        };
+
+        getQueue().enqueue(func, converted, in);
+
+    } else if (src == AF_STORAGE_COO && dest == AF_STORAGE_CSR) {
+
+        auto func = [=] (SparseArray<T> out, const SparseArray<T> in_) {
+            Array<T  > ovalues = out.getValues();
+            Array<int> orowIdx = out.getRowIdx();
+            Array<int> ocolIdx = out.getColIdx();
+
+            Array<T  > ivalues = in_.getValues();
+            Array<int> irowIdx = in_.getRowIdx();
+            Array<int> icolIdx = in_.getColIdx();
+
+            kernel::coo_csr<T>()(ovalues, orowIdx, ocolIdx, ivalues, irowIdx, icolIdx);
+        };
+
+        getQueue().enqueue(func, converted, in);
+
+    } else {
+        // Should never come here
+        AF_ERROR("CPU Backend invalid conversion combination", AF_ERR_NOT_SUPPORTED);
+    }
+
+    return converted;
 }
 
 #define INSTANTIATE_TO_STORAGE(T, S)                                                                        \
--- a/src/backend/cpu/sparse.cpp
+++ b/src/backend/cpu/sparse.cpp
@@ -119,7 +119,7 @@ SPARSE_FUNC(csrcsc, cdouble,z)
 #endif // USE_MKL
 
 ////////////////////////////////////////////////////////////////////////////////
-// Common to MKL and Not MKL
+// Common Funcs for MKL and Non-MKL Code Paths
 ////////////////////////////////////////////////////////////////////////////////
 
 // Partial template specialization of sparseConvertDenseToStorage for COO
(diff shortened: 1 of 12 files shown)


shehzan10:
build arrayfire windows ci

src/backend/cpu/sparse.cpp
```diff
@@ -339,18 +339,52 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in_)
...
-    AF_ERROR("CPU Backend only supports Dense to CSR or COO", AF_ERR_NOT_SUPPORTED);
-
     in.eval();
 
-    SparseArray<T> dense = createEmptySparseArray<T>(in.dims(), (int)in.getNNZ(), dest);
+    SparseArray<T> converted = createEmptySparseArray<T>(in.dims(), (int)in.getNNZ(), dest);
+    converted.eval();
 
-    return dense;
+    if(src == AF_STORAGE_CSR && dest == AF_STORAGE_COO) {
```
umar456: This whole function could be replaced with something like this:
```
function<void, Array<T>, Array<int>, Array<int>,
 +                    Array<T> const, Array<int> const, Array<int> const> func;
if(src == AF_STORAGE_CSR && dest == AF_STORAGE_COO) {
  func = kernel::csr_coo<T>::operator();
} else if (src == AF_STORAGE_COO && dest == AF_STORAGE_CSR) {
  func = kernel::coo_csr<T>::operator();
}
...

getQueue().enqueue(func, converted.getValues(), converted.getRowIdx(), ...);
```

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+        size_t pBufferSizeInBytes = 0;
+        CUSPARSE_CHECK(cusparseXcoosort_bufferSizeExt(
+                        getHandle(),
+                        in.dims()[0], in.dims()[1], nNZ,
+                        converted.getRowIdx().get(), converted.getColIdx().get(),
+                        &pBufferSizeInBytes));
+        char *pBuffer = memAlloc<char>(pBufferSizeInBytes);
+
+        int *P = memAlloc<int>(nNZ);
+        CUSPARSE_CHECK(cusparseCreateIdentityPermutation(getHandle(), nNZ, P));
```
umar456: Memory leak on error

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+
+        CUSPARSE_CHECK(cusparseXcoosortByColumn(
```
umar456: Memory leak on error

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+                        getHandle(),
+                        in.dims()[0], in.dims()[1], nNZ,
+                        converted.getRowIdx().get(), converted.getColIdx().get(),
+                        P, (void*)pBuffer));
+
+        CUSPARSE_CHECK(gthr_func<T>()(
```
umar456: memory leak on error

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+            size_t pBufferSizeInBytes = 0;
+            CUSPARSE_CHECK(cusparseXcoosort_bufferSizeExt(
+                            getHandle(),
+                            cooT.dims()[0], cooT.dims()[1], nNZ,
+                            cooT.getRowIdx().get(), cooT.getColIdx().get(),
+                            &pBufferSizeInBytes));
+            char *pBuffer = memAlloc<char>(pBufferSizeInBytes);
+
+            int *P = memAlloc<int>(nNZ);
+            CUSPARSE_CHECK(cusparseCreateIdentityPermutation(getHandle(), nNZ, P));
```
umar456: Memory leak on error.

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+
+            CUSPARSE_CHECK(cusparseXcoosortByRow(
```
umar456: memory leak on error

src/backend/cuda/sparse.cu
```diff
@@ -340,14 +361,121 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in)
...
+                            getHandle(),
+                            cooT.dims()[0], cooT.dims()[1], nNZ,
+                            cooT.getRowIdx().get(), cooT.getColIdx().get(),
+                            P, (void*)pBuffer));
+
+            CUSPARSE_CHECK(gthr_func<T>()(
```
umar456: Memory leak on error

src/backend/cpu/sparse.cpp
```diff
@@ -339,18 +339,52 @@ Array<T> sparseConvertStorageToDense(const SparseArray<T> &in_)
 ////////////////////////////////////////////////////////////////////////////////
 // Common to MKL and Not MKL
```
shehzan10: There are 2 code paths: One that uses MKL and one that is used when compiled without MKL support. This section is common to both as it does not depend on MKL.

src/backend/cuda/sparse.cu
```diff
@@ -145,6 +145,21 @@ struct nnz_func_def_t
                                               int *, int *);
 };
 
+//cusparseStatus_t cusparseZgthr(cusparseHandle_t handle,
```
shehzan10: This is in line with other libraries/code we are using where we keep a commented version of the function declaration around as reference.

src/backend/cuda/sparse.cu
```diff
@@ -145,6 +145,21 @@ struct nnz_func_def_t
...
+//                               int nnz,
+//                               const cuDoubleComplex *y,
+//                               cuDoubleComplex *xVal, const int *xInd,
+//                               cusparseIndexBase_t idxBase)
+template<typename T>
+struct gthr_func_def_t
+{
+    typedef cusparseStatus_t (*gthr_func_def)(cusparseHandle_t,
```
shehzan10: The above comments are there for this purpose.

src/backend/cuda/sparse.cu
shehzan10: The memory manager should delete it on error right?

test/sparse.cpp
```diff
@@ -226,3 +226,77 @@ CREATE_TESTS(AF_STORAGE_CSR)
...
+    TEST(SPARSE_CONVERT, T##_##DTYPE##_##STYPE##_##SUFFIX)                      \
+    {                                                                           \
+        sparseConvertTester<T, DTYPE, STYPE>(M, N, F);                          \
+    }                                                                           \
+
+#define CONVERT_TESTS(T, STYPE, DTYPE)                                          \
+    CONVERT_TESTS_TYPES(T, STYPE, DTYPE, 1, 1000, 1000,  5)                     \
+    CONVERT_TESTS_TYPES(T, STYPE, DTYPE, 2,  512,  512,  1)                     \
+    CONVERT_TESTS_TYPES(T, STYPE, DTYPE, 3,  512, 1024,  2)                     \
+    CONVERT_TESTS_TYPES(T, STYPE, DTYPE, 4, 2048, 1024, 10)                     \
```
shehzan10: The first test is 1000x1000. Or do you mean non-multiples of 2?

src/backend/cuda/sparse.cu
umar456: No its a pointer. Memory manager doesn't know when it should be deleted.

test/sparse.cpp
umar456: yes. I just mean something more diverse.

src/backend/cpu/sparse.cpp
umar456: I guess non-MKL would be more meaningful. I understand now.

shehzan10:
build arrayfire windows ci

PR #1651, opened 2016-11-29
Minor fix in the documentation of the unified backend

marbre:
Only a minor fix. The command now does what mentioned in the comment.


shehzan10:
Thanks for the fix @marbre.
Since this a documentation fix, it can go into hotfix-3.4.2 branch. Can you branch off of that and then open the PR. Currently, if I simply switch the target branch in this PR, then it will make devel commits go into hotfix-3.4.2.

marbre:
@shehzan10 I rebased onto hotfix-3.4.2, since I though I could change the target branch in this PR, but it seems this is not possible. Will redo the PR.

PR #1657, opened 2016-11-30
Fix FindCBLAS.cmake

marbre:
If PkgConfig is used, the path to the library is determined correctly. However, the include path is not adopted from the cblas.pc file. This is fixed with this commit.


shehzan10:
build arrayfire linux minimal ci

PR #1658, opened 2016-12-01
optimization: inline some dim4 functions

vakopian:
In my profiles `dim4::operator[]` always registers as consuming some non-zero time. This patch inlines that operator and a few others in `dim4`. It also changes `ndims()` to avoid calling `elements()` which does multiplication.

### DIFF
--- a/include/af/dim4.hpp
+++ b/include/af/dim4.hpp
@@ -38,19 +38,26 @@ public:
             dim_t fourth = 1);
     dim4(const dim4& other);
     dim4(const unsigned ndims, const dim_t * const dims);
-    dim_t elements();
-    dim_t elements() const;
-    dim_t ndims();
-    dim_t ndims() const;
+    inline dim_t elements() const { return dims[0] * dims[1] * dims[2] * dims[3]; }
+    inline dim_t ndims() const {
+        if (dims[0] == 0 || dims[1] == 0 || dims[2] == 0 || dims[3] == 0) {
+            return 0;
+        }
+        if (dims[3] != 1) return 4;
+        if (dims[2] != 1) return 3;
+        if (dims[1] != 1) return 2;
+
+        return 1;
+    }
     bool operator==(const dim4& other) const;
     bool operator!=(const dim4& other) const;
     dim4& operator*=(const dim4& other);
     dim4& operator+=(const dim4& other);
     dim4& operator-=(const dim4& other);
-    dim_t& operator[](const unsigned dim);
-    const dim_t& operator[](const unsigned dim) const;
-            dim_t* get()         { return dims; }
-    const   dim_t* get() const   { return dims; }
+    inline dim_t& operator[](const unsigned dim) { return dims[dim]; }
+    inline const dim_t& operator[](const unsigned dim) const { return dims[dim]; }
+    inline dim_t* get() { return dims; }
+    inline const dim_t* get() const { return dims; }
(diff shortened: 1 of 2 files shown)


shehzan10:
@umar456 I believe this would be counted as breaking API?

include/af/dim4.hpp
```diff
@@ -38,19 +38,26 @@ class AFAPI dim4
...
     dim4& operator+=(const dim4& other);
     dim4& operator-=(const dim4& other);
-    dim_t& operator[](const unsigned dim);
-    const dim_t& operator[](const unsigned dim) const;
-            dim_t* get()         { return dims; }
-    const   dim_t* get() const   { return dims; }
+    inline dim_t& operator[](const unsigned dim) { return dims[dim]; }
+    inline const dim_t& operator[](const unsigned dim) const { return dims[dim]; }
+    inline dim_t* get() { return dims; }
+    inline const dim_t* get() const { return dims; }
```
pavanky: @umar456 Any chance this may break the ABI ? This is implemented in the header so I don't think it is an issue, but I am not entirely sure.

include/af/dim4.hpp
```diff
@@ -38,19 +38,26 @@ class AFAPI dim4
...
+    inline dim_t ndims() const {
+        if (dims[0] == 0 || dims[1] == 0 || dims[2] == 0 || dims[3] == 0) {
+            return 0;
+        }
+        if (dims[3] != 1) return 4;
+        if (dims[2] != 1) return 3;
+        if (dims[1] != 1) return 2;
+
+        return 1;
+    }
```
pavanky: This will most definitely break the ABI, we can not merge this.

include/af/dim4.hpp
```diff
@@ -38,19 +38,26 @@ class AFAPI dim4
...
     bool operator==(const dim4& other) const;
     bool operator!=(const dim4& other) const;
     dim4& operator*=(const dim4& other);
     dim4& operator+=(const dim4& other);
     dim4& operator-=(const dim4& other);
-    dim_t& operator[](const unsigned dim);
-    const dim_t& operator[](const unsigned dim) const;
-            dim_t* get()         { return dims; }
-    const   dim_t* get() const   { return dims; }
+    inline dim_t& operator[](const unsigned dim) { return dims[dim]; }
```
pavanky: This will break the ABI too.

include/af/dim4.hpp
```diff
@@ -38,19 +38,26 @@ class AFAPI dim4
...
+    inline const dim_t& operator[](const unsigned dim) const { return dims[dim]; }
```
pavanky: This will break the ABI too.

include/af/dim4.hpp
umar456: I am not sure about the inline keyword but the implementation will not be available in the library binary which may be expected by libraries linked with older version of ArrayFire. This may be something we can do for 4.0

vakopian:
I see. I agree with @umar456, postpone for 4.0

include/af/dim4.hpp
pavanky: I think these particular functions are OK  because they were previously implemented in the header (so they are not part of the lib).

The other functions that moved to the header will break the ABI.

umar456:
Postpone change to ArrayFire 4.0

PR #1659, opened 2016-12-01
optimization: avoid unnecessarily copying ArrayInfo objects

vakopian:
The `getInfo` method returns a const reference to `ArrayInfo`, but almost all the calls to it do a copy of the object. Unless I'm completely missing the point for this, the copy should be avoided since the `ArrayInfo` object (on my system) takes `88` bytes vs the pointer which takes `8` bytes.

### DIFF
--- a/src/api/c/approx.cpp
+++ b/src/api/c/approx.cpp
@@ -38,8 +38,8 @@ af_err af_approx1(af_array *out, const af_array in, const af_array pos,
                   const af_interp_type method, const float offGrid)
 {
     try {
-        ArrayInfo i_info = getInfo(in);
-        ArrayInfo p_info = getInfo(pos);
+        const ArrayInfo& i_info = getInfo(in);
+        const ArrayInfo& p_info = getInfo(pos);
 
         dim4 idims = i_info.dims();
         dim4 pdims = p_info.dims();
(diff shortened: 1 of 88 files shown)


shehzan10:
build arrayfire linux ci
build arrayfire osx opencl ci
build arrayfire windows ci

shehzan10:
It looks like the tests for the PR are failing due to segfaults. Have you been able to run the tests on your end to ensure everything is passing?

shehzan10:
I agree with the analysis. You can make the fix (put the release after the printer call). If all the tests pass at your end, you can push it and we'll run the tests again.

vakopian:
After the fix in `print` tests pass now on my end

shehzan10:
build arrayfire linux ci
build arrayfire osx opencl ci
build arrayfire windows ci

shehzan10:
@vakopian I have removed the `ready to merge` tag from this PR for 2 reasons:

1. Can you rebase your changes to the hotfix-3.4.2 branch and the force push. This way, we can simply change the target branch. We cannot change the target branch directly as it would get the devel commits into hotfix too.
2. We will be doing a bunch of merges in the next few days. Once those are done, you can update those new commits with the `ArrayInfo` change too (if any).

If by any chance you are unable to do these, then we'll merge this PR manually into hotfix.

vakopian:
I just created a different PR for 3.4.2: https://github.com/arrayfire/arrayfire/pull/1665, you can close this one

PR #1661, opened 2016-12-05
getting_started.md: `u8` is supported data type

mlloreda:
Also re-ordered type list by increasing size


docs/pages/getting_started.md
```diff
@@ -20,17 +20,18 @@ ArrayFire provides one generic container object, the [array](\ref af::array)
 on which functions and mathematical operations are performed. The `array`
 can represent one of many different [basic data types](\ref af::af_dtype):
 
+* [u8](\ref u8) 8-bit unsigned values (`unsigned char`)
```
shehzan10: The order should match https://github.com/arrayfire/arrayfire/blob/devel/include/af/defines.h#L195-L212

PR #1668, opened 2016-12-08
Fix input array validations that are based on isVector meta data

### LINKED ISSUE #1660: setUnion does not work on arrays of size 1
Unions of size 1 give the following error: 
```
ERROR: ArrayFire Error (202) : Invalid input argument
```

It works for arrays with 2 or more elements though. 

Ref https://github.com/JuliaComputing/ArrayFire.jl/issues/99

### LINKED ISSUE #1675: setIntersect over singletons throws exception
The following code fails with an abort. The exception is some sort of memory access violation. This occurs regardless of the use of iota() or constant() to create the singleton array. Using an array containing at least two elements works fine. This happens with both the CPU and the OPENCL backend. 

```
#include <arrayfire.h>

using namespace af;

{
	array a1 = iota(dim4(1), dim4(1), s32);
	array a2 = iota(dim4(5), dim4(1), s32);
	
	array a3 = setIntersect(a1, a2);

	af_print(a3);

	return 0;
}
```

9prady9:
- Fixes #1660 
- Fixes #1675


src/api/c/convolve.cpp
```diff
@@ -113,9 +115,20 @@ af_err convolve2_sep(af_array *out, af_array col_filter, af_array row_filter, co
...
+
+        if (cfInfo.isScalar() && rfInfo.isScalar()) {
+            af_array colArray, rowArray, filter;
+
+            AF_CHECK(af_tile(&colArray, col_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_tile(&rowArray, row_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_mul (&filter, colArray, rowArray, false));
+
+            return af_mul(out, signal, filter, false);
+        }
```
pavanky: It should handle the case when even one of them is scalar. We should perhaps handle this in each backend.

9prady9: Yes, that can't be handled here. It would be a larger change in each of the backends.

9prady9: But may be this particular case is something like a once in a blue moon case for separable convolution. I am not saying we shouldn't do it, but just that this can be much more tersely handled by the user if one checks if filter is scalar before calling separable convolution.

pavanky: I just noticed there is a memory leak. `colArray`, `rowArray` and `filter` are not released.

9prady9: My bad, fixing it.

9prady9: Shall i add code to handle one scalar array and one vector array case to each backend now ?

src/api/c/convolve.cpp
```diff
@@ -113,9 +115,26 @@ af_err convolve2_sep(af_array *out, af_array col_filter, af_array row_filter, co
...
+        if (cfInfo.isScalar() && rfInfo.isScalar()) {
+            af_array colArray, rowArray, filter;
+
+            AF_CHECK(af_tile(&colArray, col_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_tile(&rowArray, row_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_mul (&filter, colArray, rowArray, false));
+
+            AF_CHECK(af_mul(out, signal, filter, false));
+
+            AF_CHECK(af_release_array(colArray));
```
shehzan10:
```
if(colArray != 0) AF_CHECK(af_release_array(colArray));
```

Similarly for other lines below.

src/api/c/convolve.cpp
```diff
@@ -113,9 +115,26 @@ af_err convolve2_sep(af_array *out, af_array col_filter, af_array row_filter, co
 
         af_dtype signalType  = sInfo.getType();
 
-        dim4 signalDims = sInfo.dims();
+        dim4 sdims = sInfo.dims();
+
+        ARG_ASSERT(1, (sdims.ndims()>=2));
+
+        if (cfInfo.isScalar() && rfInfo.isScalar()) {
+            af_array colArray, rowArray, filter;
```
shehzan10: Initialize to 0.

src/api/c/convolve.cpp
```diff
@@ -113,9 +115,28 @@ af_err convolve2_sep(af_array *out, af_array col_filter, af_array row_filter, co
...
+        if (cfInfo.isScalar() && rfInfo.isScalar()) {
+            af_array colArray = 0;
+            af_array rowArray = 0;
+            af_array filter = 0;
+
+            AF_CHECK(af_tile(&colArray, col_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_tile(&rowArray, row_filter, sdims[0], sdims[1], sdims[2], sdims[3]));
+            AF_CHECK(af_mul (&filter, colArray, rowArray, false));
+
+            AF_CHECK(af_mul(out, signal, filter, false));
```
shehzan10: I think you should use a temporary `af_array output = 0` variable and then do a `std::swap` on it to be consistent with all the other C-API functions. This is also done a little lower in this function too.

9prady9: I don't think its needed here, it is just too much redundancy. It is the final af_* call in that branch of code flow and `af_mul` does the swap you are talking about.

9prady9:
@pavanky what if we change the A_CHECK macro to accept variable list of
af_array asking with the call to execute. Those af_arrays will be the ones
we want cleared if the af_ call fails for any reason. If all AF_CHECK calls
take care of their dependencies, we can remove potential leaks?

pavanky:
@9prady9 That is a terrible way to implement it. We already have a function that is working on internal C++ arrays. Use that instead of using the C API.

9prady9:
Sure, I shall use internal API. I can't give your examples right away, i
remember that there were places where we had to use C api instead of
internal backend API for some reason - may be these were fixed later, I
don't know about that. That is the I suggested the above alternative.

9prady9:
@pavanky Look at [these](https://github.com/arrayfire/arrayfire/blob/devel/src/api/c/unary.cpp#L560-L646) functions, those were the ones i was referring to earlier (That is not the comprehensive list of such functions although).

I fixed the following files to use backend/internal API already(Haven't pushed it to github yet) so far.
- `src/api/c/convolve.cpp`
- `src/api/c/vector_field.cpp`

src/api/c/convolve.cpp
```diff
@@ -29,9 +33,23 @@ inline static af_array convolve(const af_array &s, const af_array &f, AF_BATCH_K
...
+        Array<accT> colArray = detail::tile(colFilter, signal.dims());
+        Array<accT> rowArray = detail::tile(rowFilter, signal.dims());
+
+        Array<accT> filter = arithOp<accT, af_mul_t>(colArray, rowArray, signal.dims());
+
+        return getHandle(cast<T, accT>(arithOp<accT, af_mul_t>(signal, filter, signal.dims())));
+    }
+
+    ARG_ASSERT(2, colFilter.isVector());
+    ARG_ASSERT(3, rowFilter.isVector());
```
pavanky: Isn't this redundant ?

PR #1670, opened 2016-12-09
fix initilization of chart type in set axes titles fn

9prady9:


src/api/c/window.cpp
```diff
@@ -371,10 +371,12 @@ af_err af_set_axes_titles(const af_window wind,
         ForgeManager& fgMngr = ForgeManager::getInstance();
 
         forge::Chart* chart = NULL;
+
         // The ctype here below doesn't really matter as it is only fetching
         // the chart. It will not set it.
         // If this is actually being done, then it is extremely bad.
-        fg_chart_type ctype = FG_CHART_2D;
+        // But lets have a check anyway.
+        fg_chart_type ctype = (ztitle == NULL || ztitle == 0) ? FG_CHART_2D : FG_CHART_3D;
```
shehzan10: This can be
```
fg_chart_type ctype = ztitle ? FG_CHART_2D : FG_CHART_3D;
```

Checking for 0 or NULL is the same.

PR #1671, opened 2016-12-11
add missing CholeskyInPlace example

gaika:


shehzan10:
Merged into hotfix-3.4.2. I'm guessing because this is from an "unknown repository", it didnt figure it out or something.

PR #1676, opened 2016-12-15
Bug fixes for Sparse Convert on AMD and Intel

shehzan10:
* Fixes out of bounds memory access on AMD/Intel

[skip arrayfire ci]

### DIFF
--- a/src/backend/opencl/kernel/csr2coo.cl
+++ b/src/backend/opencl/kernel/csr2coo.cl
@@ -51,21 +51,30 @@ void csrReduce_kernel(__global       int *orowIdx,
 
     if(id >= nNZ) return;
 
+    // Read COO row indices
     int iRId  = irowIdx[id];
     int iRId1 = 0;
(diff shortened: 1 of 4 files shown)


shehzan10:
build arrayfire linux ci
build arrayfire osx opencl ci

pavanky:
@shehzan10 while you are at it, can you change the kernel caching to match what I did instead of the old method ?

We should move on to using the new method eventually. It'll be nice to have consistency in a single file when we start the cleanup at a later point.

shehzan10:
I actually had the new method in but I thought it was part of the problem and reverted to the old method.
I think it's better to convert all the kernels into the new method at once, maybe for v3.5.0. Maybe we should create an issue for this.

I'll change the kernels in the sparse file to the new format.

pavanky:
> I'll change the kernels in the sparse file to the new format.

Yeah that is all I was asking for now :)

shehzan10:
@pavanky it had been done. give it a review.

PR #1680, opened 2016-12-19
performance fixes in fftconvolve kernels

### LINKED ISSUE #1679: ComplexMultiply redundant operations
In the fftconvolve.hpp (and perhaps elsewhere) for CUDA (arrayfire/src/backend/cuda/kernel/fftconvolve.hpp) complex multiplies are performed like so:

```
        out.ptr[ridx2].x = c1.x*c2.x - c1.y*c2.y;
        out.ptr[ridx2].y = (c1.x+c1.y) * (c2.x+c2.y) - c1.x*c2.x - c1.y*c2.y;
```

Is there a reason for this form for the multiply? It would appear to have several redundant operations that may reduce performance and do increase floating point propagation error.

The multiplies should be:

```
        out.ptr[ridx2].x = c1.x*c2.x - c1.y*c2.y;
        out.ptr[ridx2].y = c1.x*c2.y + c1.y*c2.x;
```

9prady9:
Fixes #1679

### DIFF
--- a/src/backend/cpu/kernel/fftconvolve.hpp
+++ b/src/backend/cpu/kernel/fftconvolve.hpp
@@ -114,11 +114,8 @@ void complexMultiply(Array<T> packed, const af::dim4 sig_dims, const af::dim4 si
                         T c = in2_ptr[ridx];
                         T d = in2_ptr[iidx];
 
-                        T ac = a*c;
-                        T bd = b*d;
-
-                        out_ptr[ridx] = ac - bd;
-                        out_ptr[iidx] = (a+b) * (c+d) - ac - bd;
+                        out_ptr[ridx] = a*c - b*d;
+                        out_ptr[iidx] = a*d + b*c;
                     }
                     else if (kind == AF_BATCH_LHS) {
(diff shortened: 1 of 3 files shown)


src/backend/cpu/kernel/fftconvolve.hpp
```diff
@@ -114,11 +114,8 @@ void complexMultiply(Array<T> packed, const af::dim4 sig_dims, const af::dim4 si
                         T c = in2_ptr[ridx];
                         T d = in2_ptr[iidx];
 
-                        T ac = a*c;
-                        T bd = b*d;
-
-                        out_ptr[ridx] = ac - bd;
-                        out_ptr[iidx] = (a+b) * (c+d) - ac - bd;
+                        out_ptr[ridx] = a*c - b*d;
+                        out_ptr[iidx] = a*d + b*c; // (a+b) * (c+d) - ac - bd;
```
shehzan10: Remove the comment. The actual computation is a standard way to do it. The comment is more confusing.

Also remove it from everywhere else in this PR.

PR #1684, opened 2016-12-20
Release notes for v3.4.2

mlloreda:


docs/pages/release_notes.md
```diff
@@ -1,6 +1,105 @@
 Release Notes {#releasenotes}
 ==============
 
+v3.4.2
+==============
+
+Deprecation
```
shehzan10: Change this to:
```

Deprecation Announcement
------------------------
This release supports CUDA 6.5 and higher. The next ArrayFire relase will support CUDA 7.0 and higher, dropping support for CUDA 6.5. Reasons for no longer supporting CUDA 6.5 include:
* CUDA 7.0 NVCC supports the C++11 standard (whereas CUDA 6.5 does not), which is used by ArrayFire's CPU and OpenCL backends.
* Very few ArrayFire users still use CUDA 6.5.

As a result, the older Jetson TK1 / Tegra K1 will no longer be supported in the next ArrayFire release. The newer Jetson TX1 / Tegra X1 will continue to have full capability with ArrayFire.
```

PR #1687, opened 2016-12-21
Remove CUDA 6.5 Support

shehzan10:
This PR deals with removing CUDA 6.5 support and asserting a minimum CUDA version of 7.0.
It does NOT enable features such as C++11 etc that are associated with using CUDA 7 as minimum version.


9prady9:
@shehzan10 Is there any reason you refrained from enabling C++11 support ?

shehzan10:
I wanted that to be done when C++11 features start getting used. That is, the boost stuff is replaced by STL.

src/backend/cuda/Array.hpp
```diff
@@ -8,26 +8,24 @@
...
 #include <Param.hpp>
 #include <JIT/Node.hpp>
 #include <boost/shared_ptr.hpp>
 #include <vector>
 #include <memory.hpp>
 
+#if CUDA_VERSION < 7000
+    #error "ArrayFire CUDA requires CUDA Toolkit Version 7.0 or newer."
+#endif
+
```
pavanky: Any reason you want to have this here instead of cmake?

In CMake it is as simple as requesting a minimum version in `FIND_PACKAGE(CUDA, 7.0)` or something similar.

shehzan10: Yeah, we could do that.

pavanky:
@shehzan10 merge ?

shehzan10:
Let me run the PR builds. Somehow forgot about it.

PS. OSX CUDA Might fail on tests - But thats a system issue.

build arrayfire linux ci
build arrayfire osx cuda ci
build arrayfire windows ci

shehzan10:
I think it can be merged now

PR #1690, opened 2016-12-23
Remove CL_TO_AF_ERROR macro

umar456:
This removes the CL_TO_AF_ERROR macro and the try/catch blocks from the OpenCL backend. This is done by redefining the CATCHALL macro to look for cl::Error exceptions in addition to the standard ArrayFire Exceptions.

### DIFF
--- a/src/api/c/err_common.hpp
+++ b/src/api/c/err_common.hpp
@@ -204,6 +204,5 @@ void print_error(const std::string &msg);
                       "\n", __err);                         \
     } while(0)
 
-
 static const int MAX_ERR_SIZE = 1024;
 std::string& get_global_error_string();
--- a/src/backend/opencl/err_opencl.hpp
+++ b/src/backend/opencl/err_opencl.hpp
@@ -19,21 +19,6 @@
                 __AF_FILENAME__, __LINE__, "OpenCL");   \
     } while(0)
 
-#define CL_TO_AF_ERROR(ERR) do {                                \
-        char opencl_err_msg[1024];                              \
-        snprintf(opencl_err_msg,                                \
-                 sizeof(opencl_err_msg),                        \
-                 "OpenCL Error (%d): %s when calling %s",       \
-                 ERR.err(), getErrorMessage(ERR.err()).c_str(), \
-                 ERR.what());                                   \
-        if (ERR.err() == CL_MEM_OBJECT_ALLOCATION_FAILURE) {    \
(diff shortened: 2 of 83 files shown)


shehzan10:
build arrayfire osx cuda ci
build arrayfire windows ci

src/api/c/err_common.hpp
```diff
@@ -207,3 +207,7 @@ void print_error(const std::string &msg);
 
 static const int MAX_ERR_SIZE = 1024;
 std::string& get_global_error_string();
+
+#ifdef AF_OPENCL
+#include <err_opencl.hpp>
+#endif
```
pavanky: This is a circular dependency. `err_opencl.hpp` is including `err_common.hpp`. Not sure if this is a good idea. I think the CATCHALL should go in a separate file.

src/backend/opencl/err_opencl.hpp
```diff
@@ -14,26 +14,31 @@
...
+    if (err.err() == CL_MEM_OBJECT_ALLOCATION_FAILURE) {    \
+      AF_ERROR(opencl_err_msg, AF_ERR_NO_MEM);              \
+    } else {                                                \
+      AF_ERROR(opencl_err_msg, AF_ERR_INTERNAL);            \
+    }                                                       \
+  }                                                         \
+  catch(...) {                                              \
+    return processException();                              \
+  }
+
```
pavanky: May be instead of doing `#undef CATCHALL` and defining the whole macro to include the original `CATCHALL` functionaliy, wouldn't it be better to do the following:

```
#define COMMON_CATCHAL CATCHALL
#undef CATCHALL

#define CATCHALL \
   catch (const cl::Error &err) {\
       //handle the error \
    } \
   COMMON_CATCHALL
```

This is to ensure any changes we make in `err_common.hpp` will be reflected immediately here.

umar456: Good idea. This is a better approach.

umar456:
Addressed all feedback.

Moved the logic for catching cl::Error into err_common.cpp. This seems to make the most sense and does not require additional macros to handle the same behavior.

@pavanky Yeah, this would provide us with less information if the user is running into an issue regarding the exception. We can ask them to run the application using gdb and break the program on exception (using the `catch throw` command).

BTW. This change reduced the binary size by 7MB

pavanky:
looks good 👍

shehzan10:
build arrayfire osx opencl ci
build arrayfire linux ci
build arrayfire windows ci

PR #1695, opened 2016-12-29
Add genetic algorithm example

kaatish:


examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+using namespace af;
+
+array searchSpace_x_display = 0;
+array searchSpace_y_display = 0;
+array searchSpace_z_display = 0;
+array searchSpace;
+
+array samplex;
+array sampley;
+array samplez;
```
shehzan10: Global are bad. The only reason to use them should be for input data for benchmarking with `timeit`.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+#include <arrayfire.h>
+#include <stdio.h>
+#include <string.h>
+#include <math.h>
+#include <limits.h>
+
+#include <iostream>
+
+#define OFFSET n/100
```
shehzan10: This seems out of place and small enough not to be a macro.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+array searchSpace_y_display = 0;
+array searchSpace_z_display = 0;
+array searchSpace;
+
+array samplex;
+array sampley;
+array samplez;
+
+int n = 256;
+int nsamples = 16;
```
shehzan10: static const

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+                        sampley(where_m).scalar<unsigned int>(), maximum, truemax);
+            }
+
+            win(0,0).setAxesLimits(searchSpace_x_display, searchSpace_y_display, searchSpace);
+            win(0,0).setAxesTitles("idx", "idy", "Search Space");
+            win(0,1).setAxesLimits(searchSpace_x_display, searchSpace_y_display, searchSpace);
+            win(0,1).setAxesTitles("idx", "idy", "Search Space");
+            win(0,0).surface(searchSpace_x_display, searchSpace_y_display, searchSpace);
+            win(0,1).scatter(samplex.as(f32), sampley.as(f32), samplez.as(f32));
+            win.show();
```
shehzan10: Just FYI, setAxesLimits has been reverted to automatic detection of limits in case you want to use that. You can test it by removing the `setAxesLimit` lines.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+        } while(!win.close());
+
+    } else {
+
+        do {
+            reproduce();
+            float maximum = max<float>(samplez);
+            array where_m = where(samplez == maximum);
```
shehzan10: Copying to host is expensive. Unless you are using maximum elsewhere, do something like:

array maxScalarArray = max(samplez); // Choose the correct dimension. this should give a scalar array
array where_m = where(samplez == tile(maxScalarArray, samplez.dims()));

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+            printf("Current max at (%d,%d): %f (truemax %f)\n",
+                    samplex(where_m).scalar<unsigned int>(),
+                    sampley(where_m).scalar<unsigned int>(), maximum, truemax);
+        } while (max<float>(samplez) < truemax * 0.99);
```
shehzan10: Repeated copy to host. Do this only in one place.

examples/machine_learning/geneticalgorithm.cpp
shehzan10: Alternately, since two of these contain the same limits, you can determine the limits here and then pass them as host side values. This will avoid double computation. Either is fine.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+    float truemax = max<float>(searchSpace);
+    bool found_max = false;
+
+    if (!console) {
+        af::Window win(1600, 800, "ArrayFire Genetic Algorithm Search Demo");
+        win.grid(1, 2);
+
+        do{
+            float maximum = max<float>(samplez);
+            array where_m = where(samplez == maximum);
```
shehzan10: See my comment below about using max.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+    setSeed(time(NULL));
+    samplex = randu(nsamples, u32) % n;
+    sampley = randu(nsamples, u32) % n;
+    update();
+}
+
+
+void init() {
+    //initialize space
+    searchSpace = tile(array(seq(n / 2)), 1, n / 2) + tile(moddims(array(seq(n / 2)), 1, n / 2), n / 2);
```
shehzan10: `tile(array(seq(n / 2)), 1, n / 2)` = `range(dim4(n / 2, n / 2), 0)`
`tile(moddims(array(seq(n / 2)), 1, n / 2), n / 2)` = `range(dim4(n / 2, n / 2), 1)`

Use range.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#include <arrayfire.h>
+#include <stdio.h>
+#include <string.h>
+#include <math.h>
+#include <limits.h>
```
shehzan10:
```
<cstdio>
<cstring>
<cmath>
<climits>
```

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,202 @@
...
+
```
shehzan10: Remove space

examples/machine_learning/geneticalgorithm.cpp
shehzan10: Actually on second review, this isn't even used.

examples/machine_learning/geneticalgorithm.cpp
kaatish: @shehzan10 While max is not used by the algorithm, the demo displays it at every time step. Therefore a host side transfer is necessary. What do you think about a command line option to display it only when the user requests it?

kaatish:
Addressed feedback. Rebased against devel. Please review.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+    sampley = join(0, children_y1, children_y2);
+
+    //Create mutant children
+    array mutantx = samplex.copy();
+    array mutanty = sampley.copy();
+
+    //Flip a random bit to vary the gene pool
+    int bits = (int)(log((double)n) / log(2.0));
+    mutantx = mutantx ^ (pow2((randu(nsamples / 2, u32) % bits).as(f32))).as(u32);
+    mutanty = mutanty ^ (pow2((randu(nsamples / 2, u32) % bits).as(f32))).as(u32);
```
pavanky: Use this instead
```
mutant = mutant ^ (1 << randu(nsamples / 2, u32))
```

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+ ********************************************************/
+
+#include <arrayfire.h>
+#include <cstdio>
+
+using namespace af;
+
+array update(const array& searchSpace, const array& samplex, const array& sampley, const int nsamples, const int n)
+{
+    array result(nsamples, 1, f32);
```
pavanky: No need to initialize.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+array selectFittest(const array& samplez, const int nsamples) {
+    //pick top 50% fittest
+    array sorted = sort(samplez);
+    array indices = where(samplez >= sorted.row(nsamples / 2).scalar<float>());
+
+    if (indices.elements() > nsamples / 2) {
+        indices = indices.rows(indices.elements() - nsamples / 2, indices.elements() - 1);
+    }
+
+    return indices;
```
pavanky: You can always pick to top 50% without doing where. If there are a bunch of variables around the midway point that have the same value, some of them may not get selected, but considering that this is a genetic algorithm I think that is fine. Just do this instead.

```
array indices, values;
sort(values, indices, samplez);
int n = indices.elements();
return (n > nsamples / 2) ? indices(seq(n - nsamples/2, n -1)) : indices;
```

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    //Divide selection in two
+    array parentsx1 = parentsx.rows(0, parentsx.elements() / 2 - 1);
+    array parentsx2 = parentsx.rows(parentsx.elements() / 2, parentsx.elements() - 1);
+    array parentsy1 = parentsy.rows(0, parentsy.elements() / 2 - 1);
+    array parentsy2 = parentsy.rows(parentsy.elements() / 2, parentsy.elements() - 1);
+
+    //Get crossover points (at which bit to crossover) and construct bit masks from them
+    array crossover = randu(nsamples / 4, u32) % (log((double)n) / log(2.0));
+    array lowermask = pow2(crossover.as(f32)).as(u32) - 1;
```
pavanky: Use bitshifting like shown below.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+    array parentsy = sampley(selection);
+
+    //Divide selection in two
+    array parentsx1 = parentsx.rows(0, parentsx.elements() / 2 - 1);
+    array parentsx2 = parentsx.rows(parentsx.elements() / 2, parentsx.elements() - 1);
+    array parentsy1 = parentsy.rows(0, parentsy.elements() / 2 - 1);
+    array parentsy2 = parentsy.rows(parentsy.elements() / 2, parentsy.elements() - 1);
+
+    //Get crossover points (at which bit to crossover) and construct bit masks from them
+    array crossover = randu(nsamples / 4, u32) % (log((double)n) / log(2.0));
```
pavanky:
```
(int)(log2(n));
```

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+    array children_x2 = (parentsx2 & uppermask) + (parentsx1 & lowermask);
+    array children_y2 = (parentsy2 & uppermask) + (parentsy1 & lowermask);
+
+    //Join two new sets
+    samplex = join(0, children_x1, children_x2);
+    sampley = join(0, children_y1, children_y2);
+
+    //Create mutant children
+    array mutantx = samplex.copy();
+    array mutanty = sampley.copy();
```
pavanky: No reason for copy. arrayfire is copy on write.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    //Flip a random bit to vary the gene pool
+    int bits = (int)(log((double)n) / log(2.0));
```
pavanky: Use log2, move this to the beginning and use `bits` in the other place I mentioned log2.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+    init(searchspace, searchspace_x_display, searchspace_y_display,
+            samplex, sampley, samplez, nsamples, n);
+
+    float truemax = max<float>(searchspace);
+    bool found_max = false;
+
+    if (!console) {
+        af::Window win(1600, 800, "arrayfire genetic algorithm search demo");
+        win.grid(1, 2);
+
```
pavanky: End if condition here.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+                found_max = true;
+                printf("max found at (%d,%d): %f (truemax %f)\n",
+                        samplex(where_m).scalar<unsigned int>(),
+                        sampley(where_m).scalar<unsigned int>(), maximum, truemax);
+            }
+            win(0,0).setAxesTitles("idx", "idy", "search space");
+            win(0,1).setAxesTitles("idx", "idy", "search space");
+            win(0,0).surface(searchspace_x_display, searchspace_y_display, searchspace);
+            win(0,1).scatter(samplex.as(f32), sampley.as(f32), samplez.as(f32));
+            win.show();
```
pavanky: add `if (!console)` around these. Also add a break condition when window is closed.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+            printf("current max at (%d,%d): %f (truemax %f)\n",
+                    samplex(where_m).scalar<unsigned int>(),
+                    sampley(where_m).scalar<unsigned int>(), maximum, truemax);
+        } while (maximum < truemax * 0.99);
+        maximum = max<float>(samplez);
+        array where_m = where(samplez == maximum);
+        printf("max found at (%d,%d): %f (truemax %f)\n",
+                samplex(where_m).scalar<unsigned int>(),
+                sampley(where_m).scalar<unsigned int>(), maximum, truemax);
+    }
```
pavanky: Else is redundant if the conditions are restructured.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const int nsamples = 16;
+    if (argc > 2 || (argc == 2 && strcmp(argv[1], "-"))) {
+        printf("usage: %s [-]\n", argv[0]);
+        return -1;
+    } else if (argc == 2 && argv[1][0] == '-') {
+        console = true;
+    }
+
+    try {
+        printf("** ArrayFire Genetic Algorithm Search Demo **\n\n");
```
pavanky: May be add a little bit more info on what is being searched.

examples/machine_learning/geneticalgorithm.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+array update(const array& searchSpace, const array& samplex, const array& sampley, const int nsamples, const int n)
+{
+    array result(nsamples, 1, f32);
+    result = searchSpace(sampley*n + samplex);
+    return result;
+}
+
+array selectFittest(const array& samplez, const int nsamples) {
+    //pick top 50% fittest
```
pavanky: May be accept the percent fittest as a parameter so it can be easily tweaked.

mlloreda:
@kaatish is this ready to be merged?

pavanky:
@kaatish you will need to checkout from devel, cherry pick the commit that has your example and create a new PR>

PR #1696, opened 2016-12-29
Sparse Arithmetic

shehzan10:
Operators +, -, * and / for sparse-dense operations.

Notes:
* Addition and Subtraction return dense arrays
* Multiplication, division also return sparse array. This sparse array will not handle cases of 0s in the dense array which will result in 0s in the output sparse array and NaN in case of division.
* Division by sparse is not going to be supported. This has been implemented but will not be supported because of Inf/NaN chaos.


pavanky:
@shehzan10 what's happening here ?

mlloreda:
build arrayfire linux ci

shehzan10:
build arrayfire windows ci

Had to merge the `CL_TO_AF_ERROR` changes. It should be ready now.

Issue #483, opened 2015-03-10
3D and 4D support for matmul and dot


pavanky:
Do this inside a for loop with offsets. This will be needed for GFOR

pavanky:
These functions error out in 3.0. Pushing the feature request for a later release.

ckeitz:
Pavan, can you clarify explicitly whether this means that matrix multiply is not available within gfor? I am currently getting incorrect results when I use it inside gfor for batch sizes larger than ~5 (small matricies on the order of [40,20] x [20, 40]) - interestingly it seems to work correctly in small batches and I'm nowhere near out of memory.

An aside - is there a registry of functions that are and are not supported in gfor that's available somewhere? Maybe we could put something together for easy reference.

Thanks!

pavanky:
It is most definitely not supported in the open source version. We were not throwing errors before. I changed this behavior couple of days ago. We will add support for it before next release.

ckeitz:
Thanks for the quick reply! Glad to know the issue :+1:

mchandra:
Will release 3.2 allow for solve() inside a gfor() loop?

Thanks,
Mani

Botev:
Maybe adding and batched version of `gaussian kernel` would be good as well. Again by batch I mean having 3D tensor inputs, where one dimension is the same.

InBetweenNames:
What would it take to get this implemented?  I could really use this feature in my code.  Right now I'm using a workaround that involves tiling out matrices (and transposing one) and doing:

af::sum(A * B)

followed by a moddims to get the proper shape.  It does work, although memory usage is a concern and I know this probably isn't the most efficient method.

Thormidable:
Could I also request this feature

WilliamTambellini:
CPU matmul seems now to be done from 3.6.2, right ?

umar456:
Yes, Updated the list

PR #1705, opened 2017-01-09
Batch support for solve function in all backends

### LINKED ISSUE #1075: solve() inside gfor() for a large number of independent matrices
Right now, `solve()` is not optimized to solve for a large number of independent linear systems `A_i x_i = b_i`, where `i = 1,..., N`

`solve()` uses `magma_getrf_gpu()` (https://github.com/maxhutch/magma/blob/master/src/sgetrs_gpu.cpp) from Magma as its backend. Magma also has `magma_sgetrs_batched()` (https://github.com/maxhutch/magma/blob/master/src/sgetrs_batched.cpp) which could be used for batch computations (like those inside gfor?)

Possibly connected issue: https://github.com/arrayfire/arrayfire/issues/483

Relevant post on forums: https://groups.google.com/forum/#!topic/arrayfire-users/RZLKGjsJR3g

Details of the batched implementation in Magma:
1) http://icl.cs.utk.edu/news_pub/submissions/batchedlu.pdf
2) http://on-demand.gputechconf.com/gtc/2015/presentation/S5476-Stanimire-Tomov.pdf (starts from slide 36)

pavanky:
When completed:

- Fixes #1075 
- Related to #483

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ include(Version)
 include(platform)
 include(GetPrerequisites)
 include(CheckCXXCompilerFlag)
+include(CheckSymbolExists)
 include(SplitDebugInfo)
 
 # Use the function generate_product_version on Windows
@@ -170,6 +171,7 @@ mark_as_advanced(CLEAR CUDA_VERSION)
 if(${AF_BUILD_CPU} OR ${AF_BUILD_OPENCL})
   if("${AF_COMPUTE_LIBRARY}" STREQUAL "Intel-MKL"
       OR "${AF_COMPUTE_LIBRARY}" STREQUAL "MKL")
+    af_mkl_batch_check()
     dependency_check(MKL_FOUND "Please ensure Intel-MKL / oneAPI-oneMKL is installed")
     set(BUILD_WITH_MKL ON)
   elseif("${AF_COMPUTE_LIBRARY}" STREQUAL "FFTW/LAPACK/BLAS")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,8 @@ mark_as_advanced(CLEAR CUDA_VERSION)
 # Note that the default value of AF_COMPUTE_LIBRARY is Intel-MKL.
 # Also, cmake doesn't have short-circuit of OR/AND conditions in if
 if(${AF_BUILD_CPU} OR ${AF_BUILD_OPENCL})
-  if("${AF_COMPUTE_LIBRARY}" STREQUAL "Intel-MKL")
+  if("${AF_COMPUTE_LIBRARY}" STREQUAL "Intel-MKL"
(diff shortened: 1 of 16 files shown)


mchandra:
Hi, could this be merged into master? Would really help. Thanks.

9prady9:
@pavanky How much more work is left for this feature to be complete ?

mchandra:
I think this PR implements batch support for CUDA, not for OpenCL and the CPU backends (which don't have batch support implemented at the backend library level). I vote for merging as is -- with just CUDA batch support.

WilliamTambellini:
Watch up batched gemm in recent mkl : 
https://github.com/arrayfire/arrayfire/issues/2146

test/solve_dense.cpp
```diff
@@ -12,9 +12,163 @@
...
+    //! [ex_solve_recon]
+
+    ASSERT_NEAR(
+        0,
+        sum<typename dtype_traits<T>::base_type>(abs(real(B0 - B1))) / (m * k),
+        eps);
+    ASSERT_NEAR(
+        0,
+        sum<typename dtype_traits<T>::base_type>(abs(imag(B0 - B1))) / (m * k),
+        eps);
```
9prady9, line 80: I have a local change where I am checking why we don't use `ASSERT_ARRAYS_NEAR` in solve tests. Are you facing any issues using those macros with solve ? because I do, hence I didn't go ahead with my change - https://github.com/9prady9/arrayfire/commit/a683c8858e64ab9d22a0a2af7dfe9185cd64cd0b

CMakeLists.txt
```diff
@@ -168,7 +168,8 @@ mark_as_advanced(CLEAR CUDA_VERSION)
 # Note that the default value of AF_COMPUTE_LIBRARY is Intel-MKL.
 # Also, cmake doesn't have short-circuit of OR/AND conditions in if
 if(${AF_BUILD_CPU} OR ${AF_BUILD_OPENCL})
-  if("${AF_COMPUTE_LIBRARY}" STREQUAL "Intel-MKL")
+  if("${AF_COMPUTE_LIBRARY}" STREQUAL "Intel-MKL"
+      OR "${AF_COMPUTE_LIBRARY}" STREQUAL "MKL")
```
9prady9, line 172: I don't think this is needed at all because `AF_COMPUTE_LIBRARY` only allows a predetermined list of strings. It is not possible to set it to `MKL`.

9prady9 (changes requested):
Can you please also merge your solve related commits right after pavan's.  The `MKL` option change is not needed, so that commit can be removed - please check the relevant change comment.

If the solve batch support change doesn't work without pinned memory commit, you can merge that as well into the commit that has all the solve batch support changes you added after pavan.

I am changing the title to reflect that this adds batch support for solve alone.

test/solve_dense.cpp
umar456: I picked your commit.

9prady9 (approved):

PR #1706, opened 2017-01-10
Thread safety

### LINKED ISSUE #907: Make sure all backends are thread safe
Hi team,

Currently I'm saturating my GPU, which is cool. But on the same server I can have more GPU's - eg. an Amazon g2.8xlarge instance has 4 Nvidia K520's. I want to be able to use them too.

I want to be able to launch 4 threads and allocate each one a GPU for them to work on. I was reading the documentation for cudaSetDevice() and I believe it is possible to do this, but I think it requires some work on the ArrayFire memory manager and device manager. I really want to be able to do something like this:

```
Main thread
(launch # of gpu's of std::threads, eg. 4 threads)

Worker thread
af::setDevice(threadID);       // threadID is 0-3, passed in as a parameter by the main thread
```

Then the rest of the worker thread uses the GPU for the device id it set.

The traditional solution of a cluster of GPU hosts would be ideal, but I'm processing huge amounts of data very fast and distributing the work to multiple hosts adds too much latency. There's just too much data to serialise, transfer and deserialise to make the calculation worthwhile on a cluster. So it needs to be done on the same server. 

Currently my only alternative is creating multiple processes o the server each having their dedicated GPU and having 1 coordinator process managing the work and shared memory for the data structures. It can be massively simplified if ArrayFire could support multiple GPU's used through dedicated threads.

Please could you add this to the library?

### LINKED ISSUE #1728: Adding an external device after a Manager is created causes segmentation faults
The following code shows the failure occurring in MemoryManager, but it can happen in other managers that depend on deviceId.

```
TEST(OCLExtContext, set)
{
    cl_device_id deviceId = NULL;
    cl_context context = NULL;
    cl_command_queue queue = NULL;

    af_print(af::randu(5,5)); // Remove this line and code works
    getExternals(deviceId, context, queue);
    afcl::addDevice(deviceId, context, queue);
    afcl::setDevice(deviceId, context);
    af::info();

    const int x = 5;
    const int y = 5;
    const int s = x * y;
    af::array a = af::constant(1, x, y);
    vector<float> host(s);
    a.host((void*)host.data());
    for (int i=0; i<s; ++i)
        ASSERT_EQ(host[i], 1.0f);
}
```

### LINKED ISSUE #1754: Memory remains on device after ArrayFire array's are cleaned up
In my codes, it calls af::fft3 and cleans up all arrayfire arrays but memory allocated on GPU remains. For example, there is 711.16MB allocated on GPU at the beginning but it is increased to 1063.04MB at the end.

AF Memory @start
    Allocated [ Bytes | Buffers ] = [ 0 | 0 ]
    In Use    [ Bytes | Buffers ] = [ 0 | 0 ]
CUDA Memory @start [0] = {**711.16MB**/6144.00MB}
AF Memory @end of program
    Allocated [ Bytes | Buffers ] = [ 0 | 0 ]
    In Use    [ Bytes | Buffers ] = [ 0 | 0 ]
CUDA Memory @end of program [0] = {**1063.04MB**/6144.00MB}

In the real application with big data size,  it causes the memory to run out from the second execution because some of the allocated memory at the first iteration is not freed up.

```
#include <iostream>

#include <arrayfire.h>
#include <af/data.h>

#include <nvml.h>

#define TC_MEM_INFO(msg) do {                                                           \
    size_t abytes = 0, abuffs = 0, lbytes = 0, lbuffs = 0;                              \
    af_err err = af_device_mem_info(&abytes, &abuffs, &lbytes, &lbuffs);                \
    if(err == AF_SUCCESS) {                                                             \
        printf("AF Memory %s\n", msg);                    \
        printf("    Allocated [ Bytes | Buffers ] = [ %lld | %lld ]\n", abytes, abuffs);    \
        printf("    In Use    [ Bytes | Buffers ] = [ %lld | %lld ]\n", lbytes, lbuffs);    \
						    } else {                                                                \
        fprintf(stderr, "AF Memory at %s:%d: " msg "\nAF Error %d\n",                   \
                __FILE__, __LINE__, err);                                               \
						    }                                                                       \
} while(0)


#define TC_PRINT(tag, idx, device) do {																												\
	nvmlMemory_t mem;																																								\
	TC_MEM_INFO(tag);																																								\
	nvmlDeviceGetMemoryInfo(device, &mem);																													\
	printf("CUDA Memory %s [%d] = {%.2fMB/%.2fMB}\n", tag, idx, mem.used/1024.0/1024.0, mem.total/1024.0/1024.0);		\
} while(0)

using namespace std;


int main(int argc, char* argv[])
{
	nvmlDevice_t device;

	nvmlInit();
	nvmlDeviceGetHandleByIndex(0, &device);

	TC_PRINT("@start", 0, device);

	const int d0 = 444;
	const int d1 = d0;
	const int d2 = 206;

	for (int i = 0; i < 2; i++)
	{
		af::array a0 = af::constant(0, d0, d1, d2, f32);
		a0.eval();

		af::array a1 = af::fft3(a0);
		a1.eval();
	}

	af::sync();
	af::deviceGC();

	TC_PRINT("@end of program", 0, device);

	nvmlShutdown();
		
	return 0;
}
```

9prady9:
Fixes #907 
Fixes #1728 
Fixes #1754 

- [x] Remove boost::shared_mutex usage in favor of thread local storage for cache management objects.
- [x] ~Implement `thread_local` alternative for OSX Platform~ Making XCode 8.0 the minimum requirement, which has `thread_local` support.
- [x] Change kernel cache management to be thread local storage.
- [x] Change fft plan caches to use per thread local storage.
- [x] Maintain per thread active device
- [x] Unit tests


src/backend/cpu/platform.hpp
```diff
@@ -9,34 +9,134 @@
...
+        static const int MAX_QUEUES = 1;
+        static const int NUM_DEVICES = 1;
+        static const int ACTIVE_DEVICE_ID = 0;
+        static const bool IS_DOUBLE_SUPPORTED = true;
+
+        static DeviceManager& getInstance();
+
+        friend queue& getQueue(int device);
+
+        CPUInfo getCPUInfo() const;
```
pavanky: Where else (outside of platform.cpp) is `getCPUInfo` going to be used?

src/backend/cuda/CMakeLists.txt
```diff
@@ -338,6 +338,10 @@ ELSE()
   ENDIF()
 ENDIF()
 
+IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_MWAITXINTRIN_H_INCLUDED -D_FORCE_INLINES")
+ENDIF()
```
pavanky: Can you check if this is required for gcc 4.x ?

src/backend/cuda/CMakeLists.txt
```diff
@@ -429,7 +433,8 @@ TARGET_LINK_LIBRARIES(afcuda    PRIVATE ${CUDA_CUBLAS_LIBRARIES}
                                 PRIVATE ${CUDA_cusparse_LIBRARY}
                                 PRIVATE ${CUDA_cusolver_LIBRARY}
                                 PRIVATE ${CUDA_nvvm_LIBRARY}
-                                PRIVATE ${CUDA_CUDA_LIBRARY})
+                                PRIVATE ${CUDA_CUDA_LIBRARY}
+                                PRIVATE ${Boost_LIBRARIES})
```
pavanky: Is boost still required?

src/backend/cuda/blas.cpp
```diff
@@ -173,7 +171,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs,
     if(rDims[bColDim] == 1) {
         N = lDims[aColDim];
         CUBLAS_CHECK(gemv_func<T>()(
-                         getHandle(),
+                         getcublasHandle(),
```
pavanky: may be just call the function `cublasHandle()` ?

src/backend/cuda/cholesky.cu
```diff
@@ -112,7 +110,7 @@ int cholesky_inplace(Array<T> &in, const bool is_upper)
     if(is_upper)
         uplo = CUBLAS_FILL_MODE_UPPER;
 
-    CUSOLVER_CHECK(potrf_buf_func<T>()(getDnHandle(),
+    CUSOLVER_CHECK(potrf_buf_func<T>()(getcusolverDnHandle(),
```
pavanky: may be call it `cusolverDnHandle()` ?

src/backend/cuda/cublasManager.cpp
```diff
@@ -7,68 +7,47 @@
...
+    case    CUBLAS_STATUS_MAPPING_ERROR:        return "CUBLAS_STATUS_MAPPING_ERROR";
+    case    CUBLAS_STATUS_EXECUTION_FAILED:     return "CUBLAS_STATUS_EXECUTION_FAILED";
+    case    CUBLAS_STATUS_INTERNAL_ERROR:       return "CUBLAS_STATUS_INTERNAL_ERROR";
 #if CUDA_VERSION > 5050
-        case    CUBLAS_STATUS_NOT_SUPPORTED:        return "CUBLAS_STATUS_NOT_SUPPORTED";
+    case    CUBLAS_STATUS_NOT_SUPPORTED:        return "CUBLAS_STATUS_NOT_SUPPORTED";
 #endif
-        default:                                    return "UNKNOWN";
-        }
+    default:                                    return "UNKNOWN";
```
pavanky: Since CUDA 6.5 is dropped, may be remove the condition.

src/backend/cuda/cufftManager.cpp
```diff
@@ -0,0 +1,149 @@
...
+        case CUFFT_NO_WORKSPACE:
+            return "cuFFT: no workspace provided";
+
+#if CUDA_VERSION >= 6050
+        case CUFFT_NOT_IMPLEMENTED:
+            return "cuFFT: not implemented";
+
+        case CUFFT_LICENSE_ERROR:
+            return "cuFFT: license error";
+#endif
```
pavanky: Remove the macro checks as 7.0 is minimum.

src/backend/cuda/cusparseManager.cpp
```diff
@@ -10,66 +10,45 @@
...
-    cusparseHandle_t getHandle()
-    {
-        static std::unique_ptr<cusparseHandle> handle[cuda::DeviceManager::MAX_DEVICES];
-
-        int id = cuda::getActiveDeviceId();
+cusparseHandle::cusparseHandle()
+    : handle(0)
+{
+    CUSPARSE_CHECK(cusparseCreate(&handle));
+    CUSPARSE_CHECK(cusparseSetStream(handle, cuda::getStream(cuda::getActiveDeviceId())));
```
pavanky: May be have a function called `getActiveStream()` so we don't have to repeatedly do this.

src/backend/cuda/memoryManager.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+void MemoryManagerPinned::nativeFree(void *ptr)
+{
+    cudaError_t err = cudaFreeHost(ptr);
+    if (err != cudaErrorCudartUnloading) {
+        CUDA_CHECK(err);
+    }
+}
+
+}
```
pavanky: Why is this split from memory.cpp ?

src/backend/cuda/memoryManager.hpp
```diff
@@ -0,0 +1,63 @@
...
+    void *nativeAlloc(const size_t bytes);
+    void nativeFree(void *ptr);
+    ~MemoryManagerPinned()
+    {
+        common::lock_guard_t lock(this->memory_mutex);
+        this->garbageCollect();
+    }
+};
+
+}
```
pavanky: Why is this split from memory.hpp?

src/backend/cuda/platform.cpp
```diff
@@ -478,6 +598,21 @@ int DeviceManager::setActiveDevice(int device, int nId)
...
+void DeviceManager::resetcusolverHandle(int device)
+{
+    cusolverHandles[device].reset(new cusolver::cusolverDnHandle());
+}
+
+void DeviceManager::resetcusparseHandle(int device)
+{
+    cusparseHandles[device].reset(new cusparse::cusparseHandle());
+}
+
```
pavanky: Why unique pointers? There's no reason for any of them to be unique pointers.

src/backend/cuda/platform.hpp
```diff
@@ -116,9 +159,29 @@ class DeviceManager
...
+
+        std::unique_ptr<InteropManager> gfxManager;
+
+        cufft::cuFFTPlanner cufftManagers[MAX_DEVICES];
+
+        std::unique_ptr<cublas::cublasHandle> cublasHandles[MAX_DEVICES];
+
+        std::unique_ptr<cusolver::cusolverDnHandle> cusolverHandles[MAX_DEVICES];
+
+        std::unique_ptr<cusparse::cusparseHandle> cusparseHandles[MAX_DEVICES];
```
pavanky: Why are some unique_ptr while others are not?

src/backend/cpu/platform.cpp
```diff
@@ -249,43 +169,50 @@ std::string getDeviceInfo()
...
 #endif
     info << std::endl;
+
     return info.str();
 }
 
 bool isDoubleSupported(int device)
 {
-    return true;
+    return DeviceManager::IS_DOUBLE_SUPPORTED;
```
9prady9: Just moved all(possible) hard coded constants into single location(static data members of DeviceManager class).

src/backend/cpu/platform.cpp
```diff
@@ -166,8 +98,8 @@ CPUInfo::CPUInfo()
             mNumCores = mNumLogCpus = 1;
         }
     } else {
-        mVendorId = "Unkown, probably ARM";
-        cout<< "Unexpected vendor id" <<endl;
+        mVendorId = "Unkown";
+        std::cout<< "Unexpected vendor id" << std::endl;
```
9prady9: A left over i think, i will remove it.

src/backend/cpu/platform.hpp
9prady9: Probably nowhere else. But i can't say we wont in future, may be while checking SSE support availability in hardware when specialising CPU kernels for respective hardware.

src/backend/cuda/CMakeLists.txt
9prady9: Okay

src/backend/cuda/CMakeLists.txt
9prady9: I did a clean build and it worked, so i don't think boost is needed unless we would need shared_mutex in for any other purpose because that is not there in C++ until C++17 i think.

src/backend/cuda/cusparseManager.cpp
9prady9: Makes sense

src/backend/cuda/platform.hpp
9prady9: In some cases, the respective resource manager class itself uses some information that is initialised during the creation of DeviceManager class which is causing infinite loop. Therefore, for such cases, i am lazying initialising them using unique_ptrs.

src/backend/cuda/platform.cpp
9prady9: I will go through them again. The sections of code that was managing sparse, solve and blas handles was using boost::scoped_ptr earlier, which i replaced with it's std equivalent unique_ptr.

src/backend/cuda/memoryManager.cpp
9prady9: In order to use the memory manager classes for variable declaration in DeviceManager class, i moved the class declaration into hpp file. Also, I thought it would be consistent with other managers being in their own source files already - `*Managers.[hpp|cpp]`.

src/backend/cuda/CMakeLists.txt
pavanky: I say remove it. It can be added back if it is necessary.

src/backend/cuda/platform.hpp
```diff
@@ -74,14 +72,14 @@ bool& evalFlag();
...
 class MemoryManagerPinned;
-class InteropManager;
 ///////////////////////// BEGIN Sub-Managers ///////////////////
 //
 MemoryManager& getMemoryManager();
 
 MemoryManagerPinned& getMemoryManagerPinned();
 
-InteropManager& getGfxInteropManager();
+typedef common::InteropManager<GraphicsResourceManager, CGR_t> GraphicsManager;
```
pavanky: I don't think this is necessary. Can you try renaming `GraphicsResourceManager` to `GraphicsManager` and removing this line to see if it works?

9prady9: That has been addressed in later commits.

9prady9:
That may be true, but with the `-std=c++11` from "src/backend/cuda/CMakeLists.txt" i get the following error on Ubuntu 16.04. Removing it makes it compile and run fine on Ubuntu.

```
[ 57%] Building NVCC (Device) object src/backend/cuda/CMakeFiles/afcuda.dir/afcuda_generated_approx.cu.o
nvcc fatal   : redefinition of argument 'std'
nvcc fatal   : redefinition of argument 'std'
nvcc fatal   : redefinition of argument 'std'
CMake Error at afcuda_generated_all.cu.o.cmake:207 (message):
  Error generating
  /home/pradeep/Downloads/arrayfire/build/src/backend/cuda/CMakeFiles/afcuda.dir//./afcuda_generated_all.cu.o

CMake Error at afcuda_generated_approx.cu.o.cmake:207 (message):
  Error generating
  /home/pradeep/Downloads/arrayfire/build/src/backend/cuda/CMakeFiles/afcuda.dir//./afcuda_generated_approx.cu.o

CMake Error at afcuda_generated_any.cu.o.cmake:207 (message):
  Error generating
  /home/pradeep/Downloads/arrayfire/build/src/backend/cuda/CMakeFiles/afcuda.dir//./afcuda_generated_any.cu.o
```

test/threading.cpp
```diff
@@ -479,3 +480,97 @@ TEST(Threading, FFT)
...
+            std::cout << "Failed test " << i << "\nCalculated: " << std::endl;
+            std::copy(h_out.begin(), h_out.end(), std::ostream_iterator<T>(std::cout, ", "));
+            std::cout << "Expected: " << std::endl;
+            std::copy(tests[i].begin(), tests[i].end(), std::ostream_iterator<T>(std::cout, ", "));
+            FAIL();
+        }
+    }
+}
+
+#define TEST_FOR_TYPE(TypeName)                                   \
```
umar456: Can you break up this test into multiple tests that test the different scenarios? You shouldn't have multiple tests in a test case because if one were to fail you will have to do additional digging to figure out the root of the problem. I noticed that this is happening in other tests as well.

pavanky:
@9prady9 any reason why threading on the CPU is failing for OSX ?

pavanky:
I found this: http://mirror.informatimago.com/next/developer.apple.com/hardware/ve/errata.html#lapack_threads which says LAPACK is not thread safe for Accelerate framework.  But this seems to be for an older version.

@mlloreda can you find more info about the newer versions ?

May be we should just have a mutex lock for CPU LAPACK calls on OSX when using accelerate framework.

9prady9:
@pavanky You should note that Test_threading_cpu doesn't always fail. So, this could be related to the following statement from the link you shared. We use 10.11.6 on Jesse, I am assuming it is the same on build machines. 

```
In MacOS X Release 10.2, LAPACK is not thread-safe.
Applications that intend to call the LAPACK APIs from multiple 
threads must implement their own locking discipline to prevent 
simultaneous execution of LAPACK routines.

In MacOS X Release 10.3, LAPACK thread-safety is greatly 
enhanced. Applications that intend to call the LAPACK APIs 
from multiple threads must ensure that the following two 
initialization calls are completed before commencing 
simultaneous execution of LAPACK routines.
```

pavanky:
@9prady9 do we call those initialization functions ?

9prady9:
@pavanky We don't call `slamch_` or `dlamch_` anywhere in cpu backend. I am testing it now.

9prady9:
Neither the mutex nor `slamch_` calls helped.

pavanky:
@9prady9 can you check if OSX CUDA version is using Accelerate framework or if it is using MKL ?

9prady9:
@pavanky I did a `grep -ri` for terms `mkl`, `lapack`, `USE_MKL` and `Accelerate` in `src/backend/cuda` and it gave zero results. I don't think we use either MKL or Accelerate framework in CUDA backend.

9prady9:
@pavanky Both, CUDA and OpenCL builds are using Accelerate framework for OSX. For some reason, it is failing on OpenCL build. May be it happens randomly with CUDA and OpenCL both, but I can't say for certain.

otool output on CUDA-OSX build.
```
octane-osx-01:lib((HEAD detached at 53e1fd8))$ otool -L libafcpu.dylib pu.dylib 
libafcpu.dylib:
        @rpath/libafcpu.3.dylib (compatibility version 3.0.0, current version 3.5.0)
        /System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate (compatibility version 1.0.0, current version 4.0.0)
        /usr/local/opt/fftw/lib/libfftw3.3.dylib (compatibility version 9.0.0, current version 9.5.0)
        /usr/local/opt/fftw/lib/libfftw3f.3.dylib (compatibility version 9.0.0, current version 9.5.0)
        @rpath/libforge.0.dylib (compatibility version 0.0.0, current version 0.9.2)
        /System/Library/Frameworks/OpenGL.framework/Versions/A/OpenGL (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

umar456:
build arrayfire windows ci

umar456:
@9prady9 Looks like there are a couple of failures on Windows. Can you check it out?

pavanky:
build arrayfire windows ci

pavanky:
@umar456 @9prady9 windows is still broken.

9prady9:
@pavanky I believe canny's narrowing error is still present in this branch.

pavanky:
@9prady9 can you merge from devel?

9prady9:
@pavanky Will do and push along with other Windows fixes.

umar456:
build arrayfire windows ci

umar456:
build arrayfire linux ci
build arrayfire osx ci
build arrayfire windows ci

umar456:
build arrayfire windows ci

PR #1708, opened 2017-01-11
Fix for CUDA usage with gcc > 5.x

9prady9:
[skip arrayfire ci]


pavanky:
build arrayfire linux ci

PR #1723, opened 2017-02-05
Fixes for Windows for CUDA after C++11 introduction

shehzan10:


shehzan10:
build arrayfire windows ci

pavanky:
@shehzan10 Doesn't windows need specific options to enable C++11 ? If it is automatically enabled, may be there should be checks to see if the compiler version is correct?

shehzan10:
> @shehzan10 Doesn't windows need specific options to enable C++11 ? If it is automatically enabled, may be there should be checks to see if the compiler version is correct?

Nope, that's why CPU/OpenCL never needed it. The existing [C++11 flag in CMakeLists.txt](https://github.com/arrayfire/arrayfire/blob/devel/CMakeLists.txt#L134) is under `if(UNIX)` too.

Visual Studio 2013 has (partial) C++11 support by default. No flags are required.


Ps:
build arrayfire windows ci

pavanky:
Any idea why Windows OpenCL is segfaulting ?

shehzan10:
build arrayfire windows ci

According to @mlloreda, it was a matter of rebooting. Not sure why it happened as I did not touch the OpenCL backend.

Issue #1611, opened 2016-10-06
Adding support for CLBlast


CNugteren:
For the past year I have been working on [CLBlast](https://github.com/CNugteren/CLBlast). I believe it will soon be mature enough to replace the clBLAS back-end which you are currently using for the OpenCL version of ArrayFire. I also saw issue #1374 which discusses some problems that CLBlast could solve (better performance on non-AMD GPUs for example). However, CLBlast still relies on the community to run and report tuning results on a variety of OpenCL devices.

Are you open to having CLBlast as an (optional) replacement for clBLAS? If so, is there anything I can do to make integration as easy as possible?

shehzan10:
We are definitely open to providing alternates based on user choices.

Some of the requirements check out immediately, like:
* License - APACHE 2.0 is compatible.
* Cross platform
* C/C++ API is great
* float, double, cfloat, cdouble are all there

Going through your README, I have the following questions for you:
* We like to keep ArrayFire's minimum requirement to Visual Studio 2013. I see that CLBlast requirement is VS 2015. Is this a hard requirement based on some C++11 features that are not available in VS 2013?
* Do you allow static linking across all platforms? This would be important if we are to ship CLBlast with ArrayFire binaries that are hosted on our website.
* You wrote about the performance tuning. Is this something that is compile time or can it also be done when we ship ArrayFire to our users (because we have no control over which devices they use).

If you are interested in starting off a PR, the first step we would like to do is to get a `build_CLBlast.cmake` file to build CLBlast as an external project as an alternate to clBLAS. We can add an option in `src/backend/opencl/CMakeLists` to switch between the two libraries.

Once this is good, we can work on source code switching between clBLAS and CLBlast.

CNugteren:
Thanks for your quick response! Here are answers to your questions:

1. VS 2015 is indeed a requirement because of C++11. A while back I made an attempt to start 'fixing' the code to make it compile under VS 2013. But there is just too much that isn't available - it would require re-writing parts of the code. If this is really needed I can check again what can be done, but it'll cost time. If possible I would like to do it in a separate header/source-file - not polluting the main code to much with VS 2013 specific hacks.

2. I don't think I have tested static linking - it isn't part of the current CMake build. But I don't see why it wouldn't possible. Anything that you know that might be blocking this?

[...]

PR #1727, opened 2017-02-11
Added CLBlast as an optional back-end for OpenCL BLAS

CNugteren:
These changes add the CLBlast OpenCL BLAS library as an optional alternative to clBLAS. This PR is based on the discussions in #1611.

There are basically two changes:
* All clBLAS-specific code is moved to separate files with compilation guarded by the `USE_CLBLAS` define and the `USE_CLBLAS` CMake option. This is done for the regular 'blas' part and the 'magma' part. These changes make it easy to add any other OpenCL BLAS libraries as a replacement for clBLAS.
* The CLBlast-specific code is added in separate files with compilation guarded by the `USE_CLBLAST` define and the `USE_CLBLAST` CMake option.

So, for example we now have:
* `src/backend/opencl/blas.hpp`: A back-end agnostic OpenCL BLAS header
* `src/backend/opencl/blas_clblas.cpp`: clBLAS-specific implementation of the `blas.hpp` header
* `src/backend/opencl/blas_clblast.cpp`: CLBlast-specific implementation of the `blas.hpp` header

Things that are still missing perhaps:
* Documentation on the new CMake options
* Proper testing to see whether the clBLAS back-end behaves as it did before
* TRSV and TRSM support in CLBlast ([under construction](https://github.com/CNugteren/CLBlast/tree/triangular_solvers))


pavanky:
@CNugteren 

This is great first push. But would it be possible to wrap clBLAS into a clBLAST like API inside a single file in arrayfire ? This way you don't have to create multiple files like `blas_clblas` and `blas_clblast`. You can just include the middle layer when compiling clblas or use clblast directly if it is enabled?

pavanky:
@CNugteren

ArrayFire has a wide range of hardware. Some of which are not present in the pre-tuned versions included in clblast. If you can provide instructions, I think one of  @umar456 @mlloreda may be willing to send you a PR for some additional hardware.

CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ IF(${OpenCL_FOUND})
     SET(BUILD_OPENCL ON CACHE BOOL "")
 ENDIF(${OpenCL_FOUND})
 OPTION(BUILD_OPENCL "Build ArrayFire with a OpenCL backend" OFF)
+OPTION(USE_CLBLAST "Build ArrayFire with the CLBlast BLAS library for the OpenCL backend" OFF)
+OPTION(USE_CLBLAS "Build ArrayFire with the clBLAS BLAS library for the OpenCL backend" ON)
```
CNugteren: Makes sense. It's done.

src/backend/opencl/CMakeLists.txt
```diff
@@ -60,14 +60,39 @@ ENDIF()
...
+IF(USE_CLBLAST AND USE_CLBLAS)
+    MESSAGE(SEND_ERROR "Cannot use both CLBlast and clBLAS, please select only one of them using USE_CLBLAST=OFF or USE_CLBLAS=OFF")
+ENDIF()
+
+IF(NOT USE_CLBLAST AND NOT USE_CLBLAS)
+    MESSAGE(SEND_ERROR "The OpenCL backend requires either CLBlast or clBLAS, please select one of them using USE_CLBLAST=ON or USE_CLBLAS=ON")
+ENDIF()
+
+IF(USE_CLBLAST)
+    OPTION(USE_SYSTEM_CLBLAST "Use system CLBlast" OFF)
```
CNugteren: OK, I moved it (and the clBLAS one) outside of the condition and added a guard to make sure this is only set if USE_CLBLAST is also set.

CNugteren:
> But would it be possible to wrap clBLAS into a clBLAST like API inside a single file in arrayfire ? This way you don't have to create multiple files like blas_clblas and blas_clblast. You can just include the middle layer when compiling clblas or use clblast directly if it is enabled?

Could be done of course, but I thought that the current way it is more flexible and cleaner. Currently you can add a new back-end for another (future) OpenCL BLAS easily by just supplying one file. I'm not sure if it makes it cleaner to re-write the ArrayFire code to use CLBlast syntax and then provide a wrapper for clBLAS. And why not the other way around? Actually, that would not require most of my changes in this PR? And what if you want to support other back-ends later on? By the way I already have a [CLBlast to clBLAS wrapper](https://github.com/CNugteren/CLBlast/blob/master/test/wrapper_clblas.hpp) implemented as part of the CLBlast tests in case we decide that we need it.

> ArrayFire has a wide range of hardware. Some of which are not present in the pre-tuned versions included in clblast. If you can provide instructions, I think one of umar456 mlloreda may be willing to send you a PR for some additional hardware.

Additional tuning results would be more than welcome! However, I think that for now we should focus on correctness and later look at performance? If you do want to provide tuning results, please read the [relevant README section](https://github.com/CNugteren/CLBlast#using-the-tuners-optional) and upload the results [here](https://github.com/CNugteren/CLBlast/issues/1). Note that CLBlast might not necessarily be slow on un-seen devices, it can benefit from other tuning results on similar devices of the same vendor.

pavanky:
>Could be done of course, but I thought that the current way it is more flexible and cleaner. Currently you can add a new back-end for another (future) OpenCL BLAS easily by just supplying one file. I'm not sure if it makes it cleaner to re-write the ArrayFire code to use CLBlast syntax and then provide a wrapper for clBLAS. And why not the other way around? Actually, that would not require most of my changes in this PR? And what if you want to support other back-ends later on? 

My suggestion came because of what is happening for CPU [cblas](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cblas.cpp) and [lapace](https://github.com/arrayfire/arrayfire/blob/devel/src/backend/cpu/lapack_helper.hpp). The CLBLAST API looks clean enough so it can be hijacked if there are other OpenCL blas implementations :) 

> By the way I already have a CLBlast to clBLAS wrapper implemented as part of the CLBlast tests in case we decide that we need it. 

That's great. That is what I was thinking about :)

> Additional tuning results would be more than welcome! However, I think that for now we should focus on correctness and later look at performance? If you do want to provide tuning results, please read the relevant README section and upload the results here. Note that CLBlast might not necessarily be slow on un-seen devices, it can benefit from other tuning results on similar devices of the same vendor.

That was more of a side note :) It doesn't need to be part of the PR.

CNugteren:
So, are you sure that's what you want to have? To me the current PR seems to be a much cleaner solution, all the library-specific (e.g. clBLAS) code is now in one file. Encapsulation is good, right? And it would also be a bit strange to have clBLAS as a default and then see CLBlast enums and functions everywhere in the code. Or am I misunderstanding you?

If this is really what you want I would even suggest to cancel the whole PR and just add a clBLAS-compatible header to the CLBlast project instead... Not so clean as the current solution imho, because what if you want to add support for another back-end later on?

pavanky:
@CNugteren If CLBlast performs better than clBLAS there is no reason to keep clBLAS as the default :) 

As for the two methods discussed, I don't have strong feelings about the integration either way. I mainly want to reduce the code duplication.

CNugteren:
> If CLBlast performs better than clBLAS there is no reason to keep clBLAS as the default :)

Don't think that's a good idea yet. Better to give it as an option first and test it thoroughly before making it a default. clBLAS still has a lot of issues, but it is also still much wider in use and thus better tested on a wide variety of software/hardware. More so than CLBlast still.

> As for the two methods discussed, I don't have strong feelings about the integration either way. I mainly want to reduce the code duplication.

I don't think duplication is the right word? But there is indeed extra code now due to the additional layer in between. Other than that I only see advantages of my current approach...
* All clBLAS-specific code is in a single place: encapsulation, easy to maintain, make changes only in one place, etc.
* Same for the CLBlast specific part
* It is now easy to add another OpenCL BLAS library as a back-end

... and I also see disadvantages of your suggestion using either the clBLAS or CLBlast as the main API:
* If one back-end is selected it is very confusing to see (and use!) data-types and function calls to the other back-end in the code.
* Function calls and data-types are in multiple places: changing it (for whatever reason) will require multiple changes.
* The APIs are not the same, so choosing one back-end will not work for all cases. Perhaps only clBLAS would work, because (so far) CLBlast has fewer arguments and doesn't need an initialization and clean-up phase.

Hopefully I can still convince you :slightly_smiling_face: But if it's just the amount of lines of code and the amount of changes that count for you then I guess I'll cancel this PR and work on a clBLAS-compatible interface for CLBlast as well (could come in for other cases as well of course).

Let me know what you think.

pavanky:
I am deferring the decision to @umar456

src/backend/opencl/blas_clblast.cpp
```diff
@@ -0,0 +1,153 @@
...
+    const auto alpha = scalar<T>(1);
+    const auto beta  = scalar<T>(0);
+    const auto alpha_clblast = toCLBlastConstant(alpha);
+    const auto beta_clblast = toCLBlastConstant(beta);
+
+    const dim4 lStrides = lhs.strides();
+    const dim4 rStrides = rhs.strides();
+    if(rDims[bColDim] == 1) {
+        CLBLAST_CHECK(
+            clblast::Gemv(clblast::Layout::kColMajor, lOpts,
```
umar456: Can't the gpu_blas_gemm_func functions be used here? I would like to minimize the specialized logic in these functions.

src/backend/opencl/blas_clblas.cpp
```diff
@@ -7,6 +7,8 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
+#if defined(USE_CLBLAS)
```
CNugteren: I agree, that would be my preference as well. But it seems ArrayFire is using the *.cpp GLOB in CMake: `FILE(GLOB opencl_sources "*.cpp")`, so I'm not sure how to handle that properly without explicitly listing all files...

src/backend/opencl/magma/getrf.cpp
```diff
@@ -219,23 +219,22 @@ magma_int_t magma_getrf_gpu(
...
-                                 dAT(j-1,j+1), lddat,
-                                 1, &queue, 0, nullptr, &event));
-
-                if (m > j * nb)  {
-                    CLBLAS_CHECK(gpu_blas_gemm( clblasNoTrans, clblasNoTrans,
-                                                n-(j+1)*nb, m-j*nb, nb,
-                                                c_neg_one,
-                                                dAT(j-1,j+1), lddat,
-                                                dAT(j,  j-1), lddat,
+                OPENCL_BLAS_CHECK(gpu_blas_trsm(OPENCL_BLAS_SIDE_RIGHT, OPENCL_BLAS_TRIANGLE_UPPER, OPENCL_BLAS_NO_TRANS, OPENCL_BLAS_UNIT_DIAGONAL,
```
CNugteren: I believe they are aligned, but your IDE (or GitHub) is showing it differently because the lines are now quite long. I've split it into smaller lines to make it look better (also for the cases below).

src/backend/opencl/blas_clblas.cpp
9prady9: You should be able to use the following in CMake.
```
list(REMOVE_ITEM <list> <value> [<value> ...])
```

src/backend/opencl/magma/magma_blas_clblas.h
```diff
@@ -0,0 +1,74 @@
...
+clblasTranspose      clblas_trans_const( magma_trans_t trans );
+clblasUplo           clblas_uplo_const ( magma_uplo_t  uplo  );
+clblasDiag           clblas_diag_const ( magma_diag_t  diag  );
+clblasSide           clblas_side_const ( magma_side_t  side  );
+
+// Error checking
+#define OPENCL_BLAS_CHECK CLBLAS_CHECK
+
+// Transposing
+#define OPENCL_BLAS_TRANS_TYPE clblasTranspose // the type
```
CNugteren: I'm also against using `#define`, but I saw quite a lot of it already in the codebase, so I thought it would be OK. But still better to get rid of it :-)

Before I make the changes everywhere let's make sure we understand each other. For example for the OPENCL_BLAS_TRANS_TYPE define I could replace it with (for the CLBlast version):
```
enum class OPENCL_BLAS_TRANS_T {
    NO = static_cast<int>(clblast::Transpose::kNo),
    YES = static_cast<int>(clblast::Transpose::kYes),
    CONJUGATE = static_cast<int>(clblast::Transpose::kConjugate)
};
```
Then in the code we would replace for example `OPENCL_BLAS_CONJ_TRANS` with `OPENCL_BLAS_TRANS_T::CONJUGATE`. We now have a scoped enum that is convertible to the CLBlast version, but that's not going to happen automatically. So there will quite some extra code in the 2 back-ends to handle this conversion. Are there any better ways of handling this? We have to keep in mind that although CLBlast provides a scoped enum (`clblast::Transpose`), clBLAS doesn't since it has a C API.

src/backend/opencl/magma/magma_data.h
```diff
@@ -74,7 +74,7 @@ magma_malloc( magma_ptr* ptrPtr, int num)
         size = sizeof(T);
     cl_int err;
     *ptrPtr = clCreateBuffer(opencl::getContext()(), CL_MEM_READ_WRITE, size, NULL, &err );
-    if ( err != clblasSuccess ) {
+    if ( err != CL_SUCCESS ) {
```
CNugteren: OK, so I won't make any changes here, right?

src/backend/opencl/blas_clblast.cpp
CNugteren: So you mean that the `matmul` function can just call either `gpu_blas_gemv_func` or `gpu_blas_gemm_func`? If that's the case then we can merge `blas_clblast.cpp` and `blas_clblas.cpp` into one function. I guess that's what you mean with minimising the specialised logic? Yes, I think that's a good idea!

src/backend/opencl/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 FIND_PACKAGE(OpenCL REQUIRED)
 
+OPTION(USE_CLBLAST "Build ArrayFire with the CLBlast BLAS library for the OpenCL backend" OFF)
+OPTION(USE_CLBLAS "Build ArrayFire with the clBLAS BLAS library for the OpenCL backend" ON)
```
CNugteren: OK, so it will be an either CLBlast/clBLAS option such that we don't have verify whether only one of them is set? Makes sense. Do you perhaps have an example of this construct in the library already? Otherwise I'll figure it out myself based on the link.

umar456: We do that with the CMAKE_BUILD_TYPE option. You can find the code [here](https://github.com/arrayfire/arrayfire/blob/devel/CMakeLists.txt#L38).

src/backend/opencl/blas_clblas.cpp
umar456: Yeah, I would really like to remove those variables. It is not the recommended to use that approach in CMake. I think for now you should use what @9prady9 suggested.

src/backend/opencl/magma/getrf.cpp
umar456: I see the rest of the parameters are on a different line but it would be better if the first three parameters into different line so it doesn't span so many columns.

src/backend/opencl/magma/magma_blas_clblas.h
umar456: Hmm good point. Although it would be safer to use scoped enums, in this case, they would reduce readability. Let's just use unscoped enums for those types for now.

src/backend/opencl/magma/magma_data.h
umar456: This is just a personal preference. It is your call.

src/backend/opencl/CMakeLists.txt
CNugteren: OK, I implemented it. Perhaps you can check if this is what you wanted?

src/backend/opencl/blas_clblas.cpp
CNugteren: This is not needed anymore since the `blas_*.cpp` files are merged into a single `blas.cpp` file. The only #ifdefs left over are in `magma_blas.h`.

src/backend/opencl/magma/magma_blas_clblas.h
CNugteren: Also an unscoped enum doesn't seem so easy to implement: the clBLAS wrapper already takes clBLAS types (and thus unscoped enums), but the CLBlast wrapper takes CLBlast types and thus scoped enums. Not sure how to do this nicely without changing some things to the function interfaces and adding some casts. Perhaps you could clarify what you had in mind?

I did change the types from *_TYPE into *_T.

src/backend/opencl/blas_clblast.cpp
CNugteren: This is now done. It was a great idea, because it saves a lot of code!

CNugteren:
I've implemented most of your requests now (I think). My comments are in the in-line reviews, but they are now hidden because they are 'outdated' according to GitHub. Let me know if something is still unclear and what is still left to be done.

CNugteren:
Any updated from your side?

pavanky:
@CNugteren I am going to do a final review and merge it if everything looks good.

src/backend/opencl/blas.cpp
```diff
@@ -7,113 +7,45 @@
...
+toBlasTranspose(af_mat_prop opt)
 {
-    clblasTranspose out = clblasNoTrans;
     switch(opt) {
-        case AF_MAT_NONE    : out = clblasNoTrans;   break;
-        case AF_MAT_TRANS   : out = clblasTrans;     break;
-        case AF_MAT_CTRANS  : out = clblasConjTrans; break;
+        case AF_MAT_NONE    : return OPENCL_BLAS_NO_TRANS;
+        case AF_MAT_TRANS   : return OPENCL_BLAS_TRANS;
+        case AF_MAT_CTRANS  : return OPENCL_BLAS_CONJ_TRANS;
```
pavanky: 👍

src/backend/opencl/magma/magma_blas.h
```diff
@@ -10,47 +10,29 @@
 #ifndef __MAGMA_BLAS_H
 #define __MAGMA_BLAS_H
 
+// This file contains the common interface for Magma OpenCL BLAS
+// functions. They can be implemented in different back-ends,
+// such as CLBlast or clBLAS.
```
pavanky: 👍

pavanky:
@CNugteren looks good, let me kick off some tests before merging it in.

pavanky:
build arrayfire linux ci
build arrayfire osx opencl ci

pavanky:
@CNugteren The BLAS tests seem to be failing saying `CLBLAS` was not initialized. The OSX one passed because it falls back to using CPU BLAS.

Can you run the tests on your machines and try to see what's the issue ?

CNugteren:
OK, I'll take a look soon.

CNugteren:
I just fixed an accidental removal of an `initBlas` call. On my system the OpenCL BLAS tests now pass again, but I haven't re-ran all the other tests.

pavanky:
build arrayfire linux ci
build arrayfire osx opencl ci

mlloreda:
Thank you for the pull request!

After some simple benchmarking (`matmul`) with the new CLBlast backend, I have been unable to observe speedups over clBLAS. Benchmarks have been performed on following hardware: NVIDIA K20c, AMD S9100, Iris Pro.

k20
s9100
iris_pro

I'm hoping you can steer me in the right direction, here. Am I doing something wrong? Should I be benchmarking other functions? Can we expect to see performance improvements in the future?

CNugteren:
@mlloreda 

Cool, thanks for the benchmarks! A couple of things:
* I have an IrisPro as well and for some reason ArrayFire falls back on CPU BLAS in that case if I recall. That would explain that the performance matches exactly. 
* CLBlast benefits from per-device tuning. Those first 2 devices aren't yet in the [list of tuned devices](https://github.com/CNugteren/CLBlast#using-the-tuners-optional), so performance might still be improved after tuning. Below is an example of a benchmark for a device for which CLBlast is tuned. It shows most of the benefits for irregular sizes.
* I'm planning to publish a webpage with benchmarks for many devices. Not in all cases CLBlast is faster, but it is in most. However, I have a limited number of devices available to test on, so of course it could be that it is slower on a particular device for some reason I haven't investigated yet.

gtx750ti_gemm

pavanky:
@CNugteren are the devices tuned based on full names ?  May be we can figure out if there's a way to tune them based on the architecture. For example there's no reason a K20x and K20c should have different parameters. Even better, would it be possible to *programatically* specify the tuning parameters ? This way the burden can be on the downstream library like arrayfire while

CNugteren:
@pavanky There is an [API](https://github.com/CNugteren/CLBlast/blob/development/include/clblast.h#L648) to manually override parameters, however, I think it is better if it is all arranged properly within CLBlast. Indeed, you might want to have an extra 'architecture' layer with defaults. Currently there is a over-all default, and then a device-type-vendor default (e.g. AMD GPU, Intel CPU). An extra default would e.g. be NVIDIA Pascal GPU. Or perhaps some LUT: some devices now have different OpenCL names based on the OpenCL implementation used (e.g. different on name of an AMD GPU on OS X versus with AMD APP). It's on my list (among many other things) to improve.

shehzan10:
> I have an IrisPro as well and for some reason ArrayFire falls back on CPU BLAS in that case if I recall. That would explain that the performance matches exactly.

@CNugteren This is because of the OpenCL CPU offloading. You can use the environment variable `AF_OPENCL_CPU_OFFLOAD=0` at runtime to disable this behavior.

PR #1743, opened 2017-03-20
canny edge detector

### LINKED ISSUE #16: Canny Edge Detector
Implement Canny Edge Detector for all backends.

9prady9:
Fixes #16 

* CPU backend implementation
* CUDA backend implementation
* OpenCL backend implementation
* Unit tests for Array of less than block size and equals block size.
* Added canny edge to exiting edge detection filters demo example

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -504,7 +504,7 @@ namespace af
     typedef af_random_engine_type randomEngineType;
 #endif
 #if AF_API_VERSION >= 35
-    typedef af_canny_threshold CannyThreshold;
+    typedef af_canny_threshold cannyThreshold;
 #endif
 }
 
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -446,6 +446,13 @@ typedef enum {
 #endif
 ////////////////////////////////////////////////////////////////////////////////
 
+#if AF_API_VERSION >= 35
+typedef enum {
+    AF_MANUAL_THRESHOLD    = 0, ///< User has to define canny thresholds manually
+    AF_AUTO_OTSU_THRESHOLD = 1, ///< Determine canny algorithm thresholds using Otsu algorithm
+} af_canny_threshold;
+#endif
+
 #if AF_API_VERSION >= 34
 typedef enum {
     AF_STORAGE_DENSE     = 0,   ///< Storage type is dense
@@ -495,6 +502,9 @@ namespace af
 #endif
 #if AF_API_VERSION >= 34
     typedef af_random_engine_type randomEngineType;
+#endif
+#if AF_API_VERSION >= 35
+    typedef af_canny_threshold CannyThreshold;
 #endif
 }
 
--- a/include/af/image.h
+++ b/include/af/image.h
@@ -693,8 +693,8 @@ AFAPI array moments(const array& in, const momentType moment=AF_MOMENT_FIRST_ORD
    C++ Interface for canny edge detector
 
    \param[in] in is the input image
-   \param[in] lowThresholdRatio is the lower threshold % of maximum or auto-derived high threshold
    \param[in] thresholdType determines if user set high threshold is to be used or not. It can take values defined by the enum \ref af_canny_threshold
+   \param[in] lowThresholdRatio is the lower threshold % of maximum or auto-derived high threshold
    \param[in] highThresholdRatio is the higher threshold % of maximum value in gradient image used in hysteresis procedure. This value is ignored if \ref AF_AUTO_OTSU_THRESHOLD is chosen as \ref af_canny_threshold
    \param[in] sobelWindow is the window size of sobel kernel for computing gradient direction and magnitude
    \param[in] isFast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
@@ -702,8 +702,8 @@ AFAPI array moments(const array& in, const momentType moment=AF_MOMENT_FIRST_ORD
 
    \ingroup image_func_canny
 */
-AFAPI array canny(const array& in, const float lowThresholdRatio,
-                  const cannyThreshold thresholdType, const float highThresholdRatio,
+AFAPI array canny(const array& in, const cannyThreshold thresholdType,
+                  const float lowThresholdRatio, const float highThresholdRatio,
                   const unsigned sobelWindow = 3, const bool isFast = false);
 #endif
 }
@@ -1399,8 +1399,8 @@ extern "C" {
 
        \param[out] out is an binary array containing edges
        \param[in] in is the input image
-       \param[in] low_threshold_ratio is the lower threshold % of the maximum or auto-derived high threshold
        \param[in] threshold_type determines if user set high threshold is to be used or not. It can take values defined by the enum \ref af_canny_threshold
+       \param[in] low_threshold_ratio is the lower threshold % of the maximum or auto-derived high threshold
        \param[in] high_threshold_ratio is the higher threshold % of maximum value in gradient image used in hysteresis procedure. This value is ignored if \ref AF_AUTO_OTSU_THRESHOLD is chosen as \ref af_canny_threshold
        \param[in] sobel_window is the window size of sobel kernel for computing gradient direction and magnitude
        \param[in] is_fast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
@@ -1410,8 +1410,8 @@ extern "C" {
        \ingroup image_func_canny
     */
     AFAPI af_err af_canny(af_array* out, const af_array in,
-                          const float low_threshold_ratio,
                           const af_canny_threshold threshold_type,
+                          const float low_threshold_ratio,
                           const float high_threshold_ratio,
                           const unsigned sobel_window, const bool is_fast);
 #endif
--- a/include/af/image.h
+++ b/include/af/image.h
@@ -703,7 +703,7 @@ AFAPI array moments(const array& in, const momentType moment=AF_MOMENT_FIRST_ORD
    \ingroup image_func_canny
 */
 AFAPI array canny(const array& in, const float lowThresholdRatio,
-                  const CannyThreshold thresholdType, const float highThresholdRatio,
+                  const cannyThreshold thresholdType, const float highThresholdRatio,
                   const unsigned sobelWindow = 3, const bool isFast = false);
 #endif
 }
--- a/include/af/image.h
+++ b/include/af/image.h
@@ -693,16 +693,18 @@ AFAPI array moments(const array& in, const momentType moment=AF_MOMENT_FIRST_ORD
    C++ Interface for canny edge detector
 
    \param[in] in is the input image
-   \param[in] threshold1 is the lower threshold used in hysteresis procedure.
-   \param[in] threshold2 is the higher threshold used in hysteresis procedure.
-   \param[in] sobel_window is the window size of sobel kernel for computing gradient direction and magnitude
-   \param[in] isfast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
+   \param[in] lowThresholdRatio is the lower threshold % of maximum or auto-derived high threshold
+   \param[in] thresholdType determines if user set high threshold is to be used or not. It can take values defined by the enum \ref af_canny_threshold
+   \param[in] highThresholdRatio is the higher threshold % of maximum value in gradient image used in hysteresis procedure. This value is ignored if \ref AF_AUTO_OTSU_THRESHOLD is chosen as \ref af_canny_threshold
+   \param[in] sobelWindow is the window size of sobel kernel for computing gradient direction and magnitude
+   \param[in] isFast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
    \return binary array containing edges
 
    \ingroup image_func_canny
 */
-AFAPI array canny(const array& in, const float threshold1, const float threshold2,
-                  const unsigned sobel_window = 3, const bool isfast = false);
+AFAPI array canny(const array& in, const float lowThresholdRatio,
+                  const CannyThreshold thresholdType, const float highThresholdRatio,
+                  const unsigned sobelWindow = 3, const bool isFast = false);
 #endif
 }
 #endif
@@ -1397,18 +1399,21 @@ extern "C" {
 
        \param[out] out is an binary array containing edges
        \param[in] in is the input image
-       \param[in] threshold1 is the lower threshold used in hysteresis procedure.
-       \param[in] threshold2 is the higher threshold used in hysteresis procedure.
+       \param[in] low_threshold_ratio is the lower threshold % of the maximum or auto-derived high threshold
+       \param[in] threshold_type determines if user set high threshold is to be used or not. It can take values defined by the enum \ref af_canny_threshold
+       \param[in] high_threshold_ratio is the higher threshold % of maximum value in gradient image used in hysteresis procedure. This value is ignored if \ref AF_AUTO_OTSU_THRESHOLD is chosen as \ref af_canny_threshold
        \param[in] sobel_window is the window size of sobel kernel for computing gradient direction and magnitude
-       \param[in] isfast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
-       \return     ref AF_SUCCESS if the moment calculation is successful,
+       \param[in] is_fast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
+       \return    \ref AF_SUCCESS if the moment calculation is successful,
        otherwise an appropriate error code is returned.
 
        \ingroup image_func_canny
     */
     AFAPI af_err af_canny(af_array* out, const af_array in,
-                          const float threshold1, const float threshold2,
-                          const unsigned sobel_window, const bool isfast);
+                          const float low_threshold_ratio,
+                          const af_canny_threshold threshold_type,
+                          const float high_threshold_ratio,
+                          const unsigned sobel_window, const bool is_fast);
 #endif
 
 #ifdef __cplusplus
--- a/include/af/image.h
+++ b/include/af/image.h
@@ -688,6 +688,22 @@ AFAPI void moments(double* out, const array& in, const momentType moment=AF_MOME
 AFAPI array moments(const array& in, const momentType moment=AF_MOMENT_FIRST_ORDER);
 #endif
 
+#if AF_API_VERSION >= 35
+/**
+   C++ Interface for canny edge detector
+
+   \param[in] in is the input image
+   \param[in] threshold1 is the lower threshold used in hysteresis procedure.
+   \param[in] threshold2 is the higher threshold used in hysteresis procedure.
+   \param[in] sobel_window is the window size of sobel kernel for computing gradient direction and magnitude
+   \param[in] isfast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
+   \return binary array containing edges
+
+   \ingroup image_func_canny
+*/
+AFAPI array canny(const array& in, const float threshold1, const float threshold2,
+                  const unsigned sobel_window = 3, const bool isfast = false);
+#endif
 }
 #endif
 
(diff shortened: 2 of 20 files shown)


src/backend/cpu/kernel/canny.hpp
```diff
@@ -0,0 +1,237 @@
...
+                                    mag1  = (1-alpha)*mag[offset-1] + alpha*mag[offset-dims[0]-1];
+                                    mag2  = (1-alpha)*mag[offset+1] + alpha*mag[offset+dims[0]+1];
+                                } else {
+                                    // North of North West
+                                    alpha = (float)dX[offset]/-dY[offset];
+                                    mag1  = (1-alpha)*mag[offset-dims[0]] + alpha*mag[offset-dims[0]-1];
+                                    mag2  = (1-alpha)*mag[offset+dims[0]] + alpha*mag[offset+dims[0]+1];
+                                }
+                            }
+                        }
```
pavanky: This loop can be collapsed further. There are a few redundant operations.

src/backend/cpu/kernel/canny.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+                        if (mag[offset]>mag1 && mag[offset]>mag2) {
+                            out[offset] = mag[offset] > 255 ? EDGE : (unsigned char)mag[offset];
```
pavanky: Don't hard code it to 255. It should depend on the image type.

src/backend/cpu/kernel/canny.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+template<typename T>
+void traceEdge(T* out, const T* strong, const T* weak, int t, int width)
+{
+    if (!out || !strong || !weak)
+        return;
+
+    const T EDGE = 1;
+
+    std::list<dim_t> edges; // list of edges to be checked
```
pavanky: Why is this a list ? All insertions and deletions are at the back. This should be a vector.

src/backend/cpu/kernel/canny.hpp
```diff
@@ -0,0 +1,237 @@
...
+        // get indices of 8 neighbours
+        dim_t nw = t - width - 1;  // north-west
+        dim_t no = nw + 1;         // north
+        dim_t ne = no + 1;         // north-east
+        dim_t we = t - 1;          // west
+        dim_t ea = t + 1;          // east
+        dim_t sw = t + width - 1;  // south-west
+        dim_t so = sw + 1;         // south
+        dim_t se = so + 1;         // south-east
+
```
pavanky: Put this in an `std::array<dim_t, 8>`

src/backend/cpu/kernel/canny.hpp
```diff
@@ -0,0 +1,237 @@
...
+        if (weak[so] > 0 && out[so] != EDGE)
+        {
+            out[so] = EDGE;
+            edges.push_back(so);
+        }
+        if (weak[se] > 0 && out[se] != EDGE)
+        {
+            out[se] = EDGE;
+            edges.push_back(se);
+        }
```
pavanky: Iterate over array instead of a bunch of if conditions.

src/backend/cuda/kernel/canny.hpp
```diff
@@ -0,0 +1,372 @@
...
+                }
+            }
+
+            if (cmag>mag1 && cmag>mag2) {
+                out[idx] = cmag > 255 ? EDGE : (unsigned char)cmag;
+            } else {
+                out[idx] = SUPPRESS;
+            }
+        }
+    }
```
pavanky: Same issues as the CPU implementation. Try to reduce code. Try not to ard code 255.

src/backend/cuda/kernel/canny.hpp
```diff
@@ -0,0 +1,372 @@
...
+        int sw = outMem[j+1][i-1];
+        int we = outMem[j  ][i-1];
+
+        bool hasStrongNeighbour = nw==STRONG || no==STRONG || ne==STRONG || ea==STRONG ||
+                                  se==STRONG || so==STRONG || sw==STRONG || we==STRONG;
+
+        if (cu==WEAK && hasStrongNeighbour)
+        {
+            outMem[j][i] = STRONG;
+            atomicAdd(&numChanges, 1);
```
pavanky: You are not using the actual count anywhere. You can just do `if (hasChanged == 0) atomicAdd(&hasChanged, 1);`

src/backend/cuda/kernel/canny.hpp
```diff
@@ -0,0 +1,372 @@
...
+        bool hasWeakNeighbour = _nw || _no || _ne || _ea || _se || _so || _sw || _we;
+
+        continueIter = __syncthreads_or(cu==STRONG && hasWeakNeighbour);
+    };
+
+    // If there are any changes in the block
+    // vote for another iteration of processing
+    if (numChanges>0 && lx==0 && ly==0)
+        atomicAdd(&hasChanged, numChanges);
+
```
pavanky: Continuing from previous comment, you can remove this line.

src/backend/cuda/canny.cu
```diff
@@ -0,0 +1,44 @@
...
+using af::dim4;
+
+namespace cuda
+{
+Array<uchar> nonMaximumSuppression(const Array<float>& mag,
+                                   const Array<float>& gx, const Array<float>& gy)
+{
+    mag.eval();
+    gx.eval();
+    gy.eval();
```
pavanky: You don't need to call these for CUDA backend. Same for the rest.

src/backend/opencl/canny.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace opencl
+{
+Array<uchar> nonMaximumSuppression(const Array<float>& mag,
+                                   const Array<float>& gx, const Array<float>& gy)
+{
+    mag.eval();
+    gx.eval();
+    gy.eval();
+
```
pavanky: You don't need to call these for OpenCL backend. Same for the rest.

src/backend/opencl/kernel/canny.cl
```diff
@@ -0,0 +1,321 @@
...
+
+        continueIter = predicates[0];
+    };
+
+    // If there are any changes in the block
+    // vote for another iteration of processing
+    int count = numChanges;
+
+    if (count>0 && lx==0 && ly==0)
+        atomic_add(hasChanged, count);
```
pavanky: Same suggestions from CUDA.

src/backend/cpu/kernel/canny.hpp
9prady9: I intended to peek and pop from front but didn't notice i wrote back. I will change that to front now.

src/backend/cpu/kernel/canny.hpp
9prady9: True. code will look more concise.

src/backend/cpu/kernel/canny.hpp
9prady9: Agree.

src/backend/cuda/kernel/canny.hpp
9prady9: Isn't `hasChanged` a variable from global memory ? Won't doing atomic on shared variable within block and later doing one `atomicAdd` per block if required towards the end be better than directly modifying `hasChanged` - Doesn't `hasChanged==0` result lot of global reads ?

src/backend/cuda/canny.cu
9prady9: Will remove them.

src/backend/opencl/canny.cpp
9prady9: Will remove them.

src/backend/cpu/kernel/canny.hpp
9prady9: I am not sure, may be i have to revisit the algorithm details once again. The output array is of `uchar` type, therefore i didn't think it would be a problem earlier.

pavanky: But you are comparing magnitude with 255. magnitude depends on input type.

src/backend/cpu/kernel/canny.hpp
pavanky: It really doesn't matter if it is front or back. you should be using a vector.

src/backend/cuda/kernel/canny.hpp
pavanky: But you are doing atomicAdd fewer times if you are checking its current value before writing to it.

src/backend/cpu/kernel/canny.hpp
9prady9: I understand what you are saying, I will look into it.

src/backend/cpu/kernel/canny.hpp
9prady9: May be. The reason I used `std::list` is that If we push to `std::vector` frequently, every time the size is increased beyond current capacity, it has to allocate and move elements. With list, the insertion and removal have constant times and the elements doesn't need to be moved at all.

src/backend/cuda/kernel/canny.hpp
9prady9: May be I wasn't clear, my question is about reading the value of `hasChanged`.  Isn't that global variable and hence result in global memory(which has higher latency) read ? 

Since it is single variable, do you think it will be a broadcast ?

9prady9: Okay

9prady9:
@pavanky Please review again, i have updated the kernels in all backends.

9prady9:
@pavanky I have done some digging into colour space ranges and per channel bit representations.

Image's per channel value representation can range from 2-bit to 16-bit, sometimes 32-bit as well - i believe these are HDR images, but other formats may exist.

I shall explain my dilemma with an example. Let's take an `af::array` of float type, the maximum per-channel value this image represents can't be surmised from `af::array` meta data - unless we do a max operation and guess the the channel value range - the theoretical maximum for specific bit representation. So, we can't assume it will be 255, as you have pointed out earlier, but we can guess it.

The above approach will work for types `uint`, `int`, `short`, `ushort` and `uchar` types. But for `float` and `double` types, the image channel values can be of any range: [0-1], [0-255], [0-65535] etc. We can have an `af::array` image whose actual channel range is [0-65535] but the image data has only values from [0-255]. In such cases, our guess based on max will turn out to be incorrect.

I think we should have channel max value parameter for the function which defaults to 255(8-bit channels).

What do you think ?

9prady9:
I checked the CUDA-OSX build log, functions that I haven't touched are failing for some reason with the error message "CUDA launch failure".

canny failed on OpenCL-OSX. I shall check this on Jesse later.

9prady9:
@pavanky Looks like canny kernel on OSX is behaving stupidly. Build log doesn't print anything other than `parse error`.  However, splitting one of the kernel functions into a separate file fixed the error on OSX. Very weird!!!

include/af/defines.h
```diff
@@ -446,6 +446,13 @@ typedef enum {
 #endif
 ////////////////////////////////////////////////////////////////////////////////
 
+#if AF_API_VERSION >= 35
+typedef enum {
+    AF_MANUAL_THRESHOLD    = 0, ///< User has to define canny thresholds manually
+    AF_AUTO_OTSU_THRESHOLD = 1, ///< Determine canny algorithm thresholds using Otsu algorithm
+} af_canny_threshold;
```
pavanky: I don't think Canny edge detection uses any other type of thresholding. Do we really need an enum?

include/af/image.h
```diff
@@ -688,6 +688,24 @@ AFAPI void moments(double* out, const array& in, const momentType moment=AF_MOME
...
+   \param[in] highThresholdRatio is the higher threshold % of maximum value in gradient image used in hysteresis procedure. This value is ignored if \ref AF_AUTO_OTSU_THRESHOLD is chosen as \ref af_canny_threshold
+   \param[in] sobelWindow is the window size of sobel kernel for computing gradient direction and magnitude
+   \param[in] isFast indicates if L<SUB>1</SUB> norm(faster but less accurate) is used to compute image gradient magnitude instead of L<SUB>2</SUB> norm.
+   \return binary array containing edges
+
+   \ingroup image_func_canny
+*/
+AFAPI array canny(const array& in, const float lowThresholdRatio,
+                  const CannyThreshold thresholdType, const float highThresholdRatio,
+                  const unsigned sobelWindow = 3, const bool isFast = false);
```
pavanky: Why are lowerThresholdRatio and highThresholdRatio separated by the type?

src/api/c/canny.cpp
```diff
@@ -0,0 +1,248 @@
...
+    const af::dim4 hDims = hist.dims();
+
+    // reduce along histogram dimension i.e. 0th dimension
+    auto totals  = reduce<af_add_t, uint, float>(hist, 0);
+
+    // tile histogram total along 0th dimension
+    auto ttotals = tile(totals, af::dim4(hDims[0]));
+
+    // pixel frequency probabilities
+    auto probability = arithOp<float, af_div_t>(cast<float, uint>(hist), ttotals, hDims);
```
pavanky: Use reduce_all, then createScalar followed by div. That is more efficient than what is happening here.

src/api/c/canny.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    af_index_t seqBegin[4];
+    af_index_t seqRest[4];
+
+    seqBegin[0].idx.seq = af_make_seq(0, hDims[0]-1, 1);
+    seqBegin[0].isSeq   = true;
+    seqBegin[0].isBatch = false;
+    seqRest[0].idx.seq = af_make_seq(0, hDims[0]-1, 1);
+    seqRest[0].isSeq   = true;
+    seqRest[0].isBatch = false;
```
pavanky: No reason to call `af_make_seq`. You are just storing them in the struct.

src/api/c/canny.cpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        seqBegin[s].idx.seq = af_span;
+        seqBegin[s].isSeq   = true;
+        seqBegin[s].isBatch = false;
+
+        seqRest[s].idx.seq = af_span;
+        seqRest[s].isSeq   = true;
+        seqRest[s].isBatch = false;
+    }
+
```
pavanky: None of this is required, You are only using sequences! No reason to store it in `af_index_t`

src/api/c/canny.cpp
```diff
@@ -0,0 +1,248 @@
...
+    Array<float> sigmas = detail::createEmptyArray<float>(hDims);
+
+    for (unsigned b=0; b<(NUM_BINS-1); ++b)
+    {
+        seqBegin[0].idx.seq.end = (double)b;
+        seqRest[0].idx.seq.begin = (double)(b+1);
+
+        auto frontPartition = index(probability, seqBegin);
+        auto endPartition   = index(probability, seqRest);
```
pavanky: These need to be `createSubArray`

src/api/c/canny.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    GradientPair g  = detail::sobelDerivatives<float, float>(smt, sw);
+    Array<float> gx = g.first;
+    Array<float> gy = g.second;
+
+    Array<float> gmag = gradientMagnitude(gx, gy, isf);
+
+    Array<float> supEdges = detail::nonMaximumSuppression(gmag, gx, gy);
+
+    auto swpair = computeCandidates(supEdges, t1, ct, t2);
```
pavanky: Why auto here and not at the previous pair ?

src/backend/cpu/canny.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+Array<float> nonMaximumSuppression(const Array<float>& mag,
+                                   const Array<float>& gx, const Array<float>& gy)
+{
+    mag.eval();
+    gx.eval();
+    gy.eval();
+
+    Array<float> out = createValueArray<float>(mag.dims(), 0);
+    out.eval();
```
pavanky: Does it need to be a value array ?

src/backend/cpu/canny.cpp
```diff
@@ -0,0 +1,47 @@
...
+    return out;
+}
+
+Array<char> edgeTrackingByHysteresis(const Array<char>& strong, const Array<char>& weak)
+{
+    strong.eval();
+    weak.eval();
+
+    Array<char> out = createValueArray<char>(strong.dims(), 0);
+    out.eval();
```
pavanky: Does this need to be a value array?

include/af/defines.h
9prady9: Based on what I have found in the literature, some frameworks/libraries does provide a version(not necessarily another function, could be overloaded API) of the function where threshold values are chosen automatically. Hence, the current API provides that flexibility of choosing high threshold ratio either to be computed by Otsu's algorithm or set manually.

Providing this choice via an enum instead of bool basically gives us the opportunity to add additional auto thresholding options in the future.

include/af/image.h
9prady9: As I said in the above comment, high threshold ratio will be either set manually by the user or computed using Otsu's algorithm - current options available via enum. 

Lower threshold ratio is a mandatory input from the user. I have put the threshold type enum input before high threshold parameter so that the user will notice the input value of threshold has an effect on the following high threshold ratio parameter.

If this is unsettling order, we can swap them. As long as we don't provide default value to threshold type parameter in C++ API, I don't have an objection with swapping the order of these variables. Because adding a default threshold type would confuse the user on how the algorithm behaves.

src/backend/cpu/canny.cpp
9prady9: Yes, all border pixels need to be set to zero.

src/backend/cpu/canny.cpp
9prady9: Yes, all output pixels need to be set to zero initially. While the edge tracking is in progress, Only the pixels that are edges will be set.

src/api/c/canny.cpp
9prady9: A side effect of working on it after a break. I will remove the typedef and qualify GradientPair declartion with auto.

src/api/c/canny.cpp
9prady9: Agree. createSubArray didn't occur to me at that moment.

src/api/c/canny.cpp
9prady9: Agree.

src/api/c/canny.cpp
9prady9: Agree.

src/api/c/canny.cpp
9prady9: That won't take care of batch of images scenario right ?

pavanky: Hmm. ok.

include/af/defines.h
```diff
@@ -496,6 +503,9 @@ namespace af
 #if AF_API_VERSION >= 34
     typedef af_random_engine_type randomEngineType;
 #endif
+#if AF_API_VERSION >= 35
+    typedef af_canny_threshold CannyThreshold;
```
mlloreda: @9prady9 what do you think about making this typedef lower camel case to stay consistent with the surrounding code?

src/backend/opencl/kernel/nonmax_suppression.cl
```diff
@@ -0,0 +1,119 @@
...
+    // global indices
+    const int gx = get_local_size(0) * (get_group_id(0)-b2*nBBS0) + lx;
+    const int gy = get_local_size(1) * (get_group_id(1)-b3*nBBS1) + ly;
+
+    __local T localMem[SHRD_MEM_HEIGHT][SHRD_MEM_WIDTH];
+
+    __global const T* mag = in + (b2 * inInfo.strides[2] + b3 * inInfo.strides[3] + inInfo.offset) + inInfo.strides[1] + 1;
+    __global const T* dX  = dx + (b2 * dxInfo.strides[2] + b3 * dxInfo.strides[3] + dxInfo.offset) + dxInfo.strides[1] + 1;
+    __global const T* dY  = dy + (b2 * dyInfo.strides[2] + b3 * dyInfo.strides[3] + dyInfo.offset) + dyInfo.strides[1] + 1;
+    __global     T*   out = output + (b2 * oInfo.strides[2] + b3 * oInfo.strides[3]) + oInfo.strides[1] + 1;
```
mlloreda: @9prady9 can you reduce the length of the lines above?

include/af/defines.h
9prady9: We use CamelCase(lower camel case) for the function names in C++ API, Wouldn't enum names in same style be confusing ? Shouldn't we move away to PascalCase(Upper camel case) for enums ?

pavanky: @9prady9 We have been using enums in snakeCase forever. You can't change the style on a whim.

include/af/image.h
pavanky: May be move the thresholdType to be before lowThresholdRatio ?

mlloreda: Agree with @pavanky, here.

9prady9:
@mlloreda @pavanky Check it out now.

PR #1767, opened 2017-04-26
Fixed PVS-Studio issues

IlyaPVS:
We have found and fixed some bugs using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/
We suggests having a look at the emails, sent from @pvs-studio.com.
V501 There are identical sub-expressions to the left and to the right of the '==' operator: yddims[0] == yddims[0] afcpu homography.cpp 74
V522 Dereferencing of the null pointer 'imag' might take place. afcpu reduce.cpp 216
V522 Dereferencing of the null pointer 'imag_val' might take place. afcpu reduce.cpp 464
V523 The 'then' statement is equivalent to the 'else' statement. afcpu median.cpp 47
V656 Variables 'adims', 'bdims' are initialized through the call to the same function. It's probably an error or un-optimized code. Consider inspecting the 'a_info.dims()' expression. Check lines: 42, 43. afcpu solve.cpp 43
V656 Variables 'adims', 'bdims' are initialized through the call to the same function. It's probably an error or un-optimized code. Consider inspecting the 'a_info.dims()' expression. Check lines: 112, 113. afcpu solve.cpp 113

### DIFF
--- a/src/api/c/homography.cpp
+++ b/src/api/c/homography.cpp
@@ -71,7 +71,7 @@ af_err af_homography(af_array *H, int *inliers,
         ARG_ASSERT(1, (xsdims[0] > 0));
         ARG_ASSERT(2, (ysdims[0] == xsdims[0]));
         ARG_ASSERT(3, (xddims[0] > 0));
-        ARG_ASSERT(4, (yddims[0] == yddims[0]));
+        ARG_ASSERT(4, (yddims[0] == xddims[0]));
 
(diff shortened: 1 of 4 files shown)


9prady9:
Thank you for the fixes, we will check out the static analyzer as soon as we can.

9prady9:
build arrayfire linux ci
build arrayfire osx ci
build arrayfire windows ci

PR #1770, opened 2017-05-02
Cache backend symbols in the unified backend

umar456:
unified

Test code:

```cpp
#include <arrayfire.h>
#include <cstdio>
#include <cstdlib>
#include <chrono>

using namespace af;

{
    try {
      int backends = af::getAvailableBackends();
      af::info();

      printf("\nRunning %s...\n", argv[0]);

      for(int i = 500; i < 100000; i+= 500) {
        auto start = std::chrono::high_resolution_clock::now();
        for(int iter = 0; iter < 10000; iter++) {
          af::array blah = af::randu(i);
          af::array constant_val = af::constant(1, i);
          af::array output = dot(blah, constant_val);
        }
        af::sync();
        auto end = std::chrono::high_resolution_clock::now();
        af::deviceGC();
        printf("| %7d | %5llu us |\n", i, std::chrono::duration_cast<std::chrono::microseconds>(end-start).count());
      }
    } catch (af::exception& e) {

        throw;
    }

    return 0;
}
```

### DIFF
--- a/src/api/unified/symbol_manager.hpp
+++ b/src/api/unified/symbol_manager.hpp
@@ -22,6 +22,9 @@ typedef HMODULE LibHandle;
 typedef void* LibHandle;
 #endif
 
+#include <array>
+#include <unordered_map>
+
 namespace unified
 {
 
@@ -34,6 +37,14 @@ const int NUM_ENV_VARS = 2;
                     "for instructions to set up environment for Unified backend.", \
                     AF_ERR_LOAD_LIB)
 
+static int backend_index(af::Backend be) {
+    switch (be) {
+        case AF_BACKEND_CPU: return 0;
+        case AF_BACKEND_CUDA: return 1;
+        case AF_BACKEND_OPENCL: return 2;
+        default: return -1;
+    }
+}
 
 class AFSymbolManager {
     public:
@@ -51,16 +62,22 @@ class AFSymbolManager {
 
         template<typename... CalleeArgs>
         af_err call(const char* symbolName, CalleeArgs... args) {
+            typedef af_err(*af_func)(CalleeArgs...);
             if (!activeHandle) {
                 UNIFIED_ERROR_LOAD_LIB();
             }
-            typedef af_err(*af_func)(CalleeArgs...);
-            af_func funcHandle;
+            static std::array<std::unordered_map<const char*, af_func>, NUM_BACKENDS> funcHandles;
+
+            int index = backend_index(getActiveBackend());
+            af_func& funcHandle = funcHandles[index][symbolName];
+
+            if (!funcHandle) {
 #if defined(OS_WIN)
-            funcHandle = (af_func)GetProcAddress(activeHandle, symbolName);
+                funcHandle = (af_func)GetProcAddress(activeHandle, symbolName);
 #else
-            funcHandle = (af_func)dlsym(activeHandle, symbolName);
+                funcHandle = (af_func)dlsym(activeHandle, symbolName);
 #endif
+            }
             if (!funcHandle) {
                 std::string str = "Failed to load symbol: ";
                 str += symbolName;


src/api/unified/symbol_manager.hpp
```diff
@@ -54,13 +57,18 @@ class AFSymbolManager {
             if (!activeHandle) {
                 UNIFIED_ERROR_LOAD_LIB();
             }
+
             typedef af_err(*af_func)(CalleeArgs...);
-            af_func funcHandle;
+            static std::unordered_map<const char*, std::array<af_func, NUM_BACKENDS>> funcHandles;
+            af_func& funcHandle = funcHandles[symbolName][getActiveBackend()];
```
pavanky: Does it always return `NULL` when it is not found ?

src/api/unified/symbol_manager.hpp
```diff
@@ -54,13 +57,18 @@ class AFSymbolManager {
             if (!activeHandle) {
                 UNIFIED_ERROR_LOAD_LIB();
             }
+
             typedef af_err(*af_func)(CalleeArgs...);
-            af_func funcHandle;
+            static std::unordered_map<const char*, std::array<af_func, NUM_BACKENDS>> funcHandles;
```
pavanky: Move the array outside ?

src/api/unified/symbol_manager.hpp
umar456: I am 96.3% sure it does.

src/api/unified/symbol_manager.hpp
umar456: 👍

src/api/unified/symbol_manager.hpp
```diff
@@ -59,14 +59,14 @@ class AFSymbolManager {
             }
 
             typedef af_err(*af_func)(CalleeArgs...);
-            static std::unordered_map<const char*, std::array<af_func, NUM_BACKENDS>> funcHandles;
-            af_func& funcHandle = funcHandles[symbolName][getActiveBackend()];
+            static std::array<std::unordered_map<const char*, af_func>, NUM_BACKENDS> funcHandles;
+            af_func& funcHandle = funcHandles[getActiveBackend()][symbolName];
```
pavanky: Is `getActiveBackend` returning the enum or is it returning the index ?

umar456: oops. It's a flag. Fixing.

src/api/unified/symbol_manager.hpp
```diff
@@ -37,6 +37,14 @@ const int NUM_ENV_VARS = 2;
                     "for instructions to set up environment for Unified backend.", \
                     AF_ERR_LOAD_LIB)
 
+static int backend_index(af::Backend be) {
+    switch (be) {
+        case AF_BACKEND_CPU: return 0;
+        case AF_BACKEND_CUDA: return 1;
+        case AF_BACKEND_OPENCL: return 2;
+        default: return -1;
```
pavanky: Hmm, we have something called `AF_BACKEND_DEFAULT`. Please check if it works with that.

pavanky:
@umar456 Alternative suggestion: Instead of querying from an `unordered_map` each time, can't the function pointer be cached inside the caller function ? 

This reduces the lookup time from the map. You will need change the definition of `CALL` macros.

src/api/unified/symbol_manager.hpp
umar456: The [SymbolManager](https://github.com/arrayfire/arrayfire/blob/devel/src/api/unified/symbol_manager.cpp#L158) and [setBackend](https://github.com/arrayfire/arrayfire/blob/devel/src/api/unified/symbol_manager.cpp#L203) function sets the activeBackend to one of those backends.

umar456:
I tried that first but the call function's template parameters are based on the type of arguments and not the function name itself. I could change the macro so that we only have an array instead of a map but I didn't want to break anything.

I can change it if you think its worthwhile.

pavanky:
We can still pass the arguments, but instead of calling the function, it will just return the function pointer. The call can be made externally. No reason to cache it inside the `call` function.

umar456:
build arrayfire windows ci

PR #1773, opened 2017-05-08
fix inconsistent behaviour of replace & replace_scalar

### LINKED ISSUE #1772: Bug in replace_scalar
Inconsistent behaviour of `replace` and `replace_scalar` functions although [documentation](http://arrayfire.org/docs/group__data__func__replace.htm#ga13a51c418de0201102413f05aae45ccd) claims similar behaviour.

```
#include <arrayfire.h>
#include <cstdio>
#include <cstdlib>
using namespace af;

{
    try {
        af::setDevice(0);
        array a = constant(2, 10, 10);
        array d = constant(2, 10, 10);
        array b = constant(1, 10, 10, b8);
        array c = constant(3, 10, 10);
        replace(a, b, 3.0);
        replace(d, b, c);
        af_print(a);
        af_print(d);
    } catch (af::exception& e) {
        throw;
    }
    return 0;
}
```

9prady9:
Fixes #1772 

replace(output, cond, input) where input is af::array works as
expected(output has input where cond is false). However, if the
input is a scalar parameter, then the output has inverse behaviour.
This change fixes this inconsistency.

Afer this change, output will have values from input when cond is
false irrespective of whether input is af::array or scalar.

### DIFF
--- a/src/api/c/replace.cpp
+++ b/src/api/c/replace.cpp
@@ -77,7 +77,7 @@ af_err af_replace(af_array a, const af_array cond, const af_array b)
 template<typename T>
 void replace_scalar(af_array a, const af_array cond, const double b)
 {
-    select_scalar<T, true>(getWritableArray<T>(a), getArray<char>(cond), getArray<T>(a), b);
+    select_scalar<T, false>(getWritableArray<T>(a), getArray<char>(cond), getArray<T>(a), b);
 }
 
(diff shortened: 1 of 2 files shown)


pavanky:
@9prady9 can you ensure the behavior and the documentation matchup?

9prady9:
@pavanky The documentation was actual correct, only the function behaviour was inconsistent.

PR #1776, opened 2017-05-15
Fix approx tests for single precision. Reduce memory requirements

umar456:
Tests on OSX builds were sometimes failing because the approx tests were using af::sum<double> and Iris Pro does not support double precision.


umar456:
build arrayfire osx ci

PR #1785, opened 2017-05-18
Fix cpu errors caused by order of deletion in the device manager

umar456:
The order of deletion of the queues and the memory manager should be reversed. The memory manager may still be required by the functions in the async_queue and therefore the memory manager should be deleted after the queue. This PR will change the order of deletion.


pavanky:
@umar456 can you also check this for other backends ?

umar456:
I did. The other backends do not implement their own async queues so this kind of thing will not be a problem there.

PR #1787, opened 2017-05-19
Converting a few static variables to thread_local

pavanky:


src/backend/opencl/cpu/cpu_blas.cpp
```diff
@@ -54,9 +54,9 @@ using std::conditional;
 //                  const void *B, const int ldb, const void *beta,
 //                  void *C, const int ldc);
 #if defined(IS_OPENBLAS)
-    static const bool cplx_void_ptr = false;
+    const bool cplx_void_ptr = false;
```
umar456: We should keep these as static since they are not changing.

src/api/c/graphics_common.cpp
```diff
@@ -174,7 +174,7 @@ namespace graphics
 {
 ForgeManager& ForgeManager::getInstance()
 {
-    static ForgeManager my_instance;
+    thread_local ForgeManager my_instance;
```
9prady9: Why is graphics class thread_local ? We wanted all rendering calls queuing into a single thread, most probably main thread right ? Did the graphics example execute successfully after this change?

pavanky: I didn't see any mutex locks around the calls for the ForgeManager methods. I thought it was safer to make this thread local.

src/backend/opencl/platform.cpp
```diff
@@ -200,10 +200,10 @@ static inline std::string &ltrim(std::string &s)
 
 static std::string platformMap(std::string &platStr)
 {
-    static bool isFirst = true;
+    thread_local bool isFirst = true;
 
     typedef std::map<std::string, std::string> strmap_t;
-    static strmap_t platMap;
+    thread_local strmap_t platMap;
```
umar456: we can leave this as static and get rid of the isFirst variable by doing this:

```cpp
static strmap_t platMap = {
std::make_pair("AMD Accelerated Parallel Processing", "AMD"),
std::make_pair("Intel(R) OpenCL", "INTEL"),
std::make_pair("Portable Computing Language", "POCL"),
...
};

File: src/backend/opencl/platform.cpp [hunk 4]
```diff
@@ -644,7 +644,7 @@ void removeDeviceContext(cl_device_id dev, cl_context ctx)
 
 bool synchronize_calls()
 {
-    static bool sync = getEnvVar("AF_SYNCHRONOUS_CALLS") == "1";
+    thread_local bool sync = getEnvVar("AF_SYNCHRONOUS_CALLS") == "1";
```
umar456: static const would be fine here

src/api/c/device.cpp
```diff
@@ -74,11 +74,10 @@ af_err af_get_active_backend(af_backend *result)
 af_err af_init()
 {
     try {
-        static bool first = true;
-        if(first) {
-            getDeviceInfo();
-            first = false;
-        }
+        static std::once_flag flag;
+        std::call_once(flag, []() {
```
umar456: I would think `std::call_once(flag, getDeviceInfo)` would work here. no need for a lambda.

src/api/c/transform.cpp
```diff
@@ -160,12 +160,12 @@ af_err af_translate(af_array *out, const af_array in, const float trans0, const
 {
 
     try {
-        static float trans_mat[6] = {1, 0, 0,
-                                     0, 1, 0};
+        thread_local float trans_mat[6] = {1, 0, 0,
```
umar456: there is almost no point to make this thread local. We aren't gaining much by saving its state.

src/api/unified/symbol_manager.cpp
```diff
@@ -113,7 +113,7 @@ LibHandle openDynLibrary(const int bknd_idx, int flag=RTLD_LAZY)
          * /usr/local/arrayfire-3/lib
         */
         if (retVal == NULL) {
-            static
+            const
```
umar456: Static const?

src/backend/opencl/cpu/cpu_blas.cpp
pavanky: Wouldn't `const` make more sense if they are not changing?

umar456: As far as I remember, static const are compile-time variables and const is a runtime variable. I don't know if the compiler can make assumptions about the value if it is just const.

src/api/c/device.cpp
pavanky: This did not work.

src/backend/opencl/platform.cpp
pavanky: I changed this but it didnt update here

pavanky:
build arrayfire linux ci
build arrayfire osx opencl ci
build arrayfire osx cuda ci

PR #1803, opened 2017-05-25
Added clFFT and clBLAS deps to `{scan,sort}_by_key` modules.

mlloreda:
These missing dependencies were causing ArrayFire build failures under certain
configurations. For example, the issue often manifested itself when building
ArrayFire in parallel with graphics disabled.


pavanky:
This doesn't make sense. There should be another issue at play here.

9prady9:
@pavanky I think what @mlloreda means is that due to lack of the target dependency we faced issues such as `clFFT.h` not found when we do `make -j8` - where sort_by_key went ahead to build even before clFFT external project build finished. 

He is not referring to the other issues we have been debugging that are related to thread safety.

mlloreda:
Thanks for clarifying, @9prady9.

pavanky:
@9prady9 The question is why is `sort_by_key` including `clFFT.h` ? That is the broken part, not the missing dependency.

mlloreda:
@pavanky 

```
In file included from /tmp/arrayfire/src/backend/opencl/platform.hpp:32:0,
                 from /tmp/arrayfire/src/backend/opencl/Param.hpp:11,
                 from /tmp/arrayfire/src/backend/opencl/kernel/sort_by_key.hpp:13,
                 from /tmp/arrayfire/src/backend/opencl/kernel/sort_by_key_impl.hpp:11,
                 from /tmp/arrayfire/src/backend/opencl/kernel/sort_by_key/sort_by_key_impl.cpp:10:
/tmp/arrayfire/src/backend/opencl/clfft.hpp:12:19: fatal error: clFFT.h: No such file or directory
``

**Comment by 9prady9** [comment]
The order(top-to-bottom) of header inclusions that lead to this problem I think is below
```
~/arrayfire (devel) $vim src/backend/opencl/kernel/sort_by_key.hpp 
~/arrayfire (devel) $vim src/backend/opencl/debug_opencl.hpp 
~/arrayfire (devel) $vim src/backend/opencl/err_opencl.hpp 
~/arrayfire (devel) $vim src/backend/opencl/err_opencl.hpp 
~/arrayfire (devel) $vim src/backend/opencl/platform.hpp 
~/arrayfire (devel) $vim src/backend/opencl/clfft.hpp 
```

pavanky:
Hmm it seems `platform.hpp` seems to be included indiscriminately. But solving that takes a longer time. So I suppose this is fine.

PR #1804, opened 2017-05-25
add missing template specialisations for sparse arith ops

9prady9:
* Add af_mul_t, af_div_t op specialisation for opencl::arithOpD
* Add af_add_t, af_sub_t op specialisation for opencl::arithOpS

The above missing specialisations are causing linking errors on Windows platform in Debug mode builds.


umar456:
Could you add tests for these specializations?

pavanky:
@umar456 They aren't being (shouldn't be)  used by external facing code. The specializations were added to avoid issues when compiling in debug mode in windows.

Issue #1760, opened 2017-04-17
Internal Error 998 caused by af::moddims() function


pavolmarak:
Hello everyone,

I have very simple code snippet which I am not able to run because everytime I receive **998 Internal error exception**. 

**This is how my code looks like:**

```cpp
try{
        int w = 17;
        int tx = 30;
        int ty = 30;

        af::array A(w*w,w*w*ty*tx);
        af::array B(w*w,1,w*ty*tx);

        A= af::moddims(A,w*w,w,w*ty*tx);   
        A=  A* af::tile(B,1,w); // multiplication - when I comment this line, it works without 998 exception
        A= af::moddims(A,w*w,w*w*ty*tx);
}
    catch(af::exception e)
    {
        qDebug() << e.what();
    }
```

**This is my error message:**

```cpp
ArrayFire Exception (Internal error:998):
In function void __cdecl cuda::evalNodes<float>(class std::vector<struct cuda::Param<float>,class std::allocator<struct cuda::Param<float> > > &,class std::vector<class cuda::JIT::Node *,class std::allocator<class cuda::JIT::Node *> >)
In file src\backend\cuda\jit.cpp:658
CU Error (1)

In function class af::array __cdecl af::moddims(const class af::array &,const unsigned int,const __int64 *const )
In file src\api\cpp\data.cpp:191
```

I tried doing `af::eval(A)` before and after the line with multiplication with no effect, I thought it has something to do with lazy evaluation. However, when I leave out the line with multiplication, it works without any problem. 

Is there someone who could help me with this?

Thank you.

`af::info():`

```cpp
ArrayFire v3.4.2 (CUDA, 64-bit Windows, build 2da9967)
Platform: CUDA Toolkit 8, Driver: CUDA Driver Version: 8000
[0] GeForce GT 730M, 1024 MB, CUDA Compute 3.5
```

pavanky:
Can you run your program from commandline with `AF_PRINT_ERRORS=1` exported ?

pavolmarak:
I launched it from CMD and got this:

```
In function void __cdecl cuda::evalNodes<float>(class std::vector<struct cuda::Param<float>,class std::allocator<struct cuda::Param<float> > > &,class std::vector<class cuda::JIT::Node *,class std::allocator<class cuda::JIT::Node *> >)
In file src\backend\cuda\jit.cpp:658
CU Error (1)
```

[...]

Issue #1702, opened 2017-01-05
Reduce failure due to large second dimension


mlloreda:
First reported [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/arrayfire-users/I8g-Ec9oK8s/UIxaZAhoEgAJ).

Failure is due to exceeding launch configuration limits on CUDA. This appears to be related to issue #133.

Code to reproduce

```
TEST(Reduce, ISSUE_1702)
{
  af::array A = af::randu(1, 65535*8 + 1);
  af::array B = af::sum(A, 0);
  SUCCEED();
}
```

Error message:

```
Running main() from gtest_main.cc
Note: Google Test filter = *ISSUE_1702*
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Reduce
[ RUN      ] Reduce.ISSUE_1702
unknown file: Failure
C++ exception with description "ArrayFire Exception (Internal error:998):
In function void cuda::kernel::reduce_first_launcher(cuda::Param<To>, cuda::CParam<T>, cuda::uint, cuda::uint, cuda::uint, bool, double) [with Ti = float; To = float; af_op_t op = (af_op_t)0u; cuda::uint = uns\
igned int]
In file src/backend/cuda/kernel/reduce.hpp:328
CUDA Error (9): invalid configuration argument


In function af::array af::sum(const af::array&, int)
In file src/api/cpp/reduce.cpp:21" thrown in the test body.
[  FAILED  ] Reduce.ISSUE_1702 (694 ms)
[----------] 1 test from Reduce (694 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (694 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Reduce.ISSUE_1702

 1 FAILED TEST

```

pavanky:
Not a bug in reduce. Bug in JIT. Duplicate of #133

pavanky:
@mlloreda reopen the issue if it fails with randu instead of constant.

mlloreda:
@pavanky, failing with `randu` as well.

mlloreda:
@pavanky done.

pavanky:
Fixed in v3.5

Issue #1700, opened 2017-01-04
JIT is not exception safe.


gaika:
I'm mucking with a Julia wrapper for ArrayFire, and I get these weird dumps any time over ~10 million elements are in a vector.  Same thing would work correctly with a 1-2 million elements `s` vector. Any idea what I'm doing wrong? Underneath it is just calling `af_sum_all` over `af_mul` with batch flag on.
````
julia> afInfo()
ArrayFire v3.5.0 (CUDA, 64-bit Linux, build 1251007)
Platform: CUDA Toolkit 8, Driver: 375.26
[0] GeForce GTX 1080, 8114 MB, CUDA Compute 6.1
-1- Quadro K1200, 4042 MB, CUDA Compute 5.0

julia> typeof(s)
ArrayFire.AFArray{Float32,1}
julia> sum(0.5f0 * s * s)
Only PHI nodes may reference their own value!
  %val1 = fmul float %val0, %val1
Broken module found, compilation terminated!
4115, 1
LOG:

target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:1
28:128-n16:32:64"                                                                                                                    


target triple = "nvptx64-unknown-cuda"

define void @KER8644411623765721933 (
float* %in0,
i32 %dim00,i32 %dim10,i32 %dim20,i32 %dim30,
i32 %str10,i32 %str20,i32 %str30,
float* %out2,

i32 %ostr0, i32 %ostr1, i32 %ostr2, i32 %ostr3,
i32 %odim0, i32 %odim1, i32 %odim2, i32 %odim3,
i32 %blkx, i32 %blky, i32 %ndims )

{

entry:


%tidx = call i32 @llvm.nvvm.read.ptx.sreg.tid.x()
%bdmx = call i32 @llvm.nvvm.read.ptx.sreg.ntid.x()
%bidx = call i32 @llvm.nvvm.read.ptx.sreg.ctaid.x()
%bidy = call i32 @llvm.nvvm.read.ptx.sreg.ctaid.y()
%gdmx = call i32 @llvm.nvvm.read.ptx.sreg.nctaid.x()



%tidy = call i32 @llvm.nvvm.read.ptx.sreg.tid.y()
%bdmy = call i32 @llvm.nvvm.read.ptx.sreg.ntid.y()
[...]

PR #1807, opened 2017-05-30
Max grid dimensions for cuda backend

syurkevi:
This is a WIP branch with updates to each kernel fixing the CUDA API's limitation for the maximum number of y-blocks.  
Related to #133 #1513 #1702 #1700 #1760
Currently, tests only test for ability to handle larger array sizes but do not check for correctness. Tests for correctness will be added as more kernels are fixed since kernels like reduce() or index() are necessary to test, but themselves don't work with larger dimensions. Lots of changes to come!

### DIFF
--- a/src/backend/cuda/kernel/approx.hpp
+++ b/src/backend/cuda/kernel/approx.hpp
@@ -23,104 +23,88 @@ namespace cuda
         static const int TY = 16;
         static const int THREADS = 256;
 
-        // \param largeYWDim true denotes 2nd & 4th dimensions greater than device limits
-        // \param iterPerBlockY number of iterations along grid.Y per block
-        template<typename Ty, typename Tp, int order, bool largeYWDim>
+        template<typename Ty, typename Tp, int order>
         __global__
         void approx1_kernel(Param<Ty> out, CParam<Ty> in, CParam<Tp> xpos,
                             const float offGrid, const int blocksMatX, const bool batch,
-                            af_interp_type method, int iterPerBlockY)
+                            af_interp_type method)
         {
             const int idy = blockIdx.x / blocksMatX;
             const int blockIdx_x = blockIdx.x - idy * blocksMatX;
             const int idx = blockIdx_x * blockDim.x + threadIdx.x;
 
-            // For smaller kernels statically set iterPerPlockY
-            // to 1 (register count optimization)
-            if(!largeYWDim) { iterPerBlockY = 1; }
+            const int idw = (blockIdx.y + blockIdx.z * gridDim.y) / out.dims[2];
+            const int idz = (blockIdx.y + blockIdx.z * gridDim.y) - idw * out.dims[2];
 
-            for(int ib = 0; ib < iterPerBlockY; ++ib) {
-                const int idw = (blockIdx.y + ib * gridDim.y) / out.dims[2];
-                const int idz = (blockIdx.y + ib * gridDim.y) - idw * out.dims[2];
+            if (idx >= out.dims[0] || idy >= out.dims[1] ||
+                idz >= out.dims[2] || idw >= out.dims[3])
+                return;
 
-                if (idx >= out.dims[0] || idy >= out.dims[1] ||
-                    idz >= out.dims[2] || idw >= out.dims[3])
-                    return;
+            const int omId = idw * out.strides[3] + idz * out.strides[2]
+                                               + idy * out.strides[1] + idx;
+            int xmid = idx;
+            if(batch) xmid += idw * xpos.strides[3] + idz * xpos.strides[2] + idy * xpos.strides[1];
 
-                const int omId = idw * out.strides[3] + idz * out.strides[2]
-                                                   + idy * out.strides[1] + idx;
-                int xmid = idx;
-                if(batch) xmid += idw * xpos.strides[3] + idz * xpos.strides[2] + idy * xpos.strides[1];
-
-                const Tp x = xpos.ptr[xmid];
-                if (x < 0 || in.dims[0] < x+1) {
-                    out.ptr[omId] = scalar<Ty>(offGrid);
-                    return;
-                }
-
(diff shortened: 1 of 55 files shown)


pavanky:
@syurkevi The simplest solution is to start using `gridDim.z` when `gridDim.y`  gets too big. You can then replace `blockIdx.y` with `blockIdx_y` where `blockIdx_y == gridDim.z == 1 ? blockIdx.y : (blockIdx.z * gridDim.y + blockIdx.y);`

Creating unnecessary templates and parameters is unnecessary and just increases the compilation time.

umar456:
@pavanky The template parameter was done because of performance reasons. The non-template version was using was nearly 4-5ms slower than the templated version ( something like 14ms vs. 19ms). This was because the for loop for this change requires a few extra registers.

pavanky:
@umar456 That seems a bit too much. I am not sure if it is just the registers. I still think the 3D grid option is better. You don't need a loop for this.

src/backend/cuda/kernel/approx.hpp
```diff
@@ -23,22 +23,19 @@ namespace cuda
...
         {
-            const int idw = blockIdx.y / out.dims[2];
-            const int idz = blockIdx.y - idw * out.dims[2];
-
             const int idy = blockIdx.x / blocksMatX;
             const int blockIdx_x = blockIdx.x - idy * blocksMatX;
             const int idx = blockIdx_x * blockDim.x + threadIdx.x;
 
+            const int idw = (blockIdx.y + blockIdx.z * gridDim.y) / out.dims[2];
+            const int idz = (blockIdx.y + blockIdx.z * gridDim.y) - idw * out.dims[2];
```
pavanky: Calculate `blockIdx.y + blockIdx.z * gridDim.y` only once.

src/backend/cuda/kernel/approx.hpp
```diff
@@ -73,13 +70,12 @@ namespace cuda
...
+            const int blockIdx_x = blockIdx.x - idz * blocksMatX;
+            const int idx = threadIdx.x + blockIdx_x * blockDim.x;
 
-            int blockIdx_x = blockIdx.x - idz * blocksMatX;
-            int blockIdx_y = blockIdx.y - idw * blocksMatY;
-
-            int idx = threadIdx.x + blockIdx_x * blockDim.x;
-            int idy = threadIdx.y + blockIdx_y * blockDim.y;
+            const int idw = (blockIdx.y + blockIdx.z * gridDim.y) / blocksMatY;
+            const int blockIdx_y = (blockIdx.y + blockIdx.z * gridDim.y) - idw * blocksMatY;
```
pavanky: Same comment as before.

9prady9:
@syurkevi @pavanky Shouldn't this go into hotfixes instead of devel ?

pavanky:
@syurkevi can you also modify the blocks.y and blocks.z calculation from

```
            const int maxBlocksY    = cuda::getDeviceProp(cuda::getActiveDeviceId()).maxGridSize[1];
            const int blocksPerMatZ = divup(blocks.y, maxBlocksY);
            if(blocksPerMatZ > 1) {
                blocks.y = maxBlocksY;
                blocks.z = blocksPerMatZ;
            }
```
to
```
            const int maxBlocksY    = cuda::getDeviceProp(cuda::getActiveDeviceId()).maxGridSize[1];
            blocks.z = divup(blocks.y, maxBlocksY);
            blocks.y = divup(blocks.y, blocks.z);
    ```

This will result in fewer blocks being launched. For example if original blocks.y == 100,000 and maxBlocksY == 65535, the original would launch (blocks.x, 65535, 2), while the new code would launch (blocks.x, 50000, 2);

**Comment by mlloreda** [comment]
build arrayfire linux cuda ci
build arrayfire osx cuda ci

**Comment by syurkevi** [comment]
build arrayfire linux cuda ci
build arrayfire osx cuda ci

File: src/backend/cuda/kernel/ireduce.hpp [hunk 3]
```diff
@@ -194,6 +194,11 @@ namespace kernel
         dim3 blocks(blocks_dim[0] * blocks_dim[2],
                     blocks_dim[1] * blocks_dim[3]);
 
+        printf("dim [%d %d %d]\n", blocks.x, blocks.y, blocks.z);
```
pavanky: Can you remove the printf?

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -430,6 +438,7 @@ namespace kernel
     template<typename T, af_op_t op>
     void ireduce(Param<T> out, uint *olptr, CParam<T> in, int dim)
     {
+        printf("AMNIHRERERE");
```
pavanky: Remove the printf.

src/backend/cuda/kernel/join.hpp
```diff
@@ -12,6 +12,7 @@
 #include <Param.hpp>
 #include <err_cuda.hpp>
 #include <debug_cuda.hpp>
+#include <iostream>
```
pavanky: remove iostream.

test/join.cpp
```diff
@@ -114,6 +114,32 @@ void joinTest(string pTestFile, const unsigned dim, const unsigned in0, const un
...
+    af::deviceGC();
+    {
+        af::array in = af::randu(nx, ny, u8);
+        af::array joined = af::join(0, in, in);
+        af::dim4 in_dims = in.dims();
+        af::dim4 joined_dims = joined.dims();
+
+        ASSERT_EQ(2*in_dims[0], joined_dims[0]);
+        //todo: uncomment as assert
+        //printf("%f\n", af::sum<float>((joined(0, af::span) - joined(1, af::span)).as(f32)));
```
pavanky: Add the assert back ?

umar456:
@pavanky Are you happy with this PR?

PR #1809, opened 2017-06-01
Add function to get scalar in the C API

### LINKED ISSUE #1725: Function to return scalar value from the C api
The C++ API is copying the entire array to host and returning one value. This is entirely unnecessary. 
https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/array.cpp#L973

It'll also be great if we can specify the n'th element to be copied.

9prady9:
Fixes #1725

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -1668,10 +1668,23 @@ extern "C" {
         \returns error codes
     */
     AFAPI af_err af_is_sparse       (bool *result, const af_array arr);
+#endif
+
+#if AF_API_VERSION >= 35
+    /**
+        \brief Get first element from an array
+
+        \param[out] output_value is the element requested
+        \param[in] arr is the input array
+        \return \ref AF_SUCCESS if the execution completes properly
+    */
+    AFAPI af_err af_get_scalar(void* output_value, const af_array arr);
+#endif
+
     /**
         @}
     */
-#endif
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/api/c/array.cpp
+++ b/src/api/c/array.cpp
@@ -11,6 +11,7 @@
 #include <platform.hpp>
 #include <handle.hpp>
 #include <backend.hpp>
+#include <copy.hpp>
 #include <sparse_handle.hpp>
 
 using namespace detail;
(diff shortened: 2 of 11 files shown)


9prady9:
build arrayfire linux ci
build arrayfire osx opencl ci

9prady9:
build arrayfire windows ci

pavanky:
@9prady9 This does not break the user code, but it will break the ABI. The symbol generated for `array::scalar` will be different.

9prady9:
Changing the function avoid ABI breakage.

src/backend/cpu/copy.cpp
```diff
@@ -131,4 +131,33 @@ SPECILIAZE_UNUSED_COPYARRAY(cdouble, uintl)
...
+template<typename T>
+T getScalar(const Array<T> &in, const unsigned idx)
+{
+    in.eval();
+    getQueue().sync();
+
+    const T* inPtr = in.get() + idx;
+
+    T retVal;
+    std::copy(inPtr, inPtr+1, &retVal);
```
pavanky: Don't launch a copy for this. This is a simple index operation.

include/af/array.h
```diff
@@ -1668,10 +1668,24 @@ extern "C" {
...
     /**
-        @}
+        \brief Get nth element element from an array
+
+        \param[out] out is the nth element requested
+        \param[in] arr is the input array
+        \param[in] idx is the index of the nth element we want from the input array
+        \return \ref AF_SUCCESS if the execution completes properly
     */
+    AFAPI af_err af_get_scalar(double* real, double* imag, const af_array arr, const unsigned idx);
```
pavanky: Remove `idx`. The C++ API is using `idx == 0` and using anything else would be a problem if the array was indexed previously.

mlloreda:
build arrayfire linux ci
build arrayfire osx ci

test/array.cpp
```diff
@@ -503,3 +504,13 @@ TEST(Device, JIT)
     array a = constant(1, 5, 5);
     ASSERT_EQ(a.device<float>() != NULL, 1);
 }
+
+TYPED_TEST(Array, Scalar)
+{
+    if (noDoubleTests<TypeParam>()) return;
+
+    dtype type = (dtype)af::dtype_traits<TypeParam>::af_type;
+    array a = constant(10.0, dim4(1), type);
```
pavanky: perhaps use randu instead of iota or constant.

src/api/c/array.cpp
```diff
@@ -391,3 +392,47 @@ INSTANTIATE(af_is_bool        , isBool        )
 INSTANTIATE(af_is_sparse      , isSparse      )
 
 #undef INSTANTIATE
+
+
+af_err af_get_scalar(double* real_val, double* imag_val, const af_array arr)
```
9prady9: Yes I guess, but I was merely following what other functions such as reductions are currently doing.

pavanky: This follows what we were doing for reductions. 

@9prady9 We should probably fix this to be `void *` since it is similar to `af_get_host_ptr`. Also only return a single `void *` instead of two outputs.

9prady9: Got it.

Issue #1806, opened 2017-05-30
Problem when building with -DUSE_SYSTEM_CLBLAST


bzcheeseman:
I get the following error when I try to use a CLBlast that I've already installed. This is during the arrayfire `make -j8` step. I ran
```
cmake .. -DBUILD_OPENCL=ON -DBUILD_CUDA=OFF -DCMAKE_BUILD_TYPE=Release -DOPENCL_BLAS_LIBRARY=CLBlast -DUSE_SYSTEM_CLBLAST=ON
```
 I'm running OSX 10.12.4, clang -v outputs
```
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
The error is
```
Undefined symbols for architecture x86_64:
  "clblast::StatusCode clblast::Gemm<std::__1::complex<double> >(clblast::Layout, clblast::Transpose, clblast::Transpose, unsigned long, unsigned long, unsigned long, std::__1::complex<double>, _cl_mem*, unsigned long, unsigned long, _cl_mem*, unsigned long, unsigned long, std::__1::complex<double>, _cl_mem*, unsigned long, unsigned long, _cl_command_queue**, _cl_event**)", referenced from:
      opencl::Array<cl_double2> opencl::matmul<cl_double2>(opencl::Array<cl_double2> const&, opencl::Array<cl_double2> const&, af_mat_prop, af_mat_prop) in blas.cpp.o
      int magma_gebrd_hybrid<cl_double2>(int, int, cl_double2*, int, _cl_mem*, unsigned long, int, void*, void*, cl_double2*, cl_double2*, cl_double2*, int, _cl_command_queue*, int*, bool) in gebrd.cpp.o
      int magma_getrf_gpu<cl_double2>(int, int, _cl_mem*, unsigned long, int, int*, _cl_command_queue*, int*) in getrf.cpp.o
      int magma_larfb_gpu<cl_double2>(magma_side_t, magma_trans_t, magma_direct_t, magma_storev_t, int, int, int, _cl_mem*, unsigned long, int, _cl_mem*, unsigned long, int, _cl_mem*, unsigned long, int, _cl_mem*, unsigned long, int, _cl_command_queue*) in larfb.cpp.o
[...]

PR #1810, opened 2017-06-01
Add FindCLBlast.cmake

bzcheeseman:
Fixes a bug in issue #1806 and finds CLBlast when it's already installed on the system

### DIFF
--- a/CMakeModules/FindCLBlast.cmake
+++ b/CMakeModules/FindCLBlast.cmake
@@ -1,0 +1,57 @@
+# Modification of FindclBLAS from the following repo
+# https://github.com/CNugteren/CLBlast
+
+# - Find CLBlast library
+#
+# This module finds an installed CLBlast library.
+#
+# This module sets the following variables:
+#  CLBLAST_FOUND - set to true if a library implementing the CBLAS interface is found
+#  CLBLAST_LIBRARIES - list of libraries (using full path name) to link against to use CBLAS
+#  CLBLAST_INCLUDE_DIR - path to includes
+
+# Sets the possible install locations
+set(CLBLAST_HINTS
+  ${CLBLAST_ROOT}
+  $ENV{CLBLAST_ROOT}
+)
+set(CLBLAST_PATHS
+  /usr/
+  /usr/local/
+)
+
+# Finds the include directories
+find_path(CLBLAST_INCLUDE_DIRS
+  NAMES clblast.h
+  HINTS ${CLBLAS_HINTS}
+  PATH_SUFFIXES include inc include/x86_64 include/x64
+  PATHS ${CLBLAST_PATHS}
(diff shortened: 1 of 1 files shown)


pavanky:
@bzcheeseman @CNugteren can this be made part of the CLBlast package so that downstream libraries like arrayfire can simply include `FIND_PACKAGE(CLBLAST)` ?

CNugteren:
I believe something like this is automatically generated by CMake when installing CLBlast, but to be honest I never tried that myself. I can take a look in a week from now or so, but I'm a bit busy right now.

CNugteren:
Sorry for my late reply, I completely forgot about this. Yes, CLBlast generates a CMake [package configuration file](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-configuration-file) at installation (see [this PR](https://github.com/CNugteren/CLBlast/pull/86)). This should make it possible to use `FIND_PACKAGE(CLBLAST)`.

I have just tested this. The following command:
```
cmake .. -DBUILD_OPENCL=ON -DBUILD_CUDA=OFF -DCMAKE_BUILD_TYPE=Release -DOPENCL_BLAS_LIBRARY=CLBlast -DUSE_SYSTEM_CLBLAST=ON
```
results in an error when CLBlast is not installed:
```
CMake Error at src/backend/opencl/CMakeLists.txt:75 (FIND_PACKAGE):
  By not providing "FindCLBlast.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "CLBlast", but
  CMake did not find one.

  Could not find a package configuration file provided by "CLBlast" with any
  of the following names:

    CLBlastConfig.cmake
    clblast-config.cmake

  Add the installation prefix of "CLBlast" to CMAKE_PREFIX_PATH or set
  "CLBlast_DIR" to a directory containing one of the above files.  If
  "CLBlast" provides a separate development package or SDK, be sure it has
  been installed.
```
However, by installing CLBlast in a system location or by setting CMAKE_PREFIX_PATH to include the path of the CLBlast installation, CMake finds CLBlast's `CLBlastConfig.cmake` file and continues happily. However, it seems not to help in finding `clblast.h`.

I tried reading the documentation, but it is not entirely clear to me. Anyone around here who has done this before? The relevant CLBlast CMake line is [here](https://github.com/CNugteren/CLBlast/blob/master/CMakeLists.txt#L276). Or maybe we have to change the lines `INCLUDE_DIRECTORIES(${CLBLAST_INCLUDE_DIRS})` and `LINK_DIRECTORIES(${CLBLAST_LIBRARY_DIR})` in the ArrayFire CMake configuration to something else? But to what?

umar456:
The only reason to create Find* files is for packages that do not use CMake as their build system. The CLBlast package is correctly creating the config files which CMake should be able to find if the package was installed correctly.

PR #1815, opened 2017-06-06
Add locks around cuda device api calls

umar456:
CUDA Driver API is not thread-safe. This PR protects these calls in the JIT kernel compilation and launches.

### DIFF
--- a/src/api/c/device.cpp
+++ b/src/api/c/device.cpp
@@ -74,7 +74,7 @@ af_err af_get_active_backend(af_backend *result)
 af_err af_init()
 {
     try {
-        static std::once_flag flag;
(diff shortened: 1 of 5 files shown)


pavanky:
I think it is cleaner to have device specific mutexes.

WilliamTambellini:
Sounds good to me. Just sad that we need the sparse_matmul lock.

pavanky:
@WilliamTambellini all the calls are asynchronous. The lock time should be fairly minimal.

WilliamTambellini:
Hi guys, @umar456  when do you plan to merge this one ? Tks.

umar456:
build arrayfire windows ci

PR #1820, opened 2017-06-12
Fix jit lock to before the find function in getKernel.

umar456:
The scope of the lock was too narrow.

### DIFF
--- a/src/backend/cuda/jit.cpp
+++ b/src/backend/cuda/jit.cpp
@@ -490,6 +490,8 @@ char linkError[size];
 
 static kc_entry_t compileKernel(const char *ker_name, string jit_ker)
 {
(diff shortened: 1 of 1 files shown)


WilliamTambellini:
Works also if moving the lock to the very beginning of compileKernel(...).

pavanky:
Looks like we missed the `nvvm*` calls for the lock. Putting it at the top of `compileKernel` takes care of this issue.

@WilliamTambellini thanks for testing this.

src/backend/cuda/jit.cpp
```diff
@@ -490,6 +490,8 @@ char linkError[size];
 
 static kc_entry_t compileKernel(const char *ker_name, string jit_ker)
 {
+    std::lock_guard<std::mutex> lock(getDriverApiMutex(getActiveDeviceId()));
+
```
pavanky: :+1:

PR #1824, opened 2017-06-15
Add the ability to deep copy for sparse arrays

umar456:


src/backend/SparseArray.cpp
```diff
@@ -140,6 +146,11 @@ SparseArray<T> createArrayDataSparseArray(
 }
 
 template<typename T>
+SparseArray<T> copySparseArray(const SparseArray<T>& other, bool copy) {
+  return SparseArray<T>(other, copy);
+}
+
```
pavanky: Why do you need a `copy` parameter ? I don't see it being set to `false` anywhere.

test/sparse.cpp
```diff
@@ -118,3 +118,55 @@ TEST(Sparse, ISSUE_1745)
...
+
+    // Make sure that the deep copy allocated additional memory
+    af::deviceMemInfo(&alloc_bytes, &alloc_buffers,
+                      &lock_bytes, &lock_buffers);
+
+    ASSERT_NE(s.get(), s2.get()) << "The sparse arrays point to the same "
+                                    "af_array object.";
+    ASSERT_EQ(size_of_alloc * 2,
+              lock_bytes) << "The number of bytes allocated by the deep copy do "
+                             "not match the original array";
```
pavanky: This may not always be true if other tests are run before this. You may need to call `af::deviceGC()` at the beginning of the test.

test/sparse.cpp
```diff
@@ -118,3 +118,55 @@ TEST(Sparse, ISSUE_1745)
...
+
+    ASSERT_EQ(buffers_per_sparse * 2,
+              lock_buffers) << "The number of buffers allocated by the deep "
+                               "copy do not match the original array";
+  }
```
pavanky: Check the values as well.

test/sparse.cpp
pavanky: You will also need to call it before getting the memory info.

src/backend/SparseArray.cpp
umar456: Good point. You only need it for the constructors.

test/sparse.cpp
umar456: I am only comparing against the locked bytes. I thought that would be deterministic.

test/sparse.cpp
umar456: Doh!

test/sparse.cpp
umar456: The tests are failing. You called it.

src/backend/SparseArray.cpp
```diff
@@ -140,6 +146,11 @@ SparseArray<T> createArrayDataSparseArray(
 }
 
 template<typename T>
+SparseArray<T> copySparseArray(const SparseArray<T>& other) {
```
mlloreda: Can we think of a better name for the argument, here?

src/backend/SparseArray.cpp
```diff
@@ -196,20 +201,15 @@ SparseArray<T>::SparseArray(af::dim4 _dims,
...
-                      "SparseArray<T>::base must be the first member variable of SparseArray<T>");
-#endif
-}
+    values(_copy ? copyArray<T>(_values): _values) {}
 
 template<typename T>
-SparseArray<T>::~SparseArray()
-{
-}
+SparseArray<T>::SparseArray(const SparseArray<T> &other, bool copy):
```
mlloreda: Same here.

src/backend/SparseArray.cpp
```diff
@@ -230,6 +230,7 @@ SparseArray<T>::~SparseArray()
             const Array<int> &_rowIdx, const Array<int> &_colIdx,                                   \
             const af::storage _storage, const bool _copy);                                          \
     template SparseArray<T> *initSparseArray<T>();                                                  \
+    template SparseArray<T> copySparseArray<T>(const SparseArray<T>& other);                        \
```
mlloreda: And here.

umar456:
Addressed all feedback

[skip+ci]

PR #1830, opened 2017-06-16
Allow af_create_handle to accept nullptr for dims

umar456:


src/api/c/array.cpp
```diff
@@ -94,11 +94,11 @@ af_err af_create_handle(af_array *result, const unsigned ndims, const dim_t * co
                         const af_dtype type)
 {
     try {
-        af_array out;
+        af_array out = 0;
         AF_CHECK(af_init());
 
-        dim4 d((size_t)dims[0]);
```
pavanky: Add an assert before the for loop to check if `dims` is not `NULL` when `ndils > 0`

umar456: done

src/api/c/array.cpp
```diff
@@ -97,7 +97,10 @@ af_err af_create_handle(af_array *result, const unsigned ndims, const dim_t * co
         af_array out = 0;
         AF_CHECK(af_init());
 
-        dim4 d(0, 0, 0, 0);
+        if (ndims > 0) {
+            ARG_ASSERT(2, ndims > 0 && dims != NULL);
+        }
```
pavanky: You don't need the if condition.

PR #1831, opened 2017-06-17
Add matching af_free_host and use vector instead of new in tests

umar456:


src/api/c/memory.cpp
```diff
@@ -241,17 +241,15 @@ af_err af_free_pinned(void *ptr)
...
+    }
+    return AF_ERR_NO_MEM;
 }
 
 af_err af_free_host(void *ptr)
 {
-    try {
-        free(ptr);
-    } CATCHALL;
+    free(ptr);
```
pavanky: Can you wrap it under `if (ptr)`

PR #1832, opened 2017-06-18
Allow creation of empty sparse arrays. Allow sparse deep copies

umar456:


pavanky:
This is fine but can you change the PR name ? The changes only affect copy as far as I can see. Allowing creation of sparse empty arrays seems to be already supported.

PR #1839, opened 2017-06-20
Fix variable name(IN) conflicts with macro defined in Windows

umar456:
The IN name is defined in the minwindef.h header. Transform also defines
an IN variable which conflicts with this macro and causing compilation
errors.


9prady9:
@umar456 I looked at the Windows-CUDA build job log but couldn't figure why the job final state says it is failure.

Something is wrong with Windows-02 node which is running Windows-OpenCL job, there aren't any OpenCL devices as per the log file errors.

pavanky:
Can you change the name of the PR

PR #1840, opened 2017-06-20
Add driver locks around cudaMalloc and cudaFree

umar456:


umar456:
@pavanky Doh! Addressed all feedback.

PR #1847, opened 2017-06-21
Quick Fix for issues arising with step size in sequences > 1

### LINKED ISSUE #1845: Bug in indexing after reorder
The following code will cause the opencl backend to crash:
```c++
#include <iostream>
#include <arrayfire.h>

using namespace std;
using namespace af;
int main()
{
    setBackend(AF_BACKEND_OPENCL);

    array a = array(1,512,1);
    array b = reorder(a,0,2,1);
    // array c = reorder(b(0,0,0),2,1,0);
    array d = reorder(b(0,0,span),2,1,0);
}

```
After this, all future calls to reorder will fail.

512 is the smallest number for which this occurs.
The bug only occurs when using span on the large dimension of b. 

However, the if you access some other parts of b first, the code will not crash. For example, if you uncomment the commented line, this code will run as expected. 

The reported error is:

```
OpenCL Error (-36): Invalid Command Queue when calling clFinish
In function af::array af::reorder(const af::array&, unsigned int, unsigned int, unsigned int, unsigned int)
In file src/api/cpp/data.cpp:253
Aborted

```

### LINKED ISSUE #1846: Bug in Indexing an already indexed array
The following code should print all 1s, but it doesn't.

```
#include <iostream>
#include <arrayfire.h>

using namespace std;
using namespace af;
int main()
{
    af::info();
    af::array a = af::randu(3, 12);
    af::array b = a(span, af::seq(0, af::end, 2));
    af::array c = b(span, af::seq(0, af::end, 3));
    af::array d = a(span, af::seq(0, af::end, 6));
    af_print(c == d);
    return 1;
}
```

Related to #1845

pavanky:
- Added relevant tests
- Updated release notes

- Fixes #1846 
- Fixes #1845

### DIFF
--- a/src/backend/opencl/Array.cpp
+++ b/src/backend/opencl/Array.cpp
@@ -268,6 +268,14 @@ namespace opencl
         parent.eval();
 
         dim4 dDims = parent.getDataDims();
+        dim4 dStrides = calcStrides(dDims);
+        dim4 parent_strides = parent.strides();
+
+        if (dStrides != parent_strides) {
+            const Array<T> parentCopy = copyArray(parent);
+            return createSubArray(parentCopy, index, copy);
+        }
+
         dim4 pDims = parent.dims();
 
         dim4 dims    = toDims  (index, pDims);
@@ -275,7 +283,6 @@ namespace opencl
 
         // Find total offsets after indexing
         dim4 offsets = toOffset(index, pDims);
-        dim4 parent_strides = parent.strides();
         dim_t offset = parent.getOffset();
         for (int i = 0; i < 4; i++) offset += offsets[i] * parent_strides[i];
 
--- a/test/index.cpp
+++ b/test/index.cpp
@@ -1536,3 +1536,22 @@ TEST(Index, ISSUE_1101_MODDIMS)
         ASSERT_EQ(ha[i + st], hc[i]);
     }
 }
+
+TEST(Index, ISSUE_1846_Index_Step_Cascade)
+{
+    using namespace af;
+    array a = randu(3, 12);
+    array b = a(span, seq(0, af::end, 2));
+    array c = b(span, seq(0, af::end, 3));
+    array d = a(span, seq(0, af::end, 6));
+    EXPECT_EQ(allTrue<bool>(c == d), true);
+}
+
+TEST(Index, ISSUE_1845_Index_Step_reorder)
+{
+    using namespace af;
+    array a = randu(1,8,1);
+    array b = reorder(a,0,2,1);
+    array d = reorder(b(0,0,span),2,1,0);
+    EXPECT_EQ(allTrue<bool>(a.T() == d), true);
+}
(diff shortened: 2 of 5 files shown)


test/index.cpp
```diff
@@ -1536,3 +1536,22 @@ TEST(Index, ISSUE_1101_MODDIMS)
...
+{
+    using namespace af;
+    af::array a = af::randu(3, 12);
+    af::array b = a(span, af::seq(0, af::end, 2));
+    af::array c = b(span, af::seq(0, af::end, 3));
+    af::array d = a(span, af::seq(0, af::end, 6));
+    ASSERT_EQ(af::allTrue<bool>(c == d), true);
+}
+
+TEST(Index, ISSUE_184_Index_Step_reorder)
```
umar456: ISSUE_1845_Index_Step_reorder

src/backend/opencl/Array.cpp
```diff
@@ -268,14 +268,21 @@ namespace opencl
         parent.eval();
 
         dim4 dDims = parent.getDataDims();
+        dim4 dStrides = calcStrides(dDims);
+        dim4 parent_strides = parent.strides();
+
+        if (dStrides != parent_strides) {
+            const Array<T> &parentCopy = copyArray(parent);
```
umar456: copyArray returns an Array<T> by value. I would just make this a value Array<T> rather than a reference array. The createSubArray function should increase the reference count anyway. It would be a good idea to check.

test/index.cpp
```diff
@@ -1536,3 +1536,22 @@ TEST(Index, ISSUE_1101_MODDIMS)
...
 }
+
+TEST(Index, ISSUE_1846_Index_Step_Cascade)
+{
+    using namespace af;
+    af::array a = af::randu(3, 12);
+    af::array b = a(span, af::seq(0, af::end, 2));
+    af::array c = b(span, af::seq(0, af::end, 3));
+    af::array d = a(span, af::seq(0, af::end, 6));
+    ASSERT_EQ(af::allTrue<bool>(c == d), true);
```
umar456: `EXPECT_TRUE(af::allTrue<bool>(c == d));`

test/index.cpp
```diff
@@ -1536,3 +1536,22 @@ TEST(Index, ISSUE_1101_MODDIMS)
...
+}
+
+TEST(Index, ISSUE_184_Index_Step_reorder)
+{
+    using namespace af;
+    array a = af::randu(1,8,1);
+    array b = reorder(a,0,2,1);
+    array d = reorder(b(0,0,span),2,1,0);
+    ASSERT_EQ(af::allTrue<bool>(a.T() == d), true);
```
umar456: `EXPECT_TRUE(af::allTrue<bool>(a.T() == d));`

test/index.cpp
```diff
@@ -1536,3 +1536,22 @@ TEST(Index, ISSUE_1101_MODDIMS)
         ASSERT_EQ(ha[i + st], hc[i]);
     }
 }
+
+TEST(Index, ISSUE_1846_Index_Step_Cascade)
+{
+    using namespace af;
+    af::array a = af::randu(3, 12);
+    af::array b = a(span, af::seq(0, af::end, 2));
```
umar456: af:: is unnecessary.

PR #1853, opened 2017-06-23
Fix max allowable window size in af_unwrap

### LINKED ISSUE #1852: af_unwrap contains incorrect assertions
af_unwrap() in src/api/c/unwrap.cpp contains wrong assertions on lines 37 and 38. The maximum size for a window dimension should be dim+2*padding as the padding is applied on both sides of the dimension. 

For example, to following code should be valid but fails the assertion:

```
#include <arrayfire.h>

using namespace af;
int main(){
    array a = constant(1,2,2);
    array b = unwrap(a, 4, 2, 4, 2, 1, 0);
}
```

plavin:
As padding is added a both sides of a dimension, the max allowable window size
should be dim_size + 2 * padding

 fixes #1852


pavanky:
build arrayfire ci

PR #1857, opened 2017-06-24
Remove boost dependency from CPU & CUDA backends

### LINKED ISSUE #1855: Remove boost as a REQUIRED dependency
It is no longer needed for CPU and CUDA backends. For the OpenCL backend only boost compute is used and it is built separately anyway.

9prady9:
Fixes #1855


src/backend/opencl/CMakeLists.txt
```diff
@@ -108,12 +108,26 @@ INCLUDE_DIRECTORIES(${CLFFT_INCLUDE_DIRS})
...
+    FIND_PACKAGE(Boost 1.61.0)
+    IF(Boost_FOUND)
+        MESSAGE(STATUS "Found Boost ${Boost_VERSION}")
+        SET(Compute_INCLUDE_DIRS ${Boost_INCLUDE_DIRS})
+    ELSE()
+        MESSAGE(STATUS "Boost version with Compute headers not found.")
+        MESSAGE(STATUS "Looking for BoostCompute package directly ...")
         FIND_PACKAGE(BoostCompute REQUIRED)
+        SET(Compute_INCLUDE_DIRS ${BoostCompute_INCLUDE_DIRS})
+        MESSAGE(STATUS "Found BoostCompute.")
```
pavanky: I feel like there are a bit too many status messages to see what is actually happening in the CMake code.

9prady9:
@pavanky Any idea why `Windows-OpenCL` job failed with `boost/config.hpp` not found. Does `Boost.Compute` itself use `Boost` anywhere internally ?

pavanky:
@9prady9 no idea.

9prady9:

9prady9:
`config.hpp` header is from `Boost` and `BoostCompute` is using it. It includes other platform and compiler specific checks and more header inclusions associated to that.

9prady9:
Please don't merge yet. All though ci builds worked fine, my local machine is having an issue. Looking into it.

src/backend/opencl/CMakeLists.txt
```diff
@@ -108,12 +108,22 @@ INCLUDE_DIRECTORIES(${CLFFT_INCLUDE_DIRS})
 LINK_DIRECTORIES(${CLFFT_LIBRARY_DIR})
 
 OPTION(USE_SYSTEM_BOOST_COMPUTE "Use system BoostCompute" OFF)
-IF(USE_SYSTEM_BOOST_COMPUTE)
-    IF(Boost_VERSION VERSION_LESS "1.61")
+
+ADD_DEFINITIONS(-DBOOST_ALL_NO_LIB)
+SET(Boost_USE_STATIC_LIBS OFF)
+FIND_PACKAGE(Boost REQUIRED)
+
```
pavanky: We are not using anything else other than Boost.Compute in opencl backend. Is there any point i making this REQUIRED if we are not asking for 1.61?

9prady9: Boost.Compute is using Boost's platform specific headers(`config.hpp` which is not part of Compute), so we can't get rid of it whether the version of Boost is 1.61 or lesser.

pavanky: Alright that makes sense 👍

pavanky:
@umar456 this can be merged.

umar456:
Can't we just do this as part of #1861 ?

9prady9:
@umar456 If we do that this dependency will still be a requirement for other backends until we do a 3.6 which is not soon.

umar456:
#1861 is not changing API. It can go in a patch release.

9prady9:
@umar456 You were the one who suggested it should go into devel. Hence my earlier comment.

pavanky:
@umar456 I would rather have #1861 be 3.6. Even if it does not change the API, we are changing the (cmake) dependency requirement by several versions.

PR #1871, opened 2017-07-14
Change Array<T> variables to Param<T> on the queue for CPU backend

pavanky:
This change allows us to efficiently reuse memory allocated on the same queue for cpu backend

### DIFF
--- a/src/backend/cpu/Param.hpp
+++ b/src/backend/cpu/Param.hpp
@@ -1,5 +1,5 @@
 /*******************************************************
- * Copyright (c) 2014, ArrayFire
+ * Copyright (c) 2017, ArrayFire
  * All rights reserved.
  *
  * This file is distributed under 3-clause BSD license.
@@ -74,5 +74,27 @@ public:
         return CParam<T>(const_cast<T *>(ptr), dims, strides);
     }
 };
+
+template<typename T> class Array;
+
+template<typename T>
+T toParam(const T &val)
+{
+    return val;
+}
+
+
+template<typename T>
+Param<T> toParam(Array<T> &val)
+{
+    return (Param<T>)(val);
+}
+
+
+template<typename T>
+CParam<T> toParam(const Array<T> &val)
+{
+    return (CParam<T>)(val);
+}
 
 }
--- a/src/backend/cpu/Param.hpp
+++ b/src/backend/cpu/Param.hpp
@@ -116,19 +116,20 @@ public:
 
 template<typename T> class Array;
 
+// These functions are needed to convert Array<T> to Param<T> when queueing up functions.
+// This is necessary because the memory used by Array<T> can be put back into the queue faster.
+// This is fine becacuse we only have 1 compute queue. This ensures there's no race conditions.
 template<typename T>
 T toParam(const T &val)
 {
     return val;
 }
 
-
 template<typename T>
 Param<T> toParam(Array<T> &val)
 {
     return (Param<T>)(val);
 }
-
 
 template<typename T>
 CParam<T> toParam(const Array<T> &val)
--- a/src/backend/cpu/Param.hpp
+++ b/src/backend/cpu/Param.hpp
@@ -1,0 +1,78 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#pragma once
+#include <af/defines.h>
+#include <af/dim4.hpp>
+#include <backend.hpp>
+
+namespace cpu
+{
+
+using af::dim4;
+
+template<typename T>
+class CParam
+{
+private:
+    const T *ptr;
+
+public:
+    dim4 dims;
+    dim4 strides;
+
+    CParam(const T *iptr, const dim4 &idims, const dim4 &istrides) :
+        ptr(iptr)
+    {
+        for (int i = 0; i < 4; i++) {
+            dims[i] = idims[i];
+            strides[i] = istrides[i];
+        }
+    }
+    const T *get() const
+    {
+        return ptr;
+    }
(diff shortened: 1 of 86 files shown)


pavanky:
approx2 test is failing, I'll be investigating that.

pavanky:
All tests pass now. @umar456 can you review please.

pavanky:
God damn it the threading tests are failing now.

pavanky:
OK they should pass the tests now..

src/backend/cpu/queue.hpp
```diff
@@ -48,6 +48,30 @@ typedef async_queue queue_impl;
 
 namespace cpu {
 
+template<typename T> class Array;
+template<typename T> class Param;
+template<typename T> class CParam;
+
+template<typename T>
+T toParam(const T &val)
```
umar456: Shouldn't these be defined in Param.hpp?

src/backend/cpu/kernel/sift_nonfree.hpp
```diff
@@ -177,12 +177,12 @@ void gaussian1D(T* out, const int dim, double sigma=0.0)
...
 
 template<typename T>
-Array<T> gauss_filter(float sigma)
+Param<T> gauss_filter(float sigma)
 {
     // Using 6-sigma rule
     unsigned gauss_len = std::min((unsigned)round(sigma * 6 + 1) | 1, 31u);
 
-    Array<T> filter = createEmptyArray<T>(gauss_len);
+    Param<T> filter = createEmptyParam<T>(gauss_len);
```
umar456: Params should never be created without associated Array<T> objects. They are only used to represent a full array in a more compact structure. They do not manage memory and this is very unsafe. We shouldn't even have the createEmptyParam functions.

src/backend/cpu/queue.hpp
pavanky: Sure.

src/backend/cpu/Param.hpp
```diff
@@ -0,0 +1,78 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
pavanky: Will fix.

src/backend/cpu/kernel/sift_nonfree.hpp
pavanky: Yeah this was a mistake with sed. This file isnt even using any queue stuff. I'll just probably move the function into src/backend/cpu

src/backend/cpu/cholesky.cpp
```diff
@@ -70,13 +71,7 @@ int cholesky_inplace(Array<T> &in, const bool is_upper)
...
         uplo = 'U';
 
-    int info = 0;
-    auto func = [&] (int& info, Array<T>& in) {
-        info = potrf_func<T>()(AF_LAPACK_COL_MAJOR, uplo, N, in.get(), in.strides()[1]);
-    };
-
-    getQueue().enqueue(func, info, in);
-    getQueue().sync();
+    int info = potrf_func<T>()(AF_LAPACK_COL_MAJOR, uplo, N, in.get(), in.strides()[1]);
```
umar456: This work will be performed on the calling thread and all other functions will be called on the queue thread. While this is a synchronous function, it might be beneficial to keep all work on the queue thread for consistency.

src/backend/cpu/Param.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+template<typename T>
+class CParam
+{
+private:
+    const T *ptr;
+
+public:
+    dim4 dims;
+    dim4 strides;
```
umar456: I would like these to be private or at the very least const. There is no reason for the dims and the strides to change once the [C]Param object is created.

src/backend/cpu/Param.hpp
```diff
@@ -0,0 +1,100 @@
...
+};
+
+template<typename T>
+class Param
+{
+private:
+    T *ptr;
+public:
+    dim4 dims;
+    dim4 strides;
```
umar456: Private or const. Same argument as above.

src/backend/cpu/Param.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    operator CParam<T>() const
+    {
+        return CParam<T>(const_cast<T *>(ptr), dims, strides);
+    }
+};
+
+template<typename T> class Array;
+
+template<typename T>
```
umar456: Can you comment on why this function specialization is necessary?

src/backend/cpu/Param.hpp
```diff
@@ -0,0 +1,100 @@
...
+using af::dim4;
+
+template<typename T>
+class CParam
+{
+private:
+    const T *ptr;
+
+public:
+    dim4 dims;
```
umar456: Perhaps also add data_dims. Seems to be necessary for some functions.

src/backend/cpu/cholesky.cpp
pavanky: This is simpler than dealing with references in `toParam`. Besides calling this in a queue and then calling sync seems unnecessary.

src/backend/cpu/Param.hpp
pavanky: data_dims shouldn't be needed inside kernel.

src/backend/cpu/cholesky.cpp
umar456: I don't understand what you mean by references. I thought Param were not reference counted so none of that should be going on. I just think its better to be consistant in case there is an issue with thread_local stuff in the future(although its probably unlikely).

pavanky: Sorry, I should have clarified better. The `info` needs be assigned a value from inside the queue.

umar456: That's why the lambda was being used(although from the looks of it, slightly ambiguously).

pavanky:
@umar456 squash this when merging.

pavanky:
@umar456 any idea why the non free tests may be failing ?

PR #1872, opened 2017-07-17
Pthread fix

### LINKED ISSUE #1856: Missing symbol when building on gcc 7.1
Command used to reproduce:

```
mkdir build && cd build
cmake .. -DBUILD_CUDA=OFF -DBUILD_OPENCL=OFF -DBUILD_GRAPHICS=OFF && make example_helloworld_cpu
```

Error message:

```
[100%] Linking CXX executable helloworld/helloworld_cpu
../src/backend/cpu/libafcpu.so.3.5.0: undefined reference to `pthread_create'
collect2: error: ld returned 1 exit status
```

umar456:
Fixes #1856


pavanky:
Shouldn't the CUDA backend also be linked with the threading library ? We can not compile CUDA with gcc 7 at the moment, but it would be nice if we can fix it before it breaks in the future.

umar456:
I thought it was only an issue when linking with lapacke. That library is only required for the CPU and OpenCL backends.

pavanky:
@umar456 it is only manifesting when linking with lapacke that itself wasnt linked against pthread causing a missing symbol error.

pthread_create is still found in libafcuda.so

```
$ nm -C ~/Workspace/local/lib/libafcuda.so.3.5.0  | grep pthread_create
                 U pthread_create@@GLIBC_2.2.5
```

Issue #1867, opened 2017-07-11
Memory leak using rows and cols


Thormidable:
Have produced a memory leak when using rows and cols.
I have a temporary work around, but thought you would be interested in it.

Attached is a Minimum Working Example

[AF Minimal working Example memory leak.zip](https://github.com/arrayfire/arrayfire/files/1139608/AF.Minimal.working.Example.memory.leak.zip)

FloopCZ:
I am going to paste your code in here, so that others do not have to extract the VC++ project. The code indeed seems to eat up an increasing amount of memory (tested on build `ArrayFire v3.5.0 (CPU, 64-bit Linux, build 05999f3)`)

```c++
#include <arrayfire.h>
using namespace af;

void NoLeak(af::array lImageI)
{
    af::array lRows = lImageI.rows(0, 1023);
    af::array Q3 = lRows.cols(0, 1023);
    Q3.eval();

    af::sync();
}

void CausesLeak(af::array lImageI)
{
    af::array Q3 = lImageI.rows(0, 1023).cols(0, 1023);
    Q3.eval();

    af::sync();
}

void GenerateLeak()
{
    af::array lInput = af::randn(4000, 4000, f32);
    //NoLeak(lInput);
    CausesLeak(lInput);
}

int main()
{
    for (long int lCounter = 0; lCounter < 10000; ++lCounter)
    {
        GenerateLeak();
        af::deviceGC();
    }
}
```

pavanky:
@FloopCZ Thanks for this!

pavanky:
@umar456 replacing `.rows().cols()` with `(af::seq(), af::seq())` doesn't cause the leak. This makes me think something's up with proxy ? We may be basing it on some assumption that we broke in src/api/c..

Here's a slightly modified test to see the memory leak immediately.

```
using namespace af;

void NoLeak(af::array lImageI)
{
    af::array lRows = lImageI.rows(0, 1023);
    af::array Q3 = lRows.cols(0, 1023);
    Q3.eval();
    af::sync();
}

void NoLeak2(af::array lImageI)
{
    af::array Q3 = lImageI(af::seq(0, 1023), af::seq(0, 1023));
    Q3.eval();
    af::sync();
}

void CausesLeak(af::array lImageI)
{
    af::array Q3 = lImageI.rows(0, 1023).cols(0, 1023);
    Q3.eval();
    af::sync();
}

void GenerateLeak()
{
[...]

PR #1879, opened 2017-07-20
Fix leak when chaining indexing operations

umar456:
Indexing operations leaked when chained(i.e. arr.rows(10, 20).cols(1, 4).
This was happening because the array_proxy object's member functions created
an array pointer when indexing operations were performed. This array was
not freed when the indexing operation was evaluated on conversion back to
af::array.

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -22,9 +22,6 @@
 #include <stdexcept>
 #include <memory>
 
-
-using std::unique_ptr;
-
 namespace af
 {
     static int gforDim(af_index_t *indices)
@@ -96,6 +93,9 @@ namespace af
         array *       parent_;        //< The original array
         af_index_t    indices_[4];    //< Indexing array or seq objects
         bool          is_linear_;
+
+        // if true the parent_ object will be deleted on distruction. This is
+        // necessary only when calling indexing functions in array_proxy objects.
         bool          delete_on_destruction_;
         array_proxy_impl(array &parent, af_index_t *idx, bool linear)
             : parent_(&parent)
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -6,7 +6,6 @@
  * The complete license agreement can be obtained at:
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
-#include <stdexcept>
 #include <af/array.h>
 #include <af/arith.h>
 #include <af/blas.h>
@@ -19,6 +18,12 @@
 #include <af/algorithm.h>
 #include <af/internal.h>
 #include "error.hpp"
+
+#include <stdexcept>
+#include <memory>
+
+
+using std::unique_ptr;
 
 namespace af
 {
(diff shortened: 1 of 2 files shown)


umar456:
Fixes #1867

src/api/cpp/array.cpp
```diff
@@ -20,6 +19,12 @@
 #include <af/internal.h>
 #include "error.hpp"
 
+#include <stdexcept>
+#include <memory>
+
+
+using std::unique_ptr;
```
pavanky: Where are you using this?

src/api/cpp/array.cpp
```diff
@@ -88,16 +93,31 @@ namespace af
 
     struct array::array_proxy::array_proxy_impl
     {
-        array       *parent;        // The original array
-        af_index_t  indices[4];     // Indexing array or seq objects
-        bool        lin;
+        array *       parent_;        //< The original array
+        af_index_t    indices_[4];    //< Indexing array or seq objects
+        bool          is_linear_;
+        bool          delete_on_destruction_;
```
pavanky: Add a comment about why this is needed now.

src/api/cpp/array.cpp
```diff
@@ -88,16 +93,31 @@ namespace af
...
+        }
+
+        ~array_proxy_impl() {
+            if (delete_on_destruction_) delete parent_;
         }
+      private:
+        array_proxy_impl(const array_proxy_impl&);
+        array_proxy_impl(const array_proxy_impl&&);
+        array_proxy_impl operator=(const array_proxy_impl&);
+        array_proxy_impl operator=(const array_proxy_impl&&);
```
pavanky: Where are they implemented?

test/index.cpp
```diff
@@ -1547,11 +1547,26 @@ TEST(Index, ISSUE_1846_Index_Step_Cascade)
...
+    {
+        array lInput = randn(100, 100, f32);
+        array Q3 = lInput.rows(0, 3).cols(0, 3);
+        Q3.eval();
+        af::sync();
+    }
+    size_t alloc_bytes, alloc_buffers, lock_bytes, lock_buffers;
+    af::deviceMemInfo(&alloc_bytes, &alloc_buffers,
+                      &lock_bytes, &lock_buffers);
+    ASSERT_EQ(0u, lock_buffers);
```
pavanky: Was this test failing before the fix? The issue had a particular sequence of events when it was failing.

umar456: This issue exposed a leak in the implementation. It was failing earlier because the chained indexing was causing leaks so there was a locked array in the memory manager after the code block.

src/api/cpp/array.cpp
umar456: They are private. They shouldn't be used.

src/api/cpp/array.cpp
umar456: I originally implemented array_proxy_impl using unique ptr but it became more complicated than it needed to be. Removing.

umar456:
@pavanky Addressed all feedback.

test/index.cpp
pavanky: @umar456 I meant "make sure this is a valid test for the fix" because the code looks different from the code in the bug report.

umar456: How so? I am just using smaller arrays. The issue is with the chaining of the `arr.rows(x, x).cols(x, x)` call.

pavanky: Nevermind, I thought it was failing in a very particular setup. I made the following comment in the issue.

> Also to consider, the leak doesnt happen if the array is in the main loop and you call CausesLeak from the main loop. I am not sure why this is happening.

Tested this again and it doesn't seem to be the case. It is just as reproducible using the test you have here. :+1:

WilliamTambellini:
Should nt we backport this one to AF 3.4 ?

PR #1891, opened 2017-08-01
Fix pixel tests in fast kernels

### LINKED ISSUE #1842: bug in fast docs / possible bug in implementation
https://github.com/arrayfire/arrayfire/blob/2528d711f371b18c0f51d5f545737d5c62ea284f/src/backend/opencl/kernel/fast.cl#L29-L41

- the documentation does not agree with the comments
- the implementation does not agree with the function name
- it likely should be <,>, instead of <=,>=.

I was working on a fast implementation outside of this project and noticed this. Thought I'd drop a quick line.

9prady9:
Fixes #1842


9prady9:
build arrayfire ci

PR #1899, opened 2017-08-04
Change static objects into pointers to fix release issues on Windows

umar456:
Windows terminates threads before the queue threads and other resources are
released. This causes deadlocks with the condition_variables in the async_queue
objects. This is a bug in Visual Studio/Windows that is documented here:

https://connect.microsoft.com/VisualStudio/feedback/details/747145

This will leak some resources but these resources will be released by the
operating system on exit.

### DIFF
--- a/src/backend/common/MemoryManager.hpp
+++ b/src/backend/common/MemoryManager.hpp
@@ -14,6 +14,8 @@
 #include <util.hpp>
 
 #include <algorithm>
+
+// TODO(umar): Remove iostream
 #include <iomanip>
 #include <iostream>
 #include <mutex>
@@ -298,9 +300,9 @@ class MemoryManager
 
         std::cout << msg << std::endl;
 
-        static const std::string head("|     POINTER      |    SIZE    |  AF LOCK  | USER LOCK |");
-        static const std::string line(head.size(), '-');
-        std::cout << line << std::endl << head << std::endl << line << std::endl;
+        printf("---------------------------------------------------------\n"
+               "|     POINTER      |    SIZE    |  AF LOCK  | USER LOCK |\n"
+               "---------------------------------------------------------\n");
 
         for(auto& kv : current.locked_map) {
             std::string status_mngr("Yes");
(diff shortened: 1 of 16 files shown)


src/backend/common/MemoryManager.hpp
```diff
@@ -14,6 +14,8 @@
 #include <util.hpp>
 
 #include <algorithm>
+
+// TODO(umar): Remove iostream
 #include <iomanip>
 #include <iostream>
```
9prady9: @umar456 What are your concerns about `iostream` header ?

9prady9: @umar456 What are your concerns about `iostream` header ?

src/backend/common/MemoryManager.hpp
```diff
@@ -298,9 +300,9 @@ class MemoryManager
 
         std::cout << msg << std::endl;
 
-        static const std::string head("|     POINTER      |    SIZE    |  AF LOCK  | USER LOCK |");
-        static const std::string line(head.size(), '-');
-        std::cout << line << std::endl << head << std::endl << line << std::endl;
+        printf("---------------------------------------------------------\n"
+               "|     POINTER      |    SIZE    |  AF LOCK  | USER LOCK |\n"
+               "---------------------------------------------------------\n");
```
pavanky: Can you store this as a string literal and then print it ?

src/backend/common/MemoryManager.hpp
umar456: `iostream` is a pretty big header.  Since this header is going to be included nearly everywhere it is important we keep this header as small as possible.

9prady9: isn't that only a cause of concern if we are doing `using namespace std` inside this header(which we are not right now) ?

pavanky: @9prady9 just including templated headers adds a lot of compilation overhead.

PR #1903, opened 2017-08-09
Re-enable cuda texture object based access in regions kernel

9prady9:
There is no improvement except for some specific sizes. Note that I have run the benchmark on only 2^n sizes.

|Size | TextureObject | Plain Old Memory |
|---|----|---|
128 x  128 | 0.00247927 | 0.00299974
256 x  256 | 0.00671363 | 0.006602
384 x  384 | 0.00620648 | 0.00795144
512 x  512 | 0.00989444 | 0.0123784
640 x  640 | 0.0136365 | 0.0155043
768 x  768 | 0.0186786 | 0.0196094
896 x  896 | 0.0253983 | 0.0266321
1024 x 1024 | 0.0322667 | 0.0337688
1152 x 1152 | 0.0411638 | 0.0431613
1280 x 1280 | 0.0529367 | 0.0560853
1408 x 1408 | 0.0612112 | 0.0643147
1536 x 1536 | 0.0872647 | 0.0769113
1664 x 1664 | 0.105343 | 0.0999272
1792 x 1792 | 0.12006 | 0.112325
1920 x 1920 | 0.112999 | 0.118628
2048 x 2048 | 0.129985 | 0.136978

chart

[skip arrayfire ci]

### DIFF
--- a/src/backend/cuda/kernel/regions.hpp
+++ b/src/backend/cuda/kernel/regions.hpp
@@ -23,8 +23,6 @@
 #include <thrust/sort.h>
 #include <thrust/transform_scan.h>
 
-#if __CUDACC__
-
 static const int THREADS_X = 16;
 static const int THREADS_Y = 16;
 
@@ -35,19 +33,18 @@ __device__ static int continue_flag = 1;
 
 // Wrapper function for texture fetch
 template<typename T>
-__device__ __inline__
-static T fetch(const int n,
-               cuda::Param<T> equiv_map,
-               cudaTextureObject_t tex)
+static inline __device__
+T fetch(const int n, cuda::Param<T> equiv_map, cudaTextureObject_t tex)
 {
-// FIXME: Enable capability >= 3.0
-//#if (__CUDA_ARCH__ >= 300)
-#if 0
-    // Kepler bindless texture objects
     return tex1Dfetch<T>(tex, n);
-#else
+}
+
+template<> __device__
+STATIC_ double fetch<double>(const int n,
+                             cuda::Param<double> equiv_map,
+                             cudaTextureObject_t tex)
+{
     return equiv_map.ptr[n];
-#endif
 }
 
 // The initial label kernel distinguishes between valid (nonzero)
@@ -128,7 +125,7 @@ static void update_equiv(cuda::Param<T> equiv_map, const cudaTextureObject_t tex
 {
 
     typedef warp_count<block_dim*block_dim> num_warps;
-#if (__CUDA_ARCH__ >= 120) // This function uses warp ballot instructions
+
     // Basic coordinates
     const int base_x = (blockIdx.x * blockDim.x * n_per_thread) + threadIdx.x;
     const int base_y = (blockIdx.y * blockDim.y * n_per_thread) + threadIdx.y;
@@ -160,13 +157,9 @@ static void update_equiv(cuda::Param<T> equiv_map, const cudaTextureObject_t tex
     s_changed[warpIdx] = (T)0;
     __syncthreads();
 
-#if (__CUDA_ARCH__ >= 130)
     #pragma unroll
-#endif
     for (int xb = 0; xb < n_per_thread; ++xb) {
-#if (__CUDA_ARCH__ >= 130)
         #pragma unroll
-#endif
         for (int yb = 0; yb < n_per_thread; ++yb) {
 
             // Indexing variables
@@ -251,9 +244,7 @@ static void update_equiv(cuda::Param<T> equiv_map, const cudaTextureObject_t tex
     s_changed[warpIdx] = __any((int)tid_changed);
     __syncthreads();
 
-#if (__CUDA_ARCH__ >= 130)
     #pragma unroll
-#endif
     for (int i = 0; i < num_warps::value; i++)
         continue_iter = continue_iter || (s_changed[i] != 0);
 
@@ -263,13 +254,9 @@ static void update_equiv(cuda::Param<T> equiv_map, const cudaTextureObject_t tex
         // Reset whether or not this thread's pixels have changed.
         tid_changed = false;
 
-#if (__CUDA_ARCH__ >= 130)
         #pragma unroll
-#endif
         for (int xb = 0; xb < n_per_thread; ++xb) {
-#if (__CUDA_ARCH__ >= 130)
             #pragma unroll
-#endif
             for (int yb = 0; yb < n_per_thread; ++yb) {
 
                 // Indexing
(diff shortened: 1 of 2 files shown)


src/backend/cuda/kernel/regions.hpp
```diff
@@ -35,21 +36,24 @@ __device__ static int continue_flag = 1;
...
-               cuda::Param<T> equiv_map,
-               cudaTextureObject_t tex)
+static inline __device__
+T fetch(const int n, cuda::Param<T> equiv_map, cudaTextureObject_t tex)
 {
-// FIXME: Enable capability >= 3.0
-//#if (__CUDA_ARCH__ >= 300)
-#if 0
-    // Kepler bindless texture objects
+#if __CUDA_ARCH__ >= 300
```
pavanky: Can you actually drop this ? We don't support pre 3.0 compute anymore.

src/backend/cuda/regions.cu
```diff
@@ -29,20 +29,30 @@ Array<T>  regions(const Array<char> &in, af_connectivity connectivity)
...
-    //    cudaTextureDesc texDesc;
-    //    memset(&texDesc, 0, sizeof(texDesc));
-    //    texDesc.readMode = cudaReadModeElementType;
-    //    CUDA_CHECK(cudaCreateTextureObject(&tex, &resDesc, &texDesc, NULL));
-    //}
+
+    cudaDeviceProp prop = getDeviceProp(getActiveDeviceId());
+
+    //Use texture objects with compute 3.0 or higher
+    if (prop.major>=3 && prop.minor>=0 && !std::is_same<T,double>::value) {
```
pavanky: Same as above.

9prady9:
@arrayfire/core-devel This is ready for merge.

examples/benchmarks/regions.cpp
```diff
@@ -0,0 +1,47 @@
...
+            printf("Avg. Time Taken:  %g \n", timeit(benchmarkRegions));
+            fflush(stdout);
+        }
+    } catch (af::exception& e) {
+        fprintf(stderr, "%s\n", e.what());
+        throw;
+    }
+
+    return 0;
+}
```
pavanky: Do we need to check this in?

src/backend/cuda/kernel/regions.hpp
```diff
@@ -23,8 +23,9 @@
 #include <thrust/sort.h>
 #include <thrust/transform_scan.h>
 
-#if __CUDACC__
+#include <type_traits>
```
pavanky: Where is this used ?

src/backend/cuda/kernel/regions.hpp
```diff
@@ -23,8 +23,9 @@
...
 
+#if __CUDACC__
```
pavanky: Is this necessary?

src/backend/cuda/regions.cu
```diff
@@ -29,20 +29,28 @@ Array<T>  regions(const Array<char> &in, af_connectivity connectivity)
...
+            resDesc.res.linear.desc.f = cudaChannelFormatKindUnsigned;
+        else
+            resDesc.res.linear.desc.f = cudaChannelFormatKindFloat;
+
+        resDesc.res.linear.desc.x = sizeof(T)*8; // bits per channel
+        resDesc.res.linear.sizeInBytes = dims[0] * dims[1] * sizeof(T);
+        cudaTextureDesc texDesc;
+        memset(&texDesc, 0, sizeof(texDesc));
+        texDesc.readMode = cudaReadModeElementType;
+        CUDA_CHECK(cudaCreateTextureObject(&tex, &resDesc, &texDesc, NULL));
```
pavanky: You are not releasing the texture.

src/backend/cuda/kernel/regions.hpp
9prady9: Not anymore, it was used in first commit but I removed it now. Will remove it.

src/backend/cuda/kernel/regions.hpp
9prady9: Not really. It wasn't added in this change, so I didn't notice this.

9prady9:
@pavanky The the performance looks good. 
chart 1

9prady9:
@pavanky If there is no additional feedback, I think this is ready for merge.

pavanky:
@9prady9 can you close this PR and create a new one targeting v3.5 by cherry picking the appropriate commits?

9prady9:
@umar456 / @mlloreda  please kill the ci jobs for this task.

pavanky:
@9prady9 Nevermind, the problem is with the base branch. I am fixing v3.5

pavanky:
@9prady9 fixed now.

umar456:
The performance of this function depends on the input image. What input image did you use for this function? Can you test the performance where more iterations are required? How does it compare to the performance where there are only few iterations?

Also, I am having trouble seeing how this access pattern takes advantage of the texture objects. Looks like you are loading into the shared memory so you are not taking advantage of the additional texture cache. You are using linear textures so you aren't taking advantage of the spatial locality accesses of Texture caches. I don't see a huge advantage to using this approach. Maybe it will be more pronounced in inputs with more iterations.

pavanky:
Also we should probably be using cudaArray instead of cudaTextureObject.

9prady9:
@umar456 There is an advantage to using texture objects for [this section of code](https://github.com/arrayfire/arrayfire/blob/0db608154d5e5a5ed30ab74d4e6026bf8d12237f/src/backend/cuda/kernel/regions.hpp#L163-L244) because it is doing 2D spatial access(reading more than one element per thread - I am not talking about per thread load) and textures surely have an advantage compared to global memory in that aspect.

Does having the additional layer of shared memory reduce the performance ? It depends on whether shared memory has less latency compared to texture access. I think shared memory latency is better compared to texture access, but I am not certain of it.

I had some benchmark sample using randomized input commited earlier, but removed it later and did a force push. Do you have any suggestion for the input you were talking about - the one that will cause regions to run more iterations ?

@pavanky Is cudaArray similar to texture in performance ? I have seen couple of [example](https://github.com/odlgroup/odl-cpp-utils/blob/master/cuda/texture.h)s online where cudaArray_t is again wrapped around with cudaTextureObject_t. Not sure if it is equivalent in performance with textures.

pavanky:
The benefits coming from texture memory seem to be fairly minimal at the moment.

9prady9:
@pavanky @umar456 I have cleaned up more of it. I have tested the changes with two actual images(using timeit) and the improvement is around 20+%.

One of them has lot of smaller regions(image with lot of text) and the other has huge but fewer regions.
- 21% improvement for image with lot of individual components but smaller regions - took **8 iterations**
- 26% improvement for image with less individual components but big regions - took **22 iterations**.

wtext
sciie

PR #1917, opened 2017-09-04
Fix Meanshift filter kernel implementation in all backends

9prady9:


src/backend/cuda/kernel/meanshift.hpp
```diff
@@ -12,200 +12,152 @@
...
-#include "shared.hpp"
+
+#include <type_traits>
 
 namespace cuda
 {
-
 namespace kernel
 {
+static const int THREADS_X = 32;
```
umar456: did you see a noticeable speedup with these parameters? It might be better to make this change in another PR.

src/backend/cuda/kernel/meanshift.hpp
```diff
@@ -12,200 +12,152 @@
...
-        const int mean_x = (int)(shift_x*fcount+0.5f);
-        const int mean_y = (int)(shift_y*fcount+0.5f);
 #pragma unroll
-        for(int ch=0; ch<channels; ++ch)
-            means[ch] *= fcount;
+        for (int ch=0; ch<channels; ++ch)
+            means[ch] = __float2int_rz(means[ch]*fcount);
 
-        float norm = 0.f;
+        AccType norm = 0;
```
umar456: AccType can only be a floating point type, right? This should be 0.0 or 0.0f like the old code.

src/backend/opencl/kernel/meanshift.cl
```diff
@@ -7,149 +7,115 @@
...
                 for(int wi=-radius; wi<=radius; ++wi) {
 
-                    int tj = j + wj;
                     int ti = i + wi;
 
-                    // proceed
-                    float norm = 0.0f;
+                    if (ti<0 || ti>dim0LenLmt) break;
+
+                    AccType norm = 0;
```
umar456: 0.0f

src/backend/opencl/kernel/meanshift.cl
```diff
@@ -7,149 +7,115 @@
...
-            load2LocalMem(localMem, iptr, a, b, shrdLen, schStride, channels,
-                    d0, d1, gx2-radius, gy2-radius, ichStride, s1, s0);
-        }
-    }
-
-    int i   = lx + radius;
-    int j   = ly + radius;
-
-    barrier(CLK_LOCAL_MEM_FENCE);
+    const int gx = get_local_size(0) * (get_group_id(0)-b2*nBBS0) + get_local_id(0);
```
umar456: What is the point of removing local memory here? Did you benchmark this change?

src/backend/cuda/kernel/meanshift.hpp
```diff
@@ -12,200 +12,152 @@
...
-    for (int b=ly, gy2=gy; b<shrdLen; b+=blockDim.y, gy2+=blockDim.y) {
-        // move row_set get_local_size(1) along coloumns
-        for (int a=lx, gx2=gx; a<shrdLen; a+=blockDim.x, gx2+=blockDim.x) {
-            load2ShrdMem<T, channels>(shrdMem, iptr, a, b, shrdLen, schStride,
-                    in.dims[0], in.dims[1], gx2-radius, gy2-radius, ichStride,
-                    in.strides[1], in.strides[0]);
-        }
-    }
+    int i = gx;
+    int j = gy;
```
umar456: Why is shared mem removed? Do you see better performance accessing global memory directly?

examples/image_processing/image_segmentation.cpp
```diff
@@ -0,0 +1,58 @@
...
+        af::setDevice(device);
+        af::info();
+
+        printf("** ArrayFire Image Segmentation Demo **\n");
+        af::Window myWindow("Image Segmentation Demo");
+
+        isColor = true;
+        array in = loadImage(ASSETS_DIR "/examples/images/atlantis.png", isColor);
+
+        array out = meanShift(in, 15.f, 30.f, 10, isColor);
```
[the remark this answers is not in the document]
9prady9: I have removed it from this PR but I think we need to have an example that show cases different image segmentations possible in ArrayFire - they can be just over-the-top of AF implementations as well.

src/backend/cpu/kernel/meanshift.hpp
```diff
@@ -11,125 +11,133 @@
...
-                                // proceed
-                                float norm = 0.0f;
-                                for(dim_t ch=0; ch<channels; ++ch) {
-                                    tmpclrs[ch] = inData[ tj*istrides[1] + ti*istrides[0] + ch*istrides[2]];
-                                    norm += (centers[ch]-tmpclrs[ch]) * (centers[ch]-tmpclrs[ch]);
+                                std::vector<T> tmpclrs(channels, 0);
+
+                                AccType norm = 0;
+                                for (unsigned ch=0; ch<channels; ++ch) {
+                                    tmpclrs[ch] = inData[ tistride + tjstride + ch*istrides[2] ];
```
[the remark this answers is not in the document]
9prady9: The value of `channels` can be only 3 or 1, and the input validation checks for this are in place under `src/api/c/meanshift.cpp` file already. Also, each color channel contributes to the `norm` variable which is in term used to calculate the mean position and color. I am not sure this is causing the problem you think it is.

src/backend/cuda/kernel/meanshift.hpp
9prady9: `AccType` can be double as well, if the input array is double.

src/backend/opencl/kernel/meanshift.cl
9prady9: `AccType` can be double as well, if the input array is double.

9prady9:
The main change in this PR is not about removing local/shared memory, that happened only as a side effect of the original change. 

The main change is that each iteration of meanshift filter is actually shifting the center of the window to the new color mean color which was not happening with the earlier code. Notice that, each mean shift can be to a different position in different threads, thus I couldn't use shared memory. I am open to suggestions/ideas w.r.t that. I think, using texture memory would give better performance here.

src/backend/cpu/kernel/meanshift.hpp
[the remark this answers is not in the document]
umar456: I don't think this is an input issue. I understand that you can only have one or three channels. In the one channel scenario, this will not be a problem because the access will be linear. This will be a pretty big problem when we are dealing with a 3 channel image because you are accessing data down the third dimension. This type of accesses are not problems in CUDA or OpenCL because each access is linear wrt the threads.

A better approach would be to read the data into a std::array and perform the norm on it before doing a write. This ensures that the reads will be contiguous along the first dimension. Test with a few different sizes to get the optimal value.

src/backend/cuda/kernel/meanshift.hpp
```diff
@@ -12,200 +12,153 @@
...
-        shrd[lIdx(lx, ly, shrdStride, 1)+ch*schStride] = in[lIdx(gx_, gy_, inStride1, inStride0)+ch*ichStride];
-}
-
-template<typename T, int channels>
+template<typename AccType, typename T, int channels>
 static __global__
-void meanshiftKernel(Param<T> out, CParam<T> in,
-                     float space_, int radius, float cvar,
-                     uint iter, int nBBS0, int nBBS1)
+void meanshiftKernel(Param<T> out, CParam<T> in, int radius, float cvar, uint numIters,
```
mlloreda: How come the integral type parameters are missing `const`?

9prady9:
@mlloreda The failures on OSX-OpenCL, I have looked into them. 

- I  haven't been able to replicate medfilt_opencl failure even i run the test for multiple iterations consecutively using `--gtest_repeat` flag.
- The meanshift_opencl failure is happening while calling `clEnqueueReadBuffer`, I have tried the fix in the https://github.com/arrayfire/arrayfire/pull/1921 but that didn't work here. 

`clEnqueueReadBuffer` is happening inside `reduce` kernel, thus this bug is not related to meanshift fix I did. So, I think we can go ahead and merge this fix and deal with OSX-OpenCL driver bugs later. @umar456 What do you say ?

umar456:
build arrayfire osx ci

test/canny.cpp
```diff
@@ -95,31 +96,31 @@ void cannyImageOtsuTest(string pTestFile, bool isColor)
...
-        af_array goldArray= 0;
-        dim_t nElems   = 0;
-
         inFiles[testId].insert(0,string(TEST_DIR "/CannyEdgeDetector/"));
         outFiles[testId].insert(0,string(TEST_DIR "/CannyEdgeDetector/"));
 
-        ASSERT_EQ(AF_SUCCESS, af_load_image(&inArray, inFiles[testId].c_str(), isColor));
-        ASSERT_EQ(AF_SUCCESS, af_load_image_native(&goldArray, outFiles[testId].c_str()));
-        ASSERT_EQ(AF_SUCCESS, af_get_elements(&nElems, goldArray));
+        try {
```
umar456: It would have been good if you didn't refactored this file. The tests are failing and I don't know the cause. I noticed that the original test is using af_load_image_native. and your example isn't. Is that intentional?

test/canny.cpp
```diff
@@ -95,31 +96,31 @@ void cannyImageOtsuTest(string pTestFile, bool isColor)
...
-        ASSERT_EQ(AF_SUCCESS, af_canny(&outArray, inArray, AF_CANNY_THRESHOLD_AUTO_OTSU, 0.08, 0.32, 3, false));
+            array output = canny(input, AF_CANNY_THRESHOLD_AUTO_OTSU, 0.08, 0.32, 3, false);
 
-        std::vector<char> outData(nElems);
-        ASSERT_EQ(AF_SUCCESS, af_get_data_ptr((void*)outData.data(), outArray));
+            output = (255.0*normalize<T>(output)).as(type);
 
-        std::vector<char> goldData(nElems);
-        ASSERT_EQ(AF_SUCCESS, af_get_data_ptr((void*)goldData.data(), goldArray));
+            std::vector<T> outData(nElems);
```
umar456: T is char in the original. Again I don't know if this is the cause of the failures. Probably not.

test/meanshift.cpp
```diff
@@ -65,38 +65,29 @@ void meanshiftTest(string pTestFile)
...
+        outFiles[testId].insert(0,string(TEST_DIR "/meanshift/"));
 
-        inFiles[testId].insert(0,string(TEST_DIR"/meanshift/"));
-        outFiles[testId].insert(0,string(TEST_DIR"/meanshift/"));
+        try {
+            af_dtype type = (af_dtype)af::dtype_traits<T>::af_type;
 
-        ASSERT_EQ(AF_SUCCESS, af_load_image(&inArray_f32, inFiles[testId].c_str(), isColor));
-        ASSERT_EQ(AF_SUCCESS, conv_image<T>(&inArray, inArray_f32));
+            array input = loadImage(inFiles[testId].c_str(), isColor).as(type);
```
umar456: Tests should use the C API whenever possible. The C++ API should be tested separately.

9prady9: I can refactor this to use C API.

test/testHelpers.hpp
```diff
@@ -307,29 +307,37 @@ template<typename T>
 bool compareArraysRMSD(dim_t data_size, T *gold, T *data, double tolerance)
 {
     double accum  = 0.0;
-    double maxion = FLT_MAX;//(double)std::numeric_limits<T>::lowest();
+    double maxion = FLT_MIN;//(double)std::numeric_limits<T>::lowest();
```
umar456: Are you sure you don't mean `-FLT_MAX`. FLT_MIN will give you a value close to 0.

9prady9: True, I have question of previous to mine. Why are the `std::numeric_limits` commented in the first place ? They point to the right values and we don't need to bother about signs.

test/canny.cpp
9prady9: I don't see how this is related the `clEqueueReadBuffer` failure on OSX. I have verified the inputs manually.

May be these are stemming from using C++ API. I will refactor this to C and check what happens.

test/canny.cpp
9prady9: Failures are not related to this. I modified the test because it was not properly comparing values. I came across this issue while cross-referencing our output with ITK's.

test/canny.cpp
umar456: I want to have commits that serve one purpose. In this PR you are refactoring canny and modifying mean shift. Having separate commits allows us to pinpoint issues if there are errors in the future and it gives you an understanding why a certain change was made. For example, I don't know why you changed this test. If I looked at the commit message I would see that you modified the mean shift filter which will not give me information about this change.

9prady9:
Closing this PR, will reopen new one with right commit order.

PR #1921, opened 2017-09-10
Workaround for Apple bug in OpenCL driver. Fixes Canny

umar456:
This commit implements a workaround for a Apple bug in their Iris OpenCL driver
where clEnqueueWriteBuffer fails when you pass in static C arrays. This change
fixes canny on OSX.

### DIFF
--- a/src/api/c/canny.cpp
+++ b/src/api/c/canny.cpp
@@ -30,9 +30,10 @@
 #include <tile.hpp>
 #include <iota.hpp>
 #include <utility>
-#include <iostream>
+#include <vector>
 
 using af::dim4;
+using std::vector;
 using namespace detail;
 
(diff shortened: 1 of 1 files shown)


9prady9:
This is the most absurd fix we had to do. @umar456 Is there a way to know if this bug is specific to any one OSX version ? I guess what I want to know is - Can OSX users update their OpenCL libs independently of their OSX update ?

umar456:
The OpenCL framework is part of the OSX install. I am not sure how or when Apple sends out driver updates but I imagine it is done during system updates and those are pretty rare. I remember running into this a couple of years ago but I forgot about it until now. I wouldn't count on them fixing this anytime soon. This is a particularly simple workaround so I don't think its a big issue.

PR #1924, opened 2017-09-10
Fix Pure Virtual Call errors with FFT

umar456:
On Windows, the resources that are released after the main function have exited
cause "Pure Virtual Function Called" errors. It seems that Windows releases all
resources when exiting the main without calling their destructors. When the
destructors are called this error is thrown. This is related to #1899


pavanky:
Earlier when we tried this, it resulted in clFFT complaining when it is cleaning up. Can you double check if that is not happening?

umar456:
I didn't see anything when I ran the tests on Windows. I think we were seeing those errors because we were releasing resources on exit at some point.

build arrayfire ci

umar456:
build arrayfire osx cuda ci

PR #1926, opened 2017-09-11
Mean fixes

pavanky:
- Fixes segaults in OpenCL
- Improves performance of mean slightly.

### DIFF
--- a/src/backend/cuda/kernel/mean.hpp
+++ b/src/backend/cuda/kernel/mean.hpp
@@ -27,34 +27,25 @@ namespace cuda
 namespace kernel
 {
 
-    template<typename T, typename Tw>
-    struct MeanOp
+    template<typename To, typename Tw>
+    __device__ __host__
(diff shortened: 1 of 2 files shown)


9prady9:
@pavanky I am not sure if it's already broken earlier, but covariance is still broken in CUDA backend on OSX and Linux. Probably on Windows too, but there is not job for it. The error is a value error.

```
[ RUN      ] Covariance/0.Matrix
/var/lib/jenkins-slave/workspace/arrayfire-linux/pr-build/test/covariance.cpp:104: Failure
The difference between ::real(currGoldBar[elIter]) and ::real(outData[elIter]) is 0.017891168594360352, which exceeds 1.0e-3, where
::real(currGoldBar[elIter]) evaluates to 1.1246999502182007,
::real(outData[elIter]) evaluates to 1.142591118812561, and
1.0e-3 evaluates to 0.001.
at: 0

[  FAILED  ] Covariance/0.Matrix, where TypeParam = float (2 ms)
```

pavanky:
I think this is because of my change. I'll look into it.

PR #1934, opened 2017-09-18
BUGFIX: Add support for broadcasting weights when using mean.

pavanky:


umar456:
build arrayfire osx opencl ci

PR #1948, opened 2017-10-06
Fixes for meanShift kernels in all backends

9prady9:
[skip arrayfire ci]

### DIFF
--- a/src/api/c/meanshift.cpp
+++ b/src/api/c/meanshift.cpp
@@ -18,19 +18,21 @@
 using af::dim4;
 using namespace detail;
 
-template<typename T, bool is_color>
-static inline af_array mean_shift(const af_array &in, const float &s_sigma, const float &c_sigma, const unsigned iter)
+template<typename T>
+static inline af_array mean_shift(const af_array &in, const float &s_sigma, const float &c_sigma,
+                                  const unsigned niters, const bool is_color)
 {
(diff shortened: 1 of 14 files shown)


src/api/c/meanshift.cpp
```diff
@@ -18,39 +18,39 @@
 using af::dim4;
 using namespace detail;
 
-template<typename T, bool is_color>
-static inline af_array mean_shift(const af_array &in, const float &s_sigma, const float &c_sigma, const unsigned iter)
+template<typename T, bool IsColor>
```
umar456: Could you pass IsColor as an argument until it is required by the kernel? There is no reason to instantiate all these functions for something that is only needed by the kernel.

src/backend/cuda/meanshift.cu
```diff
@@ -17,22 +17,24 @@ using af::dim4;
...
+Array<T>  meanshift(const Array<T> &in,
+        const float &spatialSigma, const float &chromaticSigma,
+        const unsigned numIterations)
 {
     const dim4 dims = in.dims();
 
     Array<T> out   = createEmptyArray<T>(dims);
 
-    kernel::meanshift<T, is_color>(out, in, s_sigma, c_sigma, iter);
+    kernel::meanshift<T, IsColor>(out, in,
```
umar456: you can put the IsColor condition here.

Issue #1949, opened 2017-10-06
af::array::scalar<T> produces wrong numbers


FloopCZ:
## Current behavior:
When using `a.scalar<double>()` on an array of type `af::dtype::f32`, the result is wrong.

## Expected behavior:
`scalar<T>` should behave similarly to e.g., `af::sum<T>` and `af::mean<T>` and produce a correct result even if the array dtype and the C++ type do not match or throw an exception.

## Code to reproduce the issue:
```c++
#include <arrayfire.h>
#include <iostream>
int main()
{
    af::setBackend(AF_BACKEND_CPU);
    af::info();
    af::array a = af::constant(0.2, 1, af::dtype::f32);
    std::cout << "sum: " << af::sum<double>(a) << '\n';
    std::cout << "float scalar: " << a.scalar<float>() << '\n';
    std::cout << "double scalar: " << a.scalar<double>() << '\n';
}
```

Output:
```
ArrayFire v3.5.0 (CPU, 64-bit Linux, build 05999f3)
[0] Intel: Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz, 15988 MB, Max threads(8)
sum: 0.2
float scalar: 0.2
double scalar: 5.16408e-315
```

## Arrayfire version:
ArrayFire v3.5.0 (CPU, 64-bit Linux, build 05999f3)
Tested on CPU and CUDA backends.

9prady9:
`sum<T>` is doing a reduction operation and the template parameter you give to it is the output variable's type. Thus, whatever the result of reduction is cast to the user provide template type.

`scalar<T>` on the other hand is trying to read the first element of the array a.k.a the first element in the GPU data. In your case, the af::array is of type `f32`(float) while you are trying to access it as double. `scalar` method doesn't handle the type coercion here. You have to do that explicitly as follows.

`(double)a.scalar<float>()`.

Having said that, please do note that scalar is made available for debugging purposes and it is quite expensive call.

pavanky:
@9prady9 This shouldn't fail silently. This should throw an exception or return the correct value by copying implicitly.

9prady9:
@FloopCZ We have fixed the type check in the PR https://github.com/arrayfire/arrayfire/pull/1951

FloopCZ:
Thank you, the type check is definitely useful. Anyway, I still suggest to consider the automatic type conversion including the check whether the array is actually a scalar. Why?

Consider a function like this:

```c++
[...]

PR #1951, opened 2017-10-07
Throw error if scalar template type doesn't match array type

9prady9:
Related to #1949

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -980,6 +980,10 @@ af::dtype implicit_dtype(af::dtype scalar_type, af::dtype array_type)
     }                                                               \
     template<> AFAPI T array::scalar() const                        \
     {                                                               \
+        af_dtype type = (af_dtype)af::dtype_traits<T>::af_type;     \
+        if (type != this->type())                                   \
+            AF_THROW_ERR("Requested type doesn't match array type", \
+                         AF_ERR_TYPE);                              \
         T val;                                                      \
         AF_THROW(af_get_scalar(&val, get()));                       \
         return val;                                                 \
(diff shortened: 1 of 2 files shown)


src/api/cpp/array.cpp
```diff
@@ -980,6 +980,10 @@ af::dtype implicit_dtype(af::dtype scalar_type, af::dtype array_type)
     }                                                               \
     template<> AFAPI T array::scalar() const                        \
     {                                                               \
+        af_dtype type = (af_dtype)af::dtype_traits<T>::af_type;     \
+        if (type != this->type())                                   \
+            AF_THROW_ERR("Requested type doesn't match array type", \
```
umar456: `TYPE_ERROR(0, type);`?

test/array.cpp
```diff
@@ -520,3 +520,16 @@ TYPED_TEST(Array, Scalar)
 
     EXPECT_EQ(true, gold[0]==a.scalar<TypeParam>());
 }
+
+TEST(Array, ScalarTypeMismatch)
+{
+    if (noDoubleTests<float>()) return;
+
+    array a = randu(dim4(1), f32);
```
umar456: Nit: Maybe faster to do a constant or create an af::array from data.

test/array.cpp
```diff
@@ -520,3 +520,16 @@ TYPED_TEST(Array, Scalar)
...
+
+    std::vector<float> gold(a.elements());
```
umar456: No need to create a vector for one value.

src/api/cpp/array.cpp
9prady9: There are no arguments for this method right?

PR #1958, opened 2017-10-11
Fix missing example targets in CMake

umar456:


umar456:
build arrayfire osx cpu ci

Issue #1957, opened 2017-10-11
moments calculation 


nikkadim:
Example below shows calculation first-order moment M00 for 4x4 binary array, which should represent area for binary image, but it shows M00 = 1.0;

```
		int h_buffer[] = {0, 1, 1, 0,    // host array with 16 elements
				  1, 1, 1, 1,    
				  1, 1, 1, 1,
				  0, 1, 1, 0};
		dim4 dims(4, 4);
		array A(dims, h_buffer); 

		double m00;
		af_print(A);
		moments(&m00, A.as(b8), AF_MOMENT_M00);
		printf("Moment M00 = %f\n", m00);
```
output:

```
A
[4 4 1 1]
         0          1          1          0
         1          1          1          1
         1          1          1          1
         0          1          1          0

Moment M00 = 1.000000
```

pavanky:
@nikkadim This should be fixed in master. Let me know if it works for you.

PR #1960, opened 2017-10-11
Fixes bug casting moments to bool

syurkevi:
Addresses #1957

### DIFF
--- a/src/api/c/moments.cpp
+++ b/src/api/c/moments.cpp
@@ -80,20 +80,7 @@ af_err af_moments_all(double* out, const af_array in, const af_moment_type momen
 
         af_array moments_arr;
         af_moments(&moments_arr, in, moment);
-
-        const ArrayInfo& m_info = getInfo(moments_arr);
-        af_dtype type = m_info.getType();
-
-        switch(type) {
-            case f32: moment_copy<float>          (out, moments_arr); break;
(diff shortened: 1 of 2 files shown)


src/api/c/moments.cpp
```diff
@@ -29,14 +29,14 @@
...
+template<typename T, typename To>
 static inline void moments(af_array *out, const af_array in, af_moment_type moment)
 {
     Array<float> temp = moments<T>(getArray<T>(in), moment);
     af_array tarr = getHandle<float>(temp);
-    Array<T> output = castArray<T>(tarr);
+    Array<To> output = castArray<To>(tarr);
 
-    *out = getHandle<T>(output);
+    *out = getHandle<To>(output);
```
pavanky: @syurkevi I think the output should always be float. Is there any reason to cast it to `To` ? Casting it to `To` (or `T` for that matter) results in incorrect values for higher order moments.

syurkevi: Leaving everything as floats caused tests to fail for ints and doubles. This is because the tests expect the same output type as input type. I changed the tests, let me know if it's alright or if I should revert.

pavanky:
@syurkevi May be change `af_moments_all` to only check for floats ? You don't need a switch case in there.

pavanky:
build arrayfire ci

PR #1962, opened 2017-10-12
Convert memAlloc to return unique_ptr

umar456:
memAlloc currently returns raw pointers. This is unsafe because it requires an explicit free operation to release memory. This PR converts memAlloc to return a unique_ptr which releases memory on exit.


umar456:
@pavanky The homograph_cuda test still fails on Windows with the M6000. This test was always passing before this PR. I am not sure what is the cause for this. Could you make sure I didn't make a mistake anywhere?

pavanky:
@umar456 can this be merged?

Issue #1952, opened 2017-10-08
Opencl error "double requires cl_khr_fp64" on Nvidia fermi


fzimmermann89:
The following tests fail for datatype=double using a GTX 580 (fermi, opencl 1.1) on windows 7

- cholesky_dense_opencl
- inverse_dense_opencl
- lu_dense_opencl
- solve_dense_opencl

> 
> [----------] 2 tests from Inverse/2, where TypeParam = double
> [ RUN      ] Inverse/2.Square
> OpenCL error -59 on line 1192
> OpenCL error -59 on line 1365
> OpenCL error -11 on line 228
> clBuildProgram Failed
> err = -11
> Error: Failed to build program executable!
> 
> Build Log:
> 
> <kernel>:14:11: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __global double const * restrict A, 
>           ^
> <kernel>:16:11: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __global double *d_dinvA, 
>           ^
> <kernel>:21:2: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  double Ystx = 0;
>  ^
> <kernel>:22:10: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __local double *y = 0;
>          ^
> <kernel>:23:2: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  double switcher;
>  ^
> <kernel>:24:2: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  double neg_switcher;
>  ^
> <kernel>:29:17: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __global const double *Aoff = A + bx*lda*BLOCK_SIZE + bx*BLOCK_SIZE;
>                 ^
> <kernel>:32:10: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __local double Bs[BLOCK_SIZE*BLOCK_SIZE];
>          ^
> <kernel>:33:10: error: use of type 'double' requires cl_khr_fp64 extension to be enabled
>  __local double workspace[BLOCK_SIZE]; 
>          ^
> 
> OpenCL error -45 on line 253
> OpenCL error -48 on line 132
> OpenCL error -48 on line 990
> OpenCL error -48 on line 992
> OpenCL error -48 on line 994
> OpenCL error -48 on line 996
> OpenCL error -48 on line 998
[...]

PR #1964, opened 2017-10-13
Fix mean kernel for OpenCL 1.1 devices

fzimmermann89:
Enable usage of double if output is of type double 
As mentioned in https://github.com/arrayfire/arrayfire/issues/1952, cl_khr_fp32 has to be enabled to use double. This is already done if the input is double, but should be done as well if the output is double (because for int64-input the output is double, so double is used inside the kernel)

### DIFF
--- a/src/backend/opencl/kernel/mean.hpp
+++ b/src/backend/opencl/kernel/mean.hpp
@@ -174,7 +174,8 @@ void mean_dim_launcher(Param out, Param owt,
         if (output_weight) { options << " -D OUTPUT_WEIGHT"; }
 
         if (std::is_same<Ti, double>::value ||
-                std::is_same<Ti, cdouble>::value) {
+                std::is_same<Ti, cdouble>::value ||
+                std::is_same<To, double>::value) {
             options << " -D USE_DOUBLE";
         }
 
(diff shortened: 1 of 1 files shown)


pavanky:
@fzimmermann89 Can you enable if `To` is double as well ?

src/backend/opencl/kernel/mean.hpp
```diff
@@ -174,7 +174,8 @@ void mean_dim_launcher(Param out, Param owt,
         if (output_weight) { options << " -D OUTPUT_WEIGHT"; }
 
         if (std::is_same<Ti, double>::value ||
-                std::is_same<Ti, cdouble>::value) {
+                std::is_same<Ti, cdouble>::value ||
+                std::is_same<To, cdouble>::value) {
```
pavanky: Did you mean to use cdouble here or double ?

src/backend/opencl/kernel/mean.hpp
```diff
@@ -340,7 +341,8 @@ void mean_first_launcher(Param out, Param owt,
...
```
pavanky: Did you mean to use cdouble here or double ?

fzimmermann89: argh, yes...

src/backend/opencl/kernel/mean.hpp
fzimmermann89: ...yes...

pavanky:
@fzimmermann89 Can you add the failing code from #1952 as a test ?

fzimmermann89:
Test_mean_opencl on opencl1.1 covers it. Otherwise it is difficult to test for correct setting of cl_khr_fp32 without messing with the compiler

pavanky:
@fzimmermann89 Thanks for letting me know!

pavanky:
build arrayfire ci

PR #1967, opened 2017-10-15
Disable native_exp for double precision in OpenCL

### LINKED ISSUE #1966: Bilateral uses exp_native(double) even though it's not in the OpenCL specifications
The bilataral_opencl test fails for double on Nvidia Fermi
In arrayfire/src/backend/opencl/kernel/bilateral.cl `exp_native` is used for double inputs. But this is not specified in either of the OpenCL specifications even though some implementations seem to support it, the Nvidia OpenCL implementation (at least for Fermi) doesn't

@pavanky already found this issue for POCL (pocl/pocl#385) and @shehzan10 for apple. The current solution is checking for those two platforms before enabling `exp_native` https://github.com/arrayfire/arrayfire/blob/125100705e4e47e1854735feebce891c103a1b0c/src/backend/opencl/kernel/bilateral.hpp#L52

Maybe there is a better way of fixing it instead of just checking for the platform?
I don't know if there is a way  to check for existence of `exp_native(double)`.
Because `expnative(double)` is not specified I would suggest not using it if USE_DOUBLE is defined.

fzimmermann89:
Change to not using native_exp for type double instead of checking the platform.

native_exp is not officially defined for input type double, this seems cleaner than adding more platforms for disabling it. 
This will work for POCL and will probably work for OSX - will have to see the ci tests.

### DIFF
--- a/src/backend/opencl/kernel/bilateral.hpp
+++ b/src/backend/opencl/kernel/bilateral.hpp
@@ -47,16 +47,15 @@ void bilateral(Param out, const Param in, float s_sigma, float c_sigma)
     kc_entry_t entry = kernelCache(device, refName);
 
     if (entry.prog==0 && entry.ker==0) {
-        bool use_native_exp = (getActivePlatform() != AFCL_PLATFORM_POCL
-                && getActivePlatform() != AFCL_PLATFORM_APPLE);
         std::ostringstream options;
         options << " -D inType=" << dtype_traits<inType>::getName()
             << " -D outType=" << dtype_traits<outType>::getName();
         if (std::is_same<inType, double>::value ||
                 std::is_same<inType, cdouble>::value) {
             options << " -D USE_DOUBLE";
+        } else {
+            options << " -D USE_NATIVE_EXP";
         }
-        options << " -D USE_NATIVE_EXP=" << (int)use_native_exp;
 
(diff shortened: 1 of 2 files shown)


src/backend/opencl/kernel/bilateral.hpp
```diff
@@ -47,16 +47,16 @@ void bilateral(Param out, const Param in, float s_sigma, float c_sigma)
...
-                && getActivePlatform() != AFCL_PLATFORM_APPLE);
         std::ostringstream options;
         options << " -D inType=" << dtype_traits<inType>::getName()
             << " -D outType=" << dtype_traits<outType>::getName();
         if (std::is_same<inType, double>::value ||
                 std::is_same<inType, cdouble>::value) {
             options << " -D USE_DOUBLE";
         }
-        options << " -D USE_NATIVE_EXP=" << (int)use_native_exp;
+        else {
```
pavanky: Nit: Can you move `else`  to line 56 ? (i.e. make it look like `} else {`

fzimmermann89: ok

pavanky:
build arrayfire ci

pavanky:
@umar456 Any idea why windows opencl is failing?

umar456:
I forgot to update a path in the test script. Its rerunning now.

PR #1969, opened 2017-10-17
Remove old build status links from readme

umar456:
[skip ci]


pavanky:
@umar456 Is there a plan to replace this with something new?

9prady9:
Yes, we just removed the obsolete devel branch links. We will add new buildbot status hyperlinks again later.

PR #1979, opened 2017-10-25
Use default FindBLAS instead of FindCBLAS

umar456:
The default FindBLAS library seems to be working fine. I need to test this on additional platforms. Please feel free to send me any feedback regarding compatibility on your systems.


fzimmermann89:
...It will probably make a change to the windows build scripts necessary. But those in arrayfire/arrayfire-windows-scripts are already broken.
I would appreciate an update to those scripts, otherwise I will create a PR once your done with changing the build system. Right know whenever I have a version ready for a PR something else changes - and my test setup is quite slow ;)

umar456:
Hey I am sorry this has been causing issues for you. My goal with these changes is to make things easier to build and setup while at the same time simplify the logic within the code. I have looked at the scripts and they are not exactly user friendly. It differentiated the different libraries and required the user to modify those scripts to work. The changes I am making now should allow you to put all paths(to glbinding/blas/fft etc.) into the CMAKE_PREFIX_PATH variable. I have been using this workflow to build/test on windows:

1. Open Visual Studio 2015 Command Prompt
2. Run CMake:
```
cmake -G Ninja .. -DCMAKE_PREFIX_PATH:PATH=C:/workspace/dependencies\mkl;C:/workspace/dependencies\freeimage-3.17.0_x64-vs2015;C:/workspace/dependencies\boost_1_56_0  
-DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo -DFFTW_ROOT:PATH=C:/workspace/dependencies\mkl -DGLFW_ROOT_DIR:PATH=C:/workspace/dependencies\glfw-vs2015 ..
```
3. run `cmake --build .`

So because of forge I need to add the GLFW_ROOT_DIR path in there so that forge can find that library but it is not necessary. You can get the latest command I use by going to the cmake step in the buildbot config. Here is the [link](http://67.207.87.39:8010/#/builders/38/builds/63/steps/2/logs/stdio).

Again I am sorry that this has caused you issues. I want to get to a point where you do not need those scripts and everything is automatically found. I have tried creating native [NuGet](https://www.nuget.org/) packages but the package manager is not managed correctly and I don't want to make things more difficult by creating ArrayFire dependencies packages. [vcpkg](https://github.com/Microsoft/vcpkg) looks promising but the last time I checked it didn't have all the dependencies we needed for ArrayFire. Anyway, I will try to minimize the changes.

umar456:
FindBLAS is missing a lot of functionallity. It seems it doesn't find the header files so it will have to be modified anyway. There have been talk about updating the FindBLAS to take into account more modern libraries.

PR #2002, opened 2017-12-01
BUGFIX: Fixes ireduce functions when the reduced dimension is of size one

umar456:
The index array in the ireduce dimension was not assigned values
when the reduced dimension size was one. This commit addresses
that and unifies the naming convention of the IReduced test cases

### DIFF
--- a/src/api/c/reduce.cpp
+++ b/src/api/c/reduce.cpp
@@ -395,14 +395,15 @@ static af_err ireduce_common(af_array *val, af_array *idx, const af_array in, co
 {
     try {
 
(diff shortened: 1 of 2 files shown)


9prady9:
Everything looks good.

@umar456 Don't you think MinReduceDimensionHasSingleValue &
MaxReduceDimensionHasSingleValue testing the same functionality? We could
have just one test instead.

umar456:
They are testing different functions. I could create an enum and parameterize the tests but I didn't think it was worth it for 2 functions.

PR #2005, opened 2017-12-04
Set homebrew glbinding dir

bzcheeseman:
If `glbinding` gets installed with homebrew we need to tell cmake where the config file is. I was not able to get it to work with 
```
execute_process(COMMAND brew list glbinding
                             COMMAND grep config 
                             COMMAND xargs dirname
                             RESULT_VARIABLE glbinding_DIR
)
```
which is why we have a hardcoded path here. However, if the lib is installed via homebrew on OSX 10.12 this is where the config file lives (as tested on 2 mac machines). I don't like the hardcoded path, and if someone has better ideas by all means!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ find_package(Doxygen)
 
 # Graphics dependencies
 if (APPLE)
-  set(glbinding_DIR /usr/local/opt/glbinding)
+  set(glbinding_DIR dirname $(brew list glbinding | grep config))
 endif ()
 find_package(glbinding QUIET)
 find_package(Boost)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,9 @@ find_package(LAPACKE)
 find_package(Doxygen)
 
 # Graphics dependencies
+if (APPLE)
+  set(glbinding_DIR /usr/local/opt/glbinding)
+endif ()
 find_package(glbinding QUIET)
 find_package(Boost)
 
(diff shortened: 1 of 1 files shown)


bzcheeseman:
CMake is not able to find glbinding when I run that command actually - on either machine I tested it on. I was setting up the second machine just today when I ran into the bug! I did the following:
```
git clone arrayfire
cd arrayfire
mkdir build; cd build
cmake ..  // errored out because of missing glbinding
brew install glbinding
cmake .. // errored out again, same thing
// add the above to the CMakeLists.txt file
cmake .. // configured successfully
```
My os is 10.12 and I was using clang-3.7 and clang-4.0, same error on both. CMake versions are 3.8 and 3.10, respectively.

I haven't tried on 10.13 because I didn't want to update quite yet. If it's a problem with the OS then I understand why it might need to simply wait.

9prady9:
@arrayfire/core-devel  Do any one of you still use OSX 10.12 version ?

umar456:
This is how I run the CMake command on the CI:

```
cmake CMAKE_PREFIX_PATH:PATH="/usr/local/opt"
```

This is a more flexible way of specifying the path because all find_* commands will use this path to search for dependencies. Also brew's location for the glbinding-config.cmake paths are not installed in the correct location. It should be installed into [one of the default directories](https://cmake.org/cmake/help/v3.6/command/find_package.html) specified by CMake so that we don't need to specify the prefix path for this library. It would be nice if someone could fix it :)

umar456:
I am not sure what is the best way to fix it. I think it should be brew that needs to be updated to fix this issue.

umar456:
Addressed via  #2018. @bzcheeseman can you try that branch to see if you are experiencing the same issue. It also address some issues with MKL on all platforms.

PR #2010, opened 2017-12-08
BUGFIX: make af_lookup behavior consistent across backends

### LINKED ISSUE #2009: af::lookup fails on CUDA when looking up in a single column array
Hi team,

I think I've found a bug in the af::lookup() method on the CUDA backend. This example demonstrates it:

```
    af::array a = af::range(af::dim4(5, 1));
    af::array idx = af::constant(0, 1, u32);
    af::array l = lookup(a, idx, 1);

    af::print("a", a);
    af::print("idx", idx);
    af::print("l", l);
```

On the CPU backend this produces the correct result:

```
a
[5 1 1 1]
    0.0000 
    1.0000 
    2.0000 
    3.0000 
    4.0000 

idx
[1 1 1 1]
         0 
l
[5 1 1 1]
    0.0000 
    1.0000 
    2.0000 
    3.0000 
    4.0000 
```

However on the CUDA backend the last array is incorrect:

```
a
[5 1 1 1]
    0.0000 
    1.0000 
    2.0000 
    3.0000 
    4.0000 

idx
[1 1 1 1]
         0 
l
[5 1 1 1]
    0.0000 
    0.0000 
    0.0000 
    0.0000 
    0.0000 
```

If the initial array a was a 2+ column array the lookup works as expected on CUDA:

```
a
[5 2 1 1]
    0.0000     0.0000 
    1.0000     1.0000 
    2.0000     2.0000 
    3.0000     3.0000 
    4.0000     4.0000 

idx
[1 1 1 1]
         0 
l
[5 1 1 1]
    0.0000 
    1.0000 
    2.0000 
    3.0000 
    4.0000 
```

It just doesn't seem to work when a is a 1 column array!!

I am using CUDA 8, NVIDIA driver 384.90 on a GTX 1070, Ubuntu 16.04 and latest master arrayfire built from source as of about a month ago.

9prady9:
Fixes #2009 

CUDA backend results of af_lookup were different from CPU/OpenCL
for Vector input arrays when the indexing dimension is not along
the same dimension as the input array data.

### DIFF
--- a/src/backend/cuda/kernel/lookup.hpp
+++ b/src/backend/cuda/kernel/lookup.hpp
@@ -79,36 +79,18 @@ void lookupND(Param<in_t> out, CParam<in_t> in, CParam<idx_t> indices,
 template<typename in_t, typename idx_t, unsigned dim>
 void lookup(Param<in_t> out, CParam<in_t> in, CParam<idx_t> indices, int nDims)
 {
-    if (nDims==1) {
-        const dim3 threads(THREADS, 1);
-        /* find which dimension has non-zero # of elements */
-        int vDim = 0;
-        for (int i=0; i<4; i++) {
-            if (in.dims[i]==1)
-                vDim++;
-            else
-                break;
-        }
+    const dim3 threads(THREADS_X, THREADS_Y);
 
-        int blks = divup(out.dims[vDim], THREADS*THRD_LOAD);
+    int blks_x = divup(out.dims[0], threads.x);
+    int blks_y = divup(out.dims[1], threads.y);
 
-        dim3 blocks(blks, 1);
+    dim3 blocks(blks_x*out.dims[2], blks_y*out.dims[3]);
 
-        CUDA_LAUNCH((lookup1D<in_t, idx_t>), blocks, threads, out, in, indices, vDim);
-    } else {
-        const dim3 threads(THREADS_X, THREADS_Y);
+    const int maxBlocksY = cuda::getDeviceProp(cuda::getActiveDeviceId()).maxGridSize[1];
+    blocks.z = divup(blocks.y, maxBlocksY);
+    blocks.y = divup(blocks.y, blocks.z);
 
-        int blks_x = divup(out.dims[0], threads.x);
-        int blks_y = divup(out.dims[1], threads.y);
-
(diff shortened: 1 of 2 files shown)


src/backend/cuda/kernel/lookup.hpp
```diff
@@ -79,36 +79,18 @@ void lookupND(Param<in_t> out, CParam<in_t> in, CParam<idx_t> indices,
...
+    const dim3 threads(THREADS_X, THREADS_Y);
 
-        int blks = divup(out.dims[vDim], THREADS*THRD_LOAD);
+    int blks_x = divup(out.dims[0], threads.x);
+    int blks_y = divup(out.dims[1], threads.y);
 
-        dim3 blocks(blks, 1);
+    dim3 blocks(blks_x*out.dims[2], blks_y*out.dims[3]);
 
-        CUDA_LAUNCH((lookup1D<in_t, idx_t>), blocks, threads, out, in, indices, vDim);
```
umar456: I would like it if you fixed lookup1D instead of not using it. If it can't be fixed then you should also remove the kernel.

test/index.cpp
```diff
@@ -709,6 +709,15 @@ TEST(lookup, largeDim)
     af::array output = af::lookup(input, indices);
 }
 
+TEST(lookup, Issue2009)
+{
+    af::array a   = af::range(af::dim4(1000, 1));
+    af::array idx = af::constant(0, 1, u32);
+    af::array b   = af::lookup(a, idx, 1);
+
+    ASSERT_EQ(true, af::allTrue<bool>(a==b));
```
umar456: We need to figure out a way to print values that are not correct when a test fails so that we don't have to dig into the functions to figure out what is wrong. If this test were to fail we wouldn't know what characteristics the array has when it fails.

9prady9: I think we can generalize couple of helper functions from all the tests we currently have and use those everywhere. In those functions, whenever the test fails, we print out the the details to log.

umar456: stdout is fine. gtest has ways to print custom objects. We need to start thinking about doing that because we waste a lot of time figuring out how something is failing.

9prady9: Yeah, stdout is good, by log I meant the ci logs. True.

9prady9:
Accidentally created branch on remote repository.

PR #2012, opened 2017-12-08
Docs fix in cmake

### LINKED ISSUE #2008: Update unified backend doc
Seen on
http://arrayfire.org/docs/unifiedbackend.htm
"On Linux and OSX, add /usr/local/cuda/nvvm/(lib or lib64) to LD_LIBRARY_PATH or DYLD_LIBRARY_PATH."
Is nt nvvm no more needed from 3.5+ ?
Kind

9prady9:
Fixes #2008

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,3 +266,4 @@ endif()
 
 conditional_directory(BUILD_TESTING test)
 conditional_directory(BUILD_EXAMPLES examples)
+conditional_directory(BUILD_DOCS docs)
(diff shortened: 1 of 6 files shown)


9prady9:
I have one more commit to docs of af_lookup but that is not rendering after I do code block in documentation for some reason. Also, there are still more warnings about syntax error with out AF_API_VERSION preprocessor directives. Thus, none of `\ref` to enums/functions who has API guards are not working.

PR #2014, opened 2017-12-11
BUGFIX: correct assets path in examples

### LINKED ISSUE #2013: ASSETS_DIR variable not set for examples in CMake
This an issue only in `image_processing` and `machine_learning` examples where images are being used but the `ASSETS_DIR` was not properly set.

9prady9:
Fixes #2013 

ASSETS_DIR was not set to a proper path earlier in image_processing
and machine_learning examples due to which the examples which use
images failed during image loads.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -4,6 +4,18 @@ project(ArrayFire-Examples
   VERSION 3.5.0
   LANGUAGES CXX)
 
+if (NOT ASSETS_DIR)
+    set(ASSETS_DIR ""
+        CACHE PATH
+        "
+        Assets are the images and data files required by the examples. ASSETS_DIR
+        should point to the path where theses files are available on the build machine.
+
+        You can download or clone the assets git repository from the following
+        url: https://github.com/arrayfire/assets
+        ")
+endif (NOT ASSETS_DIR)
+
 if(WIN32)
   add_definitions(-DWIN32_LEAN_AND_MEAN)
   unset(CMAKE_RUNTIME_OUTPUT_DIRECTORY)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,4 +265,6 @@ if(BUILD_TEST)
 endif()
 
 conditional_directory(BUILD_TESTING test)
+
+set(ASSETS_DIR "${ArrayFire_SOURCE_DIR}/assets")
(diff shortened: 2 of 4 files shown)


examples/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ project(ArrayFire-Examples
   VERSION 3.5.0
   LANGUAGES CXX)
 
+if (NOT ASSETS_DIR)
+    option(ASSETS_DIR "Please provide the path to assets for ArrayFire examples")
```
umar456: Option creates a boolean variable. You want to create a variable using the set command that is cached if the examples are not built as part of the ArrayFire library.

examples/image_processing/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(ArrayFire-Example-Image-Processing
 
 find_package(ArrayFire)
 
-add_definitions("-DASSETS_DIR=\"${ASSETS_DIR}\"")
+add_definitions(-DASSETS_DIR=\"${ASSETS_DIR}\")
```
umar456: I don't know about this. I was almost certain you needed to quotes for CMake to handle the escape sequence. Can you make sure this is working by examining the command line?

examples/CMakeLists.txt
9prady9: Fixing it.

examples/image_processing/CMakeLists.txt
9prady9: It is working fine without quotes. In fact, I only made this consistent another file where quotes weren't used earlier. So, I tested it without quotes and it worked fine.

examples/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ project(ArrayFire-Examples
   VERSION 3.5.0
   LANGUAGES CXX)
 
+if (NOT ASSETS_DIR)
+    set(ASSETS_DIR "" CACHE PATH "Please provide the path to assets for ArrayFire examples")
```
umar456: The Doc string should describe the variable not instruct or ask for info from the user. Use something like

The assets directory containing images and other data for examples. Download from blah...

examples/image_processing/CMakeLists.txt
9prady9: Just realised, without quotes, windows causes issues if the path has spaces.windo

PR #2022, opened 2017-12-17
BUGFIX: Fix Boost compile definitions on Windows

9prady9:
Without BOOST_CHRONO_HEADER_ONLY compile definition for boost
on Windows, the Boost::boost target will try to link chrono lib
also - which is not needed and thus avoided.


umar456:
Wouldn't that definition be required for older versions of boost?

9prady9:
@umar456 Yes, but you are handling that inside `build_boost_compute`. I am fixing the the other case, newer versions.

umar456:
Perhaps remove the flags from boost compute and add it in one location. That way if we decide to add additional flags in the future that target boost we only need to change it in one location.

9prady9:
@umar456 Please check now.

PR #2024, opened 2017-12-18
BUGFIX: Fix OpenCL tests on Windows for CUDA9

9prady9:
The issue fixed in this commit started happening on Windows platform with CUDA9.

Both boost::compute::program_cache and ArrayFire contexts are static variables. The boost cache was being cleaned up after ArrayFire released the corresponding OpenCL contexts.

This fix keeps a pointer to the shared_ptr to boost::compute::program_cache and leaks it on Windows platform to circumvent the problem.

### DIFF
--- a/src/backend/opencl/platform.hpp
+++ b/src/backend/opencl/platform.hpp
@@ -30,6 +30,14 @@
 #include <memory.hpp>
 #include <GraphicsResourceManager.hpp>
 
+namespace boost {
+  template<typename T> class shared_ptr;
+
+  namespace compute {
+    class program_cache;
+  }
+}
+
 // Forward declaration from clFFT.h
 struct clfftSetupData_;
 typedef clfftSetupData_ clfftSetupData;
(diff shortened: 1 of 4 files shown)


src/backend/opencl/platform.hpp
```diff
@@ -198,5 +201,8 @@ class DeviceManager
         std::unique_ptr<GraphicsResourceManager> gfxManagers[MAX_DEVICES];
 #endif
         std::unique_ptr<clfftSetupData> mFFTSetup;
+
+        using BoostProgCache = boost::shared_ptr<boost::compute::program_cache>;
+        std::vector<BoostProgCache*> mBoostProgCacheVector;
```
umar456: Since we are using a pointer here. Can you use a forward declaration instead of including the headers. I want to minimize the includes in this header since it is used everywhere.

umar456: Addressed

PR #2030, opened 2017-12-25
Update configuring_arrayfire_environment.md: Fixed wrong description of default behavior when AF_MEM_DEBUG is not set.

HoneyPatouceul:
Fixed wrong description of default behavior when AF_MEM_DEBUG is not set.


umar456:
Thanks for the fix! I will merge this soon.

PR #2031, opened 2017-12-26
Add support for precompiled headers and unity builds

umar456:
Precompiled headers allow you to precompile some headers so that the same
headers are not compiled multiple times. This commit adds support for
precompiled headers using the cotire module.

Cotire also provides support for unit builds which combine multiple source files
together to reduce recompilation of the same headers.

Current limitations
* Unity builds only work with the CPU backend
  * OpenCL has several static variables that cause conflicts when the souces are combined
* Precompiled headers does not work well with CUDA files. Only *.cpp files are compiled
* Interface library sources are not included in the precompiled headers


umar456:
Not a significant improvement in the build times and it adds a lot of complexity. We may revisit this in the future but for now I am closing this PR.

PR #2032, opened 2017-12-27
Fixed variable name typo in vectorization.md

HoneyPatouceul:
Fixed typo in variable name defined as g_coef but used as f_coef.


mlloreda:
Thanks for PR, @HoneyPatouceul!

PR #2058, opened 2018-02-19
Fix assertions for select test failures

umar456:


test/select.cpp
```diff
@@ -133,7 +133,7 @@ TEST(Select, NaN)
     c.host(&hc[0]);
 
     for (int i = 0; i < num; i++) {
-        ASSERT_EQ(hc[i], std::isnan(ha[i]) ? b : ha[i]);
+        ASSERT_FLOAT_EQ(hc[i], std::isnan(ha[i]) ? b : ha[i]);
```
9prady9: Did this work? I have tried these on `linux-01` and it didn't work.

umar456: Yes. I tested this on linux-01. 

http://67.207.87.39:8010/#/builders/92/builds/120

PR #2061, opened 2018-02-22
FEAT: topk function in all backends.

umar456:
CUDA backend alone implements a custom kernel to fetch top k elements
without sorting all the values. CPU backends sorts the data and fetch
the top k elements. The OpenCL backend is optimized for CPU devices
to map the memory and perform a partial sort to get the results.

Sponsored by [SDL](https://www.sdl.com/)


src/api/c/topk.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <handle.hpp>
+#include <topk.hpp>
+#include <common/ArrayInfo.hpp>
+#include <common/err_common.hpp>
+#include <backend.hpp>
+
+using namespace detail;
+
+template<typename T>
+af_err topkWithIndices(af_array *v, af_array* i, const af_array in,
```
9prady9: can you change this function name to be topk as well since we are not explicitly saying with indices anywhere now.

docs/details/statistics.dox
```diff
@@ -62,6 +62,21 @@ Find the correlation coefficient of values in the input
...
+========================================================
+\defgroup stat_func_topk topk
+
+\ingroup basicstats_mat
+
+This function returns the top k values along a given dimension of the input
+array. The indices along with their values are returned. If the input is a
+multi-dimensional array, the indices will be the index of the value in that
+dimension. Order of duplicate values are not preserved. This function is
+optimized for small values of k.
```
9prady9: It would be nice if we can add some sort of guideline as to from which value of k does the performance deteriorate.

test/topk.cpp
```diff
@@ -0,0 +1,286 @@
...
+//Enable char, uchar, short, ushort types only after figuring out
+//how to do stable sort with CUB so that duplicate data elements
+//are handled properly
+//typedef ::testing::Types<float, double, int, uint, short, ushort, char, uchar> TestTypes;
+typedef ::testing::Types<float, double, int, uint> TestTypes;
+
+TYPED_TEST_CASE(TopK, TestTypes);
+
+template<typename T>
+void topkIndexTest(const unsigned ndims, const dim_t* dims,
```
9prady9: We can just make it `topkTest` instead of `topkIndexTest` now.

pavanky:
@umar456 why not partial_sort for cpu devices?

docs/details/statistics.dox
umar456: This would be different for different devices and implementations.

Issue #2062, opened 2018-02-22
OpenCL HelloWorld examples produce Segfault without display


reichardtj:
Thanks for making Arrayfire such an awesome library! I've installed Arrayfire on a google compute engine node with ubuntu 16.04 LTS, cuda 8.0
I've used the binary installer from http://arrayfire.s3.amazonaws.com/3.5.1/ArrayFire-v3.5.1_Linux_x86_64.sh and built the examples as described in http://arrayfire.org/docs/installing.htm .

running the examples leads to
```
~/examples/build/helloworld$ ./helloworld_opencl
ERROR: GLFW wasn't able to initalize
ArrayFire v3.5.1 (OpenCL, 64-bit Linux, build 0a675e8)
[0] NVIDIA: Tesla K80, 11441 MB
Create a 5-by-3 matrix of random floats on the GPU
Segmentation fault (core dumped)
```

As there is no display attached to the server, I can fix it with

```
export AF_DISABLE_GRAPHICS=1
```

and then I get the output
```
~/examples/build/helloworld$ ./helloworld_opencl
ArrayFire v3.5.1 (OpenCL, 64-bit Linux, build 0a675e8)
[0] NVIDIA: Tesla K80, 11441 MB
Create a 5-by-3 matrix of random floats on the GPU
A
... lots of more numbers ...
Warning:  Program terminating, but clFFT resources not freed.
Please consider explicitly calling clfftTeardown( ).
```

I suppose the hello world examples should not break just because there is no display attached and I believe this is a bug.

All other HelloWorld examples (for cuda, cpu and unified perform flawlessly - also without the clFFT warning)

----------------------------------------------------------------------------------------------- 

For reference, my entire installation procedure from a clean Ubuntu 16.04. LTS image was

follow https://cloud.google.com/compute/docs/gpus/add-gpus#install-driver-script

```
sudo apt-get update
sudo apt-get upgrade
sudo apt-get clean
sudo apt-get remove unattended-upgrades
echo "Checking for CUDA and installing."
# Check for CUDA and try to install.
if ! dpkg-query -W cuda-8-0; then
  # The 16.04 installer works with 16.10.
  curl -O http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/cuda-repo-ubuntu1604_8.0.61-1_amd64.deb
  dpkg -i ./cuda-repo-ubuntu1604_8.0.61-1_amd64.deb
  apt-get update
[...]

PR #2063, opened 2018-02-22
Note about installation on headless systems in docs

9prady9:


umar456:
The AF_DISABLE_DISPLAY variable should not be required on headless nodes. There is a separate issue that is causing problems on the machine. I was only able to reproduce this using the installers and not when I built the same commit on a Google Cloud instance described in #2062. We need to investigate this further.

PR #2064, opened 2018-02-22
removes warnings from doxygen

syurkevi:


docs/details/index.dox
```diff
@@ -14,7 +14,7 @@
 
 \brief Index an array using another array.
 
-Lets look at an example of how \ref lookup function does indexing.
+Lets look at an example of how \ref af::lookup function does indexing.
```
mlloreda: s/Lets/Let's/

PR #2068, opened 2018-02-25
Fixes linking examples with targets that are not built

umar456:
CMake built ArrayFireConfig files for each of the backend when they are built.
If you disabled a backend after it was built, the config file created a target
for the backend even though it doesn't exist. This commit modifys the config
file so that it only creates the target if the target's library also exists.

The side effect of this change is that if the library exists but the
BUILD_<backend> is off then it will build the examples anyway.


9prady9:
@umar456 
I tested(a clean build) this on Arch Linux with CMake 3.10.2 and the fix doesn't seem to be working. Here are the steps for you to reproduce on Arch.

1. `cmake .. -DBUILD_CUDA:BOOL=OFF` - that build cpu and opencl backend along with the examples.
2. `cmake .. -DBUILD_CPU:BOOL=OFF` - to turn off CPU build later, after building everything from step 1 - doesn't stop cpu examples from building again - same goes for opencl examples.

umar456:
@9prady9 Addressed feedback. The examples should be disabled if the appropriate AF_BUILD_* variable is not set.

9prady9:
@umar456 The problem is still present.

umar456:
@9prady9 Should be fixed. I was using the incorrect case for OpenCL.

PR #2070, opened 2018-02-26
Use namespaced CMake variables. Deprecated old values

umar456:
This commit deprecates the old CMake variables in favor of variables with `AF_` prefix to avoid name conflicts with user code. If the old variables are used a warning is displayed and the new variable is set to the value. The old variable is unset to avoid showing the warning again.


9prady9:
@umar456 Why are the below tests failing with this change? They don't fail anywhere usually.
```
gloh_nonfree_opencl
sift_nonfree_opencl
```

CMakeModules/InternalUtils.cmake
```diff
@@ -36,6 +36,22 @@ function(arrayfire_get_cuda_cxx_flags cuda_flags)
...
+function(__af_deprecate_var var access value)
+  if(access STREQUAL "READ_ACCESS")
+    message(DEPRECATION "Variable ${var} is deprecated. Use AF_${var} instead.")
+  endif()
+endfunction()
+
+function(af_deprecate var newvar)
+  if(DEFINED ${var})
+    message(DEPRECATION "Variable ${var} is deprecated. Use ${newvar} instead.")
+    get_property(doc CACHE ${newvar} PROPERTY HELPSTRING)
```
9prady9: where is this `doc` defined ? Is it predefined ? I didn't find any reference to it in root cmake script or here.

9prady9:
Changes look good.

CMakeModules/InternalUtils.cmake
umar456: Its the variable where the `get_property` command stores the `HELPSTRING` for the `${newvar}` variable.

PR #2079, opened 2018-03-08
Use built-in erfc for cumulative normal distribution

### LINKED ISSUE #2078: Example code for cummulative normal distribution is wrong
The example code for the cummulative normal distribution in
https://github.com/arrayfire/arrayfire/blob/master/examples/financial/black_scholes_options.cpp#L22
gives wrong results as can be seen by adapting the test code from http://www.johndcook.com/blog/cpp_phi/:
```
#include <iostream>
#include <math.h>
#include <arrayfire.h>

using namespace af;

// The following function is a modified version of http://www.johndcook.com/blog/cpp_phi/
// The example above references Handbook of Mathematical Functions by Abramowitz and Stegun

array cnd(array x)
{
    // constants
    const float a1 =  0.254829592;
    const float a2 = -0.284496736;
    const float a3 =  1.421413741;
    const float a4 = -1.453152027;
    const float a5 =  1.061405429;
    const float p  =  0.3275911;
    const float sqrt2 = sqrt(2.0);

    // Save the sign of x
    array xSign = sign(x);

    x = abs(x) / sqrt2;

    // A&S formula 7.1.26
    array t = 1.0f / (1.0f + p*x);
    array y = 1.0f + 0.5f * (((((a5*t + a4)*t) + a3)*t + a2)*t + a1)*t*exp(-x*x);

    return xSign * y + !xSign * (1 - y); // equivalent of (x >= 0) ? y : (1 - y);
}

void testCnd()
{
    // Select a few input values
    float x[] = {-3.0, -1.0, 0.0, 0.5, 2.1};
    array X(5, x, afHost);

    // Output computed by Mathematica
    // y = Phi[x]
    float y[] = {0.00134989803163, 0.158655253931, 0.5, 0.691462461274, 0.982135579437};
    array Y(5, y, afHost);
    
    float maxError = max<float>(abs(Y - cnd(X))); 
    std::cout << "Maximum error: " << maxError << "\n";
} 

int main() {
  testCnd();
  return 0;
}
```
This gives: `Maximum error: 1`

One could try to fix the above code, but since it uses one particular approximation to the error function `erf` which is already available, it is easier to just use [that relation](https://en.wikipedia.org/wiki/Error_function#Cumulative_distribution_function) directly:
```
array cnd(array x)
{
    const float sqrt05 = sqrt(0.5f);
    return 0.5f * erfc(- x * sqrt05);
}
```

If you want I can provide a PR for this.

rstub:
fixes #2078

### DIFF
--- a/examples/financial/black_scholes_options.cpp
+++ b/examples/financial/black_scholes_options.cpp
@@ -16,30 +16,13 @@
 #include "input.h"
 using namespace af;
 
-// The following function is a modified version of http://www.johndcook.com/blog/cpp_phi/
-// The example above references Handbook of Mathematical Functions by Abramowitz and Stegun
-
+// Use the relationship between the cumulative normal distribution and the
+// (complementary) error function:
+// https://en.wikipedia.org/wiki/Error_function#Cumulative_distribution_function
 array cnd(array x)
 {
-    // constants
-    const float a1 =  0.254829592;
-    const float a2 = -0.284496736;
-    const float a3 =  1.421413741;
-    const float a4 = -1.453152027;
-    const float a5 =  1.061405429;
-    const float p  =  0.3275911;
-    const float sqrt2 = sqrt(2.0);
-
-    // Save the sign of x
-    array xSign = sign(x);
-
-    x = abs(x) / sqrt2;
-
-    // A&S formula 7.1.26
-    array t = 1.0f / (1.0f + p*x);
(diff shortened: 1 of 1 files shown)


9prady9:
:+1: Thank you.

One minor correction. Please fix the typo in commit message - `built-in` Also, please move the `fixes #2078` to description section as the commit message is getting a bit longer. Thank you once again for catching this mistake.

PR #2082, opened 2018-03-10
Add cmaps missing in arrayfire but in Forge

afvincent:
Sync the `af_colormap` enum with the similar one in Forge.

I have to admit that this PR was written through the GH web interface because I never used arrayfire (yet ;)?), so please correct me if I am wrong! This change looks like a (very) low-hanging fruit, as it is simply a matter of adding a few missing entries compared to the [`fg_color_map` enum in the Forge `define.h`](http://arrayfire.org/forge/defines_8h.htm), is it not? By doing so, arrayfire users would gain access to the *perceptually uniform* colormaps that were originally designed for Matplotlib a while ago, and have started to spread among other projects. In my opinion, this would be a neat small additional feature for arrayfire, but I may be biased here ;) (regular Matplotlib user here...).

Being there, I also added (as comments) a couple information that were in the Forge header file, but not the arrayfire one.

PS: I opened this PR against the master branch because I did not find any `devel` branch as mentioned in the [contributing guidelines](https://github.com/arrayfire/arrayfire/wiki/Contribute-code-using-github)

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -424,12 +424,16 @@ typedef enum {
 ////////////////////////////////////////////////////////////////////////////////
 typedef enum {
     AF_COLORMAP_DEFAULT = 0,    ///< Default grayscale map
-    AF_COLORMAP_SPECTRUM= 1,    ///< Spectrum map
-    AF_COLORMAP_COLORS  = 2,    ///< Colors
+    AF_COLORMAP_SPECTRUM= 1,    ///< Spectrum map (390nm-830nm, in sRGB colorspace)
+    AF_COLORMAP_COLORS  = 2,    ///< Colors, aka. Rainbow
     AF_COLORMAP_RED     = 3,    ///< Red hue map
     AF_COLORMAP_MOOD    = 4,    ///< Mood map
     AF_COLORMAP_HEAT    = 5,    ///< Heat map
-    AF_COLORMAP_BLUE    = 6     ///< Blue hue map
+    AF_COLORMAP_BLUE    = 6,    ///< Blue hue map
+    AF_COLORMAP_INFERNO = 7,    ///< Perceptually uniform shades of black-red-yellow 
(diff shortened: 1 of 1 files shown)


9prady9:
@afvincent Adding the enums should be sufficient enough I think. I will give it a try once the ci jobs are done. Thanks for added comments, although the old comments are in line with comments from the days of ArrayFire 2.0. But I guess, they could use better descriptions.

Thanks for pointing out the mistake in contributing guidelines. I have corrected them now. We have changed gears regarding the git branching model couple of months back and looks like we missed updating the contributing guidelines. All PRs go into master now and there will be separate separate feature release branches so that we can backport fixes.

include/af/defines.h
```diff
@@ -424,12 +424,16 @@ typedef enum {
...
-    AF_COLORMAP_SPECTRUM= 1,    ///< Spectrum map
-    AF_COLORMAP_COLORS  = 2,    ///< Colors
+    AF_COLORMAP_SPECTRUM= 1,    ///< Spectrum map (390nm-830nm, in sRGB colorspace)
+    AF_COLORMAP_COLORS  = 2,    ///< Colors, aka. Rainbow
     AF_COLORMAP_RED     = 3,    ///< Red hue map
     AF_COLORMAP_MOOD    = 4,    ///< Mood map
     AF_COLORMAP_HEAT    = 5,    ///< Heat map
-    AF_COLORMAP_BLUE    = 6     ///< Blue hue map
+    AF_COLORMAP_BLUE    = 6,    ///< Blue hue map
+    AF_COLORMAP_INFERNO = 7,    ///< Perceptually uniform shades of black-red-yellow
```
9prady9: Though github doesn't seem to display it, there is an additional space towards end all newly added enums. Please remove the it.

afvincent: @9prady9 Sorry about that.... I think that I got them all.

PR #2094, opened 2018-03-22
Fix CLBLast libdir

FloopCZ:
On some systems, CLBlast libs are placed in `lib64` folder instead of `lib`, thus arrayfire won't find them and the build fails. This fixes the issue.

@CNugteren, can you confirm this will not cause some hidden issues?

The affected system is:
```
cmake version 3.10.3
Arch Linux (4.15.9-1-ARCH Sun Mar 11 17:54:33 UTC 2018 x86_64)
```


CNugteren:
@FloopCZ Not sure, I'm not familiar with `CMAKE_INSTALL_LIBDIR` and potential compatibility issues with all operating systems and use-cases. But if you are on a 64-bit Linux system, shouldn't ArrayFire look in the lib64 folder when trying to link with CLBlast?

FloopCZ:
@CNugteren That would be the case if Arrayfire would just link the library from the system library path. However, it downloads and builds its own static CLBlast as an CMake external project and than hardcodes the path to the library (including the `lib` folder; see [build_CLBlast.cmake](https://github.com/arrayfire/arrayfire/blob/master/CMakeModules/build_CLBlast.cmake#L11)).

I believe this should have no side effects, it only keeps the same behavior across all systems. An alternative would be to use [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) to detect whether to use the `64` suffix, but in this case, it is an overkill.

9prady9:
@FloopCZ 

I have tested it on my arch too and the change looks good. We probably missed it because we use the default option, clblas, most of the times. Can you please remove the merge commit and rebase your changes from master. We are trying to keep the git history as linear as possible. Thank you!

FloopCZ:
@9prady9 Sure thing. I appreciate you are taking a good care of the repository. :-)

Issue #2099, opened 2018-03-30
Inconsistency in CMake variables


FloopCZ:
This is a low priority. Recently, there has been a change in the `CMakeLists.txt` prefixing most of the variables with `AF_` prefix. Some of the variables, however, are still using their old name. I know about two - `OPENCL_BLAS_LIBRARY` and `BUILD_TESTING`. Was it intentional or just an oversight? Thank you.

9prady9:
@FloopCZ 

Regarding `BUILD_TESTING` variable - it is created by gtest suite and ArrayFire's CMake doesn't cache this variable so we can't do anything about this one. We can mask this with another variable in ArrayFire CMakeLists.txt file but it would just add one more unncessary variable I believe.

I think @umar456 left out the upstream project options like `MKL_THREAD_LAYER` & `OPENCL_BLAS_LIBRARY` to have the dependency as it's their prefix or may be he just missed them. 

But there are few variables(other than `CMAKE` prefixed) that can be marked as advanced, I believe.

FloopCZ:
Thank you for the information @9prady9. I just wanted to to make sure that this was intentional before v3.6 is released, because changing CMake variable names means some extra digging for quite a few Linux package maintainers.

9prady9:
The package maintainers would still be able to use old variables too. If
old variables are used a deprecation warning is thrown and the project file
generation wouldn't be effected.

PR #2100, opened 2018-03-31
Mark some cblas/lapacke/CUDA cmake vars as advanced

9prady9:
Addresses some changes related to #2099


CMakeLists.txt
```diff
@@ -77,6 +77,15 @@ mark_as_advanced(
   AF_USE_SYSTEM_FORGE
   AF_WITH_CPUID)
 
+# mark some CUDA_* variables from
```
umar456: This commend doesn't really add anything. Remove it.

CMakeLists.txt
```diff
@@ -77,6 +77,15 @@ mark_as_advanced(
...
+# find_package(CUDA) as advanced
+mark_as_advanced(
+    CUDA_SDK_ROOT_DIR
+    CUDA_TOOLKIT_ROOT_DIR
+    CUDA_HOST_COMPILER
```
umar456: Keep this visible. It is useful for distros where the default compiler is not compatible with the one CUDA requires.

CMakeLists.txt
```diff
@@ -77,6 +77,15 @@ mark_as_advanced(
   AF_USE_SYSTEM_FORGE
   AF_WITH_CPUID)
 
+# mark some CUDA_* variables from
+# find_package(CUDA) as advanced
+mark_as_advanced(
```
umar456: You can combine this with the previous mark_as_advanced call.

CMakeModules/FindCBLAS.cmake
```diff
@@ -46,7 +46,7 @@ IF(PC_CBLAS_FOUND)
   SET(CBLAS_INCLUDE_DIR ${CBLAS_INCLUDE_DIRS})
 
   FIND_PACKAGE_HANDLE_STANDARD_ARGS(CBLAS DEFAULT_MSG CBLAS_LIBRARIES CBLAS_INCLUDE_DIR)
-  MARK_AS_ADVANCED(CBLAS_LIBRARIES CBLAS_INCLUDE_DIR)
+  MARK_AS_ADVANCED(CBLAS_LIBRARIES CBLAS_INCLUDE_DIR CBLAS_INCLUDE_DIRS)
```
umar456: use lower case functions. One variable per line.

src/backend/opencl/CMakeLists.txt
```diff
@@ -400,13 +402,13 @@ target_link_libraries(afopencl
     Threads::Threads
     )
 
-if(OPENCL_BLAS_LIBRARY STREQUAL "clBLAS")
+if(AF_WITH_OPENCL_BLAS_LIBRARY STREQUAL "clBLAS")
```
umar456: this should be AF_OPENCL_BLAS_LIBRARY instead of AF_WITH_OPENCL_BLAS_LIBRARY

CMakeLists.txt
9prady9: Does changing `CUDA_HOST_COMPILER` change `CMAKE_CXX_COMPILER` & `CMAKE_C_COMPILER` automatically ? If not, it should be advanced variable like `CMAKE_C**_COMPILER` variables, otherwise it only adds to the confusion.

CMakeModules/FindCBLAS.cmake
9prady9: That file is quite old, I was just sticking to whatever style that was present in the file. I can't change the function to lower case in just one line, it would look odd.

This file may need a rewrite and I don't want to mix those changes with this PR.

PR #2101, opened 2018-04-02
Simplify ArrayFireConfig script and improve documentation.

umar456:
This commit improves the ArrayFireConfig.cmake scripts. These scripts
are used to find ArrayFire when the find_package command is called.
This commit makes the following changes:

* Create one target file for all backends instead of one per backend
* Add components support to warn if a particular backend was not
  installed
* Update the documentation and encurage the use of imported targets
  instead of the libraries and include variables
* ArrayFire_DIR now needs to point to the build directory instead
  of the build/cmake directory to work correctly


9prady9:
@umar456 The install files looks great on my ubuntu. I have one doubt though, right now it is generating `ArrayFireTargets-noconfig.cmake` file alone. How do we generate such file that is config specific ? Debug | Release etc.

PR #2103, opened 2018-04-03
Create installers with dependences only when AF_INSTALL_STANDALONE is set

umar456:


src/api/c/CMakeLists.txt
```diff
@@ -159,8 +159,10 @@ if(FreeImage_FOUND AND AF_WITH_IMAGEIO)
   else ()
     target_link_libraries(c_api_interface INTERFACE FreeImage::FreeImage)
     if (WIN32)
-      install(FILES $<TARGET_FILE:FreeImage::FreeImage>
-        DESTINATION ${AF_INSTALL_BIN_DIR})
+      if(AF_INSTALL_STANDALONE)
```
9prady9: Can't we just do `WIN32 AND AF_INSTALL_STANDALONE` directly ?

umar456: Addressed

PR #2109, opened 2018-04-04
Fix cpack GUI installer components dependencies

9prady9:


CMakeModules/CPackConfig.cmake
```diff
@@ -131,79 +131,69 @@ cpack_add_install_type(Extra DISPLAY_NAME "Extra")
...
+  if (NOT APPLE)
+    cpack_add_component(opencl
+      DISPLAY_NAME "OpenCL Backend"
+      DESCRIPTION "This Backend allows you to take advantage of OpenCL capable GPUs to run ArrayFire code. Currently ArrayFire does not support OpenCL for the Intel CPU on OSX."
+      DEPENDS mkl_dependencies common_backend_dependencies opencl_dependencies
+      INSTALL_TYPES Development Runtime)
+  else ()
+    cpack_add_component(opencl
+      DISPLAY_NAME "OpenCL Backend"
+      DESCRIPTION "This Backend allows you to take advantage of OpenCL capable GPUs to run ArrayFire code. Currently ArrayFire does not support OpenCL for the Intel CPU on OSX."
```
umar456: if we are going to have a separate section we should remove the Intel CPU line from here. Also you should mention CPUs and GPUs

CMakeModules/CPackConfig.cmake
```diff
@@ -131,79 +131,69 @@ cpack_add_install_type(Extra DISPLAY_NAME "Extra")
...
-    DISPLAY_NAME "CUDA Backend"
-    DESCRIPTION "This Backend allows you to take advantage of the CUDA enabled GPUs to run ArrayFire code. Please make sure you have CUDA toolkit installed or install CUDA dependencies component."
-    DEPENDS cuda_dependencies
-    INSTALL_TYPES Development Runtime)
-elseif (PACKAGE_GFX_DEPS)
+  if (NOT APPLE)
+    cpack_add_component(opencl
+      DISPLAY_NAME "OpenCL Backend"
+      DESCRIPTION "This Backend allows you to take advantage of OpenCL capable GPUs to run ArrayFire code. Currently ArrayFire does not support OpenCL for the Intel CPU on OSX."
+      DEPENDS mkl_dependencies common_backend_dependencies opencl_dependencies
```
umar456: Can't you just make a dependency list based on the CMake options instead of repeating the same thing over and over again?

CMakeModules/CPackConfig.cmake
```diff
@@ -131,81 +131,55 @@ cpack_add_install_type(Extra DISPLAY_NAME "Extra")
...
-    DEPENDS gfx_dependencies
-    INSTALL_TYPES Development Runtime)
-  cpack_add_component(cuda
-    DISPLAY_NAME "CUDA Backend"
-    DESCRIPTION "This Backend allows you to take advantage of the CUDA enabled GPUs to run ArrayFire code. Please make sure you have CUDA toolkit installed or install CUDA dependencies component."
-    DEPENDS gfx_dependencies cuda_dependencies
-    INSTALL_TYPES Development Runtime)
+cpack_add_component(cuda
+  DISPLAY_NAME "CUDA Backend"
+  DESCRIPTION "This Backend allows you to take advantage of the CUDA enabled GPUs to run ArrayFire code. Please make sure you have CUDA toolkit installed or install CUDA dependencies component."
```
umar456: I don't think the part about installing the cuda toolkit is relevant anymore

9prady9: If the user chose to not install `CUDA Dependencies` component in our installer, then they have to install the `cu*` libraries themselves. I thought that is entirely the reason you asked me to provide an option for `CUDA dependencies` component to be deselected if the user chooses to.

PR #2112, opened 2018-04-05
Do not build forge if USE_SYSTEM_FORGE

FloopCZ:
- Add missing `find_package(Forge)` call in order to be able to deduce the default value for `AF_WITH_GRAPHICS`, which uses `Forge_FOUND`.
- Only build `forge-ext` if `AF_USE_SYSTEM_FORGE` is `OFF`.
- Only add `forge-ext` dependency to individual libraries if `AF_USE_SYSTEM_FORGE` is `OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ find_package(MKL)
 
 # Graphics dependencies
 find_package(glbinding QUIET)
+find_package(Forge QUIET)
 include(boost_package)
 
 option(AF_BUILD_CPU      "Build ArrayFire with a CPU backend"        ON)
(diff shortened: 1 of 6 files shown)


9prady9:
There is a typo in `src/api/c/CMakeLists.txt` , please replace `afcommon_interface` with `c_api_interface`

FloopCZ:
Oh, really sorry for that. Thank you for the review, that was an ugly mistake. :-/

9prady9:
I think we missed adding `forge-ext` as dependency for `afcuda` and `afopencl` targets similar to what we have for `afcpu`. Can you please those as well.

It might not cause any issue unless the build is launched with too many threads which is probably the reason we didn't notice any problems so far.

FloopCZ:
Sure thing, I am going afk for a few hours so I'll do it later, though.

FloopCZ:
Done.

PR #2118, opened 2018-04-09
Add a default constructor for Param. Fixes var issue #2117

### LINKED ISSUE #2117: Crash in afopencl var/mean with 3.6.0 RC
Hi guys
I got a crash with afopencl (3.6.0 RC3) in af::var() which is calling af::mean().
My app, built with gcc 6.1, was running single threaded, release build, using the official intel opencl runtime installed in 
/opt/intel/opencl-1.2-6.4.0.25  
on CentOS 7 : 

Backtrace:
/lib64/libOpenCL.so.1(clRetainMemObject+0x37)[0x7f5996519c17]
libafopencl.so.3(_ZN2cl6detail7WrapperIP7_cl_memEC2ERKS4_+0x17)[0x7f5996bcbe27]
libafopencl.so.3(_ZN6opencl6kernel17mean_dim_launcherIfffEEvNS_5ParamES2_S2_S2_iiPKj+0xc1e)[0x7f5997feb43e]
libafopencl.so.3(_ZN6opencl6kernel8mean_dimIfffEEvNS_5ParamES2_S2_i+0x719)[0x7f5997fece39]
libafopencl.so.3(_ZN6opencl4meanIfffEENS_5ArrayIT1_EERKNS1_IT_EEi+0x3c2)[0x7f5997fed8d2]
libafopencl.so.3(+0x2475767)[0x7f5998b95767]
libafopencl.so.3(af_var+0x95)[0x7f5998b9cd55]
libaf.so.3(_ZN7unified15AFSymbolManager4callIJPPvS2_bxEEE6af_errPKcDpT_+0x1b6)[0x7f59b4924596]
libaf.so.3(af_var+0x4c0)[0x7f59b491a8f0]
libaf.so.3(_ZN2af3varERKNS_5arrayEbx+0x55)[0x7f59b4a25d15]
 
Trying to reproduce in debug build with gdb. Not yet simple codes to reproduce.

umar456:
Fixes #2117 

The issue appeared because the temporary Param struct was uninitialized.

### DIFF
--- a/src/backend/opencl/kernel/mean.hpp
+++ b/src/backend/opencl/kernel/mean.hpp
@@ -278,7 +278,7 @@ void mean_dim(Param out, Param in, Param inWeight, int dim)
         groups_all[dim] = 1;
         mean_dim_launcher<Ti, Tw, To>(out, owt, tmpOut, tmpWeight, dim, threads_y, groups_all);
     } else {
-        Array<Tw> tmpWeight = createEmptyArray<Tw>(0);
+        Param tmpWeight;
         mean_dim_launcher<Ti, Tw, To>(out, tmpWeight, in, inWeight, dim, threads_y, groups_all);
     }
 
--- a/src/backend/opencl/kernel/mean.hpp
+++ b/src/backend/opencl/kernel/mean.hpp
@@ -476,7 +476,7 @@ void mean_weighted(Param out, Param in, Param inWeight, int dim)
 template<typename Ti, typename Tw, typename To>
 void mean(Param out, Param in, int dim)
 {
-    Array<Tw> noWeight = createEmptyArray<Tw>(dim4(0, 0, 0, 0));
+    Param noWeight;
     mean_weighted<Ti, Tw, To>(out, in, noWeight, dim);
 }
 
@@ -576,8 +576,8 @@ To mean_all(Param in)
         uint groups_y = divup(in.info.dims[1], threads_y);
 
         Array<To> tmpOut = createEmptyArray<To>(groups_x);
-        Array<To> iWt = createEmptyArray<To>(0);
         Array<Tw> tmpCt = createEmptyArray<Tw>(groups_x);
+        Param iWt;
 
         mean_first_launcher<Ti, Tw, To>(tmpOut, tmpCt, in, iWt, threads_x, groups_x, groups_y);
 
--- a/test/var.cpp
+++ b/test/var.cpp
@@ -162,3 +162,12 @@ TYPED_TEST(Var, DimCPPSmall)
         }
     }
 }
+
+TEST(Var, ISSUE2117) {
+  using namespace af;
+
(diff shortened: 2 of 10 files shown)


test/var.cpp
```diff
@@ -125,40 +127,51 @@ TYPED_TEST(Var, DimCPPSmall)
...
-                // NOTE: will work for all types
-                if (is_same_type<float, outType>::value ||
-                    is_same_type<cfloat, outType>::value) {
-                    ASSERT_FLOAT_EQ(real(h_out[j][jj]), real(tests[j][jj]));
-                    ASSERT_FLOAT_EQ(imag(h_out[j][jj]), imag(tests[j][jj]));
-                } else {
-                    ASSERT_DOUBLE_EQ(real(h_out[j][jj]), real(tests[j][jj]));
-                    ASSERT_DOUBLE_EQ(imag(h_out[j][jj]), imag(tests[j][jj]));
-                }
+    array input(numDims[testNum], &in[testNum].front(), afHost);
```
9prady9: Couple of things, This is not related to the fix, is it ?

And I don't think it is doing what you are expecting it to do. `readTests` function is to be called once, but moving the for loop associated with number of tests to the caller(above `testCppVarDim`) level, you are reading the test data many times over where one read should be sufficient enough.

src/backend/opencl/kernel/mean.hpp
```diff
@@ -278,7 +278,7 @@ void mean_dim(Param out, Param in, Param inWeight, int dim)
         groups_all[dim] = 1;
         mean_dim_launcher<Ti, Tw, To>(out, owt, tmpOut, tmpWeight, dim, threads_y, groups_all);
     } else {
-        Array<Tw> tmpWeight = createEmptyArray<Tw>(0);
+        Param tmpWeight;
```
9prady9: There are similar Array declarations on Lines 479 & 579 in the same source file. Can you please correct them as well.

test/var.cpp
umar456: There is only one test so this shouldn't be an issue.
Also, we shouldn't be writing large test cases that handle multiple things. We need to be able to figure out what is going wrong just by looking at the tests that pass and tests that fail. If we can't do that then we waste time debugging tests. I tried to fix this test but it required more effort than I was willing to spend on this. I will come back to it at a later time.

src/backend/opencl/kernel/mean.hpp
umar456: Addressed

test/var.cpp
9prady9: That may be the case, but that is not the point I was trying to make. `readTests` is being used incorrectly and anyone who looks at this file for reference will be mislead as to how that function works and it can lead to whole new set of problems. This change shouldn't go in this PR and should be made separately whenever we will fix this later.

umar456: Addressed

PR #2123, opened 2018-04-12
Remove nvcc-gcc version compatibility check

9prady9:


pavanky:
gcc 7 is not compatible with cuda 9.1. But the toolkit errors out so i am not sure if that error should exist in cmake.

9prady9:
@pavanky We discussed about it and decided to not have any nvcc-gcc checks as that is taken care of by nvcc. Instead, we will udpate the development github wiki suggesting the developer to check [cuda toolkit requirements section](http://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#system-requirements) if they get an errors related to nvcc-gcc compatibility. This removes the burden of maintaining the checks for every new CUDA toolkit release. This also removes redundant checks in our cmake which nvcc is already doing.

PR #2124, opened 2018-04-12
Move nativeFree calls outside of the memory_mutex to avoid deadlock

### LINKED ISSUE #2115: Freeze in cpu::MemoryManager::nativeFree()
Hi @umar456 
We are experimenting a freeze with the cpu backend: my app, running in single thread mode, simply stops (but does nt crash).
When that happens, my app does not use any cpu/cores, probably a deadlock.
With AF 3.6.0 (master) using the spare CPU backend on CentOS 7 :  

```
#0 0x00007fffe5cee7f9 in syscall () from /lib64/libc.so.6
#1 0x00007ffff78e64d0 in std::__atomic_futex_unsigned_base::_M_futex_wait_until (this=<optimized out>, __addr=0x1827f010, __val=2147483648,
__has_timeout=<optimized out>, __s=..., __ns=...) at ../../../../../gcc-6.1.0/libstdc+-v3/src/c+11/futex.cc:55
#2 0x00007fffd67f9010 in cpu::MemoryManager::nativeFree(void*) ()
from libafcpu.so.3
#3 0x00007fffd67fb250 in common::MemoryManager<cpu::MemoryManager>::cleanDeviceMemoryManager(int) ()
from libafcpu.so.3
#4 0x00007fffd67fb812 in common::MemoryManager<cpu::MemoryManager>::alloc(unsigned long, bool) ()
from libafcpu.so.3
#5 0x00007fffd67fb912 in std::unique_ptr<float [], std::function<void (float*)> > cpu::memAlloc<float>(unsigned long const&) ()
from libafcpu.so.3
#6 0x00007fffd648d631 in cpu::Array<float>::Array(af::dim4) ()
from libafcpu.so.3
#7 0x00007fffd648d72c in cpu::Array<float> cpu::createEmptyArray<float>(af::dim4 const&) ()
from libafcpu.so.3
#8 0x00007fffd691e91b in cpu::Array<float> cpu::reduce<(af_op_t)18, float, float>(cpu::Array<float> const&, int, bool, double) ()
from libafcpu.so.3
#9 0x00007fffd7215a28 in af_max ()
from libafcpu.so.3
#10 0x00007fffe701b70d in af_err unified::AFSymbolManager::call<void*, void, int>(char const*, void*, void, int) ()
from libaf.so.3
#11 0x00007fffe700eaf4 in af_max ()
from libaf.so.3
#12 0x00007fffe71e7328 in af::max(af::array const&, int) ()
```

At the moment, sorry but no codes to reproduce.

umar456:
Narrows the scope of the memory_mutex locks to just around the
modification and reading of the internal memory data members and
elements. This means that actual allocations happen outside of this
mutex's lock.

This was done because the CPU backend requires that the free operation
performs a sync before the pointer is freed. This caused deadlocks
when the main thread was waiting on the worker thread and the
worker thread called nativeFree. This scenario happens when a buffer
node is only referenced in the worker thread and it is removed.

Fixes #2115

### DIFF
--- a/src/backend/common/MemoryManager.hpp
+++ b/src/backend/common/MemoryManager.hpp
@@ -9,12 +9,14 @@
 
 #pragma once
 
-#include <common/err_common.hpp>
 #include <common/dispatch.hpp>
+#include <common/err_common.hpp>
 #include <common/util.hpp>
 
 #include <algorithm>
+#include <functional>
 #include <iomanip>
+#include <memory>
 #include <mutex>
 #include <string>
 #include <unordered_map>
@@ -22,8 +24,8 @@
 
 namespace common
 {
-typedef std::recursive_mutex mutex_t;
-typedef std::lock_guard<mutex_t> lock_guard_t;
+using mutex_t      = std::recursive_mutex;
+using lock_guard_t = std::lock_guard<mutex_t>;
 
 const unsigned MAX_BUFFERS   = 1000;
 const size_t ONE_GB = 1 << 30;
@@ -38,11 +40,13 @@ class MemoryManager
         size_t bytes;
     } locked_info;
 
-    using locked_t = typename std::unordered_map<void *, locked_info>;
+    using locked_t    = typename std::unordered_map<void *, locked_info>;
     using locked_iter = typename locked_t::iterator;
 
-    typedef std::unordered_map<size_t, std::vector<void *> >free_t;
-    typedef free_t::iterator free_iter;
+    using free_t    = std::unordered_map<size_t, std::vector<void *> >;
+    using free_iter = free_t::iterator;
+
+    using uptr_t = std::unique_ptr<void, std::function<void(void*)>>;
 
     typedef struct memory_info
     {
@@ -91,23 +95,33 @@ class MemoryManager
     {
         if (this->debug_mode) return;
 
-        lock_guard_t lock(this->memory_mutex);
+        // This vector is used to store the pointers which will be deleted by
+        // the memory manager. We are using this to avoid calling free while
+        // the lock is being held becasue the CPU backend calls sync.
+        std::vector<void*> free_ptrs;
         memory_info& current = memory[device];
+        {
+            lock_guard_t lock(this->memory_mutex);
+            // Return if all buffers are locked
+            if (current.total_buffers == current.lock_buffers) return;
+            free_ptrs.reserve(32);
 
-        // Return if all buffers are locked
-        if (current.total_buffers == current.lock_buffers) return;
-
-        for (auto &kv : current.free_map) {
-            size_t num_ptrs = kv.second.size();
-            //Free memory by popping the last element
-            for (int n = num_ptrs-1; n >= 0; n--) {
-                this->nativeFree(kv.second[n]);
-                current.total_bytes -= kv.first;
-                current.total_buffers--;
-                kv.second.pop_back();
+            for (auto &kv : current.free_map) {
+                size_t num_ptrs = kv.second.size();
+                // Free memory by pushing the last element into the free_ptrs
+                // vector which will be freed once outside of the lock
+                for(auto p : kv.second) {
+                    free_ptrs.push_back(p);
+                }
+                current.total_bytes -= num_ptrs * kv.first;
+                current.total_buffers -= num_ptrs;
             }
+            current.free_map.clear();
         }
-        current.free_map.clear();
+        // Free memory outside of the lock
+        for(auto ptr : free_ptrs) {
+            this->nativeFree(ptr);
+        }
     }
 
     public:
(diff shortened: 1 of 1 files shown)


src/backend/common/MemoryManager.hpp
```diff
@@ -236,71 +250,75 @@ class MemoryManager
         // Shortcut for empty arrays
         if (!ptr) return;
 
-        lock_guard_t lock(this->memory_mutex);
-        memory_info& current = this->getCurrentMemoryInfo();
+        // Frees the pointer outside the lock.
+        std::unique_ptr<void, std::function<void(void*)>> freed_ptr(nullptr, [this](void* p) {
+                                          this->nativeFree(p);
+                                        });
```
FloopCZ: (OCD - this lambda seems misindented)

src/backend/common/MemoryManager.hpp
```diff
@@ -236,71 +250,75 @@ class MemoryManager
...
-        // Pointer not found in locked map
-        if (iter == current.locked_map.end()) {
-            // Probably came from user, just free it
-            this->nativeFree(ptr);
-            return;
-        }
+            // Pointer not found in locked map
+            if (iter == current.locked_map.end()) {
+                // Probably came from user, just free it
+                freed_ptr.reset(ptr);
```
FloopCZ: This is a nice trick with `unique_ptr`. :+1:

src/backend/common/MemoryManager.hpp
```diff
@@ -236,71 +250,75 @@ class MemoryManager
...
+                free_iter fiter = current.free_map.find(bytes);
+                if (fiter != current.free_map.end()) {
+                    // If found, push back
+                    fiter->second.push_back(ptr);
+                } else {
+                    // If not found, create new vector for this size
+                    std::vector<void *> ptrs;
+                    ptrs.push_back(ptr);
+                    current.free_map[bytes] = ptrs;
+                }
```
FloopCZ: I believe this whole "regular mode" branch (L291-L301) can be replaced with this line:
```c++
current.free_map[bytes].push_back(ptr);
```
because [operator[]](http://en.cppreference.com/w/cpp/container/unordered_map/operator_at) on mapping containers constructs the vector for you if the key does not exist.

src/backend/common/MemoryManager.hpp
```diff
@@ -233,74 +248,81 @@ class MemoryManager
...
         // Shortcut for empty arrays
         if (!ptr) return;
 
-        lock_guard_t lock(this->memory_mutex);
-        memory_info& current = this->getCurrentMemoryInfo();
+        // Frees the pointer outside the lock.
+        unique_ptr<void, function<void(void*)>> freed_ptr(nullptr,
+                                                          [this](void* p) {
+                                                              this->nativeFree(p);
+                                                          });
```
pavanky: Move this into the class definition of MemoryManager ? It could be useful elsewhere.

src/backend/common/MemoryManager.hpp
pavanky: Agreed :+1:

src/backend/common/MemoryManager.hpp
pavanky: I agree with this ^ I was not aware of this quirk of C++ standard containers when I wrote this code.

src/backend/common/MemoryManager.hpp
```diff
@@ -424,6 +445,7 @@ class MemoryManager
     bool checkMemoryLimit()
     {
         const memory_info& current = this->getCurrentMemoryInfo();
+        lock_guard_t lock(this->memory_mutex);
```
pavanky: Why lock here ? Was getCurrentMemoryInfo locking previously ?

pavanky:
> I thought you wanted to avoid virtual table lookup for performance (I may be wrong, though, considering that in this PR you e.g., store the call to nativeFree in std::function that is performing a similar type of type erasure anyway.). However, the common MemoryManager has a virtual destructor, is there a reason for this? Because you cannot use it purely polymorphically anyway, it always has to be parametrized using the backend-specific manager type.

This is a result of MemoryManager starting off as a polymorphic base class and then being converted to the templated version because we could not guarantee the order of destruction of certain objects which resulted in sgefaults on exits (I don't recall the specifics).

I guess the virtual destructor is something that was missed when the conversion was done.

9prady9:
@FloopCZ Yes, it was a miss as @pavanky pointed out and I missed removing that. `virtual` qualifier is not needed there and in fact we can just remove that destructor and let the compiler handle it since it doesn't anything specific over there. @umar456 Can you please remove it.

FloopCZ:
@pavanky @9prady9 Thank you for your answers and @umar456 for the code fixing this tricky bug.

PR #2125, opened 2018-04-16
Remove recursive mutexes and locks on driver APIs

umar456:
Recursive mutexs are known to promote poor design choices. This commit
removes the need for recursive mutexes in the memory manager and
OpenCL backends.

### DIFF
--- a/src/backend/common/MemoryManager.hpp
+++ b/src/backend/common/MemoryManager.hpp
@@ -24,7 +24,7 @@
 
 namespace common
 {
-using mutex_t      = std::recursive_mutex;
+using mutex_t      = std::mutex;
 using lock_guard_t = std::lock_guard<mutex_t>;
 
 const unsigned MAX_BUFFERS   = 1000;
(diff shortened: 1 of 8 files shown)


pavanky:
@umar456 are you sure there aren't any deadlocks ?

umar456:
I went all the way down the stack for each lock several times. I made sure that we are holding the locks only when we are reading from and writing to the data structures. I tried to copy the data to a local variable if we were holding the lock for too long.

9prady9:
@umar456 Is it possible to write a test that tries to create a deadlock situation(probably overloading with threads - need not be 100% fool-proof, something that does some preliminary checking) ? That can help us prevent a regression bug when something related to memory manager changes in the future.

We can probably run it on nightly jobs or something like that.

umar456:
It's hard to write unit tests for these things that would catch deadlock bugs because each interaction between functions is different. There are several approaches which have been suggested like [this](https://www.microsoft.com/en-us/research/project/chess-find-and-reproduce-heisenbugs-in-concurrent-programs/?from=http%3A%2F%2Fresearch.microsoft.com%2Fen-us%2Fprojects%2Fchess%2F) but we might explore something like this later.

PR #2126, opened 2018-04-17
Disable ignored-attributes warnings in the OpenCL backend

umar456:


CMakeModules/InternalUtils.cmake
```diff
@@ -36,6 +36,15 @@ function(arrayfire_get_cuda_cxx_flags cuda_flags)
   endif()
 endfunction()
 
+function(arrayfire_get_opencl_cxx_flags opencl_flags)
+  include(CheckCXXCompilerFlag)
+  check_cxx_compiler_flag(-Wno-ignored-attributes has_ignored_attributes_flag)
+
+  if(has_ignored_attributes_flag)
```
9prady9: I think these checks are better included directly in the `src/backend/opencl/CMakeLists.txt` file. That would give a better perspective of what are the flags without having to move to a new file altogether. Especially, when these are not a common set of flags for all backends.

Issue #2084, opened 2018-03-13
cuda::compileToPTX()  NVRTC_ERROR_COMPILATION


WilliamTambellini:
Hi
cuda::compileToPTX fails with my program using the current AF master at least the one from yesterday:
8c34aa999b4b092e51c603360b7ed203d2453955
 
In function std::vector<char> cuda::compileToPTX(const char*, std::string)
In file src/backend/cuda/jit.cpp:291
NVRTC Error(6): NVRTC_ERROR_COMPILATION
In function af::array af::moddims(const af::array&, unsigned int, const dim_t*)
In file src/api/cpp/data.cpp:192

ArrayFire v3.6.0 (CUDA, 64-bit Linux, build b7bd543)
Platform: CUDA Toolkit 8, Driver: 375.74
[0] GeForce GTX 1060, 6073 MB, CUDA Compute 6.1
NVIDIA-SMI 375.74                 Driver Version: 375.74  

AF_PRINT_ERRORS=1
does not seem to output more details.

I can hardly atm send any source code to reproduce, as it happens when running our production software. 

Is there any way at least for AF to print the faulty buggy kernel source failing to compile ?
Kind

WilliamTambellini:
Ok so we debugged with @umar456  and looks like AF is generating a too long kernel : 
KER17250219459061981991(228): Error: Formal parameter space overflowed (4096 bytes max) in function KER17250219459061981991
Lowering AF_CUDA_MAX_JIT_LEN is of course preventing that to happen but that s just a workaround.
The best would be for AF to detect this 4KB limit and only generating kernels within that limit.

Strangely table 14 on this page
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
is reporting a  "Cache working set per multiprocessor for constant memory" of 4KB for Cuda Compute 6 but my setup is 6.1 so the limit should be 8KB...

To be continued.

PR #2127, opened 2018-04-17
Fixes for parameter overflow for NVIDIA devices.

pavanky:
- Ensure the kernels are compiled quicker for non linear kernels.

- Fixes https://github.com/arrayfire/arrayfire/issues/2084

### DIFF
--- a/src/backend/cuda/Array.cpp
+++ b/src/backend/cuda/Array.cpp
@@ -116,7 +116,7 @@ namespace cuda
     void Array<T>::eval()
     {
         if (isReady()) return;
- 
+
         this->setId(getActiveDeviceId());
         this->data = shared_ptr<T>(memAlloc<T>(elements()).release(), memFree<T>);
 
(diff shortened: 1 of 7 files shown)


src/backend/opencl/Array.cpp
```diff
@@ -228,30 +228,66 @@ namespace opencl
...
+                bool isNvidia = getActivePlatform() == AFCL_PLATFORM_NVIDIA;
+                // We eval in the following cases.
+                // 1. Too many bytes are locked up by JIT causing memory pressure.
+                // Too many bytes is assumed to be half of all bytes allocated so far.
+                // 2. Too many buffers in a nonlinear kernel cause param space overflow.
+                // Too many buffers comes out to be about 48 (49 including output).
+                // Too many buffers can occur in a tree of size 24 in the worst case scenario.
+                // This error only happens on nvidia devices.
+                // TODO: Find better solution than the following emperical solution.
+                bool isParamLimit = (isNvidia && node->getHeight() > 24);
```
umar456: Shouldn't this be based on getMaxJitSize()?

src/backend/opencl/Array.cpp
```diff
@@ -228,30 +228,66 @@ namespace opencl
...
                     for(auto &jit_node : full_nodes) {
                         jit_node->getInfo(length, buf_count, bytes);
+                        is_linear &= jit_node->isLinear(dims_);
                     }
 
-                    if (2 * bytes > lock_bytes) {
+                    // Clear the vectors so you dont hold on to any shared pointers
+                    nodes_map.clear();
+                    full_nodes.clear();
+                    full_ids.clear();
```
umar456: Aren't you just referencing pointers with these variables? You shouldn't need to do this.

test/jit.cpp
```diff
@@ -403,3 +403,24 @@ TEST(JIT, LinearLarge)
...
+    for (int i = 0; i < 1000; i++) {
+        af::array b = af::randu(1, 5);
+        a += af::tile(b, 5);
+    }
+    a.eval();
+}
+
+TEST(JIT, NonLinearBuffers2)
+{
+    af::array a = af::randu(1000, 3100);
```
umar456: The second dimension can probably be smaller. It should be based on the number of nodes you want to test. I would resize it so something like:

```
TEST(JIT, NonLinearBuffers2)
{
    af::array a = af::randu(100, 310);
    af::array b = af::randu(10, 10);
    for (int i = 0; i < 300; i++) {
        b += a(seq(10), seq(i, i+9)) * randu(10, 10);
    }
    b.eval();
}
```

src/backend/cuda/Array.cpp
```diff
@@ -209,28 +209,57 @@ namespace cuda
...
                 size_t lock_bytes, lock_buffers;
 
                 deviceMemoryInfo(&alloc_bytes, &alloc_buffers,
                                  &lock_bytes, &lock_buffers);
 
-                // Check if approaching the memory limit
-                if (lock_bytes > getMaxBytes() ||
-                    lock_buffers > getMaxBuffers()) {
+                bool isBufferLimit =
+                    lock_bytes > getMaxBytes() ||
```
umar456: You can use [checkMemoryLimit()](https://github.com/arrayfire/arrayfire/blob/master/src/backend/common/MemoryManager.hpp#L432) for this.

src/backend/opencl/Array.cpp
pavanky: No. This change is trying to have different thresholds for Linear and Non Linear kernels. For Linear kernels it is still 100. For non linear it is 24.

src/backend/opencl/Array.cpp
pavanky: `Node_ptr` may contain a shared pointer.

src/backend/cuda/Array.cpp
pavanky: Will do.

src/backend/opencl/Array.cpp
umar456: Now that I am looking at the getNodesMap function you might still need this because you are doing a push_back in there. I think you just need to update the comment.

src/backend/cuda/Array.cpp
pavanky: Can't do this because lock_bytes is used elsewhere. Fixed the other issues that were raised.

FloopCZ:
The CPU JIT bug was a good catch. I just found out that this code produced wrong results on CPU backend:

```c++
#include <arrayfire.h>
int main() {
    double data_[] = {1., 2., 3., 4.};
    af::array a{2, 2, data_};
    af::array b{2, 2, data_};

    a + a;  // remove this line and everything is fine
    af::print("difference", af::flat(a) - af::flat(b));  // nonzero difference!
}
```

And while browsing through the arrayfire code, I saw this commit that fixed the issue for me. Looking forward to a new release (or backport :-) ).

PR #2139, opened 2018-04-25
Fix infinite loop in CMake when configuring with vcpkg

umar456:
There was an infinite loop when building with Microsoft's VCPKG
package manager that was caused because the ArrayFire examples
and vcpkg overloaded the find_package function. This caused a
stack overflow during the CMake configuration step.


pavanky:
> This caused a stack overflow during the CMake configuration step.

Ok that is just funny.

umar456:
This doesn't fully address the segfault. We need to think of another way to get around this issue.

PR #2141, opened 2018-04-27
Fix cmake crash with vcpkg package manager

### LINKED ISSUE #2140: CMake crashes with vcpkg toolchain file
We've been experimenting with [vcpkg](https://github.com/Microsoft/vcpkg) in order to improve the build experience on Windows. To find installed dependencies, the `CMAKE_TOOLCHAIN_FILE` variable needs to point to vcpkg's toolchain file, `$VCPKG/scripts/buildsystems/vcpkg.cmake`.

However, setting `-DCMAKE_TOOLCHAIN_FILE` when building with ArrayFire examples causes CMake to crash due to a stack overflow.

While #2139 prevents the stack overflow, the VS solution file necessary to build ArrayFire is not being generated. The solution, for now, is to set the toolchain file but disable the ArrayFire examples via `-DAF_BUILD_EXAMPLES=OFF`.

9prady9:
Fixes #2140 

The tests are still named as before, only the cmake targets are prefixed with `test_` string.


9prady9:
@umar456 The failure is from canny_cpu which randomly fails on OSX, it is not related to this PR(I am following up on that issue separately). This is ready for merge.

test/CMakeLists.txt
```diff
@@ -118,9 +118,10 @@ function(make_test)
           CXX_STANDARD 11)
     endif(${mt_args_CXX11})
 
-	  set_target_properties(${target}
-      PROPERTIES
-        FOLDER "Tests")
+    set_target_properties(${target} PROPERTIES FOLDER "Tests")
```
umar456: You should combine these into one set_target_properties

9prady9: Addressed

PR #2150, opened 2018-05-09
Disable hunter use by forge from ArrayFire

9prady9:
This would result in developer to isntall forge dependencies
manually.
The following are the dependencies required:
* glbinding
* glfw3
* glm
* boost(headers only, not need of libs) - ArrayFire also has this dependency on couple of backends.
* fontconfig(on non-Windows platforms) - freetype is installed by this dependency and thus doesn't need to be installed explicitly.
* freetype


FloopCZ:
Thanks @9prady9. Note that after merge, [this page](https://github.com/arrayfire/arrayfire/wiki/Build-Instructions-for-Linux) should be updated not to claim that the graphical dependencies are checked out automatically.

9prady9:
This is breaking builds on our ci which uses `debian 8`. Debian 8 has `libglm-dev 0.9.5.4` available through package manager which is four years old and it doesn't provide glm config file(required by CMake find_package command).

9prady9:
@umar456 I think this is ready, the failures are unrelated to the change. I am looking into the sparse failures and will update https://github.com/arrayfire/arrayfire/issues/2154

PR #2152, opened 2018-05-10
Fixes in CUDA/OpenCL regions

### LINKED ISSUE #1650: Incorrect output for regions() with CUDA if full component
Hello.

I just wanted to report a bug I encountered with the function regions().

In the case of a binary image fully composed of 1, the resulting array is only composed of 0, while it should be entirely composed of 1.

The bug disappeared if I used `-lafcpu` instead of `-lafcuda`.

Here is the minimal reproducible example:
```c
#include "arrayfire.h"
#include "af/cuda.h"

int main() {
    unsigned char in[8*8];
    for (int i = 0; i < 8*8; i++) in[i] = 1;

    af::array arr = af::array(8, 8, in);
    af::array reg = regions(arr.as(b8), AF_CONNECTIVITY_4);

    af_print(reg);
    return 0;
}
```

9prady9:
* Fixes #1650 
* Also, removes couple of redundant checks in regions backend files.

### DIFF
--- a/test/regions.cpp
+++ b/test/regions.cpp
@@ -250,3 +250,20 @@ TEST(Regions, Docs_4)
         ASSERT_EQ(gold[i], output[i])<<" mismatch at i="<<i<<std::endl;
     }
 }
+
+TEST(Regions, Issue1650)
+{
+    const int dim = 101;
+    const int sz  = dim*dim;
+    std::vector<char> input(sz, 1);
+    std::vector<float> gold(sz, 1.0f);
+
+    af::array in  = af::array(dim, dim, input.data());
+    af::array out = af::regions(in, AF_CONNECTIVITY_4);
+
+    std::vector<float> output(sz);
+    out.host((void*)output.data());
+
+    for (int i=0; i<sz; ++i)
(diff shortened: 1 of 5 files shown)


9prady9:
@umar456 I think this is ready, the failures are unrelated to the change. I am looking into the sparse failures and will update #2154

test/regions.cpp
```diff
@@ -250,3 +250,39 @@ TEST(Regions, Docs_4)
         ASSERT_EQ(gold[i], output[i])<<" mismatch at i="<<i<<std::endl;
     }
 }
+
+TEST(Regions, Issue1650)
+{
+    // input data
+    char input[64] =  {
+        1, 1, 1, 1, 1, 1, 1, 1,
```
umar456: You should probably make a test that would test the multiple block scenario. You can make a vector like this `vector<char> input(64 * 64, 1);` instead of setting the values for each element.

9prady9: By multiple block scenario, do you many components ?

9prady9: I think I got what you meant, dims spanning multiple blocks.

PR #2153, opened 2018-05-10
Fix checks in af::seq() & indexing/assignment functions 

### LINKED ISSUE #1452: NVidia driver crash on reorder with bad sequences.
The code:

```
auto testmat = af::constant(5, 20, 20, 20, 1);
testmat = af::reorder(testmat(af::seq(15,-15), af::seq(15,-15)), 3, 2, 0, 1);
```

Produces a crash on NVidia with OpenCL on driver 368.22 on a GTX 970.

9prady9:
Fixes #1452

### DIFF
--- a/src/api/c/array.cpp
+++ b/src/api/c/array.cpp
@@ -76,36 +76,21 @@ af_err af_create_array(af_array *result, const void * const data,
 }
 
 //Strong Exception Guarantee
-af_err af_create_handle(af_array *result, const unsigned ndims, const dim_t * const dims,
+af_err af_create_handle(af_array *result,
+                        const unsigned ndims, const dim_t * const dims,
                         const af_dtype type)
 {
     try {
-        af_array out = 0;
         AF_CHECK(af_init());
 
-        if (ndims > 0) {
-            ARG_ASSERT(2, ndims > 0 && dims != NULL);
-        }
+        if (ndims > 0) ARG_ASSERT(2, ndims > 0 && dims != NULL);
+
         dim4 d(0);
         for(unsigned i = 0; i < ndims; i++) {
             d[i] = dims[i];
         }
 
-        switch(type) {
-        case f32:   out = createHandle<float  >(d); break;
-        case c32:   out = createHandle<cfloat >(d); break;
-        case f64:   out = createHandle<double >(d); break;
-        case c64:   out = createHandle<cdouble>(d); break;
-        case b8:    out = createHandle<char   >(d); break;
-        case s32:   out = createHandle<int    >(d); break;
-        case u32:   out = createHandle<uint   >(d); break;
-        case u8:    out = createHandle<uchar  >(d); break;
-        case s64:   out = createHandle<intl   >(d); break;
-        case u64:   out = createHandle<uintl  >(d); break;
-        case s16:   out = createHandle<short  >(d); break;
-        case u16:   out = createHandle<ushort >(d); break;
-        default:    TYPE_ERROR(3, type);
-        }
+        af_array out = createHandle(d, type);
         std::swap(*result, out);
     }
     CATCHALL
--- a/src/api/c/assign.cpp
+++ b/src/api/c/assign.cpp
@@ -20,95 +20,92 @@
 #include <assign.hpp>
 #include <math.hpp>
 #include <tile.hpp>
+#include <indexing_common.hpp>
 
 using namespace detail;
 using std::vector;
 using std::swap;
+using std::signbit;
+using common::convert2Canonical;
+using common::createSpanIndex;
 
 template<typename Tout, typename Tin>
 static
-void assign(Array<Tout> &out, const unsigned &ndims, const af_seq *index, const Array<Tin> &in_)
+void assign(Array<Tout> &out, const vector<af_seq> seqs,
+            const Array<Tin> &in)
 {
-    dim4 const outDs = out.dims();
-    dim4 const iDims = in_.dims();
+    size_t ndims  = seqs.size();
+    const dim4& outDs = out.dims();
+    const dim4& iDims = in.dims();
 
-    // Nothing to do for empty arrays
     if (iDims.elements() == 0) return;
-
-    DIM_ASSERT(0, (outDs.ndims()>=iDims.ndims()));
-    DIM_ASSERT(0, (outDs.ndims()>=(dim_t)ndims));
 
     out.eval();
 
-    vector<af_seq> index_(index, index+ndims);
+    dim4 oDims = toDims(seqs, outDs);
 
-    dim4 oDims = toDims(index_, outDs);
-
-    bool is_vector = true;
-    for (int i = 0; is_vector && i < (int)oDims.ndims() - 1; i++) {
-        is_vector &= oDims[i] == 1;
+    bool isVec = true;
+    for (int i = 0; isVec && i < (int)oDims.ndims() - 1; i++) {
+        isVec &= oDims[i] == 1;
     }
 
-    is_vector &= in_.isVector() || in_.isScalar();
+    isVec &= in.isVector() || in.isScalar();
 
-    for (dim_t i = ndims; i < (int)in_.ndims(); i++) {
+    for (dim_t i = ndims; i < (int)in.ndims(); i++) {
         oDims[i] = 1;
     }
 
-
-    if (is_vector) {
-        if (oDims.elements() != (dim_t)in_.elements() &&
-            in_.elements() != 1) {
+    if (isVec) {
+        if (oDims.elements() != (dim_t)in.elements() &&
+            in.elements() != 1) {
             AF_ERROR("Size mismatch between input and output", AF_ERR_SIZE);
         }
 
-        // If both out and in are vectors of equal elements, reshape in to out dims
-        Array<Tin> in = in_.elements() == 1 ? tile(in_, oDims) : modDims(in_, oDims);
-        Array<Tout> dst = createSubArray<Tout>(out, index_, false);
-
-        copyArray<Tin , Tout>(dst, in);
-    } else {
-        for (int i = 0; i < 4; i++) {
-            if (oDims[i] != iDims[i]) {
-                AF_ERROR("Size mismatch between input and output", AF_ERR_SIZE);
-            }
-        }
-        Array<Tout> dst = createSubArray<Tout>(out, index_, false);
+        // If both out and in are vectors of equal elements,
+        // reshape in to out dims
+        Array<Tin> in_ = in.elements() == 1 ? tile(in, oDims)
+                                           : modDims(in, oDims);
+        auto dst = createSubArray<Tout>(out, seqs, false);
 
         copyArray<Tin , Tout>(dst, in_);
+    } else {
+        for (int i = 0; i < AF_MAX_DIMS; i++) {
+            if (oDims[i] != iDims[i])
+                AF_ERROR("Size mismatch between input and output", AF_ERR_SIZE);
+        }
+        Array<Tout> dst = createSubArray<Tout>(out, seqs, false);
+
+        copyArray<Tin , Tout>(dst, in);
     }
 }
 
 template<typename T>
 static
-void assign_helper(Array<T> &out, const unsigned &ndims, const af_seq *index, const af_array &in_)
+void assign(Array<T> &out, const vector<af_seq> iv,
+            const af_array &in)
 {
-    const ArrayInfo& iInfo = getInfo(in_);
-    af_dtype iType  = iInfo.getType();
+    const ArrayInfo& iInfo = getInfo(in);
+    af_dtype iType = iInfo.getType();
 
-    if(out.getType() == c64 || out.getType() == c32)
-    {
-
+    if(out.getType() == c64 || out.getType() == c32) {
         switch(iType) {
-            case c64: assign<T, cdouble>(out, ndims, index, getArray<cdouble  >(in_));  break;
-            case c32: assign<T, cfloat >(out, ndims, index, getArray<cfloat   >(in_));  break;
-            default : TYPE_ERROR(1, iType); break;
+        case c64: assign<T, cdouble>(out, iv, getArray<cdouble>(in));  break;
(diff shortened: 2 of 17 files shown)


src/api/cpp/seq.cpp
```diff
@@ -71,7 +71,8 @@ seq::seq(double begin, double end, double step): m_gfor(false)
     }
     if (end >= 0 && begin >= 0 && signbit(end-begin) != signbit(step))
         AF_THROW_ERR("Sequence is invalid", AF_ERR_ARG);
-        //AF_THROW("step must match direction of sequence");
+    if (begin>0 && end<0 && step>0)
```
FloopCZ: 1) Shouldn't we also catch `af::seq(0, -15)`? i.e., `begin >= 0` instead of `begin > 0`.
2) Shouldn't we also catch `af::seq(-3, -15)`? i.e., drop `begin > 0` whatsoever?
2) This leads to the fact that negative indices are currently somehow special:
```
    af::seq(-2, -1, 1);   // passes, {-2, -1} when printed
    af::seq(-2, -1, -1);  // passes, crash when printed
    af::seq(2, 1, 1);     // does not pass
    af::seq(2, 1, -1);    // passes, {2, 1} when printed
```
Is this intentional? My question is, cannot we simply solve all the situations by removing `end >= 0 && begin >= 0` from the first condition or does it have a hidden purpose?

src/api/cpp/seq.cpp
```diff
@@ -71,7 +71,10 @@ seq::seq(double begin, double end, double step): m_gfor(false)
     }
     if (end >= 0 && begin >= 0 && signbit(end-begin) != signbit(step))
         AF_THROW_ERR("Sequence is invalid", AF_ERR_ARG);
-        //AF_THROW("step must match direction of sequence");
+    if (step>0 && !(begin<end))
```
FloopCZ: This should probably say:
```c++
    if (step > 0 && begin > end)
```
Because for step 1, it is correct when `begin == end`. Still though, this will fail for e.g., `seq(2, af::end)` aka `seq(2, -1, 1)`.

Whether the sequence is a valid index depends on the shape of the array it is indexing. In other words, we have no idea whether index 15 is less than index `af::end - 10` aka `-11`. I believe only the signs can be checked during the construction of the sequence. E.g., something like:
```c++
    if (signbit(end) == signbit(begin) && signbit(end-begin) != signbit(step))
        AF_THROW_ERR("Sequence is invalid", AF_ERR_ARG);
    if (end < 0 && begin > 0 && step < 0)
        AF_THROW_ERR("Sequence is invalid", AF_ERR_ARG);
    if (end > 0 && begin < 0 && step > 0)
        AF_THROW_ERR("Sequence is invalid", AF_ERR_ARG);
```

test/index.cpp
```diff
@@ -1590,3 +1590,8 @@ TEST(Index, Issue1867ChainedIndexingLeak)
                       &lock_bytes, &lock_buffers);
     ASSERT_EQ(0u, lock_buffers);
 }
+
+TEST(Index, Issue1452)
+{
+    EXPECT_THROW(af::seq(15,-15), af::exception);
```
FloopCZ: I believe `af::seq(15, -15)` is a valid sequence. Isn't it just `af::seq(15, af::end-14)` ?

If I am not missing something (and that is a big if! :-D ), than only the signs can be tested. E.g.,

```c++
    af::seq( 1,  1,  1);
    af::seq( 1,  5,  1);
    af::seq( 1, -5,  1);
    af::seq(-1,  5, -1);
    af::seq(-1, -5, -1);

    EXPECT_THROW(af::seq( 1,  1, -1), af::exception);
    EXPECT_THROW(af::seq( 1,  5, -1), af::exception);
    EXPECT_THROW(af::seq( 1, -5, -1), af::exception);
    EXPECT_THROW(af::seq(-1,  5,  1), af::exception);
    EXPECT_THROW(af::seq(-1, -5,  1), af::exception);
```

9prady9:
I haven't finished my changes yet.

9prady9:
@umar456 Feedback please.

src/api/c/moddims.cpp
```diff
@@ -91,21 +94,61 @@ af_err af_moddims(af_array *out, const af_array in,
...
+template Array<uint   > flatten(const Array<uint   >&);
+template Array<intl   > flatten(const Array<intl   >&);
+template Array<uintl  > flatten(const Array<uintl  >&);
+template Array<short  > flatten(const Array<short  >&);
+template Array<ushort > flatten(const Array<ushort >&);
+template Array<uchar  > flatten(const Array<uchar  >&);
+template Array<char   > flatten(const Array<char   >&);
+
+template<typename T>
+inline af_array flatten(const af_array in)
```
umar456: Add these functions in anonymous namespace to prevent their use outside this file.

src/api/c/indexing_common.hpp
```diff
@@ -0,0 +1,16 @@
+/*******************************************************
```
umar456: No need to make a separate file for these functions. Just add them to index.hpp

src/api/c/assign.cpp
```diff
@@ -20,95 +20,94 @@
...
-
-    dim4 oDims = toDims(index_, outDs);
+    dim4 oDims = toDims(seqs, outDs);
 
-    bool is_vector = true;
-    for (int i = 0; is_vector && i < (int)oDims.ndims() - 1; i++) {
-        is_vector &= oDims[i] == 1;
-    }
+    bool isVec = true;
+    for (int i = 0; isVec && i < (int)oDims.ndims() - 1; i++)
```
umar456: Braces

src/api/c/assign.cpp
```diff
@@ -277,63 +265,67 @@ af_err af_assign_gen(af_array *out,
...
-                af_dtype idxType  = idxInfo.getType();
-
-                ARG_ASSERT(3, (idxType!=c32));
-                ARG_ASSERT(3, (idxType!=c64));
-                ARG_ASSERT(3, (idxType!=b8 ));
-
-                idxrs[i].idx.arr = indexs[i].idx.arr;
-                idxrs[i].isSeq = indexs[i].isSeq;
+
+        for (dim_t i=0; i<4; ++i) {
```
umar456: Use AF_MAX_DIMS

src/api/c/indexing_common.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#include <cmath>
+#include <mutex>
+#include <common/err_common.hpp>
+#include <indexing_common.hpp>
+
+af_index_t spanner()
```
umar456: spanner sounds like an action. This is just returning a span af_index_t object. This should really be a constructor of the af_index_t object or a factory function.

src/api/c/indexing_common.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    static af_index_t s;
+    static std::once_flag flag;
+    std::call_once(flag, []() {
```
umar456: Kind of complicated way of doing this. Static objects are already thread safe. If we create a factory function that creates these objects, this call_once will go away.

test/index.cpp
```diff
@@ -1591,3 +1591,17 @@ TEST(Index, Issue1867ChainedIndexingLeak)
                       &lock_bytes, &lock_buffers);
     ASSERT_EQ(0u, lock_buffers);
 }
+
+TEST(Index, InvalidSequence0)
```
umar456: InvalidSequence0,1,2 are bad names. Perhaps should be InvalidNegativeStep would be more informative.

src/api/c/assign.cpp
```diff
@@ -20,95 +20,94 @@
...
+void assign(Array<Tout> &out, const vector<af_seq> indicesV,
+            const Array<Tin> &in)
 {
-    dim4 const outDs = out.dims();
-    dim4 const iDims = in_.dims();
+    auto ndims  = indicesV.size();
+    auto& outDs = out.dims();
+    auto& iDims = in.dims();
+
+    vector<af_seq> seqs;
```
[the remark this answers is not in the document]
umar456: std::array would be even better.

src/api/c/index.cpp
```diff
@@ -147,74 +159,73 @@ af_array genIndex(const af_array& in,  const af_index_t idxrs[])
...
+            af_array in_ = 0;
+            AF_CHECK(af_flat(&in_, in));
+            AF_CHECK(af_index_gen(out, in_, ndims, indexs));
+            AF_CHECK(af_release_array(in_));
             return AF_SUCCESS;
         }
 
         int track = 0;
-        af_seq seqs[] = {af_span, af_span, af_span, af_span};
+        vector<af_seq> seqs(4, af_span);
```
umar456: std::array

src/api/c/indexing_common.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+af_seq conv2Canonical(const af_seq s, const dim_t len)
+{
+#ifndef OS_WIN
+    using std::signbit;
+#endif
+    double begin = signbit(s.begin) ? (len+s.begin) : s.begin;
+    double end   = signbit(s.end  ) ? (len+s.end)   :   s.end;
+    if (begin<0 || end <0)
+        AF_ERROR("Sequence is invalid", AF_ERR_ARG);
```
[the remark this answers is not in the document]
umar456: We need something like std::optional for this. There is probably something in boost that we could use.

src/api/c/indexing_common.cpp
9prady9: This is not about factory based implementation or otherwise. I wanted to do the following. 
```
static const af_index_t s { {.seq=af_span}, true, false };
```
I did and it works fine with GCC, but on MSVC this is failing because using VC compiler we can only assign to the first element of the union `idx` inside the struct `af_index_t`. Although it is supported in C++11, MSVC screwed it up. Therefore, I had to split the declaration and initialization which is where call_once came in.

test/index.cpp
9prady9: Not sure where you are getting with this. Next test case also has negative step, should we add another word to differentiate from the first one ? That way, we will have lengthy names and you might say not to that again.

src/api/c/indexing_common.cpp
9prady9: Check my response to your comment near `call_once` usage. Add an class(constructor) just for this case ? I don't think it is needed. I am all ears for a new name for `spanner`, I tried to think of one but didn't find any.

src/api/c/assign.cpp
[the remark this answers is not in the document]
9prady9: not possible in couple of cases, like this one where this seqs is getting passed down to toDims function defined in Arrayinfo.hpp that takes vector.

test/index.cpp
umar456: The idea is that if a test fails we need to be able to figure out why it failed without looking at the code. If the name is descriptive enough we can see a pattern and not have to go digging in the test to figure out what is going on. Long names are fine

src/api/c/assign.cpp
```diff
@@ -277,94 +270,104 @@ af_err af_assign_gen(af_array *out,
...
+
+                    idxrs[i] = { { indexs[i].idx.arr },
+                                 isSeq, indexs[i].isBatch };
+                } else {
+                    af_seq s = af_span;
+                    af_err canonicalConversion = AF_SUCCESS;
+                    canonicalConversion = convert2Canonical(s,
+                                              indexs[i].idx.seq,
+                                              lhsDims[i]);
+                    ARG_ASSERT(4, AF_SUCCESS==canonicalConversion);
```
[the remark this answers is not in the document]
9prady9: What's the alternative you have on your mind? I was thinking `AF_SUCCESS==canonicalConversion` would suggest some sequence conversion failed ?

src/backend/cpu/index.cpp
```diff
@@ -42,7 +42,7 @@ Array<T> index(const Array<T>& in, const af_index_t idxrs[])
     // retrieve
     dim4 oDims = toDims(seqs, in.dims());
 
-    vector< Array<uint> > idxArrs(4, createEmptyArray<uint>(dim4()));
+    vector< Array<uint> > idxArrs(4, *initArray<uint>());
```
umar456: I think I made this change but its buggy. please update it back.

9prady9: @umar456 Isn't this wrong based on your lastest leak fix?

src/api/c/assign.cpp
```diff
@@ -277,94 +274,105 @@ af_err af_assign_gen(af_array *out,
...
+
+                    ARG_ASSERT(3, (idxType!=c32));
+                    ARG_ASSERT(3, (idxType!=c64));
+                    ARG_ASSERT(3, (idxType!=b8 ));
+
+                    idxrs[i] = { { indexs[i].idx.arr },
+                                 isSeq, indexs[i].isBatch };
+                } else {
+                    af_seq inSeq = convert2Canonical(indexs[i].idx.seq, lhsDims[i]);
+                    ARG_ASSERT(4, (inSeq.begin >= 0 || inSeq.end >= 0));
```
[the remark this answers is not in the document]
9prady9: Similar checks are done in two locations. You haven't changed in the second instance.

src/api/c/array.cpp
```diff
@@ -76,36 +76,21 @@ af_err af_create_array(af_array *result, const void * const data,
...
                         const af_dtype type)
 {
     try {
-        af_array out = 0;
         AF_CHECK(af_init());
 
-        if (ndims > 0) {
-            ARG_ASSERT(2, ndims > 0 && dims != NULL);
-        }
+        if (ndims > 0) ARG_ASSERT(2, ndims > 0 && dims != NULL);
```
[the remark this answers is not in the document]
9prady9: I think you are mistaking this for `dim`. ndims is unsigned like you pointed out and is checking empty arrays, based on this file's history - here's what you wrote when that condition was added.
```
Empty arrays are 0, 1, 1, 1. Update isVector and tests. 
```

PR #2155, opened 2018-05-14
Fix #2130 and handle NaNs for complex types

### LINKED ISSUE #2130: Bug in min with NAN values
The following code
```
float data[] = { 1, NAN, 5, 0.1, NAN, -0.5, NAN, 0 };

af::array A( 4, 2, data );
af::print( "A:", A );

af::array min_val;
af::array min_idx;
af::min( min_val, min_idx, A );

af::array min = af::min( A );

af::print( "min_val:", min_val );
af::print( "min_idx:", min_idx );
af::print( "min:", min );
```

produces the following output for me (ArrayFire 3.5.1, Linux):
```
A:
[4 2 1 1]
    1.0000        nan 
       nan    -0.5000 
    5.0000        nan 
    0.1000     0.0000 

min_val:
[1 2 1 1]
    1.0000        nan 
min_idx:
[1 2 1 1]
         0          0 
min:
[1 2 1 1]
    0.1000    -0.5000
```

Obviously, the result of `af::min` for getting minimum values and their locations in an array is wrong. Whereas, the result of `af::min` just returning the minimum values is correct.
The documentation claims to ignore NaN values for both variants: http://arrayfire.org/docs/group__reduce__func__min.htm#ga4f4b4607803d8fd4fea8e862d3685d15

The problem does _not_ occur if _no NaN value_ is present in the array.

I would also suggest adding a new test covering this case.

mark-poscablo:
Resolves #2130 and also tries to cover the case for complex types as well. Currently, CUDA backend is still being worked on for the complex types.

The essence of the fixes are:
- Guard against the next examined value being a NaN (all backends)
- If the initial min/max value is a NaN (first element of array in CPU, first element that a thread picks up in CUDA and OpenCL) , set it instead to `Binary<T, op>::init()`
- Element index order checks are removed from CUDA backend, since they only make sense to use in a serial reduction (CPU), not in parallel
- The issue only concerns indexed min/max, but OpenCL's non-indexed min/max also had incorrect results also for complex types. Adding the guard fixes it

### DIFF
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -31,8 +31,7 @@ struct MinMaxOp
         m_val(val), m_idx(idx)
     {
         if (isNan(val)) {
-            Binary<T, op> ireduce;
-            m_val = ireduce.init();
+            m_val = Binary<T, op>::init();
         }
     }
 
@@ -57,8 +56,7 @@ struct MinMaxOp<af_max_t, T>
         m_val(val), m_idx(idx)
     {
         if (isNan(val)) {
-            Binary<T, af_max_t> ireduce;
-            m_val = ireduce.init();
+            m_val = Binary<T, af_max_t>::init();
         }
     }
 
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -39,7 +39,7 @@ struct MinMaxOp
     {
         if ((cabs(val) < cabs(m_val) ||
              (cabs(val) == cabs(m_val) &&
-              idx >= m_idx)) &&
+              idx > m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
@@ -64,7 +64,7 @@ struct MinMaxOp<af_max_t, T>
     {
         if ((cabs(val) > cabs(m_val) ||
              (cabs(val) == cabs(m_val) &&
-              idx >= m_idx)) &&
+              idx <= m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -9,6 +9,7 @@
 
 #pragma once
 #include <Param.hpp>
+#include <ops.hpp>
 
 namespace cpu
 {
@@ -19,6 +20,7 @@ template<typename T> double cabs(const T in) { return (double)in; }
 static double cabs(const char in) { return (double)(in > 0); }
 static double cabs(const cfloat &in) { return (double)abs(in); }
 static double cabs(const cdouble &in) { return (double)abs(in); }
+template<typename T> static bool isNan(T in) { return in != in; }
 
 template<af_op_t op, typename T>
 struct MinMaxOp
@@ -28,13 +30,18 @@ struct MinMaxOp
     MinMaxOp(T val, uint idx) :
         m_val(val), m_idx(idx)
     {
+        if (isNan(val)) {
+            Binary<T, op> ireduce;
+            m_val = ireduce.init();
+        }
     }
 
     void operator()(T val, uint idx)
     {
-        if (cabs(val) < cabs(m_val) ||
-            (cabs(val) == cabs(m_val) &&
-             idx > m_idx)) {
+        if (!isNan(val) &&
+            (cabs(val) < cabs(m_val) ||
+            cabs(val) == cabs(m_val)) &&
+            idx >= m_idx) {
             m_val = val;
             m_idx = idx;
         }
@@ -49,13 +56,18 @@ struct MinMaxOp<af_max_t, T>
     MinMaxOp(T val, uint idx) :
         m_val(val), m_idx(idx)
     {
+        if (isNan(val)) {
+            Binary<T, af_max_t> ireduce;
+            m_val = ireduce.init();
+        }
     }
 
     void operator()(T val, uint idx)
     {
-        if (cabs(val) > cabs(m_val) ||
-            (cabs(val) == cabs(m_val) &&
-             idx <= m_idx)) {
+        if (!isNan(val) &&
+            (cabs(val) > cabs(m_val) ||
+            cabs(val) == cabs(m_val)) &&
+            idx >= m_idx) {
             m_val = val;
             m_idx = idx;
         }
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -38,8 +38,8 @@ struct MinMaxOp
     void operator()(T val, uint idx)
     {
         if ((cabs(val) < cabs(m_val) ||
-            cabs(val) == cabs(m_val)) &&
-            idx >= m_idx &&
+             (cabs(val) == cabs(m_val) &&
+              idx >= m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
@@ -63,8 +63,8 @@ struct MinMaxOp<af_max_t, T>
     void operator()(T val, uint idx)
     {
         if ((cabs(val) > cabs(m_val) ||
-            cabs(val) == cabs(m_val)) &&
-            idx >= m_idx &&
+             (cabs(val) == cabs(m_val) &&
+              idx >= m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -38,8 +38,7 @@ struct MinMaxOp
     void operator()(T val, uint idx)
     {
         if ((cabs(val) < cabs(m_val) ||
-             (cabs(val) == cabs(m_val) &&
-              idx > m_idx)) &&
+             (cabs(val) == cabs(m_val) && idx > m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
@@ -63,8 +62,7 @@ struct MinMaxOp<af_max_t, T>
     void operator()(T val, uint idx)
     {
         if ((cabs(val) > cabs(m_val) ||
-             (cabs(val) == cabs(m_val) &&
-              idx <= m_idx)) &&
+             (cabs(val) == cabs(m_val) && idx <= m_idx)) &&
             !is_nan(val)) {
             m_val = val;
             m_idx = idx;
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -20,7 +20,7 @@ template<typename T> double cabs(const T in) { return (double)in; }
 static double cabs(const char in) { return (double)(in > 0); }
 static double cabs(const cfloat &in) { return (double)abs(in); }
 static double cabs(const cdouble &in) { return (double)abs(in); }
-template<typename T> static bool isNan(T in) { return in != in; }
+template<typename T> static bool is_nan(T in) { return in != in; }
 
 template<af_op_t op, typename T>
 struct MinMaxOp
@@ -30,17 +30,17 @@ struct MinMaxOp
     MinMaxOp(T val, uint idx) :
         m_val(val), m_idx(idx)
     {
-        if (isNan(val)) {
+        if (is_nan(val)) {
             m_val = Binary<T, op>::init();
         }
     }
 
     void operator()(T val, uint idx)
     {
-        if (!isNan(val) &&
-            (cabs(val) < cabs(m_val) ||
+        if ((cabs(val) < cabs(m_val) ||
             cabs(val) == cabs(m_val)) &&
-            idx >= m_idx) {
+            idx >= m_idx &&
+            !is_nan(val)) {
             m_val = val;
             m_idx = idx;
         }
@@ -55,17 +55,17 @@ struct MinMaxOp<af_max_t, T>
     MinMaxOp(T val, uint idx) :
         m_val(val), m_idx(idx)
     {
-        if (isNan(val)) {
+        if (is_nan(val)) {
             m_val = Binary<T, af_max_t>::init();
         }
     }
 
     void operator()(T val, uint idx)
     {
-        if (!isNan(val) &&
-            (cabs(val) > cabs(m_val) ||
+        if ((cabs(val) > cabs(m_val) ||
             cabs(val) == cabs(m_val)) &&
-            idx >= m_idx) {
+            idx >= m_idx &&
+            !is_nan(val)) {
             m_val = val;
             m_idx = idx;
         }
--- a/src/backend/cpu/kernel/ireduce.hpp
+++ b/src/backend/cpu/kernel/ireduce.hpp
@@ -38,8 +38,7 @@ struct MinMaxOp
     void operator()(T val, uint idx)
     {
         if ((cabs(val) < cabs(m_val) ||
-             (cabs(val) == cabs(m_val) && idx > m_idx)) &&
-            !is_nan(val)) {
+             (cabs(val) == cabs(m_val) && idx > m_idx))) {
             m_val = val;
             m_idx = idx;
         }
@@ -62,8 +61,7 @@ struct MinMaxOp<af_max_t, T>
     void operator()(T val, uint idx)
     {
         if ((cabs(val) > cabs(m_val) ||
-             (cabs(val) == cabs(m_val) && idx <= m_idx)) &&
-            !is_nan(val)) {
+             (cabs(val) == cabs(m_val) && idx <= m_idx))) {
             m_val = val;
             m_idx = idx;
         }
--- a/src/backend/cuda/kernel/ireduce.hpp
+++ b/src/backend/cuda/kernel/ireduce.hpp
@@ -56,8 +56,7 @@ namespace kernel
             m_val(val), m_idx(idx)
         {
             if (isNan(val)) {
-                Binary<T, op> ireduce;
-                m_val = ireduce.init();
+                m_val = Binary<T, op>::init();
             }
         }
 
@@ -81,8 +80,7 @@ namespace kernel
             m_val(val), m_idx(idx)
         {
             if (isNan(val)) {
-                Binary<T, af_max_t> ireduce;
-                m_val = ireduce.init();
+                m_val = Binary<T, af_max_t>::init();
             }
         }
 
--- a/src/backend/cuda/kernel/ireduce.hpp
+++ b/src/backend/cuda/kernel/ireduce.hpp
@@ -23,10 +23,10 @@ namespace cuda
 namespace kernel
 {
     template<typename T> __host__ __device__ 
-    static double cabs(const T in) { return (double)in; }
+    static double cabs(const T& in) { return (double)in; }
 
     template<> __host__ __device__
-    static double cabs<char>(const char in) { return (double)(in > 0); }
+    static double cabs<char>(const char& in) { return (double)(in > 0); }
 
     template<> __host__ __device__
     static double cabs<cfloat>(const cfloat &in) { return (double)abs(in); }
@@ -35,7 +35,7 @@ namespace kernel
     static double cabs<cdouble>(const cdouble &in) { return (double)abs(in); }
 
     template<typename T> __host__ __device__ 
-    static bool isNan(T in) { return in != in; }
+    static bool isNan(const T& in) { return in != in; }
 
     template<> __host__ __device__ 
     static bool isNan<cfloat>(const cfloat &in) { 
--- a/src/backend/cuda/kernel/ireduce.hpp
+++ b/src/backend/cuda/kernel/ireduce.hpp
@@ -63,8 +63,7 @@ namespace kernel
         __host__ __device__ void operator()(T val, uint idx)
         {
             if ((cabs(val) < cabs(m_val) ||
-                 (cabs(val) == cabs(m_val) && idx > m_idx)) &&
-                !is_nan(val)) {
+                 (cabs(val) == cabs(m_val) && idx > m_idx))) {
                 m_val = val;
                 m_idx = idx;
             }
@@ -87,8 +86,7 @@ namespace kernel
         __host__ __device__ void operator()(T val, uint idx)
         {
             if ((cabs(val) > cabs(m_val) ||
-                 (cabs(val) == cabs(m_val) && idx <= m_idx)) &&
-                !is_nan(val)) {
+                 (cabs(val) == cabs(m_val) && idx <= m_idx))) {
                 m_val = val;
                 m_idx = idx;
             }
--- a/src/backend/cuda/kernel/ireduce.hpp
+++ b/src/backend/cuda/kernel/ireduce.hpp
@@ -27,6 +27,7 @@ namespace kernel
     static double __host__ __device__ cabs(const char in) { return (double)(in > 0); }
     static double __host__ __device__ cabs(const cfloat &in) { return (double)abs(in); }
     static double __host__ __device__ cabs(const cdouble &in) { return (double)abs(in); }
+    template<typename T> __host__ __device__ static bool isNan(T in) { return in != in; }
 
     template<af_op_t op, typename T>
     struct MinMaxOp
@@ -36,13 +37,17 @@ namespace kernel
         __host__ __device__ MinMaxOp(T val, uint idx) :
             m_val(val), m_idx(idx)
         {
+            if (isNan(val)) {
+                Binary<T, op> ireduce;
+                m_val = ireduce.init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) < cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx > m_idx)) {
+            if (!isNan(val) &&
+                (cabs(val) < cabs(m_val) ||
+                cabs(val) == cabs(m_val))) {
                 m_val = val;
                 m_idx = idx;
             }
@@ -57,13 +62,17 @@ namespace kernel
         __host__ __device__ MinMaxOp(T val, uint idx) :
             m_val(val), m_idx(idx)
         {
+            if (isNan(val)) {
+                Binary<T, af_max_t> ireduce;
+                m_val = ireduce.init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) > cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx <= m_idx)) {
+            if (!isNan(val) &&
+                (cabs(val) > cabs(m_val) ||
+                cabs(val) == cabs(m_val))) {
                 m_val = val;
                 m_idx = idx;
             }
--- a/src/backend/cuda/kernel/ireduce.hpp
+++ b/src/backend/cuda/kernel/ireduce.hpp
@@ -22,7 +22,7 @@ namespace cuda
 {
 namespace kernel
 {
-    template<typename T> __host__ __device__ 
+    template<typename T> __host__ __device__
     static double cabs(const T& in) { return (double)in; }
 
     template<> __host__ __device__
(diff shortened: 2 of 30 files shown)


src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -27,6 +27,9 @@ namespace kernel
     static double __host__ __device__ cabs(const char in) { return (double)(in > 0); }
     static double __host__ __device__ cabs(const cfloat &in) { return (double)abs(in); }
     static double __host__ __device__ cabs(const cdouble &in) { return (double)abs(in); }
+    template<typename T> __host__ __device__ static bool isNan(T in) { return in != in; }
+    static bool __host__ __device__ isNan(const cfloat &in) { return in.x != in.x || in.y != in.y; }
+    static bool __host__ __device__ isNan(const cdouble &in) { return in.x != in.x || in.y != in.y; }
```
umar456: Can you specialize this function so that it is a template? It should look something like this.

```
template<> __host__ __device__ isNan<cfloat>(const cfloat &in) {...}
template<> __host__ __device__ isNan<cdouble>(const cdouble &in) {...}
```

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -36,13 +39,17 @@ namespace kernel
         __host__ __device__ MinMaxOp(T val, uint idx) :
             m_val(val), m_idx(idx)
         {
+            if (isNan(val)) {
+                Binary<T, op> ireduce;
+                m_val = ireduce.init();
```
umar456: I wonder if you could make `init` a static function. That way you don't have to create a binary object. It doesn't really make sense to create a Binary object here.

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+
+    // 4th element is unusually large to cover the case where
+    //  one part holds the largest value among the array,
+    //  and the other part is NaN.
+    // There's a possibility where the NaN is turned into 0
+    //  (since Binary<>::init() will initialize it to 0 in
+    //  for complex max op) during the comparisons, and so its 
+    //  magnitude will determine that that element is the max, 
+    //  whereas it should have been ignored since its other 
+    //  part is NaN
```
umar456: 👍

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+
+    std::vector<int> h_min_idx(cols);
+    min_idx.host(&h_min_idx[0]);
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_min_val[i], gold_min_val[i]);
+    }
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_min_idx[i], gold_min_idx[i]);
```
umar456: idx are integers so you can just use `ASSERT_EQ`

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+
+    std::vector<int> h_max_idx(cols);
+    max_idx.host(&h_max_idx[0]);
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_max_val[i], gold_max_val[i]);
+    }
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_max_idx[i], gold_max_idx[i]);
```
umar456: Not floats so you can just use `ASSERT_EQ`

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+    }
+}
+
+TEST(MinMax, MinCplxNaN)
+{
+    float real_wnan_data[] = {
+        849.088836, NAN, 889.189274, NAN, 506.472337,
```
umar456: Nit. Is there a significance to the non Nan(non Not a number, just number?) numbers? we can probably just make them simpler because they distract from the actual problem. Not a big deal.

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+    af::array real_wnan(rows, cols, real_wnan_data);
+    af::array imag_wnan(rows, cols, imag_wnan_data);
+    af::array a = af::complex(real_wnan, imag_wnan);
+
+    float gold_min_real[] = { 506.472337, 106.103029 };
+    float gold_min_imag[] = { 465.926785, 464.726170 };
+    
+    af::array min_val = af::min(a);
+
+    std::vector< std::complex<float> > h_min_val(cols);
```
umar456: you can define the symbols you use using the statements. Like this:
```
using std::vector;
using std::complex;
etc.
```
Then you don't have to specify the namespace.

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+    std::vector<int> h_min_idx(cols);
+    min_idx.host(&h_min_idx[0]);
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_min_val[i].real(), gold_min_real[i]);
+        ASSERT_FLOAT_EQ(h_min_val[i].imag(), gold_min_imag[i]);
+    }
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_min_idx[i], gold_min_idx[i]);
```
umar456: Index is an integer so use `ASSERT_EQ`.

test/reduce.cpp
```diff
@@ -523,6 +523,221 @@ TEST(MinMax, NaN)
...
+    std::vector<int> h_max_idx(cols);
+    max_idx.host(&h_max_idx[0]);
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_max_val[i].real(), gold_max_real[i]);
+        ASSERT_FLOAT_EQ(h_max_val[i].imag(), gold_max_imag[i]);
+    }
+
+    for (int i = 0; i < cols; i++) {
+        ASSERT_FLOAT_EQ(h_max_idx[i], gold_max_idx[i]);
```
umar456: `ASSERT_EQ`

src/backend/cuda/kernel/ireduce.hpp
9prady9: Yes, it works. I have the changes ready. @mark-poscablo I can send you them as a patch file

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: Just read up on template specialization. Haha I have a lot to learn in C++, thanks for this tip! I went ahead and did the same too for `cabs()`

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: Good idea - I applied the patch from @9prady9, thanks!

test/reduce.cpp
mark-poscablo: Lol I got carried away with using assert floats. Thanks for catching this!

test/reduce.cpp
mark-poscablo: 👍

test/reduce.cpp
mark-poscablo: Not really, yeah I'll make them simpler as to not distract

test/reduce.cpp
mark-poscablo: 👍

test/reduce.cpp
mark-poscablo: 👍

test/reduce.cpp
mark-poscablo: Thanks, this should eliminate some namespace bloat

mark-poscablo:
I'm also going to move the tests for indexed min/max to test/ireduce.cpp btw, since I think it makes more sense to put them there than mix them with regular min/max in test/reduce.cpp

src/backend/opencl/kernel/morph.hpp
```diff
@@ -44,7 +44,7 @@ template<typename T, bool isDilation, int SeLength>
 std::string generateOptionsString()
 {
     ToNumStr<T> toNumStr;
-    T init = isDilation ? Binary<T, af_max_t>().init() : Binary<T, af_min_t>().init();
+    T init = isDilation ? Binary<T, af_max_t>()::init() : Binary<T, af_min_t>()::init();
```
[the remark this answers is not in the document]
mark-poscablo: 👍

9prady9:
@umar I think all the changes are in, Can take another look ?

src/backend/opencl/kernel/iops.cl
```diff
@@ -7,13 +7,20 @@
...
+#endif
+
 #if CPLX
 #define sabs(in) ((in.x)*(in.x) + (in.y)*(in.y))
 #ifdef MIN_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((sabs(lhs[0]) > sabs(rhs)) ||
-        (sabs(lhs[0]) == sabs(rhs) && *lidx < ridx)) {
+    if (!(IS_NAN(lhs[0])) &&
```
umar456: I think if we reverse the order of the && operation we can avoid calling IS_NAN as much because of short-circuiting.

so instead of
```
if (!(IS_NAN(lhs[0])) &&
    ((sabs(lhs[0]) < sabs(rhs)) || (sabs(lhs[0]) == sabs(rhs)))) {
```
```
if ((sabs(lhs[0]) < sabs(rhs)) || (sabs(lhs[0]) == sabs(rhs)))
    && !(IS_NAN(lhs[0])))
```

src/backend/opencl/kernel/ireduce_first.cl
```diff
@@ -52,7 +52,7 @@ void ireduce_first_kernel(__global T *oData,
     uint out_idx = xid;
 
     if (cond && xid < lim) {
-        out_val = iData[xid];
+        if (!(IS_NAN(iData[xid]))) out_val = iData[xid];
```
umar456: This condition should wrap both statements in the body of this if statement. Probably a good idea to add a test for this scenario.

test/binary_ops.hpp
```diff
@@ -85,7 +85,7 @@ struct Binary<T, AF_BINARY_MAX>
     template<>                                      \
     struct Binary<T, AF_BINARY_MIN>                 \
     {                                               \
-        T init()                                    \
+        static T init()                                    \
```
umar456: The backslash is not formatted correctly.

test/ireduce.cpp
```diff
@@ -14,7 +14,16 @@
...
+using std::vector;
+using std::complex;
+using af::array;
+using af::dtype;
+using af::dtype_traits;
+using af::randu;
+using af::constant;
+using af::span;
+using af::min;
+using af::allTrue;
```
umar456: :+1:

test/ireduce.cpp
```diff
@@ -185,3 +194,146 @@ TEST(IndexedReduce, MaxReduceDimensionHasSingleValue)
...
+}
+
+TEST(IndexedReduce, MaxNaN) 
+{ 
+    float test_data[] = { 1, NAN, 5, 0.1, NAN, -0.5, NAN, 0 }; 
+    int rows = 4; 
+    int cols = 2; 
+    array a(rows, cols, test_data); 
+ 
+    float gold_max_val[] = { 5.0, 0.0 };
```
umar456: add f to the end of those constants to avoid narrowing warnings. i.e. `{5.0f, 0.f};`

umar456: Seems like there are other places where this could be applied.

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -57,13 +79,16 @@ namespace kernel
...
+            if (isNan(val)) {
+                m_val = Binary<T, af_max_t>::init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) > cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx <= m_idx)) {
```
umar456: Why was this last condition removed? I think it enforces that if the values are equal then we should prefer indices that are smaller.

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -36,13 +55,16 @@ namespace kernel
...
+            if (isNan(val)) {
+                m_val = Binary<T, op>::init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) < cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx > m_idx)) {
```
umar456: I think this condition was incorrectly removed. See comment below.

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -22,11 +22,30 @@ namespace cuda
 {
 namespace kernel
 {
+    template<typename T> __host__ __device__ 
+    static double cabs(const T& in) { return (double)in; }
```
umar456: @pavanky I am questioning the reasoning behind using doubles as the output of this function. Integral types should return the same type. I understand wanted to keep the original C return value but we are leaving a lot of performance on the table here. @mark-poscablo I am making general comments. You don't need to change anything in this PR.

src/backend/cuda/kernel/ireduce.hpp
umar456: Another good candidate for a test.

src/backend/cpu/kernel/ireduce.hpp
```diff
@@ -49,13 +55,17 @@ struct MinMaxOp<af_max_t, T>
     MinMaxOp(T val, uint idx) :
         m_val(val), m_idx(idx)
     {
+        if (isNan(val)) {
+            m_val = Binary<T, af_max_t>::init();
```
umar456: The m_idx value should be updated as well. To what is a good question.

umar456: @pavanky perhaps max uint?

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -57,13 +79,16 @@ namespace kernel
         __host__ __device__ MinMaxOp(T val, uint idx) :
             m_val(val), m_idx(idx)
         {
+            if (isNan(val)) {
+                m_val = Binary<T, af_max_t>::init();
```
umar456: Update the value of m_idx

src/backend/cpu/kernel/ireduce.hpp
```diff
@@ -28,13 +30,17 @@ struct MinMaxOp
...
+            m_val = Binary<T, op>::init();
+        }
     }
 
     void operator()(T val, uint idx)
     {
-        if (cabs(val) < cabs(m_val) ||
-            (cabs(val) == cabs(m_val) &&
-             idx > m_idx)) {
+        if (!isNan(val) &&
```
umar456: I think if we reverse the order of the && operation we can avoid calling IS_NAN as much because of short-circuiting.

so instead of
```
if (!(IS_NAN(lhs[0])) &&
    ((sabs(lhs[0]) < sabs(rhs)) || (sabs(lhs[0]) == sabs(rhs)))) {
```
```
if ((sabs(lhs[0]) < sabs(rhs)) || (sabs(lhs[0]) == sabs(rhs)))
    && !(IS_NAN(lhs[0])))
```

src/backend/cpu/kernel/ireduce.hpp
```diff
@@ -49,13 +55,17 @@ struct MinMaxOp<af_max_t, T>
...
+            m_val = Binary<T, af_max_t>::init();
+        }
     }
 
     void operator()(T val, uint idx)
     {
-        if (cabs(val) > cabs(m_val) ||
-            (cabs(val) == cabs(m_val) &&
-             idx <= m_idx)) {
+        if (!isNan(val) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -36,13 +55,16 @@ namespace kernel
...
+                m_val = Binary<T, op>::init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) < cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx > m_idx)) {
+            if (!isNan(val) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/cuda/kernel/ireduce.hpp
```diff
@@ -57,13 +79,16 @@ namespace kernel
...
+                m_val = Binary<T, af_max_t>::init();
+            }
         }
 
         __host__ __device__ void operator()(T val, uint idx)
         {
-            if (cabs(val) > cabs(m_val) ||
-                (cabs(val) == cabs(m_val) &&
-                 idx <= m_idx)) {
+            if (!isNan(val) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/opencl/kernel/iops.cl
```diff
@@ -23,8 +30,9 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 #ifdef MAX_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((sabs(lhs[0]) < sabs(rhs)) ||
-        (sabs(lhs[0]) == sabs(rhs) && *lidx > ridx)) {
+    if (!(IS_NAN(lhs[0])) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/opencl/kernel/iops.cl
```diff
@@ -35,8 +43,9 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 #ifdef MIN_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((*lhs > rhs) ||
-        (*lhs == rhs && *lidx < ridx)) {
+    if (!(IS_NAN(*lhs)) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/opencl/kernel/iops.cl
```diff
@@ -46,8 +55,9 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 #ifdef MAX_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((*lhs < rhs) ||
-        (*lhs == rhs && *lidx > ridx)) {
+    if (!(IS_NAN(*lhs)) &&
```
umar456: Switch order of isNan operation. See other comments for info.

src/backend/opencl/kernel/iops.cl
mark-poscablo: Oh yeah that makes sense - that comparison is probably more often useful than the is_nan check anyway. Btw what do you think of making the is_nan check an inline function than a macro? Just to make it safer from potential un-obvious syntactical errors next time we might have to deal with the is_nan check?

umar456: Fewer macros the better. If you don't run into any issues then you should do it.

test/binary_ops.hpp
mark-poscablo: Yup I'll fix these slashes

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: I remember that the CUDA reduce and ireduce tests were failing because of that last condition right after I added the isNan condition and the corresponding initial test cases. In any case, this works in a serial reduction, but in CUDA (and I think in a parallel setting in general), that last condition gives incorrect output in the following scenario: suppose that `val > m_val` but `idx > m_idx` - this case is very plausible since we're not going through the array sequentially. In this case, we should assign `m_val` and `m_idx` to `val` and `idx`, but that last condition will prevent that from happening since `idx > m_idx`. The min operation suffered a similar problem. Also I think this is the reason why the OpenCL min/max ireduce kernels don't have this last condition (even before I modified them). Please let me know your thoughts?

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: Addressed this concern in my comment in the related change too (the max counterpart), please let me know what you think in that thread.

src/backend/cuda/kernel/ireduce.hpp
umar456: The (idx) only comes into play when the `val` and `m_val` values are the same (notice that the right-hand side is surrounded by the parenthesis). The new condition should look something like this:
```
(cabs(val) > cabs(m_val) || (cabs(val) == cabs(m_val) && idx <= m_idx)) && !isNan(val)
```

mark-poscablo: OH that makes so much more sense. I think that confused me at first and tried to change the parenthesizing of the conditions, which led me to the failing tests I mentioned earlier. I'll the parenthesizing back (as you've written above) and I'll add the conditions back too. Thanks for the clarification!

mark-poscablo: Actually I'm wrong on another note - the opencl kernel had that last condition also. Will add that back there too.

mark-poscablo: And yep I can add tests for this specific scenario

test/binary_ops.hpp
mark-poscablo: 👍

test/ireduce.cpp
mark-poscablo: 👍

src/backend/cpu/kernel/ireduce.hpp
mark-poscablo: 👍

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: 👍

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: 👍

src/backend/opencl/kernel/iops.cl
mark-poscablo: 👍

src/backend/opencl/kernel/iops.cl
mark-poscablo: 👍

src/backend/opencl/kernel/iops.cl
mark-poscablo: 👍

src/backend/cuda/kernel/ireduce.hpp
mark-poscablo: Looking more closely at the original version of the kernels before i modified them, looks like the kernels actually prefer the larger index (the newer element) in case `val == `m_val`. I'll go with that original intent unless you want me to change it to really prefer smaller indices instead?

mark-poscablo: Actually, hold on - that above comment is true for min, but the max is the other way - it prefers the smaller index if available.

mark-poscablo: :+1: just waiting for these changes to pass checks now

src/backend/opencl/kernel/ireduce_first.cl
mark-poscablo: So this last one - let me try to see if I got what you're trying to say here. Do you mean that both assignments to `out_val` and `out_idx` should be together in the body of an if-statement?

mark-poscablo:
Just rebased this PR off the current master. Build went fine on my OS X and will push it now to my fork so it can run against the checks again.

src/backend/opencl/kernel/ireduce_first.cl
mark-poscablo: :+1: changed this based on our slack conversation. The is_nan check be moved to the outer condition so that we don't assign out_val and out_idx to new values if the current iData element is nan. just waiting for the checks now

test/scan_by_key.cpp
```diff
@@ -106,7 +106,7 @@ void verify(af::dim4 dims,
                     gold = (To)in[index];
                     ASSERT_NEAR(gold, out[index], eps);
                 } else {
-                    gold = binOp.init();
+                    gold = Binary<To, op>::init();
```
umar456: Binary is not a public facing API and should not be used in test. Please remove it.

mark-poscablo: So I think this test is using a separate header file that defines its own templated Binary operators (test/binary_ops.hpp), apart from the ArrayFire source. I accidentally made this static too in the course of scanning the whole repo for all Binary operator implementations. Let me just revert this and test/binary_ops.hpp to it's original version instead then?

mark-poscablo:
@umar456 All checks pass, please review again whenever you're free and let me know what you think - feel free to keep them coming until we get quality code for this PR

src/backend/opencl/kernel/iops.cl
```diff
@@ -7,13 +7,21 @@
...
+
 #if CPLX
 #define sabs(in) ((in.x)*(in.x) + (in.y)*(in.y))
 #ifdef MIN_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((sabs(lhs[0]) > sabs(rhs)) ||
-        (sabs(lhs[0]) == sabs(rhs) && *lidx < ridx)) {
+    if (((sabs(lhs[0]) > sabs(rhs)) ||
+         (sabs(lhs[0]) == sabs(rhs) &&
```
umar456: Nit: I liked it better when the lidx < ridx condition was on the same line as the equality condition.

src/backend/opencl/kernel/iops.cl
```diff
@@ -23,20 +31,22 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 #ifdef MAX_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((sabs(lhs[0]) < sabs(rhs)) ||
-        (sabs(lhs[0]) == sabs(rhs) && *lidx > ridx)) {
+    if (((sabs(lhs[0]) < sabs(rhs)) ||
+         (sabs(lhs[0]) == sabs(rhs) &&
```
umar456: Nit: I liked it better when the condition was on the same line.

src/backend/opencl/kernel/iops.cl
```diff
@@ -23,20 +31,22 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
...
 #else
-#define sabs(in) in
 #ifdef MIN_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((*lhs > rhs) ||
-        (*lhs == rhs && *lidx < ridx)) {
+    if (((*lhs > rhs) ||
+         (*lhs == rhs && *lidx < ridx)) &&
+        !is_nan(*lhs)) {
```
umar456: Is rhs guaranteed to be a non-nan number? Do we need to add a check for that as well?

src/backend/opencl/kernel/iops.cl
```diff
@@ -46,8 +56,9 @@ void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 #ifdef MAX_OP
 void binOp(T *lhs, uint *lidx, T rhs, uint ridx)
 {
-    if ((*lhs < rhs) ||
-        (*lhs == rhs && *lidx > ridx)) {
+    if (((*lhs < rhs) ||
+         (*lhs == rhs && *lidx > ridx)) &&
+        !is_nan(*lhs)) {
```
umar456: Is rhs guaranteed to be a non-nan number? Do we need to add a check for that as well?

src/backend/opencl/kernel/iops.cl
mark-poscablo: :+1: Yeah that makes sense. I think I was just following suit with the other backends on how this condition was originally written. I could change the others too to this style.

src/backend/opencl/kernel/iops.cl
mark-poscablo: :+1:

src/backend/opencl/kernel/iops.cl
mark-poscablo: Actually... the way this was originally written kinda made me confused because the order of lhs and rhs is flipped on the other backends. In any case though, I think I should be checking if `rhs` is nan here, instead of `*lhs`. `*lhs` (pointing to the kernel's `out_val`) is actually guaranteed to be non-nan, since it's initialized to the type's reduction initial value in the beginning (in `ireduce_first_kernel()`), and there is a guard there that prevents it from being changed to an element of the input array if that element is nan. The same guard then should be applied in the operator (in this case checking if `rhs` is nan) to maintain that guarantee.

mark-poscablo:
@umar456 I've addressed your latest comments and I just rebased my changes onto the latest master. All checks pass, please review again whenever you have time. Thanks!

test/reduce.cpp
```diff
@@ -508,7 +509,7 @@ TYPED_TEST(Reduce, Test_Any_Global)
     }
 }
 
-TEST(MinMax, NaN)
+TEST(MinMax, MinMaxNaN)
```
umar456: Nit: this doesn't really add information to the test name. I am ok with it going in.

PR #2164, opened 2018-05-26
Make FreeImage a run-time instead of a link-time dependency

umar456:
This commit will make the FreeImage library a runtime dependency
instead of a link time dependency. This means that the user does
not need to have FreeImage installed on their system if they are
not using the image input and output functions. The FreeImage
library will be loaded at runtime only when the image io functions
are used.

Also includes the DependencyModule class which will make it easier
for other classes to be loaded in a similar way.

### DIFF
--- a/src/api/c/DependencyModule.cpp
+++ b/src/api/c/DependencyModule.cpp
@@ -1,0 +1,101 @@
+
+#include <DependencyModule.hpp>
+#include <algorithm>
+#include <string>
+
+using std::string;
+namespace {
+
+#if(OS_WIN)
+    static const char* librarySuffix = ".dll";
+    static const char* libraryPrefix = "";
+
+    LibHandle loadLibrary(const char* library_name) {
+        return LoadLibrary(library_name);
+    }
+
+    void unloadLibrary(LibHandle handle) {
+        FreeLibrary(handle);
+    }
+
+    string getErrorMessage() {
+        LPVOID lpMsgBuf;
+        LPVOID lpDisplayBuf;
+        DWORD dw = GetLastError();
+
+        size_t characters_in_message;
+        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER |
+                      FORMAT_MESSAGE_FROM_SYSTEM |
+                      FORMAT_MESSAGE_IGNORE_INSERTS,
+                      NULL,
+                      dw,
+                      MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+                      (LPTSTR) &lpMsgBuf,
+                      0, NULL );
+        string error_message(lpMsgBuf);
+        return error_message;
+    }
+
+#else
+#if(OS_MAC)
+    static const char* librarySuffix = ".dylib";
+#else
+    static const char* librarySuffix = ".so";
+#endif
+
+    static const char* libraryPrefix = "lib";
+
+    LibHandle loadLibrary(const char* library_name) {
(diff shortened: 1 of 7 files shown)


9prady9:
@umar I haven't looked at the changes yet but looks like the windows jobs failed with the following error
```


C:\Users\buildbot\bb-worker\windows-nvidia-01\win-cuda-k5000\build\src\api\c\DependencyModule.hpp(2): fatal error C1083: Cannot open include file: 'dlfcn.h': No such file or directory

```

src/api/c/DependencyModule.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+#include <dlfcn.h>
+#include <utility>
+#include <type_traits>
+#include <vector>
+#include <string>
+
+#if(OS_WIN)
+#include <Windows.h>
+using LibHandle = HMODULE;
```
9prady9: Have you considered merging this class and symbol manager class at any point ? On second thought, it may be better to keep them separate, but just wondering if would reduce any repeat code!

src/api/c/DependencyModule.cpp
```diff
@@ -0,0 +1,101 @@
...
+    static const char* librarySuffix = ".dylib";
+#else
+    static const char* librarySuffix = ".so";
+#endif
+
+    static const char* libraryPrefix = "lib";
+
+    LibHandle loadLibrary(const char* library_name) {
+        return dlopen(library_name, RTLD_LAZY);
+    }
```
9prady9: I think it would be easy to understand the code if the `#if` blocks based on OS are closer to the code deviation point as follows.
```
LibHandle loadLibrary(const char* library_name)
{
+#if(OS_WIN)
        return LoadLibrary(library_name);
#else
        return dlopen(library_name, RTLD_LAZY);
#endif
}
```
instead of current approach.  This wouldn't be a big distraction when the functions have single line implementations. However, tracking where one `#if` begins and ends is sometimes a distraction and can be confusing. It is hard to loose track when the matching `#if` and `#endif` are close by.

src/api/c/DependencyModule.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
9prady9: May be more appropriate location for the `DependencyModule` header and source file is `src/backend/common` where the code common to all backends is present. The classes and functions from that location are used both at `src/api/c/` level and `src/backend/[cpu | cuda | opencl ]` level. It looks like there is a very good change that we might end up using this DependencyModule class at backend level also for other libraries - may be when you do similar changes to `forge` itself.

src/api/c/DependencyModule.hpp
umar456: It's probably a good idea. The difference is that we loading all the symbols at once whereas the unified backend is loading them as needed. I think we can merge them but I didn't want to take that on just yet.

src/api/c/DependencyModule.cpp
umar456: Yeah, that's why I separated them. It's easier for me to follow what is going on. I am thinking about having one interface file and have os specific implementations in a separate file because I don't really like the current approach either.

src/api/c/DependencyModule.hpp
umar456: Addressed.

src/api/c/DependencyModule.cpp
umar456: Moved the platform independent code in its own file. Its much easier to read. Let me know what you think.

PR #2167, opened 2018-05-27
Move memory manager definition to impl file.

umar456:
Move the definition of the memory manager to a MemoryManagerImpl.hpp
file so that the definition is not compiled multiple times. Should
improve compilation times. No source code changes were made in this
commit

### DIFF
--- a/src/backend/common/MemoryManager.hpp
+++ b/src/backend/common/MemoryManager.hpp
@@ -76,364 +76,60 @@ class MemoryManager
     std::vector<memory_info> memory;
     bool debug_mode;
 
-    memory_info& getCurrentMemoryInfo()
-    {
-        return memory[this->getActiveDeviceId()];
-    }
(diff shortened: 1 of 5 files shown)


9prady9:
Are you sure of the speed-up ? With 4 threads, the following is what I saw on my Arch Linux machine.
CPU: AMD A10-7850K Radeon R7, 12 Compute Cores 4C+8G

With this PR:
```
real	33m24.118s
user	113m23.402s
sys	3m55.884s
```
Prior to the change:
```
real	32m10.403s
user	114m7.631s
sys	3m56.753s
```

Note: I built only CPU, CUDA backends. No examples, docs, tests or OpenCL backend.

9prady9:
Although this adds a miniscual improvement in build time, this enables editing MemoryManager editing without having to build all the source files that include this header.

PR #2168, opened 2018-05-27
Set the symbol visibility to hidden

umar456:
Setting the visibility to hidden to improve link times and reduce the
binary size. This reduced the libafcpu.so size from 41MB to 36MB

### DIFF
--- a/CMakeModules/InternalUtils.cmake
+++ b/CMakeModules/InternalUtils.cmake
@@ -68,6 +68,7 @@ macro(arrayfire_set_cmake_default_variables)
 
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_EXTENSIONS OFF)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 
   # Set a default build type if none was specified
   if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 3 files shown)


9prady9:
I am getting the following error with these changes on Arch Linux
```
[ 29%] Linking CXX executable transpose_cpu
../src/backend/cpu/libafcpu.so.3.6.0: undefined reference to `dlclose'
../src/backend/cpu/libafcpu.so.3.6.0: undefined reference to `dlsym'
../src/backend/cpu/libafcpu.so.3.6.0: undefined reference to `dlopen'
../src/backend/cpu/libafcpu.so.3.6.0: undefined reference to `dlerror'
collect2: error: ld returned 1 exit status
```

9prady9:
There is a difference of 22MB in binaries in `Debug` mode builds and 5MB in `Release` mode.
:+1:

CMakeModules/InternalUtils.cmake
```diff
@@ -68,6 +68,7 @@ macro(arrayfire_set_cmake_default_variables)
 
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_EXTENSIONS OFF)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
9prady9: This is our default visibility unless we explicitly add visibility attribute for the declarations we want to expose. Is that correct ?

PR #2169, opened 2018-05-29
Added logging support for memory operations

umar456:
Added logging support to memory operations. By default this is
turned off but you can enable them by setting the AF_TRACE
environment variable to 'mem'. Other components can also be
added in a later commit

This is implemented using the spdlog library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ option(AF_BUILD_EXAMPLES "Build Examples"                            ON)
 
 option(AF_WITH_GRAPHICS "Build ArrayFire with Forge Graphics"  $<AND:${OPENGL_FOUND},${Forge_FOUND},${glbinding_FOUND}>)
 option(AF_WITH_NONFREE  "Build ArrayFire nonfree algorithms"   OFF)
+option(AF_WITH_LOGGING  "Build ArrayFire with logging support" ${spdlog_FOUND})
 
 option(AF_INSTALL_STANDALONE "Build installers that include all dependencies" OFF)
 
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -60,6 +61,9 @@ cmake_dependent_option(AF_WITH_IMAGEIO "Build ArrayFire with Image IO support" $
                        "FreeImage_FOUND" OFF)
 cmake_dependent_option(AF_BUILD_FRAMEWORK "Build an ArrayFire framework for Apple platforms.(Experimental)" OFF
                        "APPLE" OFF)
+cmake_dependent_option(AF_WITH_LOGGING "Build ArrayFire with support for logging functionallity." ON
+                       "TARGET spdlog::spdlog" OFF)
+
```
9prady9: Based on the PR description, I expected `AF_WITH_LOGGING` to be an normal option instead of dependent option. If the target build machine has spdlog installed but wasn't intended for ArrayFire, we would still build ArrayFire with logging enabled. Is that the intended behavior we want to provide ?

src/backend/common/CMakeLists.txt
```diff
@@ -14,10 +14,11 @@ target_sources(afcommon_interface
     ${CMAKE_CURRENT_SOURCE_DIR}/DependencyModule.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/DependencyModule.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/FFTPlanCache.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/Logger.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/Logger.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/MatrixAlgebraHandle.hpp
```
9prady9: Can't we avoid including the source files completely based on `AF_WITH_LOGGING`, one less translation unit that way.

CMakeLists.txt
umar456: Yeah, good point. Will fix

src/backend/common/Logger.cpp
```diff
@@ -0,0 +1,55 @@
...
+shared_ptr<logger>
+logger_factory(string name) {
+    auto logger = stdout_logger_mt(name);
+    logger->set_pattern("[%n][%t] %v");
+
+    // Log mode
+    string env_var = getEnvVar("AF_TRACE");
+    if(env_var.find_first_of("all") != string::npos ||
+       env_var.find_first_of(name) != string::npos)
+        logger->set_level(trace);
```
9prady9: I couldn't understand what's the effect of trace level, what happens if we don't set trace level ?

What is the default value of `AF_TRACE` environment variable ? If we don't set `AF_MEM`, then trace wouldn't be enabled right ? How will the log look like then ?

umar456:
Addressed all feedback

PR #2179, opened 2018-06-04
Fix memory leak in median

umar456:
An early exit code path did not free intermediate data in median

### DIFF
--- a/src/api/c/median.cpp
+++ b/src/api/c/median.cpp
@@ -104,7 +104,10 @@ static af_array median(const af_array& in, const dim_t dim)
 
     if (dimLength % 2 == 1) {
         // mid-1 is our guy
-        if (input.isFloating()) return left;
+        if (input.isFloating()) {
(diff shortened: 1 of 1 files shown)


src/api/c/median.cpp
```diff
@@ -104,7 +104,10 @@ static af_array median(const af_array& in, const dim_t dim)
 
     if (dimLength % 2 == 1) {
         // mid-1 is our guy
-        if (input.isFloating()) return left;
+        if (input.isFloating()) {
+            AF_CHECK(af_release_array(sortedIn_handle));
```
9prady9: I am not sure if that release is necessart because sortedIn_handle is just a handle to `sortedIn` which is of type `detail::Array<T>` and is destroyed once it goes out of scope. out getHandle doesn't do any memory allocation other than type casting, so there is no memory leak over there.

umar456: Get Handle increments the reference count for Array<T>.

PR #2182, opened 2018-06-06
Check for unsupported cards on a given cuda runtime

### LINKED ISSUE #2181: ArrayFire 3.6 tries to use CUDA for unsupported graphic card
The official Windows 64-bit ArrayFire 3.6 installers do not support CUDA Compute older than 3.0.
Using ArrayFire 3.6 with GeForce GTX 560, it will automatically select CUDA as the backend but throw an exception when it tries to use it.

af::info() prints:
ArrayFire v3.6.0 (CUDA, 64-bit Windows, build ee21c79)
Platform: CUDA Toolkit 9, Driver: CUDA Driver Version: 9000
[0] GeForce GTX 560, 1024 MB, CUDA Compute 2.1

af::exception, e.what() gives:
ArrayFire Exception (Internal error:998):
In function void cuda::kernel::transpose<T, conjugate>(cuda::Param<T>, cuda::CParam<T>, int)
In file C:/Users/buildbot/bb-worker/windows-nvidia-01/windows-nightly/build/src/backend/cuda\kernel/transpose.hpp:116
CUDA Error (48): no kernel image is available for execution on the device

Should the backend detection only select CUDA if the graphics card is supported - perhaps by checking the compute version against the settings used to compile ArrayFire?

9prady9:
Fixes #2181 

CUDA 9.* drops support for all Fermi(NVIDIA) GPUs. This adds
a check when the user wants to run ArrayFire built with CUDA 9.*
on a card with compute version <=2.*

### DIFF
--- a/src/backend/cuda/platform.cpp
+++ b/src/backend/cuda/platform.cpp
@@ -19,6 +19,7 @@
 #include <common/host_memory.hpp>
 
 #include <algorithm>
+#include <array>
 #include <cstdio>
 #include <mutex>
 #include <sstream>
@@ -144,6 +145,18 @@ static inline string toString(T val)
     stringstream s;
     s << val;
     return s.str();
+}
+
+static inline
+int getMinSupportedCompute(int cudaMajorVer)
+{
+    // Vector of minimum supported compute versions
+    // for CUDA toolkit (i+1).* where i is the index
+    // of the vector
+    static const std::array<int,10> minSV{1, 1, 1, 1, 1, 1, 2, 2, 3, 3};
+
+    auto CVSize = minSV.size();
+    return (cudaMajorVer>CVSize ? minSV[CVSize-1] : minSV[cudaMajorVer-1]);
 }
 
 ///////////////////////////////////////////////////////////////////////////
@@ -496,16 +509,26 @@ DeviceManager::DeviceManager()
     CUDA_CHECK(cudaGetDeviceCount(&nDevices));
     if (nDevices == 0)
         throw runtime_error("No CUDA-Capable devices found");
-
     cuDevices.reserve(nDevices);
+
+    int cudaRtVer = 0;
+    CUDA_CHECK(cudaRuntimeGetVersion(&cudaRtVer));
+    int cudaMajorVer = cudaRtVer / 1000;
 
     for(int i = 0; i < nDevices; i++) {
         cudaDevice_t dev;
         cudaGetDeviceProperties(&dev.prop, i);
-        dev.flops = dev.prop.multiProcessorCount * compute2cores(dev.prop.major, dev.prop.minor) * dev.prop.clockRate;
-        dev.nativeId = i;
-        cuDevices.push_back(dev);
+        if (dev.prop.major<getMinSupportedCompute(cudaMajorVer)) {
+            continue;
+        } else {
+            dev.flops = dev.prop.multiProcessorCount *
+                        compute2cores(dev.prop.major, dev.prop.minor) *
+                        dev.prop.clockRate;
+            dev.nativeId = i;
+            cuDevices.push_back(dev);
+        }
     }
+    nDevices = cuDevices.size();
 
     sortDevices();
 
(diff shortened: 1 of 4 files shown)


9prady9:
`ci/linux-cuda-k80 ` job is finished , not sure why github is showing it as pending. please check the details link to verify that it is indeed finished.

test/threading.cpp
```diff
@@ -346,6 +346,7 @@ TEST(Threading, FFT_R2C)
 
     int numDevices = 1;
```
mlloreda: Wouldn't it be better to initialize `numDevices` to 0?

9prady9: Probably, I didn't touch existing declaration. It wouldn't matter anyway, since af_get_device_count is going to overwrite it with 0 or available device count.

src/backend/cuda/platform.cpp
```diff
@@ -496,16 +496,26 @@ DeviceManager::DeviceManager()
...
+    CUDA_CHECK(cudaRuntimeGetVersion(&cudaRtVer));
+    int cudaMajorVer = cudaRtVer / 1000;
+
     for(int i = 0; i < nDevices; i++) {
         cudaDevice_t dev;
         cudaGetDeviceProperties(&dev.prop, i);
-        dev.flops = dev.prop.multiProcessorCount * compute2cores(dev.prop.major, dev.prop.minor) * dev.prop.clockRate;
-        dev.nativeId = i;
-        cuDevices.push_back(dev);
+        if (cudaMajorVer>8 && dev.prop.major<=2) {
```
umar456: It would be cleaner if this was implemented as a table of CUDA versions and minimum supported compute device. In the future CUDA will most likely deprecate additional hardware and that will make this implementation more complicated.

test/threading.cpp
umar456: This should be set to -1. If the af_get_device_count fails it will most likely not change the value.

src/backend/cuda/platform.cpp
```diff
@@ -555,7 +565,7 @@ int DeviceManager::setActiveDevice(int device, int nId)
 
     int numDevices = cuDevices.size();
 
-    if (device > numDevices)
+    if (device >= numDevices)
```
umar456: This seems like a bug that should be fixed in another PR.

src/backend/cuda/platform.cpp
9prady9: what kind of a table are you talking about ? Will a future deprecation be so complicated ? It would probably add another check `cudaMajorVer > 10 && dev.prop.major<=3` or something like that to the top of the if-else block at the top.

umar456: You can just a map or a vector that stores the minimum supported compute for each version of cuda. something like this:

vector<int> versions = {0, 0, 0, 0, 0, 0, 0, 00100, 00102};

or something similar.

9prady9: That is an overkill for just couple of checks. Vector or not, we can't eliminate the if condition here, it will only change the form into `dev.propr.major<=versions[cudaMajorVer-1]` which doesn't add any benefits because this versions vector isn't static and have to be updated every time CUDA toolkit drops a specific compute.

Added to that, current form is easier to understand and this code is not in a performance bottleneck zone.

src/backend/cuda/platform.cpp
9prady9: Wouldn't having it in a separate commit suffice ? I will send out another PR, however I don't get the point of it because we don't do merge commits anyway.

9prady9:
Addressed feedback.

src/backend/cuda/platform.cpp
```diff
@@ -146,6 +146,18 @@ static inline string toString(T val)
     return s.str();
 }
 
+static inline
+int getMinSupportedCompute(int cudaMajorVer)
+{
+    // Vector of minimum supported compute versions
+    // for CUDA toolkit (i+1).* where i is the index
+    // of the vector
+    static const vector<int> minSV{1, 1, 1, 1, 1, 1, 2, 2, 3, 3};
```
umar456: A std::array object would be better for this. No need to allocate on the heap if we are not sending references to the values.

PR #2183, opened 2018-06-07
Fix valgrind errors by replacing malloc with calloc

9prady9:
The difference in exeuction time for calloc and malloc is barely
noticeable. I have timed their execution times using af::timeit.

### DIFF
--- a/src/backend/cuda/driver.cpp
+++ b/src/backend/cuda/driver.cpp
@@ -31,7 +31,7 @@ int nvDriverVersion(char *result, int len)
     dwLen = GetFileVersionInfoSize(lptstrFilename, &dwHandle);
     if (dwLen == 0) return 0;
 
-    lpData = malloc(dwLen);
+    lpData = calloc(1, dwLen);
     if (!lpData) return 0;
 
     rv = GetFileVersionInfo(lptstrFilename, 0, dwLen, lpData);
--- a/src/backend/cpu/memory.cpp
+++ b/src/backend/cpu/memory.cpp
@@ -136,7 +136,7 @@ bool checkMemoryLimit()
     template void memFree(T* ptr);                                                            \
     template T* pinnedAlloc(const size_t &elements);                                          \
     template void pinnedFree(T* ptr);                                                         \
- 
+
 INSTANTIATE(float)
 INSTANTIATE(cfloat)
 INSTANTIATE(double)
@@ -181,7 +181,7 @@ size_t MemoryManager::getMaxMemorySize(int id)
 
 void *MemoryManager::nativeAlloc(const size_t bytes)
 {
-    void *ptr = malloc(bytes);
+    void *ptr = calloc(bytes, 1);
     AF_TRACE("nativeAlloc: {:>7} {}", bytesToString(bytes), ptr);
     if (!ptr) AF_ERROR("Unable to allocate memory", AF_ERR_NO_MEM);
     return ptr;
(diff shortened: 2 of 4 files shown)


src/backend/cuda/driver.cpp
```diff
@@ -31,7 +31,7 @@ int nvDriverVersion(char *result, int len)
     dwLen = GetFileVersionInfoSize(lptstrFilename, &dwHandle);
     if (dwLen == 0) return 0;
 
-    lpData = malloc(dwLen);
+    lpData = calloc(1, dwLen)
```
FloopCZ: Missing semicolon?

pavanky:
@9prady9 calloc becomes slower for large amounts of memory or when you are actually writing to it. What sizes did you benchmark ?

pavanky:
I also do not like the idea of using an allocator that zeros out memory by default. This can result in false positives when someone makes the assumption allocated memory is always 0. 

What are the warnings that valgrind is actually showing ?

9prady9:
I checked for sizes 1G to 4G

9prady9:
The errors are about uninitialised values being used.

pavanky:
Then fix the issues properly. This is just asking the problem to go away.

9prady9:
```
Size in Gigs     Malloc          Calloc
0.50G    0.46358         0.45547
1.00G    0.94403         0.97601
1.50G    1.53797         1.29497
2.00G    1.71207         1.73240
2.50G    2.15062         2.15542
3.00G    2.62374         2.64514
3.50G    3.11485         3.06558
4.00G    3.91427         4.37841
```
thats on VS2017.
```
#include <stdio.h>
#include <arrayfire.h>
using namespace af;

static const int HALFG = 500e6;
static const int ONEG = 1e9;
static size_t bytes = HALFG;

static void mallocBench()
{
    char* ptr = (char*)malloc(bytes * sizeof(char));
    for (size_t i = 0; i < bytes; ++i)
        ptr[i] = 'a';
    free(ptr);
}

static void callocBench()
{
    char* ptr = (char*)calloc(bytes, sizeof(char));
    for (size_t i = 0; i < bytes; ++i)
        ptr[i] = 'a';
    free(ptr);
}

int main(int argc, char ** argv) {
    try {
        printf("Size in Gigs\t Malloc \t Calloc\n");
        for (int i = 0; i < 8; ++i) {
            float a = (double)(bytes) / ONEG;
            printf("%.2fG \t %.5f \t %.5f\n", a, timeit(mallocBench), timeit(callocBench));
            bytes += HALFG;
        }
    } catch (exception& e) {
        throw;
    }
    return 0;
}
```
Will run similar check on linux and see how the numbers are over there.

9prady9:
```
Size in Gigs     Malloc          Calloc

0.50G    10.01780        10.23171
1.00G    20.13019        20.03820
1.50G    30.95004        31.06773
2.00G    42.19501        42.08033
2.50G    51.96417        52.09555
3.00G    62.24116        61.96884
3.50G    73.09242        71.24669
4.00G    83.07392        84.46433
```
That's on Linux, it is not clearly evident if calloc is actually slower than malloc in all cases.

umar456:
I agree with @pavanky on this. We should be using malloc in nativeAlloc. If you want to remove errors in Valgrind, then you should either look further up the stack or create a suppression file.

The other changes you have made should be wrapped in unique_ptr or something similar. We should only be using malloc rarely.

PR #2185, opened 2018-06-07
Defer fft normalization to post transform stage

### LINKED ISSUE #2050: Delay the normalization of the input when performing fft
This example shows the behavior in python but similar results can be seen in C++ as well.
https://gist.github.com/pavanky/dc64fc83e0fa298e942a3e0ca07ccd15

fft and ifft should be normalized after the transform (not before) for better accuracy.

9prady9:
Fixes #2050

### DIFF
--- a/src/api/c/fft_common.hpp
+++ b/src/api/c/fft_common.hpp
@@ -28,11 +28,13 @@ Array<outType> fft(const Array<inType> input, const double norm_factor,
 {
     dim4 pdims(1);
     computePaddedDims(pdims, input.dims(), npad, pad);
-    Array<outType> output = padArray<inType, outType>(input, pdims, scalar<outType>(0), norm_factor);
+    auto res = padArray(input, pdims, scalar<outType>(0));
 
-    fft_inplace<outType, rank, direction>(output);
+    fft_inplace<outType, rank, direction>(res);
+    if (norm_factor != 1.0)
+        multiply_inplace(res, norm_factor);
 
-    return output;
+    return res;
 }
 
 template<typename inType, typename outType, int rank>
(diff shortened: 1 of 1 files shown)


9prady9:
damn, the style fixes commit broke something. The tests are passing with earlier commit. Looking into it.

umar456:
Can you separate the style changes from the PR? Its hard differentiate the changes in functionality from style.

9prady9:
I made them in two different commits.

umar456:
It looks like you made some style changes in your first commit.

9prady9:
@umar456 Addressed feedback

src/api/c/fftconvolve.cpp
```diff
@@ -121,63 +134,70 @@ AF_BATCH_KIND identifyBatchKind(const dim4 &sDims, const dim4 &fDims)
...
         AF_BATCH_KIND convBT = identifyBatchKind<baseDim>(sdims, fdims);
 
         ARG_ASSERT(1, (convBT != AF_BATCH_UNSUPPORTED));
 
-        af_array output;
+#define FFTCNV(Ti, Tr, Tc, F0, F1) \
+fftConv<Ti, Tr, Tc, F0 , F1, baseDim>(signal, filter, expand, convBT)
+
+#define FALLBACK(T) \
+fftConvFallback<T , T , T , baseDim>(signal, filter, expand)
```
umar456: You can do the same thing using some combination of type traits instead of creating macros. You don't need macros here.

9prady9:
Rolled back style changes for the moment.

net147:
@9prady9 Does the documentation need to be updated? The documentation has norm_factor as scaling the input before the transformation is applied.

9prady9:
@net147 This doesn't require docs udpate because we didn't normalization in this commit. We merely changed the stage at which normalization is carried out. If a normalization factor other than `1.0` is provided, we use that. The API already has the normalization factor as parameter.

net147:
@9prady9 Yes the API has normalization factor as a parameter, but the documentation says that is factor is applied before the transformation but this commit changes it to be applied after the transformation:
norm_factor "is the normalization factor with which the input is scaled before the transformation is applied"

9prady9:
Sorry, I didn't notice that it was specific about when it's applied. I will change it, thanks for reporting!

9prady9:
@net147 I have addressed it in https://github.com/arrayfire/arrayfire/pull/2192 thanks once again

PR #2186, opened 2018-06-08
Improved RPM installer.

mlloreda:
The new RPM package lists FreeImage as a dependency to install (via
package manager). The "package group" attribute has also been fixed.

### DIFF
--- a/CMakeModules/CPackConfig.cmake
+++ b/CMakeModules/CPackConfig.cmake
@@ -3,7 +3,7 @@
 #
 # This file is distributed under 3-clause BSD license.
 # The complete license agreement can be obtained at:
-# http://arrayfire.com/licenses/BSD-3-Clause
+# https://arrayfire.com/licenses/BSD-3-Clause
 
 cmake_minimum_required(VERSION 3.5)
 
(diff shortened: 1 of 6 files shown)


umar456:
What other plans do you have for this PR?

mlloreda:
Finished adding necessary dependencies for gl and no-gl versions of the installer.

Needs to be tested by generating and installing both (gl, no-gl) RPM installers.

CMakeModules/CPackConfig.cmake
```diff
@@ -312,10 +312,18 @@ set(CPACK_DEBIAN_PACKAGE_HOMEPAGE http://www.arrayfire.com)
...
+
 set(CPACK_RPM_PACKAGE_AUTOREQPROV " no")
+set(CPACK_RPM_PACKAGE_GROUP "Development/Libraries")
+set(CPACK_RPM_PACKAGE_LICENSE "BSD")
+set(CPACK_RPM_PACKAGE_URL "${SITE_URL}")
+
+if(AF_WITH_GRAPHICS)
+  set(CPACK_RPM_PACKAGE_REQUIRES "freeimage >= 3, glbinding, glfw")
+else()
+  set(CPACK_RPM_PACKAGE_REQUIRES "freeimage >= 3")
```
umar456: Freeimage is an optional dependency now. It is loaded at runtime since #2164

PR #2188, opened 2018-06-12
Cleaned up and improved reduce test file.

mlloreda:


9prady9:
Looks good. Can you also please look at other files that need similar fixing. If there are too many, lets share the load.

mlloreda:
👍

PR #2196, opened 2018-06-16
Avoid 'sign' macro redefined OpenCL warning

net147:


9prady9:
Would the following be better ? instead of `undef`ing & `def`ing each time.
```
#ifndef sign
    #define sign(in) signbit((in))
#endif
```
@umar456 @pavanky

9prady9:
true, my workaround wouldn't work. Just checked the opencl docs, I thought we were defining it as it wasn't available. We are redefining it so that `sign` call does what ArrayFire's interpretation means  rather than what opencl does.

net147:
Is there a reason ArrayFire needs to redefine sign?

9prady9:
Yes, checkout for the definition of `sign` in the files `src/backend/[cpu | cuda | opencl]/unary.hpp`

You would have see how `UNARY_FN` macro is defined and functioning.

net147:
I was just thinking whether it would be better to replace all calls to sign with signbit instead of defining sign to call signbit in every backend.

9prady9:
Yeah, that would be better fix probably.

net147:
@9prady9 I am not familiar with the ArrayFire JIT codebase, so I would prefer someone more familiar with it to make the changes.

9prady9:
@net147 That is fine. We will look into it as soon as we can. I can't give an exact date but I have created the [issue ](https://github.com/arrayfire/arrayfire/issues/2200) for future updates. Closing this PR.

PR #2199, opened 2018-06-18
Improved tests.

mlloreda:
- Greatly reduced number of instances where entire std and af namespaces
  were introduced into tests.
- Utilizing `using` statements to introduce specific members from `af` and
  `std` namespaces.

### DIFF
--- a/examples/computer_vision/fast.cpp
+++ b/examples/computer_vision/fast.cpp
@@ -47,6 +47,9 @@ static void fast_demo(bool console)
         img_color(seq(y-draw_len, y+draw_len), x, 2) = 0.f;
     }
 
+    freeHost(h_x);
+    freeHost(h_y);
+
     printf("Features found: %lu\n", feat.getNumFeatures());
 
     if (!console) {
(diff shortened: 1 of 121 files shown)


9prady9:
Everything looks good, but I am unable to see why changes from commit e5225d6fa045eda4f0ffdeb5d3fc7d2baa6badf7  are necessary ? 

`Using namespace af` was being used in a cpp file, so I doubt we would have any issues there. I am trying to understand what is the advantage of having so many exact speicifiers(`using af::<name>`). I think, a single `using namespace <name>` will do in this case.

umar456:
@9prady9 Its generally a bad idea to pollute the namespace with names you don't need. You should specify using statements for everything you use instead of bringing everything in at once.

http://www.codingstandard.com/rule/7-3-1-do-not-use-using-directives/

9prady9:
I have read about it too, there is no consensus on what is good regarding it. What you said mostly stands true if the using directives are in header files. What I have gathered from multiple articles is that - it is okay to have usign directive in a cpp file.

FloopCZ:
(Just by the way, [CPP Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-using) "allow" such usage pattern. I personally tend to make a short namespace alias and specify it everywhere.)

PR #2205, opened 2018-06-20
Revert "Use CLBlast as default blas upstream for OpenCL backend"

9prady9:
This reverts commit 98e6d5e220babd52769541e87f7479526e71cebc.


9prady9:
We aren't backporting the CLBLast default to 3.6 branch, hence closing this revert to fix the problem with CLBlast instead.

Issue #2212, opened 2018-06-25
ORB: CPU backend does not detect all features that CUDA and OpenCL does


mark-poscablo:
Here's a sample program output. In this program I have calculated the features' range of coordinates in the horizontal axis, as well as on the vertical axis. Notice how on the wide image ("Two squares horizontal"), the CPU's max X (horizontal) value is significantly less than CUDA's max X value. Similarly, on the tall image, the CPU's max Y (vertical) value is significantly less than CUDA's max Y value. Also notice how the CPU detects less features (actually, half) than CUDA does:

```
*** Two Squares Horizontal ***
Rows: 400
Cols: 800
CPU features: 16
CUDA features: 32
CPU range: 
X: 100,299
Y: 100,299
CUDA range: 
X: 100,699
Y: 100,299
*** Two Squares Vertical ***
Rows: 800
Cols: 400
CPU features: 16
CUDA features: 32
CPU range: 
X: 100,299
Y: 100,299
CUDA range: 
X: 100,299
Y: 100,699
```

mark-poscablo:
By the way, here's the sample program used in this issue: https://github.com/mark-poscablo/arrayfire-featdetect-sample

syurkevi:
Addressed by #3357. Related to issues #2214, #2215, #2244. Correct resolution and release version tbd.

Issue #2214, opened 2018-06-25
SUSAN: All backends mark features beyond the image array's bounds


mark-poscablo:
Here's a sample program output. In this program I have calculated the features' range of coordinates in the horizontal axis, as well as on the vertical axis. Notice how the features' max Y coordinate is greater than the number of rows in the image. There's also an exception thrown at the end regarding seqToDims, as I use af::seq() in the program as part of marking features in the image. I think the exception arises because the program is trying to mark features outside the image's dimensions.

(This is using arrayfire at 8ecbe61)

```
*** Two Squares Horizontal ***
Rows: 400
Cols: 800
CPU features: 8
CUDA features: 8
CPU range: 
X: 100,299
Y: 100,699
CUDA range: 
X: 100,299
Y: 100,699
OpenCL range: 
X: 100,299
Y: 100,699
terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Invalid input size:203):
Invalid size
In function af::dim4 af::seqToDims(af_index_t*, af::dim4, bool)
In file src/api/cpp/array.cpp:73
```

mark-poscablo:
Closing this one because it turns out (according to the discussion in #2215) that there's nothing wrong with `af::susan()` after all. `af::susan()` has the correct convention of dimensions/coordinates, and so `af::fast()` and `af::harris` have the wrong convention - they switch the first and second dimensions in the computation and the examples.

Issue #2213, opened 2018-06-25
Harris: OpenCL backend does not detect all features that CPU and CUDA does


mark-poscablo:
Here's a sample program output. In this program I have calculated the features' range of coordinates in the horizontal axis, as well as on the vertical axis. Notice how on the wide image ("Two squares horizontal"), OpenCL's max X (horizontal) value is significantly less than CPU's max X value. Similarly, on the tall image, the OpenCL's max Y (vertical) value is significantly less than CPU's max Y value. Also notice how OpenCL detects less features (actually, half) than CPU does:

```
*** Two Squares Horizontal ***
Rows: 400
Cols: 800
CPU features: 8
OpenCL features: 4
CPU range: 
X: 100,699
Y: 100,299
OpenCL range: 
X: 100,299
Y: 100,299
*** Two Squares Vertical ***
Rows: 800
Cols: 400
CPU features: 8
OpenCL features: 4
CPU range: 
X: 100,299
Y: 100,699
OpenCL range: 
X: 100,299
Y: 100,299
```

mark-poscablo:
By the way, here's the sample program used in this issue: https://github.com/mark-poscablo/arrayfire-featdetect-sample

PR #2215, opened 2018-06-25
Fix feature detection functions' 2D to 1D indexing

mark-poscablo:
Some feature detection functions have switched the first and second dimensions when linearizing the 2D indexing of the image data. Issues #2212, #2213, and #2214 all suffer from this problem. I think that this dimension switching is due to the fact that arrayfire is column-major. I also think this was issue hasn't been detected before because the tests use a square image, in which case switching dimensions on the linear array index calculation shouldn't be a problem.

### DIFF
--- a/examples/computer_vision/fast.cpp
+++ b/examples/computer_vision/fast.cpp
@@ -36,15 +36,15 @@ static void fast_demo(bool console)
     for (size_t f = 0; f < feat.getNumFeatures(); f++) {
         int x = h_x[f];
         int y = h_y[f];
-        img_color(y, seq(x-draw_len, x+draw_len), 0) = 0.f;
-        img_color(y, seq(x-draw_len, x+draw_len), 1) = 1.f;
-        img_color(y, seq(x-draw_len, x+draw_len), 2) = 0.f;
+        img_color(x, seq(y-draw_len, y+draw_len), 0) = 0.f;
+        img_color(x, seq(y-draw_len, y+draw_len), 1) = 1.f;
+        img_color(x, seq(y-draw_len, y+draw_len), 2) = 0.f;
 
         // Draw vertical line of (draw_len * 2 + 1) pixels centered on  the corner
         // Set only the first channel to 1 (green lines)
-        img_color(seq(y-draw_len, y+draw_len), x, 0) = 0.f;
-        img_color(seq(y-draw_len, y+draw_len), x, 1) = 1.f;
-        img_color(seq(y-draw_len, y+draw_len), x, 2) = 0.f;
+        img_color(seq(x-draw_len, x+draw_len), y, 0) = 0.f;
+        img_color(seq(x-draw_len, x+draw_len), y, 1) = 1.f;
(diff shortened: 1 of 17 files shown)


mark-poscablo:
The last commit (0eeb7ad) should fix #2212 . Though the changes are actually on fast(), orb() uses fast() internally so this change actually fixes both functions.

mark-poscablo:
The last commit (cb06d94) should fix the CPU backend part of #2214 .

9prady9:
Which program did you use to generate the reference test data ? Also, both fixes seem to switch the first leading dimension with the next leading dimension which doesn't seem to be correct. Also, it is very bad in terms of performance.

mark-poscablo:
@9prady9 Here's the sample code: https://gist.github.com/mark-poscablo/c09078f80476234e0abc27bcc8376e0c. In my perspective, both fixes actually just matches x and y back to the dimensions that they should represent: x is dim1 and and y is dim0. ArrayFire is column major, so isn't this correct? Furthermore, regarding performance - I'm assuming you're talking about the changes on susan here - the index is still incremented by 1 every iteration of the nested for loop, since the inner for loop is iterating through y's, and the index calculation increments by y. This should be cache-friendly data access, no?

9prady9:
I haven't looked at ORB code, but I looked at fast and susan and they don't swap x & y anywhere through out the algorithm. x is index to the fastest moving dimension and y is index to next fastest dimension, i.e. x is dimension `0`, y is dimension `1` and so on - for any `af::array`. 

`af::loadImage` loads an image such that fastest moving direction of image(scanline) is read as a column in `af::array`, thus fastest moving direction of image is the fastest moving dimension of `af::array`. They don't need to be swapped either while processing features using ArrayFire code or inside other application because x is always pointing to the scanline of image. 

When `af::saveImage` is later used to save an image, it takes care of writing out the image to a file with correct layout.

mark-poscablo:
At this point, due to the due to the potentially overreaching nature of the changes required to fix the root causes of this issue (possibly an API change for `af::loadImage` and other functions affected by it), @umar456 suggested that he take over . Many thanks to @umar456, @9prady9, and @syurkevi for all the help regarding this PR.

mark-poscablo:
By the way, here's the sample program used in this issue: https://github.com/mark-poscablo/arrayfire-featdetect-sample. Simply change the feature detector used inside `get_feats()` to use the desired one.

9prady9:
@mark-poscablo Is this PR still valid ? If not, lets close it.

9prady9:
@mark-poscablo Any updates on this PR ?

mark-poscablo:
@9prady9 No, not really. This and #2244 are on hold. Umar said he's going to take over this one since the root issues require an ArrayFire 4.0-wide change.

9prady9:
Closing for now, please reopen when the necessary changes are done and after a rebase from latest master.

mark-poscablo:
As per my discussion with @9prady9, I'm closing this for now since it's on hold. Let's reopen this when we are ready to make some changes again.

PR #2216, opened 2018-06-27
Refactor symbolic manager and add logging support

umar456:
* Use functions defined in module_loading.hpp headers
* Add support for logging

### DIFF
--- a/src/api/unified/symbol_manager.cpp
+++ b/src/api/unified/symbol_manager.cpp
@@ -8,133 +8,120 @@
  ********************************************************/
 
 #include "symbol_manager.hpp"
-#include <algorithm>
-#include <vector>
-#include <string>
-#include <cmath>
 #include <af/version.h>
+#include <common/module_loading.hpp>
 
+#include <cmath>
+#include <string>
+#include <type_traits>
+
+
+#ifndef WIN_OS
+#include <dlfcn.h>
+#else
+#include <Windows.h>
+#endif
+
+using common::loadLibrary;
+using common::unloadLibrary;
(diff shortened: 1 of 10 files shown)


src/api/unified/symbol_manager.cpp
```diff
@@ -8,145 +8,128 @@
...
+
+template<typename... ARGS>
+string join_path(string first, ARGS... args) {
+    if(first.empty()) { return join_path(args...); }
+    else              { return first + PATH_SEPARATOR + join_path(args...); }
 }
 
 /*flag parameter is not used on windows platform */
-LibHandle openDynLibrary(const int bknd_idx, int flag=RTLD_LAZY)
+LibHandle openDynLibrary(const af_backend bknd_idx, int flag=RTLD_LAZY)
```
9prady9: you might want to change the variable `bknd_idx` name to something else, now that it is an enum and not an integer.

src/api/unified/symbol_manager.cpp
```diff
@@ -8,145 +8,128 @@
...
-                if (retVal != NULL) {
-                    if (show_load_path)
-                        printf("Using %s\n", abs_path.c_str());
-                    // if the current absolute path based dlopen
-                    // search is a success, then abandon search
-                    // and proceed for compute
-                    break;
-                }
+    LibHandle retVal = nullptr;
+    for (int i = 0; i < extent<decltype(paths)>::value; i++) {
```
9prady9: :+1:  `extent<decltype(paths)>::value`

PR #2218, opened 2018-06-27
Avoid loading backends with no devices in the unified backend

umar456:
Avoid loading a backend by the unified backend if there are no
devices available. This appears on platforms where OpenCL
runtime has not been installed and the unified backend is used.
The OpenCL backend will be selected by default but will fail to
work because no devices are available.


9prady9:
Depends on #2216 . This should be merged only after #2216 is merged.

PR #2229, opened 2018-07-06
Fix MKL environment variable name. Improve FindMKL docs

umar456:
We were using the incorrect environment variable in the `FindMKL.cmake` script. MKL sets the `MKLROOT` variable instead of the `MKL_ROOT` variable when the `mklvars.sh` script is sourced.

Updated FindMKL docs

### DIFF
--- a/CMakeModules/FindMKL.cmake
+++ b/CMakeModules/FindMKL.cmake
@@ -38,7 +38,7 @@ find_path(MKL_INCLUDE_DIR
   PATHS
     /opt/intel
     /opt/intel/mkl
-    $ENV{MKL_ROOT}
+    $ENV{MKLROOT}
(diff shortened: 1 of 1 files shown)


9prady9:
This variable is only set on non windows platforms, is that correct?

umar456:
Its set on Windows if you run the mklvars.bat script before you run cmake. See updated documentation.

9prady9:
We dont need to set it up unless MKL was installed into a non-standard location I suppose.

umar456:
Intel offers MKL with several software packages. For example on my system I have it through system studio 2018 and it is located in the `/opt/intel/system_studio_2018` directory. Instead of hardcoding the path to each package we are setting this to simplify the process.

PR #1881, opened 2017-07-21
FEAT: Image Deconvolution Functions

9prady9:
Iterative deconvolution algorithms
- [x] Landweber
- [x] Richardson-Lucy

Linear deconvolution algorithms
~- [ ] Wiener~
- [x] Tikhonov

Wiener has been disabled for now. I haven't had a break finding the bug in that. We can enable the API in future.

PR #2235, opened 2018-07-09
Remove buffer alloc calls from sparse. Pass params by value

umar456:


9prady9:
@umar456 the windows-opencl job `threading` is failing on this PR only as far as I have seen - have been monitoring the ci jobs for past two days. The job passed fine on deconvolutions PR #1881

PR #2242, opened 2018-07-16
Fix mem leaks in deconvolution tests

9prady9:


test/inverse_deconv.cpp
```diff
@@ -72,13 +72,14 @@ void imageTest(string pTestFile, const float gamma, const af_inverse_deconv_algo
 
         ASSERT_EQ(AF_SUCCESS, af_gaussian_kernel(&kerArray, 13, 13, 2.25, 2.25));
 
+        af_dtype itype = (af_dtype)af::dtype_traits<T>::af_type;
         af_dtype otype = (af_dtype)af::dtype_traits<OutType>::af_type;
 
         ASSERT_EQ(AF_SUCCESS, af_load_image(&_inArray, inFiles[testId].c_str(), isColor));
-        ASSERT_EQ(AF_SUCCESS, conv_image<T>(&inArray, _inArray));
+        ASSERT_EQ(AF_SUCCESS, af_cast(&inArray, _inArray, itype));
```
umar456: it was better earlier because it wasn't using arrayfire functions. you want to test only one function. Otherwise its hard to pinpoint what is failing by looking at the failures.

test/iterative_deconv.cpp
```diff
@@ -33,7 +33,8 @@ typedef ::testing::Types<float, uchar, short, ushort> TestTypes;
 TYPED_TEST_CASE(IterativeDeconvolution, TestTypes);
 
 template<typename T, bool isColor>
-void imageTest(string pTestFile, const unsigned iters, const float rf, const af::iterativeDeconvAlgo algo)
+void imageTest(string pTestFile, const unsigned iters, const float rf,
```
umar456: rename this to a more appropriate function

src/backend/cpu/copy.hpp
```diff
@@ -38,6 +38,7 @@ namespace cpu
                              const dim4& upperBoundPadding,
                              const af::borderType btype)
     {
+        in.eval();
```
umar456: why is this necessary?

9prady9: CPU backend implementations eval inputs prior to enqueue of their respective algorithm. I am not sure whats the fundamental criteria to that ? Followed that convention

umar456: you don't need to do that. Did you run into any issues if you didn't do this?

test/iterative_deconv.cpp
```diff
@@ -33,9 +33,10 @@ typedef ::testing::Types<float, uchar, short, ushort> TestTypes;
 TYPED_TEST_CASE(IterativeDeconvolution, TestTypes);
 
 template<typename T, bool isColor>
-void imageTest(string pTestFile, const unsigned iters, const float rf, const af::iterativeDeconvAlgo algo)
+void iterDeconvImageTest(string pTestFile, const unsigned iters, const float rf,
+                         const af::iterativeDeconvAlgo algo)
 {
-    typedef typename cond_type<is_same_type<T, double>::value, double, float>::type OutType;
+    using OutType = typename cond_type<is_same_type<T, double>::value, double, float>::type;
```
umar456: This is a c++11 thing. you should avoid using that in tests

test/inverse_deconv.cpp
```diff
@@ -33,9 +33,9 @@ typedef ::testing::Types<float, uchar, short, ushort> TestTypes;
 TYPED_TEST_CASE(InverseDeconvolution, TestTypes);
 
 template<typename T, bool isColor>
-void imageTest(string pTestFile, const float gamma, const af_inverse_deconv_algo algo)
+void invDeconvImageTest(string pTestFile, const float gamma, const af_inverse_deconv_algo algo)
 {
-    typedef typename cond_type<is_same_type<T, double>::value, double, float>::type OutType;
+    using OutType = typename cond_type<is_same_type<T, double>::value, double, float>::type;
```
umar456: This is a C++11 style syntax. Avoid using that in tests.

PR #2244, opened 2018-07-17
Fix homography tests and associated methods with it

mark-poscablo:
Related to #2215, which deals with feature detectors. The tests were initially failing because they were using `af::orb()` within the homography extraction process. At this point (fb4cb24), all the other homography tests are passing except for `Tux_RANSAC_90degrees`. We (@umar456, @syurkevi, and I) decided to disable it for now because of a possible bug in the underlying functions (like `af::transformCoordinates()`), and replace it with a simpler test that simply extracts the homography between set of points and its manually rotated version (90 degrees counter-clockwise, done through a simple matrix multiply with a hardcoded transformation matrix). The problem at this point is that on CUDA and OpenCL, the rotated points calculated through `af::homography()` could be off by more or less 2.f compared to the rotated points calculated through the hardcoded transformation matrix.

I initially worked on this, but @umar456 suggested that he take over due to the potentially overreaching nature of the changes required to fix the root causes of this issue. Many thanks to @umar456, @9prady9, and @syurkevi for all the help regarding this PR.

### DIFF
--- a/test/homography.cpp
+++ b/test/homography.cpp
@@ -289,7 +289,7 @@ TEST(Homography, CPP)
 }
 
 TEST(Homography, Rotation_90degCCW) {
-    int nfeats = 200;
+    int nfeats = 100;
 
     // Generate points to rotate
     array in_x = af::randu(nfeats);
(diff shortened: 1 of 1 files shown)


9prady9:
@mark-poscablo This seems to be out of sync with master, please rebase and push. If this PR is not valid anymore, you can close it.

9prady9:
@mark-poscablo Any updates here ?

9prady9:
Closing for now, please reopen when the necessary changes are done and after a rebase from latest master.

PR #2246, opened 2018-07-18
fix memAlloc for opencl backend

syurkevi:
Corrected use case in morph.


9prady9:
Couple of new tests are failing on `linux-opencl-k80-clblas `

PR #2248, opened 2018-07-20
Fix link errors when not compiling without Freeimage/Graphics; Clean up

umar456:
Also:
Added tests for array io operations
Compile tests with -std=c++98 by default
Allocate memory on the stack for JacobiSVD
Remove new/delete and use vector and strings.


test/arrayio.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    ASSERT_EQ(arr.type(), p.type);
+}
+
+TEST_P(ArrayIOType, ReadSize) {
+    type_params p = GetParam();
+    array arr = readArray((string(TEST_DIR) + "/arrayio/" + p.name + ".arr").c_str(), p.name.c_str());
+
+    ASSERT_EQ(arr.dims(), dim4(10, 10));
+}
```
9prady9: Would be nice to have `ReadContents` unit test too.

PR #2249, opened 2018-07-20
Add test utility functions to compare af::arrays

mark-poscablo:
A lot of tests have common code for gtest asserts. One of those common code is comparing two arrays for equality. The usual method is to get the raw data from the backend to the host and then iterate over the two host arrays for element-wise comparison. A new utility function would simplify that so that we could just pass in the af::arrays directly. It will:

- compare types _(done)_
- compare dimension sizes _(done)_
- compare elements for equality _(done, but float types are compared using simple equality for now)_
- display the difference in any of the above _(done)_
- ~~for comparing elements, display an n-dimensional window around the mismatched elements~~
- ~~for floating-point types, use a threshold of machine error to the equality operator or let the test-writer to define that threshold~~
- ~~for images, provide an option to save/display the difference between the reference and output images~~

(Strikethroughed items will be done in a separate PR)

Another utility function would be to simplify the assert for error code-returning C arrayfire calls. Instead of displaying the error code in its raw numeric form, the utility function would display the enum name for that error code, thereby making the error message more human-readable. _(done)_

p.s. I modified `test/basic.cpp` for now to try out how the error messages come out. ~~The tests will fail~~The tests should pass in the CI since we use `ASSERT_FALSE()` on the utility function within the test cases, but when one uses the actual new macro, it should fail and give back an easy-to-read error message.

### DIFF
--- a/test/basic.cpp
+++ b/test/basic.cpp
@@ -28,17 +28,17 @@ TEST(BasicTests, constant1000x1000)
 
     double valA = 3.9;
     af_array a;
-    ASSERT_EQ(AF_SUCCESS, af_constant(&a, valA, ndims, d, f32));
+    ASSERT_SUCCESS(af_constant(&a, valA, ndims, d, f32));
 
     vector<float> h_a(dim_size * dim_size, 100);
-    ASSERT_EQ(AF_SUCCESS, af_get_data_ptr((void **)&h_a[0], a));
+    ASSERT_SUCCESS(af_get_data_ptr((void **)&h_a[0], a));
 
     size_t elements = dim_size * dim_size;
     for(size_t i = 0; i < elements; i++) {
         ASSERT_FLOAT_EQ(valA, h_a[i]);
     }
 
-    ASSERT_EQ(AF_SUCCESS, af_release_array(a));
+    ASSERT_SUCCESS(af_release_array(a));
 }
 
 TEST(BasicTests, constant10x10)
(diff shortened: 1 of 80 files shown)


mark-poscablo:
Now that the basic functionalities of the new test utilities are working, I'm going to try using them in several existing tests we have so far

test/basic.cpp
```diff
@@ -320,3 +321,143 @@ TEST(BasicTests, Additionf32f64_CPP)
...
+    array gold = constant(1, 10, 9);
+    array out = constant(1, 10, 10);
+
+    // Testing this macro
+    // ASSERT_ARRAYS_EQ(gold, out);
+    ASSERT_FALSE(assertArrayEq("gold", "out", gold, out));
+}
+
+TEST(Assert, TestEqualsDiffValue) {
+    // array A = af::randu(3, 3, 3);
```
umar456: remove comment

test/basic.cpp
```diff
@@ -320,3 +321,143 @@ TEST(BasicTests, Additionf32f64_CPP)
...
+    array out = gold;
+    out(2, 2) = 2;
+
+    // Testing this macro
+    // ASSERT_ARRAYS_EQ(gold, out);
+    ASSERT_FALSE(assertArrayEq("gold", "out", gold, out));
+}
+
+TEST(Assert, TestEqualsDiffComplexValue) {
+    // array A = af::randu(3, 3, 3);
```
umar456: Remove comment

test/basic.cpp
```diff
@@ -320,3 +321,143 @@ TEST(BasicTests, Additionf32f64_CPP)
...
+    dim4 adims(3, 2);
+    fill(hA.begin(), hA.end(), 3.1f);
+
+    // Testing this macro
+    // ASSERT_ARRAYS_EQ(A, B);
+    ASSERT_FALSE(assertArrayEq("hA", "adims", "A", hA, adims, A));
+}
+
+TEST(Assert, TestVectorDiffVecSize) {
+    // array A = af::randu(3, 3, 3);
```
umar456: Remove comment

test/basic.cpp
```diff
@@ -320,3 +321,143 @@ TEST(BasicTests, Additionf32f64_CPP)
...
+    fill(gold.begin(), gold.end(), 3.1f);
+
+    // Testing this macro
+    // ASSERT_VEC_ARRAY_EQ(gold, goldDims, out);
+    ASSERT_FALSE(assertArrayEq("gold", "goldDims", "out",
+                               gold, goldDims, out));
+}
+
+TEST(Assert, TestVectorDiffDim4) {
+    // array A = af::randu(3, 3, 3);
```
umar456: Remove comment

test/testHelpers.hpp
```diff
@@ -465,4 +485,298 @@ void cleanSlate()
...
+    case s64: name = "s64"; break;
+    case u64: name = "u64"; break;
+    case s16: name = "s16"; break;
+    case u16: name = "u16"; break;
+    default: assert(false && "Invalid type");
+    }
+    return os << name;
+}
+
+af::dim4 unravelIdx(uint idx, af::dim4 dims, af::dim4 strides) {
```
umar456: A comment here explaining the function would be nice.

PR #2251, opened 2018-07-25
Increment forge version

9prady9:
I am still contemplating if we should move this version increment to 3.6.2.


9prady9:
There are couple of issues I am debugging on Windows w.r.t freetype dependency of forge. Once that is resolved I shall update the PR for merge.

PR #2253, opened 2018-07-26
Add bigobj EHsc and FS flags to CUDA_NVCC_FLAGS and test executables

umar456:


9prady9:
I will enqueue windows jobs once the test runs on another PR are done successfully.

PR #2264, opened 2018-08-07
spdlog v1.0.0 added as part of ArrayFire by default.

mlloreda:


src/backend/common/Logger.hpp
```diff
@@ -11,23 +11,7 @@
 
 #include <memory>
 #include <string>
-
-#ifdef AF_WITH_LOGGING
 #include <spdlog/spdlog.h>
```
umar456: You can forward declare spdlog::logger instead of including the entire header.

mlloreda: Good catch - addressed.

Issue #2143, opened 2018-04-27
Least squares


FloopCZ:
Hi everyone,

I would like to know what is the "official" way of calculating least squares approximation of a system of linear equations, i.e., `Ax = b` for non-square `A`.

Up to this moment, I believed that when `af::solve` gets an overdetermined system of equations (e.g., `af::solve(af::randu(4, 2), af::randu(4))`), it produces a least squares approximation of the coefficients (because in the documentation, it is in a section called [Linear solve and least squares](http://arrayfire.org/docs/group__lapack__solve__func__gen.htm)). Unfortunately, I just performed some tests and it is not the case. My questions are:

1. Is there a recommended way to solving an overdetermined system of linear equations?
2. What do the numbers computed by `af::solve` mean? Shouldn't it rather throw an exception if this functionality is not supported?

Notes:
- `af::inverse` does not work for non-squared matrices.
- Relates to #2074 and #2077.
- Thanks @Myhroun for detecting the issue in the first place.

Thank you and have a nice weekend!

pavanky:
@FloopCZ do you have some code to reproduce this ?

FloopCZ:
@pavanky Of course. :-)

The problem is when the columns are not linearly independent. Here is a minimal working example (partially by @Myhroun):

```c++
#include <arrayfire.h>
#include <iostream>

float mse(const af::array& a, const af::array& b)
{
    return af::mean(af::pow(af::flat(a) - af::flat(b), 2)).scalar<float>();
}

int main()
{
    af::setSeed(1000003);
    af::info();

    // float a_[] = {1, 1, 1, 2, 2, 1.9};      // Works fine on all backends.
    // float a_[] = {1, 1, 1, 2, 2, 2};        // Works only on OpenCL.
                                               // - CPU does not minimize MSE.
                                               // - Cuda does not minimize MSE.
    float a_[] = {1, 2, 3, 1, 1, 1, 2, 2, 2};  // Does not work at all:
                                               // - CPU does not minimize MSE.
                                               // - OpenCL crash.
                                               // - Cuda infs and nans.
    af::array a{3, std::size(a_) / 3, a_};
    af::print("a", a);

    float b_[] = {1, 2, 3};
    af::array b{3, b_};
    af::print("b", b);

    // Find x such that ax = b minimizes MSE.
    af::array x = af::solve(a, b);
    double x_mse = mse(af::matmul(a, x), b);
    af::print("x", x);
[...]

Issue #2077, opened 2018-03-02
Inverse non squared matrix


kimoes:
Hallo, 
I was told that inverse function in arrayfire works for square and non square matrix 
But, when I use it with non squared matrix it show me an error:

Inv = af.inverse(a)

in inverse safe_call(backend.get().af_inverse(c_pointer(AI.arr), A.arr, options.value))

in safe_call raise RuntimeError(to_str(err_str))

RuntimeError: In function af_err af_inverse(void**, af_array, af_mat_prop)
In file src/api/c/inverse.cpp:43
Invalid dimension for argument 1
Expected: i_info.dims()[0] == i_info.dims()[1]

The code is:
import arrayfire as af
import numpy as np

Row = 12
Col = 10

a = np.random.randn(Row, Col).astype('float32')
A = af.to_array(a) 

Inv = af.inverse(A)

rstub:
> I was told that inverse function in arrayfire works for square and non square matrix

Where did you get this? [The documentation](http://arrayfire.org/docs/group__lapack__ops__func__inv.htm) clearly states "square matrix". It might be interesting to have something like https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse in ArrayFire, though.

kimoes:
From Pavan Yalamanchili in another issue where he answered me this, then closed it.

Yes, I agree as getting the inverse of a matrix using SVD works but very slow

rstub:
That would be #2074. But note that many algorithms for the pseudo-inverse rely on SVD.

kimoes:
Yes, indeed.

yes, I understand that but when I compared between the SVD function in array fire (af.svd()) and the one in numpy (np.linalg.svd()) on a big matrix like(6000 x 5000), the numpy one is faster and the np.linalg.pinv () is faster than both. while the logic says the SVD by arrayfire should be the fastest as it would be performed using the gpu.

btw I am a beginner in cuda

mark-poscablo:
Hi @kimoes and @rstub, we have just recently implemented pseudoinverse. It's available now if you build ArrayFire from source, but it should also come with the next release with the installers. Please open a new Issue if you have any questions/feedback on it. Thanks.

Issue #2074, opened 2018-02-28
moore penrose inverse


kimoes:
Hello, I wanted to ask if Matrix pseudo inverse is supported by arrayfire as I have seen some documents using arrayfire.pinv while I cant use it as its not supported as the arrayfire.inverse for example

pavanky:
@kimoes The regular inverse should do pseudo inverse for non square matrices.

kimoes:
when I use it with non squared matrix it show me an error:

 Inv = af.inverse(a)

in inverse safe_call(backend.get().af_inverse(c_pointer(AI.arr), A.arr, options.value))

in safe_call raise RuntimeError(to_str(err_str))

RuntimeError: In function af_err af_inverse(void**, af_array, af_mat_prop)
In file src/api/c/inverse.cpp:43
Invalid dimension for argument 1
Expected: i_info.dims()[0] == i_info.dims()[1]


The code is:
import arrayfire as af
import numpy as np

Row = 12
Col = 10

a = np.random.randn(Row, Col).astype('float32')
a = af.Array(a.ctypes.data, a.shape, a.dtype.char)

Inv = af.inverse(a)

PR #2279, opened 2018-08-14
Implement af::pinverse()

mark-poscablo:
SVD-based Moore-Penrose pseudoinverse.

This has the same type restrictions as `af::inverse()`, but lifts the dims restriction of requiring the array to be square. The third dimension is also allowed for batching, although this uses naive batching for now (loop through and successively process each slice along the third dimension). No backend-specific code were added, since all the necessary pieces for the SVD-based approach are already implemented in arrayfire (thus all the code is in `src/api/c/pinverse.cpp`). 

The tests include checking if all the Moore-Penrose conditions hold (see https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse#Definition), for all the the appropriate types (f32, f64, c32, c64).

The CI tests might fail initially until the corresponding [PR on arrayfire-data](https://github.com/arrayfire/arrayfire-data/pull/11) is approved and I update the git submodule reference here. 

This addresses #2074, #2077, and #2143 (this last one indirectly).

### DIFF
--- a/docs/details/lapack.dox
+++ b/docs/details/lapack.dox
@@ -252,6 +252,31 @@ I [3 3 1 1]
 
 \endcode
 
+=======================================================================
+
+\defgroup lapack_ops_func_pinv pinverse
+
+\ingroup lapack_ops_mat
+
+\brief Pseudo-invert a matrix
+
+This function calculates the Moore-Penrose pseudoinverse of a matrix \f$A\f$,
+using \ref af::svd at its core. If \f$A\f$ is of size \f$M \times N\f$, then its
+pseudoinverse \f$A^+\f$ will be of size \f$N \times M\f$.
+
+This calculation can be batched if the input array is three or four-dimensional
+\f$(M \times N \times P \times Q\f$, with \f$Q=1\f$ for only three dimensions
+\f$)\f$. Each \f$M \times N\f$ slice along the third dimension will have its own
+pseudoinverse, for a total of \f$P \times Q\f$ pseudoinverses in the output array
+\f$(N \times M \times P \times Q)\f$.
+
+Here's an example snippet of its usage. In this example, we have a matrix \f$A\f$
+and we compute its pseudoinverse \f$A^+\f$. This condition must hold:
+\f$AA^+A=A\f$, given that the two matrices are pseudoinverses of each other (in
+fact, this is one of the Moore-Penrose conditions):
+
+\snippet test/pinverse.cpp ex_pinverse
+
 ==================================================================================
 
 \defgroup lapack_ops_func_rank rank
--- a/include/af/lapack.h
+++ b/include/af/lapack.h
@@ -200,6 +200,28 @@ namespace af
     */
     AFAPI array inverse(const array &in, const matProp options = AF_MAT_NONE);
 
+#if AF_API_VERSION >= 37
+    /**
+       C++ Interface for pseudo-inverting (Moore-Penrose) a matrix.
+       Currently uses the SVD-based approach.
+
+       \param[in] in is the input matrix
+       \param[in] tol defines the lower threshold for singular values from SVD
+       \param[in] options must be AF_MAT_NONE (more options might be supported
+                  in the future)
+       \returns the pseudo-inverse of the input matrix
+
+       \note \p tol is not the actual lower threshold, but it is passed in as
+             a parameter to the calculation of the actual threshold relative to
+             the shape and contents of \p in.
+       \note This function is not supported in GFOR
+
+       \ingroup lapack_ops_func_pinv
+    */
+    AFAPI array pinverse(const array &in, const double tol=1E-6,
+                         const matProp options = AF_MAT_NONE);
+#endif
+
     /**
        C++ Interface for finding the rank of a matrix
 
@@ -400,6 +422,30 @@ extern "C" {
        \note currently options needs to be \ref AF_MAT_NONE
     */
     AFAPI af_err af_inverse(af_array *out, const af_array in, const af_mat_prop options);
+
+#if AF_API_VERSION >= 37
+    /**
+       C Interface for pseudo-inverting (Moore-Penrose) a matrix.
+       Currently uses the SVD-based approach.
+
+       \param[out] out will contain the pseudo-inverse of matrix \p in
+       \param[in] in is the input matrix
+       \param[in] tol defines the lower threshold for singular values from SVD
+       \param[in] options must be AF_MAT_NONE (more options might be supported
+       in the future)
+
+       \note \p tol is not the actual lower threshold, but it is passed in as a
+             parameter to the calculation of the actual threshold relative to the
+             shape and contents of \p in.
+       \note At first, try setting \p tol to 1e-6 for single precision and 1e-12
+             for double.
+       \note This function is not supported in GFOR
+
+       \ingroup lapack_ops_func_pinv
+    */
+    AFAPI af_err af_pinverse(af_array *out, const af_array in, const double tol,
+                             const af_mat_prop options);
+#endif
 
     /**
        C Interface for finding the rank of a matrix
--- a/src/api/c/pinverse.cpp
+++ b/src/api/c/pinverse.cpp
@@ -1,0 +1,199 @@
+/*******************************************************
+ * Copyright (c) 2018, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#include <backend.hpp>
+#include <Array.hpp>
+
+#include <af/array.h>
+#include <af/complex.h>
+#include <af/defines.h>
+#include <af/lapack.h>
+#include <arith.hpp>
+#include <blas.hpp>
+#include <cast.hpp>
+#include <common/ArrayInfo.hpp>
+#include <common/err_common.hpp>
+#include <diagonal.hpp>
+#include <handle.hpp>
+#include <logic.hpp>
+#include <math.hpp>
+#include <reduce.hpp>
+#include <select.hpp>
+#include <svd.hpp>
+#include <tile.hpp>
+#include <transpose.hpp>
+
+using af::dim4;
+using af::dtype_traits;
+using std::vector;
+using std::swap;
+
+using namespace detail;
+
+const double dfltTol = 1e-6;
+
+template<typename T>
+Array<T> getSubArray(const Array<T> &in, const bool copy,
+                       uint dim0begin = 0, uint dim0end = 0,
+                       uint dim1begin = 0, uint dim1end = 0,
+                       uint dim2begin = 0, uint dim2end = 0,
+                       uint dim3begin = 0, uint dim3end = 0) {
+    vector<af_seq> seqs = {
+        {static_cast<double>(dim0begin), static_cast<double>(dim0end), 1.},
+        {static_cast<double>(dim1begin), static_cast<double>(dim1end), 1.},
+        {static_cast<double>(dim2begin), static_cast<double>(dim2end), 1.},
+        {static_cast<double>(dim3begin), static_cast<double>(dim3end), 1.}
+    };
+    return createSubArray<T>(in, seqs, copy);
+}
+
+// Moore-Penrose Pseudoinverse
+template<typename T>
+Array<T> pinverseSvd(const Array<T> &in, const double tol)
+{
+    in.eval();
+    int M = in.dims()[0];
+    int N = in.dims()[1];
+    int P = in.dims()[2];
+    int Q = in.dims()[3];
+
+    // Compute SVD
+    typedef typename dtype_traits<T>::base_type Tr;
+    // Ideally, these initializations should use createEmptyArray(), but for some
+    // reason, linux-opencl-k80 will produce wrong results for large arrays
+    Array<T> u = createValueArray<T>(dim4(M, M, P, Q), scalar<T>(0));
+    Array<T> vT = createValueArray<T>(dim4(N, N, P, Q), scalar<T>(0));
+    Array<Tr> sVec = createValueArray<Tr>(dim4(min(M, N), 1, P, Q), scalar<Tr>(0));
+    for (uint j = 0; j < Q; ++j) {
+        for (uint i = 0; i < P; ++i) {
+            Array<T> inSlice = getSubArray(in, false,
+                                           0, M - 1,
+                                           0, N - 1,
+                                           i, i,
+                                           j, j);
+            Array<Tr> sVecSlice = getSubArray(sVec, false,
+                                              0, sVec.dims()[0] - 1,
+                                              0, 0,
+                                              i, i,
+                                              j, j);
+            Array<T> uSlice = getSubArray(u, false,
+                                          0, u.dims()[0] - 1,
+                                          0, u.dims()[1] - 1,
+                                          i, i,
+                                          j, j);
+            Array<T> vTSlice = getSubArray(vT, false,
+                                           0, vT.dims()[0] - 1,
+                                           0, vT.dims()[1] - 1,
+                                           i, i,
+                                           j, j);
+            svd<T, Tr>(sVecSlice, uSlice, vTSlice, inSlice);
+        }
+    }
+
+    // Cast s back to original data type for matmul later
+    // (since svd() makes s' type the base type of T)
+    Array<T> sVecCast = cast<T, Tr>(sVec);
+
+    Array<T> v = transpose(vT, true);
+
+    // Build relative tolerance array
+    Array<Tr> sVecMax = reduce<af_max_t, Tr, Tr>(sVec, 0);
+    Array<T> sVecMaxCast = cast<T, Tr>(sVecMax);
+    double tolMulShape = tol * static_cast<double>(max(M, N));
+    Array<T> tolMulShapeArr = createValueArray<T>(sVecMaxCast.dims(),
+                                                  scalar<T>(tolMulShape));
+    Array<T> relTol = arithOp<T, af_mul_t>(tolMulShapeArr, sVecMaxCast,
+                                           sVecMaxCast.dims());
+    Array<T> relTolArr = tile<T>(relTol, dim4(sVecCast.dims()[0]));
+
+    // Get reciprocal of sVec's non-zero values for s pinverse, except for
+    // very small non-zero values though (< relTol), in order to avoid very
+    // large reciprocals
+    Array<T> ones = createValueArray<T>(sVecCast.dims(), scalar<T>(1.));
+    Array<T> sVecRecip = arithOp<T, af_div_t>(ones, sVecCast, sVecCast.dims());
+    Array<char> cond = logicOp<T, af_ge_t>(sVecCast, relTolArr, sVecCast.dims());
+    Array<T> zeros = createValueArray<T>(sVecCast.dims(), scalar<T>(0.));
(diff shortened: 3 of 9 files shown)


src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // (since svd() makes s' type the base type of T)
+    Array<T> sPinvCast = cast<T, Tr>(sPinv);
+
+    Array<T> uT = transpose(u, true);
+
+    // Crop v and u* for final matmul later based on s+'s size, because
+    // sVec produced by svd() has minimal dim length (no extra zeroes).
+    // Thus s+ produced by diagCreate() will have minimal dims as well,
+    // and v could have an extra dim0 or u* could have an extra dim1
+    if (v.dims()[1] > sPinvCast.dims()[0]) {
```
mark-poscablo: By the way, I'm wondering if this is the fastest implementation of what I'm trying to do in this section. At first, actually, I thought of avoiding creating new arrays (`createSubArray()`) since this might be expensive especially when the arrays are big. Instead, a reference to only the needed section of the arrays might be better to use - although this could also have the disadvantage of making the data access pattern jump near the cropped-out section of the arrays during `matmul()`.

Another way could be padding `sVec` with 0's instead to match v and u*'s dims for matmul. This creates a new array once (and a smaller one compared to the other arrays), but will waste multiplication and adding with 0's during matmul.

mark-poscablo:
I've been benchmarking this vs just `af::svd()` and `af::svdInPlace()`, and I'm thinking of adding an in-place option for pinverse too - `af::svdInPlace()` is actually really fast on CUDA. However, it has a couple of issues - I'll write about it in a separate issue.

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,134 @@
...
+    typedef typename af::dtype_traits<T>::base_type Tr;
+    Array<Tr> sVec = createEmptyArray<Tr>(dim4(min(M, N)));
+    Array<T> u = createEmptyArray<T>(dim4(M, M));
+    Array<T> vT = createEmptyArray<T>(dim4(N, N));
+    svd<T, Tr>(sVec, u, vT, in);
+
+    Array<T> v = transpose(vT, true);
+
+    // Round down small values to zero to avoid large reciprocals later
+    Array<Tr> eps = createValueArray<Tr>(sVec.dims(), scalar<Tr>(1e-6));
```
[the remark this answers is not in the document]
mark-poscablo: Implemented that relative tolerance calculation but epsilon will be 1e-6 by default instead of machine epsilon. There's an issue with SVD that prevents me from lowering this for float. I should probably write about that issue, but basically if I SVD a given array, make one of the singular values really small (1e-12 perhaps), multiply them back together, and SVD it again, the really small singular value isn't so small anymore (becomes something on the order of 1e-5). Thus I think our SVD cannot produce really small singular values and so I set the default tolerance that high.

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (v.dims()[1] > sPinvCast.dims()[0]) {
+        std::vector<af_seq> seqs = {
+            af_span,
+            {0., static_cast<double>(sPinvCast.dims()[0]), 1.}
+        };
+        v = createSubArray<T>(v, seqs);
+    }
+    if (uT.dims()[0] > sPinvCast.dims()[1]) {
+        std::vector<af_seq> seqs = {
+            {0., static_cast<double>(sPinvCast.dims()[1]), 1.},
```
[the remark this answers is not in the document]
mark-poscablo: :+1:

include/af/lapack.h
```diff
@@ -200,6 +200,25 @@ namespace af
     */
     AFAPI array inverse(const array &in, const matProp options = AF_MAT_NONE);
 
+    /**
```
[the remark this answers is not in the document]
mark-poscablo: Adding this now

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,154 @@
...
+        ARG_ASSERT(1, i_info.isFloating()); // Only floating and complex types
+
+        af_array output;
+
+        if(i_info.ndims() == 0) {
+            return af_retain_array(out, in);
+        }
+
+        double validTol = tol;
+        if (validTol < 0.) {
```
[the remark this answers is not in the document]
mark-poscablo: :+1:

pavanky:
@mark-poscablo @syurkevi Try to implement this for non-square matrices inside inverse rather than adding a new function.

umar456:
@pavanky The extra parameters are not necessary for the regular inverse. We can make this part of the regular inverse and pick good defaults so that inverse will also perform a pseudo inverse if the matrices are not square.

mark-poscablo:
@pavanky I was actually planning to do this at first, but @syurkevi advised me early on that the user most probably wants to explicitly differentiate the pseudoinverse from the regular inverse. But true, we could add it as a fallback for regular inverse in case the matrices aren't square - if we do this though I think we should add a boolean parameter to `af::inverse` that decides to allow fallback to pseudoinverse or not, setting it default to false to keep the original API's behavior. Reason is that the user might still want to be alerted that the input matrix isn't square, versus quietly selecting pseudoinverse if that's the case. At least with that extra parameter, the user will intentionally set it to true if they really want that quiet selection of pseudoinverse.

include/af/lapack.h
```diff
@@ -200,6 +200,26 @@ namespace af
...
+       C++ Interface for pseudo-inverting (Moore-Penrose) a matrix.
+       Currently uses the SVD-based approach.
+
+       \param[in] in is the input matrix
+       \param[in] tol defines the lower threshold for singular values from SVD
+       \param[in] options determining various properties of matrix \p in
+       \returns \p x, the inverse of the input matrix
+
+       \note \p tol is not the actual lower threshold, but it is passed in as a parameter to the calculation of the actual threshold relative to the shape and contents of \p in.
```
umar456: 80 characters per line. You can tab to the first character of the comment to make it more readable:

```

 \note \p tol is not the actual lower threshold, but it is passed in as a
       parameter to the calculation of the actual threshold relative to the
       shape and contents of \p in.
```

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
umar456: This is a new file so you should set this to 2018

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+using namespace detail;
+
+const double dfltTol = 1e-6;
+
+template<typename T>
+Array<T> pinverseSvd(const Array<T> &in, const double tol)
+{
+    // Moore-Penrose Pseudoinverse
+
```
umar456: Nit: Extra line

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        af_dtype type = i_info.getType();
+
+        if (options != AF_MAT_NONE) {
+            AF_ERROR("Using this property is not yet supported in inverse", AF_ERR_NOT_SUPPORTED);
+        }
+
+        ARG_ASSERT(1, i_info.isFloating()); // Only floating and complex types
+        ARG_ASSERT(2, tol >= 0.); // Only floating and complex types
```
umar456: The comment is incorrect here.

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+af_err af_pinverse(af_array *out, const af_array in, const double tol,
+                   const af_mat_prop options)
+{
+    try {
+        const ArrayInfo& i_info = getInfo(in);
+
+        if (i_info.ndims() > 3) {
```
umar456: Is this check still necessary? It looks like you are able to handle batched operations.

test/pinverse.cpp
```diff
@@ -0,0 +1,327 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
umar456: 2018

test/pinverse.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
+
+TEST(Pinverse, InvalidMatProp) {
+    array in = constant(0.f, 10, 8, f32);
+    array out;
+    ASSERT_THROW(out = pinverse(in, -1.f, AF_MAT_SYM), exception);
+}
+
+TEST(Pinverse, DocSnippet) {
+    float hA[] = {0, 1, 2, 3, 4, 5};
```
umar456: This array should be part of the example. Otherwise, the user doesn't have any context for hA.

test/pinverse.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    // Make new input array that has a small non-zero value in its SVD sigma
+    in = af::matmul(u, s, vT);
+    array inpinv = pinverse(in);
+    array out = matmul(in, inpinv, in);
+
+    ASSERT_ARRAYS_NEAR(in, out, eps<float>());
+}
+
+TEST(Pinverse, SmallSigValExistsDouble) {
```
umar456: This and the previous tests are almost identical. It would be nice if the logic was combined. Not important if it will take a long time.

test/pinverse.cpp
```diff
@@ -0,0 +1,327 @@
...
+    return std::numeric_limits<InBaseType>::epsilon()
+        * std::max(in.dims(0), in.dims(1)) * af::max<double>(in);
+}
+
+typedef ::testing::Types<float, cfloat, double, cdouble> TestTypes;
+TYPED_TEST_CASE(Pinverse, TestTypes);
+
+// Test Moore-Penrose conditions
+// See https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse#Definition
+
```
umar456: Nit: Extra line

docs/details/lapack.dox
```diff
@@ -252,6 +252,22 @@ I [3 3 1 1]
...
+
+\defgroup lapack_ops_func_pinv pinverse
+
+\ingroup lapack_ops_mat
+
+\brief Pseudo-invert a matrix
+
+This function calculates the Moore-Penrose pseudoinverse of a matrix \f$A\f$, using \ref af::svd at its core. If \f$A\f$ is of size **M x N**, then its pseudoinverse \f$A^+\f$ will be of size **N x M**.
+
+This calculation can be batched if the input array is three-dimensional (**M x N x P**). Each **M x N** slice along the third dimension will have its own pseudoinverse, for a total of **P** pseudoinverses in the output array (**N x M x P**).
```
umar456: Instead of making the M x N x P statement bold perhaps use \f$M \times N \times P$\f. 

<a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;M&space;\times&space;N&space;\times&space;P" target="_blank"></a>

include/af/lapack.h
```diff
@@ -200,6 +200,26 @@ namespace af
...
 
+#if AF_API_VERSION >= 37
+    /**
+       C++ Interface for pseudo-inverting (Moore-Penrose) a matrix.
+       Currently uses the SVD-based approach.
+
+       \param[in] in is the input matrix
+       \param[in] tol defines the lower threshold for singular values from SVD
+       \param[in] options determining various properties of matrix \p in
+       \returns \p x, the inverse of the input matrix
```
umar456: I don't see an x value in the function signature. just use
\returns the inverse of the input matrix

include/af/lapack.h
```diff
@@ -200,6 +200,26 @@ namespace af
...
     AFAPI array inverse(const array &in, const matProp options = AF_MAT_NONE);
 
+#if AF_API_VERSION >= 37
+    /**
+       C++ Interface for pseudo-inverting (Moore-Penrose) a matrix.
+       Currently uses the SVD-based approach.
+
+       \param[in] in is the input matrix
+       \param[in] tol defines the lower threshold for singular values from SVD
+       \param[in] options determining various properties of matrix \p in
```
umar456: Singular option.

Perhaps state that this must be AF_MAT_NONE

\param[in] option must be AF_MAT_NONE. For future use.

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    Array<T> out = matmul<T>(matmul<T>(v, sPinv, AF_MAT_NONE, AF_MAT_NONE),
+                             uT, AF_MAT_NONE, AF_MAT_NONE);
+
+    return out;
+}
+
+template<typename T>
+static inline af_array pinverse(const af_array in, const double tol)
```
umar456: I think you should think about moving the batching logic into the pinverseSvd function. You should use this intermediate function to convert from af_array to Array<T> objects. This makes the internal API more usable in other functions because it has the same behavior as the public API. 

So in this case this pinverse function should look like this:

```
template<typename T>
af_array pinverse(const af_array in, const double tol) {
  return getHandle(pinverse(getHandle<T>(in), tol)));
}
```

docs/details/lapack.dox
```diff
@@ -252,6 +252,22 @@ I [3 3 1 1]
...
 
+=======================================================================
+
+\defgroup lapack_ops_func_pinv pinverse
+
+\ingroup lapack_ops_mat
+
+\brief Pseudo-invert a matrix
+
+This function calculates the Moore-Penrose pseudoinverse of a matrix \f$A\f$, using \ref af::svd at its core. If \f$A\f$ is of size **M x N**, then its pseudoinverse \f$A^+\f$ will be of size **N x M**.
```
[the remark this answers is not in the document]
mark-poscablo: Yup I followed suit on the other functions' existing docs, thinking that maybe doxygen somehow requires everything on a single line or else an unnecessary line break will occur. I'll change it to 80 chars per line though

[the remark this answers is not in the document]
mark-poscablo: Alright it's good, it automatically adds a space for every new line i add on the .dox file.

include/af/lapack.h
mark-poscablo: Same case as above for this file, just added line breaks now (to pinverse and af_pinverse)

src/api/c/pinverse.cpp
mark-poscablo: Makes sense, I just changed this one and test/pinverse.cpp

src/api/c/pinverse.cpp
mark-poscablo: Here I intentionally put an extra line because the comment doesn't apply only to the code block right below it, but rather the whole function. Maybe I should put this comment above the signature instead eh?

src/api/c/pinverse.cpp
mark-poscablo: Whoops copy paste. Just changed the comment.

test/pinverse.cpp
mark-poscablo: Same reasoning as in api/c/pinverse.cpp but I just removed the extra line here and clarified that this comment just applies to the first 4 following tests

docs/details/lapack.dox
mark-poscablo: Yup I was on the fence about this one, since other functions' docs (like af::svd) just usually put bold. But it's probably better to be consistent within a doc page. I'll change this.

include/af/lapack.h
mark-poscablo: Yup just changed it.

src/api/c/pinverse.cpp
mark-poscablo: Ah I understand, yep that makes sense let me move the batching logic indeed.

src/api/c/pinverse.cpp
mark-poscablo: Even with batching, I don't handle arrays with 4 dimensions, just 3. Not sure how to go about that case, maybe we could talk more about this.

test/pinverse.cpp
mark-poscablo: Yup got it

test/pinverse.cpp
mark-poscablo: Oh whoops yup let me include it

src/api/c/pinverse.cpp
umar456: It would be better to apply that comment to the whole function. Maybe move it above the function

src/api/c/pinverse.cpp
umar456: You can flatten the 3rd and 4th dimension and then perform the inverse. so reshape it to (M, N, d3 * d4, 1)

mark-poscablo: Welp I guess I could extend batching to 4 dimensions - I simply have to enclose my current batching logic in another for loop for each 3D slice, collect all of them, and finally join them along the 4th dimension

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+    else {
+        vector<Array<T> > outputs;
+        for (int i = 0; i < batchSize; ++i) {
+            vector<af_seq> seqs = {
+                {0., static_cast<double>(inArray.dims()[0] - 1), 1.},
+                {0., static_cast<double>(inArray.dims()[1] - 1), 1.},
+                {static_cast<double>(i), static_cast<double>(i), 1.}
+            };
+            Array<T> inSlice = createSubArray<T>(inArray, seqs);
```
umar456: Can you make sure that this function is indeed making a subarray and not allocating and performing a copy? If it's copying an array then we should move this operation into the svd function and then iterate over the arrays there. You can just create a TODO if we are doing a copy and we can fix it at a later time. I suspect it is working as expected but I just want to make sure

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    // Compute SVD
+    typedef typename dtype_traits<T>::base_type Tr;
+    // Ideally, these initializations should use createEmptyArray(), but for some
+    // reason, linux-opencl-k80 will produce wrong results for large arrays
+    Array<T> u = createValueArray<T>(dim4(M, M, P, Q), scalar<T>(0));
+    Array<T> vT = createValueArray<T>(dim4(N, N, P, Q), scalar<T>(0));
+    Array<Tr> sVec = createValueArray<Tr>(dim4(min(M, N), 1, P, Q), scalar<Tr>(0));
+    for (uint j = 0; j < Q; ++j) {
+        for (uint i = 0; i < P; ++i) {
```
9prady9: Not sure if it will effect performance, but this double-loop can be merged into a single loop in this case.

src/api/c/pinverse.cpp
```diff
@@ -0,0 +1,199 @@
...
+    int P = in.dims()[2];
+    int Q = in.dims()[3];
+
+    // Compute SVD
+    typedef typename dtype_traits<T>::base_type Tr;
+    // Ideally, these initializations should use createEmptyArray(), but for some
+    // reason, linux-opencl-k80 will produce wrong results for large arrays
+    Array<T> u = createValueArray<T>(dim4(M, M, P, Q), scalar<T>(0));
+    Array<T> vT = createValueArray<T>(dim4(N, N, P, Q), scalar<T>(0));
+    Array<Tr> sVec = createValueArray<Tr>(dim4(min(M, N), 1, P, Q), scalar<Tr>(0));
```
9prady9: I understand the reason behind using createValueArray, but wondering if it justifies having three memsets because of that.

Does this happen only on k80 with OpenCL backend ?

mark-poscablo: Yes it only happens on the K80 with the OpenCL backend. I wish I knew how to solve it another way.

9prady9: I get the situation, lets deal with a work-around if at all possible later.

PR #2280, opened 2018-08-14
NearestNeighbor(s)

syurkevi:
Adds ability to return multiple values from nearest neighbor using topk functionality.

### DIFF
--- a/include/af/vision.h
+++ b/include/af/vision.h
@@ -223,7 +223,8 @@ AFAPI void hammingMatcher(array& idx, array& dist,
    \param[in]  train is the array containing the data used as training data
    \param[in]  dist_dim indicates the dimension to analyze for distance (the dimension
                indicated here must be of equal length for both query and train arrays)
-   \param[in]  n_dist is the number of smallest distances to return
+   \param[in]  n_dist is the number of smallest distances to return (currently only 
+               values <= 256 are supported)
    \param[in]  dist_type is the distance computation type. Currently \ref AF_SAD (sum
                of absolute differences), \ref AF_SSD (sum of squared differences), and
                \ref AF_SHD (hamming distances) are supported.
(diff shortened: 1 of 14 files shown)


9prady9:
@syurkevi Is this ready for review or you have more changes coming in ?

syurkevi:
> @syurkevi Is this ready for review or you have more changes coming in ?

Should be good to go.

9prady9:
@syurkevi @umar456 is going to limit the `k` from `topk` in #2291 and that is going to effect your changes and will limit the usability of nearest neighbor, wont it ?

include/af/vision.h
```diff
@@ -223,8 +223,7 @@ AFAPI void hammingMatcher(array& idx, array& dist,
    \param[in]  train is the array containing the data used as training data
    \param[in]  dist_dim indicates the dimension to analyze for distance (the dimension
                indicated here must be of equal length for both query and train arrays)
-   \param[in]  n_dist is the number of smallest distances to return (currently, only 1
-               is supported)
+   \param[in]  n_dist is the number of smallest distances to return
```
9prady9: This should add back the statement about limitation of `n_dist` as 256.

src/backend/cpu/nearest_neighbour.cpp
```diff
@@ -26,35 +27,38 @@ void nearest_neighbour(Array<uint>& idx, Array<To>& dist,
                        const uint dist_dim, const uint n_dist,
                        const af_match_type dist_type)
 {
-    if (n_dist > 1) {
-        CPU_NOT_SUPPORTED("\nNumber of smallest distances can't be <1\n");
-    }
```
9prady9: This check should be present atleast in CUDA backend where there is a limitation of `n_dist`, 256 because of topk.

syurkevi:
> @syurkevi @umar456 is going to limit the `k` from `topk` in #2291 and that is going to effect your changes and will limit the usability of nearest neighbor, wont it ?
> 
> Also, did you compare the performance with the current change w.r.t to older code ? We wouldn't want the performance to drop. Current `topk` implementation isn't optimized for all larger values of `k`. If you have some performance numbers, please do post them here.

There is a slight slowdown for k=1 on the cuda backend. With larger feature sizes and number of query/training examples it seems to be negligible.
With respect to larger values of k there are no performance changes since they have not been supported until this PR.
There is a new PR in the arrayfire-benchmarks repo if you are interested in concrete numbers.

src/backend/cpu/nearest_neighbour.cpp
syurkevi: Added as an ARG_ASSERT to api/c/ since change affects all backends

PR #2290, opened 2018-09-07
Fixes to improve the usage of ArrayFire as a subproject

umar456:


9prady9:
@umar456 Are those all such changes ? Just making we didn't miss anything anywhere else like the cmake lists file under locations like sort_by_key, scan_by_key.

umar456:
I have been able to compile ArrayFire as a subproject targeting the CPU and CUDA backends with this change. I will test OpenCL soon.

umar456:
OpenCL seems to be working as well.

PR #2298, opened 2018-09-22
Fix and improve af::accum documentation

mark-poscablo:
The accum docs has some typos, such as saying "exclusive sum" on the function docs while it is really an inclusive sum scan. Also I think the brief description needs to say "cumulative" somewhere - "inclusive sum (scan)" didn't really ring a bell to me until I started learning parallel programming (a Google search on "inclusive sum" also doesn't immediately get clear results that it means inclusive sum scan). I added an example too just so it's clearer what it does.

Here are some snapshots of the docs after the change:
image
image
image

### DIFF
--- a/docs/details/algorithm.dox
+++ b/docs/details/algorithm.dox
@@ -107,15 +107,21 @@ Return type is u32 for all input types
 
 \ingroup scan_mat
 
-Compute the cumulative sum (inclusive) along the specified dimension
+Calculate the cumulative sum (inclusive) along the specified dimension
 
-Here is a simple example for 1D arrays:
+For a 1D array \f$X\f$, the inclusive cumulative sum calculates \f$x_i =
+\sum_{p=0}^{i}x_p\f$ for every \f$x \in X\f$. Here is a simple example for the
+1D case:
 
 \snippet test/scan.cpp ex_accum_1D
 
 For 2D arrays (and higher dimensions), you can specify the dimension along which
-the cumulative sum will be performed. If none is specified, it will be performed
-along the first dimension (0):
+the cumulative sum will be calculated. Thus, the formula above will be
+calculated for all array slices along the specified dimension (in the 2D case
+for example, this looks like \f$x_{i,j} = \sum_{p=0}^{j}x_{i,p}\f$ if the second
+dimension (dim1) was chosen). If no dimension is specified, then the first
+dimension (dim0) is used by default (only in the C++ API; the dimension is
+required to be specified in the C API):
 
 \snippet test/scan.cpp ex_accum_2D
 
--- a/docs/details/algorithm.dox
+++ b/docs/details/algorithm.dox
@@ -107,9 +107,20 @@ Return type is u32 for all input types
 
 \ingroup scan_mat
 
-Perform inclusive sum along specified dimension
+Compute the cumulative sum (inclusive) along the specified dimension
 
-This table defines the return value types for the corresponding input types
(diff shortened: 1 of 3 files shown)


docs/details/algorithm.dox
```diff
@@ -107,9 +107,20 @@ Return type is u32 for all input types
...
-Perform inclusive sum along specified dimension
+Compute the cumulative sum (inclusive) along the specified dimension
 
-This table defines the return value types for the corresponding input types
+Here is a simple example for 1D arrays:
+
+\snippet test/scan.cpp ex_accum_1D
+
+For 2D arrays (and higher dimensions), you can specify the dimension along which
+the cumulative sum will be performed. If none is specified, it will be performed
```
9prady9: I think `computed/computed` sounds better than performed.

docs/details/algorithm.dox
```diff
@@ -107,9 +107,20 @@ Return type is u32 for all input types
...
+along the first dimension (0):
```
9prady9: Can you please verify if it dimension defaults to 0 or the first non-singleton dimension.

include/af/algorithm.h
```diff
@@ -306,11 +306,11 @@ namespace af
     template<typename T> void max(T *val, unsigned *idx, const array &in);
 
     /**
-       C++ Interface inclusive sum (cumulative sum) of an array
+       C++ Interface for computing the cumulative sum (inclusive) of an array
 
        \param[in] in is the input array
-       \param[in] dim The dimension along which exclusive sum is performed
-       \return the output containing exclusive sums of the input
+       \param[in] dim is the dimension along which inclusive sum is performed
```
9prady9: Similar to my first comment, `perform` is mostly with subjective contexts. Perhaps `calculated` or `computed`.

include/af/algorithm.h
```diff
@@ -761,11 +761,11 @@ extern "C" {
...
 
     /**
-       C Interface inclusive sum (cumulative sum) of an array
+       C Interface for computing the cumulative sum (inclusive) of an array
 
-       \param[out] out will contain exclusive sums of the input
+       \param[out] out will contain inclusive sums of the input
        \param[in] in is the input array
-       \param[in] dim The dimension along which exclusive sum is performed
+       \param[in] dim is the dimension along which inclusive sum is performed
```
9prady9: Same as my previous comment about `perform`

docs/details/algorithm.dox
9prady9: Sounds good. 

PS. Defaults are only available for C++ API.

mark-poscablo:
@umar456 also requested that docs include a formula for the inclusive sum in the docs. @9prady9 also requested an illustration. Both seem reasonable to include, will be working on them.

mark-poscablo:
Here's how it looks like now:
image
image
image

docs/details/algorithm.dox
umar456: This looks like a bug

9prady9: Were you expecting the default to be `-1` ? I was, but the code has `0` as default right now for whatever reason.

umar456: I expect it to work on the first non-zero dimension. Does -1 do that? If it does then yes it should be the default. If it doesn't then it should be implemented.

umar456: It should be fixed in a different PR

PR #2299, opened 2018-09-24
add set operations examples to documentation

syurkevi:
Updates documentation for setIntersect, setUnique, and setUnion

### DIFF
--- a/test/set.cpp
+++ b/test/set.cpp
@@ -179,9 +179,9 @@ TEST(Set, SNIPPET_setUniqueSorted) {
 
     //! [ex_set_unique_sorted]
 
-    int h_unique_gold[3] = { 1, 2, 3 };
-    af::array unique_gold(3, h_unique_gold);
-    ASSERT_ARRAYS_EQ(unique_gold, unique);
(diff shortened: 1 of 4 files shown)


9prady9:
@syurkevi Can you also add a line to the function description that says if batching is supported or not for these set functions.

syurkevi:
int_example
union_ex
unique_ex

test/set.cpp
```diff
@@ -161,3 +161,155 @@ SET_TESTS(short)
...
+    // allows algorithm to skip internal sorting step
+    const bool is_sorted = true;
+    af::array unique = setUnique(set, is_sorted);
+    // unique == { 1, 2, 3 };
+
+    //! [ex_set_unique_sorted]
+
+    int h_unique_gold[3] = { 1, 2, 3 };
+    af::array unique_gold(3, h_unique_gold);
+    ASSERT_ARRAYS_EQ(unique_gold, unique);
```
umar456: nit. Use ASSERT_VEC_ARRAY_EQ instead

PR #2302, opened 2018-10-02
Allow custom library path for loading dynamic backend libraries

jacobkahn:
Adds a check to the `AF_BUILD_LIB_CUSTOM_PATH` environment variable as one of the paths searched when attempting to dynamically open backends.

Important for specialized build configurations in which shared backend libraries may be built in a non-standard directory.


jacobkahn:
`AF_BUILD_PATH` and `AF_PATH` assumes shared libraries are located along either a `src/backend/**` or `lib` path. Our build configuration is slightly different; this offers an unopinionated way to specify these things with minimal additional configuration.

Issue #2273, opened 2018-08-10
reorder() before indexing produces incorrect results


jcai1:
For `inp1`, I create a 2x3x2 array, reorder it to 2x2x3, and index it along dimension 1.

For `inp2`, I prepare the reordering beforehand, and index it along dimension 1.

    int idx_host[2] = {1, 1};
    float inp1_host[12] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11};
    float inp2_host[12] = {0, 1, 6, 7, 2, 3, 8, 9, 4, 5, 10, 11};
    array idx(2, idx_host);
    array inp1 = reorder(array(2, 3, 2, inp1_host), 0, 2, 1);
    array inp2(2, 2, 3, inp2_host);
    array res1 = inp1(span, idx, span);
    array res2 = inp2(span, idx, span);
    af::print("idx", idx);
    af::print("inp1", inp1);
    af::print("inp2", inp2);
    af::print("res1", res1);
    af::print("res2", res2);

Output:

    idx
    [2 1 1 1]
             1 
             1 

    inp1
    [2 2 3 1]
        0.0000     6.0000 
        1.0000     7.0000 

        2.0000     8.0000 
        3.0000     9.0000 

        4.0000    10.0000 
        5.0000    11.0000 

    inp2
    [2 2 3 1]
        0.0000     6.0000 
        1.0000     7.0000 

        2.0000     8.0000 
        3.0000     9.0000 

        4.0000    10.0000 
        5.0000    11.0000 

    res1
    [2 2 3 1]
        2.0000     2.0000 
        3.0000     3.0000 

        8.0000     8.0000 
        9.0000     9.0000 

        0.0000     0.0000 
        0.0000     0.0000 

    res2
    [2 2 3 1]
        6.0000     6.0000 
        7.0000     7.0000 

        8.0000     8.0000 
        9.0000     9.0000 

       10.0000    10.0000 
       11.0000    11.0000 

We get `inp1` and `inp2` the same, as expected, but `res1` and `res2` differ. `res2` is what I would expect.

jcai1:
It seems that setting the `reorder()` result's dimensions to itself works:

    array my_reorder(
        const array& in,
        const unsigned x,
        const unsigned y = 1,
        const unsigned z = 2,
        const unsigned w = 3) {
      const array& result = reorder(in, x, y, z, w);
      return moddims(result, result.dims());
    }

umar456:
[...]

PR #2303, opened 2018-10-03
Reorder: Use Array::setDataDims instead of Array::modDims when x is 0

mark-poscablo:
This addresses #2273. I noticed that the user suggested that doing `moddims` on the reordered array using its new dimensions fixes the problem, but in `src/api/c/reorder.cpp`, we already call `Array::modDims` after adjusting the strides. I think that the reason why the problem still appears is that `Array::modDims` does not adjust `Array::data_dims` (which is in turn used by the internal `index` function in the backend through `Array::getDataDims`) after the fact. In comparison, I saw that `Array::setDataDims` does, and so I think it's better to use it in this case. However, if anyone thinks that there's a better approach for the solution, please let me know in the comments?

Also, if anyone has a better name for the test I just added, please suggest so. I can't think of a short, concise, yet descriptive name for the case that the issue raises, especially considering the fact that the problem doesn't appear if I use `(span, 1, span)` as the slice index. It has to be `(span, idx, span)`, where `idx` is an array that looks like (1, 1).

### DIFF
--- a/src/api/c/reorder.cpp
+++ b/src/api/c/reorder.cpp
@@ -50,7 +50,7 @@ static inline af_array reorder(const af_array in, const af::dim4 &rdims0)
             ostrides[i] = istrides[rdims[i]];
         }
         Array<T> Out = In;
-        Out.modDims(odims);
+        Out.setDataDims(odims);
         Out.modStrides(ostrides);
         out = getHandle(Out);
     } else {
--- a/test/reorder.cpp
+++ b/test/reorder.cpp
@@ -18,18 +18,18 @@
 #include <string>
 #include <testHelpers.hpp>
 
-using std::vector;
-using std::string;
-using std::cout;
-using std::endl;
 using af::allTrue;
 using af::array;
-using af::cfloat;
 using af::cdouble;
+using af::cfloat;
 using af::dim4;
 using af::dtype_traits;
 using af::reorder;
+using af::seq;
+using af::span;
 using af::tile;
+using std::string;
+using std::vector;
 
 
 template<typename T>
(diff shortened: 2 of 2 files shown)


mark-poscablo:
Turns out that I was able to get the wrong results again if I reorder a tall array to a wide array (the other direction from #2273 ), but still extracting a slice using an af::array index. What I found so far is that when an array is used for indexing a reordered array (at least in the case where x=0 so only the dims and stride are changed instead of the data layout underneath), the copying of elements happens as if the array wasn’t reordered at all (same offsets when done on non-reordered array), and hence, the slice extracted has the wrong elements.

9prady9:
@mark-poscablo I think that didn't fix it because `moddims` also does a `setDataDims`.  Check out this [line](https://github.com/arrayfire/arrayfire/blob/master/src/api/c/handle.hpp#L31)

mark-poscablo:
@9prady9 Sorry I should've been more specific, but what's really called here is `ArrayInfo::modDims` (see [here](https://github.com/arrayfire/arrayfire/blob/master/src/backend/common/ArrayInfo.cpp#L73)), which doesn't call `Array::setDataDims`. The one you referenced is the standalone `modDims`, but the internal `reorder` calls `Array::modDims` instead, which in turn calls `ArrayInfo::modDims`, which doesn't call `Array::setDataDims` inside.

mark-poscablo:
Closing this without merging for now as I have discovered that the problem lies (at least primarily) in the indexing, not in reorder itself. I'll either edit this PR to reflect the change in approach or make a new one altogether.

PR #2305, opened 2018-10-06
Fix pow precision for integral types

### LINKED ISSUE #2304: Error in pow() with integer type
Using arrayfire 3.6.0 with the CUDA backend with the following code:

```c++
#include <arrayfire.h>

int main(){
  af::array A = af::constant(-5,1, s32);
  af_print(A);
  af::array B = af::pow(A,2);
  af_print(B);
  return 0;
}
```

I get:
```shell
$ ./pow
A
[1 1 1 1]
   Offset: 0
   Strides: [1 1 1 1]
        -5
B
[1 1 1 1]
   Offset: 0
   Strides: [1 1 1 1]
        24
``` 

Instead of 25.

This problem was originally reported in https://github.com/arrayfire/arrayfire-python/issues/192

9prady9:
Fixes the issue in CUDA and OpenCL backens, CPU was working as expected
even before the fix.

Fixes #2304 

Also resolves a downstream issue https://github.com/arrayfire/arrayfire-python/issues/192


FilipeMaia:
Is it really necessary to cast to double in the case of float inputs?

9prady9:
@FilipeMaia I am modifying the test case to test with maximum values for respective type.

9prady9:
@FilipeMaia For floats and doubles, there is no cast operation at all. Can you point me to the line you are referring to ?

FilipeMaia:
@9prady9 I think I just misread something, I also can't find it now. Please ignore my previous comment.

test/binary.cpp
```diff
@@ -346,3 +346,26 @@ TEST(BinaryTests, ISSUE_1762)
...
+TYPED_TEST_CASE(BinaryOperations, IntegralTypes);
+
+TYPED_TEST(BinaryOperations, PowPrecisionIntegralTypes_Issue2304)
+{
+    af::array A = af::constant(15, 1, (af_dtype)dtype_traits<TypeParam>::af_type);
+    af::array B = af::pow(A, 2);
+    vector<TypeParam> hres(1, 0);
+    B.host(&hres[0]);
+    ASSERT_EQ(hres[0], 225);
+}
```
pavanky: Can you add more tests?

9prady9:
@umar456 I have updated pow docs in this PR itself.

PR #2311, opened 2018-10-10
Indexing: Use Param/Array::strides instead of toStride

mark-poscablo:
This addresses #2273, and is a change of approach from the original PR #2303, since the problem doesn't lie in `reorder` itself, but in indexing (one of the tests in that PR was wrong too, which led me to believe that I was initially going the right direction there). This is a special case of reordering (or in general, modifying the dims/strides of) an array before indexing it, and more specifically using an `af::array` to index. For calculating the input offsets, indexing (on all backends) call `src/backend/common/ArrayInfo.cpp:toStride`, which calculates the input stride based on the `seqs` given, but the given `af::array` index is not an `af::seq` object, and thus produces the wrong input strides. To get the correct strides (the result of the `Array::modStrides` call within the internal `reorder`), I think that we should use the `Param/Array::strides` function instead.

Please feel free to suggest any better approaches that you might think of.

### DIFF
--- a/src/api/c/reorder.cpp
+++ b/src/api/c/reorder.cpp
@@ -50,6 +50,7 @@ static inline af_array reorder(const af_array in, const af::dim4 &rdims0)
             ostrides[i] = istrides[rdims[i]];
         }
         Array<T> Out = In;
+        // Use modDims instead of setDataDims to only modify the ArrayInfo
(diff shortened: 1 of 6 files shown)


9prady9:
Looks good to me.

PR #2313, opened 2018-10-15
Use sizeof(Param<T>) to determine JIT eval. Add NodeIterator

umar456:
Uses the the size of the Param<T> object to determine when the JIT node should
be evaluated. This is required because there is a limit to the size of the
parameters you can send to a CUDA kernel.

Also created a NodeIterator which traverses the node tree. This improves the
readability of the code.

### DIFF
--- a/src/backend/cuda/Array.cpp
+++ b/src/backend/cuda/Array.cpp
@@ -7,28 +7,30 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
-#include <af/dim4.hpp>
 #include <Array.hpp>
+#include <JIT/BufferNode.hpp>
+#include <common/NodeIterator.hpp>
+#include <af/dim4.hpp>
 #include <copy.hpp>
 #include <err_cuda.hpp>
-#include <JIT/BufferNode.hpp>
-#include <scalar.hpp>
 #include <memory.hpp>
 #include <platform.hpp>
+#include <scalar.hpp>
 
 #include <cstddef>
 #include <memory>
+#include <numeric>
 
 using af::dim4;
+using cuda::JIT::BufferNode;
+using cuda::JIT::Node;
+using common::NodeIterator;
+using cuda::JIT::Node_ptr;
+using std::accumulate;
 using std::shared_ptr;
 
 namespace cuda
 {
-
-    using JIT::BufferNode;
-    using JIT::Node;
-    using JIT::Node_ptr;
-
     template<typename T>
     Node_ptr bufferNodePtr()
     {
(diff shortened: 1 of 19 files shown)


umar456:
@pavanky Please take a look

mark-poscablo:
`NodeIterator` class looks good and the way its operators work make sense to me. It does make the else clause in `createNodeArray` more readable to me (since it enabled the use of `std::accumulate` there).

9prady9:
The parameter space overflow error is still happening on k80.

src/backend/cuda/JIT/NodeIterator.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+    reference operator*() const noexcept {
+        return *tree[index];
+    }
+
+    pointer operator->() const noexcept {
+        return tree[index];
+    }
+
+    /// Creates a sentinal iterator. This is equivelant to the end iterator
```
mlloreda: Couple of misspellings: "sentinal" and "equivelant."

src/backend/cuda/JIT/NodeIterator.hpp
```diff
@@ -0,0 +1,101 @@
...
+    NodeIterator(Node* root) : tree{root}, index(0) {
+        tree.reserve(root->getHeight()*3);
+    }
+
+    /// The equality operator
+    ///
+    /// \param[in] other the rhs of the node
+    bool operator==(const NodeIterator& other) const noexcept {
+        // If the tree vector is empty in the other iterator then this means that the other
+        // iterator is a sentinal(end) node.
```
mlloreda: "sentinal" should be "sentinel."

src/backend/cuda/Array.cpp
```diff
@@ -220,46 +222,66 @@ namespace cuda
...
+                    // This is the size of the params that are passed by default
+                    constexpr int param_base_size = sizeof(Param<T>) + (4 * sizeof(uint));
+
+                    // This is the maximum size of the params that can be allowed by CUDA
+                    // NOTE: This number should have been (4096 - some_buffer_size) BUT
+                    // kernels who's kernel sizes come close to this value are not passing
+                    // and cuModuleLoadDataEx is failing with CUDA_ERROR_INVALID_IMAGE(200).
+                    // 35*sizeof(int) seems to be the magic number that passes all tests.
+                    // I have no idea why this is the case.
+                    constexpr int max_param_size = (4096 - (sizeof(Param<T>) + 35*sizeof(uint)));
```
pavanky: Why do you need to add another `sizeof(Param<T>)` ?

pavanky: We also pass in some additional metadata like the number of repeats along x and y. Number of dimensions etc. Would accounting for that make this magic constant less magical ?

umar456: I was taking into account the output Array. The 35*sizeof(uint) was supposed to take that into account(should have been 4 instead of 35) but when we create a kernel that is within the range of the allowed size the invalid image error is thrown by ModuleLoadDataEx.

PR #2315, opened 2018-10-16
Add broadcast batch for matmul

### LINKED ISSUE #2314: afcuda : 3d tensor matmul 2d tensor without tiling
This enhancement is about to take advantage of gemmStridedBatch cublas :
https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemmstridedbatched
in order to do a 3d tensor matmul a 2d tensor without having to tile the right tensor on the 3rd dimension to match the left tensor.
left 3d tensor shaped M x K x B 
matmul 
right 2d tensor shaped K x N 
outputting to a 3d tensor
M x N x B
Ref: 
https://devblogs.nvidia.com/cublas-strided-batched-matrix-multiply
seems to show that if stribeB = 0 then it s possible to do that op without having to tile the right tensor  : 
"The door is opened for even more optimizations in our implementation. For example, if a matrix stride is zero, then the batched GEMM is multiplying many matrices by a single matrix. In principle, the single matrix could be read once and reused many times. Other clever reorderings of how the computation is actually performed are now available to implementers."

9prady9:
Resolves #2314

### DIFF
--- a/src/api/c/blas.cpp
+++ b/src/api/c/blas.cpp
@@ -127,7 +127,7 @@ af_err af_matmul(af_array *out,
         dim4 lDims = lhsInfo.dims();
         dim4 rDims = rhsInfo.dims();
 
-        if (lDims.ndims() > 2 || rDims.ndims() > 2) {
+        if (lDims.ndims() > 2 && rDims.ndims() > 2) {
             DIM_ASSERT(1, lDims.ndims() == rDims.ndims());
             if (lDims[2] != rDims[2] && lDims[2] != 1 && rDims[2] != 1) {
                 AF_ERROR("Batch size mismatch along dimension 2", AF_ERR_BATCH);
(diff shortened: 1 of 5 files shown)


src/api/c/blas.cpp
```diff
@@ -127,7 +127,13 @@ af_err af_matmul(af_array *out,
         dim4 lDims = lhsInfo.dims();
         dim4 rDims = rhsInfo.dims();
 
+        //FIXME(pradeep) Move backend specific check to backend
```
mlloreda: Are your plans to resolve this FIXME in this PR or in a future one?

9prady9: I think I will add the broadcast support for CPU and OpenCL as well if doesn't take too much effort.

mlloreda: :+1:

PR #2322, opened 2018-10-21
Cleanup shift and select nodes

umar456:


src/backend/cuda/JIT/ShiftNode.hpp
```diff
@@ -15,22 +15,20 @@
...
-
 namespace JIT
 {
     template<typename T>
     class ShiftNode : public Node
     {
     private:
-
-        Node_ptr m_buffer_node;
+        std::shared_ptr<BufferNode<T>> m_buffer_node;
```
9prady9: I though `Note_ptr` just a typedef for `std::shared_ptr<BufferNode<T>>`.

umar456: Its a typedef for its base class `std::shared_ptr<Node<T>>`

PR #2325, opened 2018-10-24
Remove dims restriction from svdInPlace

### LINKED ISSUE #2282: svdInPlace() does not accept arrays with dim0 > dim1
I was playing with svd while testing and benchmarking #2279, and tried out `af::svdInPlace()`. However, I was baffled to get this exception when my array's dim0 is 3000 and dim1 is 2000:

```
ArrayFire v3.7.0 (CPU, 64-bit Linux, build 002e413e)
[0] Intel: Intel(R) Core(TM) i7-3770K CPU @ 3.50GHz, 7936 MB, Max threads(8) GNU Compiler Collection(GCC/G++) 5.5.0
ArrayFire Exception (Invalid input size:203):
In function af_err af_svd_inplace(void**, void**, void**, af_array)
In file src/api/c/svd.cpp:110
Invalid dimension for argument 3
Expected: dims[0] <= dims[1]

In function void af::svdInPlace(af::array&, af::array&, af::array&, af::array&)
In file src/api/cpp/lapack.cpp:28
``` 

I looked around svd's code and found this [dim assert that dim0 must <= dim1](https://github.com/arrayfire/arrayfire/blob/b562e974264f5338a14db37a5800e96580205586/src/api/c/svd.cpp#L110). Is there any special reason for this?

mark-poscablo:
Fixes #2282. `svdInPlace` can now take in arrays where `dim0 > dim1`. I added tests too since there weren't any for `svdInPlace`, just regular `svd`.

### DIFF
--- a/src/api/c/svd.cpp
+++ b/src/api/c/svd.cpp
@@ -107,7 +107,6 @@ af_err af_svd_inplace(af_array *u, af_array *s, af_array *vt, af_array in)
         const ArrayInfo& info = getInfo(in);
         af::dim4 dims = info.dims();
 
-        DIM_ASSERT(3, dims[0] <= dims[1]);
         ARG_ASSERT(3, (dims.ndims() >= 0 && dims.ndims() <= 3));
         af_dtype type = info.getType();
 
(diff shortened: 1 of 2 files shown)


mark-poscablo:
Found out that this only works well with the CPU backend. CUDA produces results that are way off and OpenCL segfaults in my machine. I suspect that the problem lies deeper, in the src/backend level at the very least. Closing this for now

PR #2326, opened 2018-10-24
Added samples to sparse documentation

mlloreda:
Samples added for:

```
array sparse(const dim_t nRows, const dim_t nCols,
                       const array values, const array rowIdx, const array colIdx,
                       const af::storage stype = AF_STORAGE_CSR);

array sparse(const dim_t nRows, const dim_t nCols, const dim_t nNZ,
                       const void* const values,
                       const int * const rowIdx, const int * const colIdx,
                       const dtype type = f32, const af::storage stype = AF_STORAGE_CSR,
                       const af::source src = afHost);

array sparse(const array dense, const af::storage stype = AF_STORAGE_CSR);

array dense(const array sparse);
```

Also made a few minor improvements to sparse backend and API code.

### DIFF
--- a/src/api/c/sparse.cpp
+++ b/src/api/c/sparse.cpp
@@ -45,7 +45,7 @@ const SparseArrayBase &getSparseArrayBase(const af_array in,
 // Sparse Creation
 ////////////////////////////////////////////////////////////////////////////////
 template<typename T>
-af_array createSparseArrayFromData(const af::dim4 &dims, const af_array values,
+af_array createSparseArrayFromData(const dim4 &dims, const af_array values,
                                    const af_array rowIdx, const af_array colIdx,
                                    const af::storage stype) {
     SparseArray<T> sparse = common::createArrayDataSparseArray(
@@ -96,9 +96,9 @@ af_err af_create_sparse_array(af_array *out, const dim_t nRows,
             DIM_ASSERT(5, (dim_t)cInfo.elements() == nCols + 1);
         }
 
-        af_array output = 0;
+        af_array output = nullptr;
 
-        af::dim4 dims(nRows, nCols);
+        dim4 dims(nRows, nCols);
 
         switch (vInfo.getType()) {
             case f32:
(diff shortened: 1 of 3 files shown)


test/sparse.cpp
```diff
@@ -271,3 +272,220 @@ TYPED_TEST(Sparse, EmptyDeepCopy) {
...
+
+    float dns[] = { 0, 5, 0, 0,
+                    0, 8, 0, 6,
+                    0, 0, 3, 0,
+                    0, 0, 0, 0 };
+    const int M = 4, N = 4, nnz = 4;
+
+    array dense(dim4(M,N), dns);
+    // dense
+    // [4 4 1 1]
```
umar456: I think it would be better if the sizes weren't displayed.

test/sparse.cpp
```diff
@@ -271,3 +272,220 @@ TYPED_TEST(Sparse, EmptyDeepCopy) {
...
+    array    vals = array(dim4(nnz), v);
+    array row_ptr = array(dim4(M+1), r);
+    array col_idx = array(dim4(nnz), c);
+
+    array sparse = af::sparse(M, N, vals, row_ptr, col_idx, AF_STORAGE_CSR);
+    // sparse
+    // Storage Format : AF_STORAGE_CSR
+    // [4 4 1 1]
+    // sparse: Values
+    // [4 1 1 1]
```
umar456: Maybe display the sparse arrays as rows instead of columns. This takes up a lot of space.

src/api/c/sparse.cpp
```diff
@@ -398,7 +398,7 @@ af_err af_sparse_get_values(af_array *out, const af_array in)
 af_err af_sparse_get_row_idx(af_array *out, const af_array in)
 {
     try {
-        const SparseArrayBase base = getSparseArrayBase(in);
+        const SparseArrayBase &base = getSparseArrayBase(in);
```
umar456: I am not sure this is a valid change. I think the idea is that you do want to make a copy.

This is also unrelated to the name of the PR. If you want to do this you should create a new PR.

9prady9: Shouldn't we be returning simple object in that case instead of reference to an object ? The reason I say that is `getSparseArrayBase` currently returns an object of type `SparseArrayBase&`. It is similar case with `getInfo` too - returns `const ArrayInfo`.

PR #2327, opened 2018-10-25
Modified backend transform() to accept output arg

mlloreda:


src/api/c/transform.cpp
```diff
@@ -19,10 +19,10 @@ using af::dim4;
 using namespace detail;
 
 template<typename T>
-static inline af_array transform(const af_array in, const af_array tf, const af::dim4 &odims,
+static inline void transform(af_array *out, const af_array in, const af_array tf, const dim4 &odims,
                                  const af_interp_type method, const bool inverse, const bool perspective)
```
FloopCZ:
```suggestion
                              const af_interp_type method, const bool inverse, const bool perspective)
```

src/backend/opencl/transform.cpp
```diff
@@ -38,12 +36,11 @@ namespace opencl
...
             AF_ERROR("Unsupported interpolation type", AF_ERR_ARG);
         }
-        return out;
     }
 
 
 #define INSTANTIATE(T)                                                                  \
-    template Array<T> transform(const Array<T> &in, const Array<float> &tf,             \
+    template void transform(Array<T> &out, const Array<T> &in, const Array<float> &tf,  \
                                 const af::dim4 &odims, const af_interp_type method,     \
```
FloopCZ:
```suggestion
                             const af::dim4 &odims, const af_interp_type method,     \
```

src/backend/opencl/transform.cpp
```diff
@@ -38,12 +36,11 @@ namespace opencl
...
                                 const bool inverse, const bool perspective);
```
FloopCZ:
```suggestion
                             const bool inverse, const bool perspective);
```

mlloreda:
Thanks for feedback, @FloopCZ! Addressed.

umar456:
Could you add tests to verify that these changes worked? You should look at the approx1 PR and look at the type of tests there. Ideally we want to make these tests as low costs as possible so it would be nice if you and @mark-poscablo can discuss making tests more succinct.

PR #2328, opened 2018-10-25
Modified backend wrap() to accept output arg

mlloreda:


umar456:
Please add tests.

src/backend/cpu/wrap.cpp
```diff
@@ -18,37 +18,34 @@ namespace cpu
...
 {
-    af::dim4 idims = in.dims();
-    af::dim4 odims(ox, oy, idims[2], idims[3]);
-
-    Array<T> out = createValueArray<T>(odims, scalar<T>(0));
-    out.eval();
     in.eval();
+    const dim4 idims = in.dims();
+    const dim4 odims(ox, oy, idims[2], idims[3]);
+    out = createValueArray<T>(odims, scalar<T>(0));
```
mark-poscablo: I think you meant to omit these lines, correct? Since this will make wrap write its results to the memory returned by `createValueArray` and not the already provided space in the `out` argument. I see that you omitted this in the CUDA and OpenCL backends so I guess it was just an accident that you left this here.

src/api/c/wrap.cpp
```diff
@@ -19,92 +19,91 @@ using af::dim4;
...
+                          true);
+}
+
+af_err af_wrap_v2(af_array* out, const af_array in,
+                  const dim_t ox, const dim_t oy,
+                  const dim_t wx, const dim_t wy,
+                  const dim_t sx, const dim_t sy,
+                  const dim_t px, const dim_t py,
+                  const bool is_column) {
+    if (out == 0) return AF_ERR_ARG;
```
umar456: I think you should do the same conversion you did in the approx_v2 functions here for the error messages.

src/api/c/wrap.cpp
```diff
@@ -19,92 +19,91 @@ using af::dim4;
...
-        ARG_ASSERT(4, sx > 0);
-        ARG_ASSERT(5, sy > 0);
+        const ArrayInfo& info  = getInfo(in);
+        const af_dtype in_type = info.getType();
+        const dim4 in_dims     = info.dims();
+        const dim4 out_dims(ox, oy, in_dims[2], in_dims[3]);
 
-        dim_t nx = (ox + 2 * px - wx) / sx + 1;
-        dim_t ny = (oy + 2 * py - wy) / sy + 1;
+        ARG_ASSERT(4, wx > 0);
```
umar456: These indicies will be incorrect in the older versions of approx. These checks should really be in the API level functions. Unfortunately it will be error prone to duplicate this everywhere. I am not sure how to fix this cleanly.

test/wrap.cpp
```diff
@@ -247,3 +248,311 @@ TEST(Wrap, DocSnippet) {
...
+    ASSERT_EQ(af_wrap_v2(&out, 0,
+                         4, 4,  // output dims
+                         2, 2,  // window size
+                         2, 2,  // stride
+                         0, 0,  // padding
+                         true), // is_column
+              AF_ERR_ARG);
+}
+
+class ArgDim {
```
umar456: How is this different from dim4?

test/wrap.cpp
```diff
@@ -247,3 +248,311 @@ TEST(Wrap, DocSnippet) {
...
+   public:
+    PadDims() : ArgDim(0, 0) {}
+    PadDims(dim_t d0, dim_t d1) : ArgDim(d0, d1) {}
+};
+
+class WrapArgs {
+   public:
+    WindowDims *wc_;
+    StrideDims *sc_;
+    PadDims *pc_;
```
umar456: Do these types really need to be pointers?

src/api/c/wrap.cpp
```diff
@@ -19,92 +19,91 @@ using af::dim4;
...
+        const dim_t patch_size  = is_column ? in_dims[0] : in_dims[1];
+        const dim_t num_patches = is_column ? in_dims[1] : in_dims[0];
 
         DIM_ASSERT(1, patch_size == wx * wy);
         DIM_ASSERT(1, num_patches == nx * ny);
 
-        af_array output;
+        if (allocate_out) { *out = createHandle(out_dims, in_type); }
+
+        DIM_ASSERT(1, getInfo(*out).dims() == out_dims);
```
umar456:
```suggestion
        // The out pointer can be passed in to the function by the user
        DIM_ASSERT(1, getInfo(*out).dims() == out_dims);
```

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -25,6 +29,8 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
     af::dim4 istrides = in.strides();
     af::dim4 ostrides = out.strides();
 
+    std::fill(begin(out), end(out), scalar<T>(0));
```
umar456: I don't like this. the out array should be zeroed before we get here. Instead of createHandle, you should use createValueArray.

If the out array is passed into the function, we can stipulate that the values will be summed into the output array. If they want the originals they need to pass a constant array or zero beforehand.

PR #2329, opened 2018-10-26
Move JIT classes into the common namespace for OpenCL and CUDA

umar456:
Moves JIT classes for OpenCL and CUDA into the common interface.
This does modify the CPU backend. Most of the specilization has
been moved into the JIT/kernel_generators.hpp header.


9prady9:
It would be nice we can refactor the namespace `JIT` to `jit` to be consistent with the convention followed by other namespaces.

I skimmed through it and it looks great. Trying to build it now and check out the new printnode functionality.

umar456:
Addressed feedback

PR #2331, opened 2018-10-29
svdInPlace: Flip the inequality on the dims restriction

mark-poscablo:
Following up on #2282, `svdInPlace` did have its inequality on the input array dims restriction [flipped](https://github.com/arrayfire/arrayfire/blob/master/src/api/c/svd.cpp#L110). It asserted before that the input's `dim0 <= dim1`, but I flipped it to `>=`, which is what it should've been (see the issue for my explanation of why this should be the case). I addded some tests to verify this and I also noted this down in the docs for future users of this function.

### DIFF
--- a/test/svd_dense.cpp
+++ b/test/svd_dense.cpp
@@ -18,19 +18,21 @@
 #include <string>
 #include <testHelpers.hpp>
 
-using std::vector;
-using std::string;
-using std::cout;
-using std::endl;
-using std::abs;
 using af::array;
-using af::cfloat;
 using af::cdouble;
+using af::cfloat;
+using af::dim4;
 using af::dtype;
 using af::dtype_traits;
+using af::iota;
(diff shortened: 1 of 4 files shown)


test/svd_dense.cpp
```diff
@@ -109,3 +151,35 @@ TYPED_TEST(svd, Rect1)
...
+{
+    svdInPlaceTest<TypeParam>(500, 500);
+}
+
+TYPED_TEST(svd, InPlaceRect0)
+{
+    svdInPlaceTest<TypeParam>(500, 300);
+}
+
+// dim0 < dim1 case not supported for now
```
umar456: An expect throw/error would be a better test but this is fine.

9prady9:
@mark-poscablo `empty` tests are failing everywhere, can you please look into those.

PR #2333, opened 2018-10-30
Padding and backport 36

umar456:
Add padding kernel from the de-convolution commits and backport changes from the master branch.

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -254,7 +254,12 @@ typedef enum {
     ///
     /// Out of bound values are symmetric over the edge
     ///
-    AF_PAD_SYM
+    AF_PAD_SYM,
+
+    ///
+    /// Out of bound values are clamped to the edge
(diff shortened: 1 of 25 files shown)


umar456:
Only a7bcac8 needs to be reviewed. Other commits are backports and have already been reviewed.

include/af/defines.h
```diff
@@ -254,7 +254,12 @@ typedef enum {
     ///
     /// Out of bound values are symmetric over the edge
     ///
-    AF_PAD_SYM
+    AF_PAD_SYM,
+
+    ///
+    /// Out of bound values are clamped to the edge
+    ///
+    AF_PAD_CLAMP_TO_EDGE,
```
umar456: Adding an enum will not affect the ABI. Its unfortunate we have to add this to the header but its necessary to avoid deviating from the master branch.

PR #2334, opened 2018-10-31
Added ParamIterator for the cpu backend

umar456:


src/backend/cpu/ParamIterator.hpp
```diff
@@ -0,0 +1,144 @@
...
+    // The iterator's stride
+    const af::dim4 stride;
+
+    // NOTE: This is not really the true coordinate of the iteration. It's
+    // values will go down as you move through the array.
+    std::array<dim_t, AF_MAX_DIMS> dim_index;
+
+    /// Calculates the iterator offsets. These are different from the original offsets
+    /// because they define the stride from the end of the last element in the previous
+    /// dimention to the first element on the next dimenstion.
```
syurkevi:
```suggestion
    /// dimension to the first element on the next dimension.
```

src/backend/cpu/ParamIterator.hpp
```diff
@@ -0,0 +1,144 @@
...
+    bool operator!=(const ParamIterator& other) const noexcept {
+        return other.ptr != ptr;
+    }
+
+    /// Advances the iterator
+    ParamIterator operator++() noexcept {
+        for(int i = 0; i < AF_MAX_DIMS; i++) {
+            dim_index[i]--;
+            ptr += stride[i];
+            if(dim_index[i]) return *this;
```
syurkevi: Fix

9prady9:
We should be able to test the iterator with subarray if we create a array and use `host` member function on it right ? @umar456

src/backend/cpu/ParamIterator.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <iterator>
+
+namespace cpu {
+
+/// A Param iterator that iterates through a Param object
+template<typename T>
+class ParamIterator {
+    T* ptr;
+
+    // NOTE: This is not really the true coordinate of the iteration. It's
```
mlloreda: It's -> its

PR #2344, opened 2018-11-12
Fix several -Wall -Wextra warnings

umar456:


src/api/unified/lapack.cpp
```diff
@@ -83,7 +83,7 @@ af_err af_pinverse(af_array *out, const af_array in, const double tol,
                    const af_mat_prop options)
 {
     CHECK_ARRAYS(in);
-    return CALL(out, in, options);
+    return CALL(out, in, tol, options);
```
9prady9: that an actual bug.

umar456: Yeah, we need stronger types internally. Should I submit this separately?

src/backend/common/Logger.cpp
```diff
@@ -51,10 +51,11 @@ loggerFactory(string name) {
 
 string bytesToString(size_t bytes) {
     static array<const char *, 5> units{"B", "KB", "MB", "GB", "TB"};
-    int count = 0;
+    unsigned long long count = 0;
```
pavanky: Why not size_t ?

src/backend/common/Logger.cpp
```diff
@@ -51,10 +51,11 @@ loggerFactory(string name) {
...
     double fbytes = static_cast<double>(bytes);
-    for(count = 0; count < units.size() && fbytes > 1000.0; count++) {
-        fbytes *= (1.0 / 1024.0);
+    unsigned long long num_units = units.size();
```
pavanky: Same as above.

src/api/unified/lapack.cpp
9prady9: A separate commit should be fine, I am fine with it in this PR.

src/backend/common/Logger.cpp
umar456: Yeah, size_t would be better. I am not sure what I was thinking. Fixed.

9prady9:
@umar456 There seems to be some compilation errors.

umar456:
Addressed all feedback

9prady9:
@umar456 Why didn't the CI kick in ?

PR #2347, opened 2018-11-13
Test assert macros for using existing output arrays in C API

mark-poscablo:
This provides an easy way to test a C-API arrayfire function's capability to write its output to an existing array. The macros can currently cover four kinds of output arrays: null, an existing array that's the same size as the expected output, a sub array (of an existing larger array), and a reordered array that's the same size as the expected output.


mark-poscablo:
@9prady9 Hey thanks for the feedback. Do you think renaming the functions better or structuring it somewhat differently would help make it more clear/intuitive? For now I could at least put comments on those `gen*Array` functions, I forgot to do that actually.

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    REORDERED_ARRAY
+};
+
+class TestOutputArrayInfo {
+    af_array out_arr_c; // This af_array determines what will be given back to
+                        // the af_* function, and thus what will be released
+                        // after it is used
+    af::index out_subarr_idxs[4];
+    af::array subarr;
+    TestOutputArrayType out_arr_type;
```
9prady9: I haven't tested my modifications but I think we don't need so many af::array or af_arrays here.
Please check out the following class. Using below definition for future reference, I will suggest modifications to the functions that use this class.
```C++
class TestOutputArrayInfo {
    af::array arr;
    af::index idxs[4];
    TestOutputArrayType out_arr_type;

    public:
    TestOutputArrayInfo(TestOutputArrayType arr_type) :out_arr_type(arr_type) {}

    TestOutputArrayType getOutputArrayType() { return out_arr_type; }

    void init(af::dim4 dims, af::dtype ty) { arr = af::randu(dims, ty); }

    void init(af::dim4 dims, af::dtype ty, af::index subarr_idxs[4]) {
        out_arr = af::randu(dims, ty);
        for (int i = 0; i < 4; ++i)
            idxs[i] = subarr_idxs[i];
    }

    af::array getOutputArray() { return arr; }

    void setOutputArray(af::array newArray) { arr = newArray; }

    af::index idx(dim_t dim) { return idxs[dim]; }
};
```
I have removed some unnecessary member functions that used the old member variables. Also, I have formatted it a bit densely above to have a short message.

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    af::index getSubArrayIdx(dim_t dim) {
+        return out_subarr_idxs[dim];
+    }
+};
+
+// Generates a null af_array. testWriteToOutputArray expects that the array that
+// it receives after the af_* function is a valid, allocated af_array
+void genNullArray(TestOutputArrayInfo& metadata, const af::dim4& dims,
+                  const af::dtype ty) {
+    metadata.initArrays();
```
9prady9: `genNullArray` doesn't need to call anything, it will be a no op function. We can keep it if we want to use keep the calling pattern inside switch case in the function `genTestOutputArray`. 

I think it should be removed since it doesn't do anything.

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+}
+
+// Generates a random array. testWriteToOutputArray expects that it will receive
+// the same af_array that this generates after the af_* function is called
+void genRegularArray(TestOutputArrayInfo& metadata, const af::dim4& dims,
+                  const af::dtype ty) {
+    metadata.initArrays(dims, ty);
```
9prady9: Will become `metadata.init(dims, ty)`

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    // Calculate index of sub-array. These will be used also by
+    // testWriteToOutputArray so that the gold sub array will be placed in the
+    // same location. Currently, this location is the center of the large array
+    af::index subarr_idxs[4];
+    for (int i = 0; i < 4; ++i) {
+        subarr_idxs[i] =
+            dims[i] > 1 ? af::seq(pad_size, pad_size + dims[i] - 1) : af::span;
+    }
+
+    metadata.initArrays(full_arr_dims, ty, subarr_idxs);
```
9prady9: Will become `metadata.init(full_arr_dims, ty, subarr_idxs)`

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    // The output will be checked if it is still correct even with the
+    // modified dims and strides "hack" with no data movement
+    unsigned reorder_0 = 0;
+    unsigned reorder_1 = 2;
+    unsigned reorder_2 = 1;
+
+    // Shape the output array such that the reordered output array will have
+    // the correct dimensions that the test asks for (i.e. must match dims arg)
+    af::dim4 out_dims(dims[reorder_0], dims[reorder_1], dims[reorder_2]);
+    metadata.initArrays(out_dims, ty);
```
9prady9: will become `metadata.init(out_dims, ty)`

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+        genRegularArray(*metadata, arr_dims, ty);
+        break;
+    case SUB_ARRAY:
+        genSubArray(*metadata, arr_dims, ty);
+        break;
+    case REORDERED_ARRAY:
+        genReorderedArray(*metadata, arr_dims, ty);
+        break;
+    }
+    *out_ptr = metadata->getOutputArrayC();
```
9prady9: will become
```C++
af_array out = 0;
if (metadata->getOutputArrayType() == SUB_ARRAY) {
    af::array original = metadata->getOutputArray();
    af::array subarr = original(metadata->idx(0), metadata->idx(1),
                                metadata->idx(2), metadata->idx(3));
    // may be it will be better to use C-API in above statement
    af_retain_array(&out, subarr);
} else {
    af_retain_array(&out, metadata->getOutputArray().get());
}
*out_ptr = out;
```

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+// correctly uses an existing array as its output
+void genTestOutputArray(af_array *out_ptr,
+                        const unsigned ndims, const dim_t *dims,
+                        const af::dtype ty,
+                        TestOutputArrayInfo* metadata) {
+    af::dim4 arr_dims(ndims, dims);
+    switch (metadata->getOutputArrayType()) {
+    case NULL_ARRAY:
+        genNullArray(*metadata, arr_dims, ty);
+        break;
```
9prady9: I would make null array case as the default case, that way, the whole function `genNullArray` can be removed as it won't do anything in the modified code.
```C++
switch(...) {
...
default: // do nothing as that itself is null array case
}
```

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+                       const af_array gold, const af_array out,
+                       TestOutputArrayInfo *metadata) {
+    // In the case of NULL_ARRAY, the output array starts out as null.
+    // After the af_* function is called, it shouldn't be null anymore
+    if (metadata->getOutputArrayType() == NULL_ARRAY) {
+        if (out == 0) {
+            return ::testing::AssertionFailure()
+                << "Output af_array " << result_name << " is null";
+        }
+        metadata->setOutputArrayC(out);
```
9prady9: will become
`metadata->setOutputArray(af::array())`

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    }
+    // For every other case, must check if the af_array generated by
+    // genTestOutputArray was used by the af_* function as its output array
+    else {
+        if (metadata->getOutputArrayC() != out) {
```
9prady9: `metadata->getOutputArray().get()`

test/testHelpers.hpp
```diff
@@ -979,4 +996,231 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+
+        af_array gold_retain = 0;
+        af_retain_array(&gold_retain, gold);
+        af::array gold_retain_cpp(gold_retain);
+
+        af::copy(metadata->out_arr_cpy, gold_retain_cpp,
+                 metadata->getSubArrayIdx(0),
+                 metadata->getSubArrayIdx(1),
+                 metadata->getSubArrayIdx(2),
+                 metadata->getSubArrayIdx(3));
```
9prady9: this if block will change to something like the following
```C++
af_array gold_retain = 0;
af_retain_array(&gold_retain, gold);
af::array gold_retain_cpp(gold_retain);
af::array original = metadata->getOutputArray();
af::array subout = original(metadata->idx(0), metadata->idx(1),
                            metadata->idx(2), metadata->idx(3));
af::array subgold = gold_retain_cpp(metadata->idx(0), metadata->idx(1),
                                    metadata->idx(2), metadata->idx(3));
return assertArrayEq(gold_name, result_name, subgold, subarr);
```

test/testHelpers.hpp
mark-poscablo: I like this leaner version of the class. However, with this and the changes below that this version entails, I think we're kinda straying away from the original purpose of this class and the other helper functions. So instead of me suggesting changes to this version, let me comment/ask questions on the changes below so that we can come back here and change some things if we need to.

test/testHelpers.hpp
mark-poscablo: Yes this is essentially a no-op. I debated whether to leave this here or not but I decided to leave it in order to keep the calling pattern in the switch statement in `genTestOutputArray`. But yeah we don't have to keep it (calling that empty function is a small performance hit too).

test/testHelpers.hpp
mark-poscablo: `metadata.init(dims, ty)`, according to your class definition above, simply creates a random `af::array` and stores it within the class. I think that might be too simple for this reason: We do store the `af::array` as a member data of the class, but we don't store the final `af_array` value that we give back to the test for `af_approx1` to use as its output. We need to store that `af_array` value because we need to test if `af_approx1` (or any other `af_*` function) will place its output data on the the same existing, allocated `af_array` that we give to it, and hence, we need to compare the `af_array` value we generated and the `af_array` that `af_approx1` gives as its output - they must be the same.

test/testHelpers.hpp
mark-poscablo: Same concern as above (`genRegularArray`). In addition, however, I think we also need to store a copy of the original random full array that we generated, because in this case, we test if the output of `af_approx1` modifies the subarray and only the subarray - the rest of the full array must be untouched. Thus if we make a copy of the original random array that we generated, and we compare the original and the copy after (1) `af_approx1` places its output on the subarray of the original and after (2) we inject the reference values on the subarray of the copy - the original and the copy full arrays must have the same values element-wise.

test/testHelpers.hpp
mark-poscablo: Same concern as (`genRegularArray`).

test/testHelpers.hpp
mark-poscablo: A couple of concerns here:
- I've written about storing the original, full array and making a copy of it for the actual testing later. Please see my comment on `genSubArray`.
- Maybe it is better indeed to use the C API there. Either way though we'll need to retain `subarr` or `metadata->getOutputArray().get()`. The question is where in the code do we eventually release it. My solution to this (formed from Umar's suggestion to have the destructor call the release) is to store the value of `af_retain_array`'s output in `TestOutputArrayInfo`, then eventually release that inside the destructor. This essentially makes a central point on where to release whatever was allocated/retained, regardless of whether the asserts pass or fail (since the destructor is called either way).

test/testHelpers.hpp
mark-poscablo: What if we make sure we give back 0 for the `af_array` in this case (this is what I currently do inside `genNullArray`)? Maybe I'm a little paranoid, but if the user doesn't initialize the output `af_array` (passed in here as `*out_ptr`) before calling `genTestOutputArray`, then I'd guess that it could be something other than 0 and we'd get undefined/ugly behavior.

test/testHelpers.hpp
mark-poscablo: Actually I don't think this line will make any difference, even in my original code. In the `NULL_ARRAY` case, we only check two things - that the output of `af_approx1` isn't null, and that the output matches the reference, element-wise. We do the latter in the very last else statement in this function (`testWriteToOutputArray`), and in there we just use the arguments given in the call, not whatever data we've stored in `TestOutputArrayInfo` (it'll be a different story for `SUB_ARRAY`, however).

test/testHelpers.hpp
mark-poscablo: `metadata->getOutputArray().get()` will not equal `out`. The latter is the output of `af_retain_array` in `genTestOutputArray`, and that is going to be different from the result of `af::array::get()`. This is one of the reasons why I currently store the output of the retain call as a member of `TestOutputArrayInfo`.

test/testHelpers.hpp
mark-poscablo: I think you've got this already from our conversation in Slack, but basically my concern here is reflected in my response to your comment in `genSubArray`. Anyway, `gold` and `gold_retain_cpp` are both the reference sub-array. We want to inject that into a duplicate of the original random array that we generated in `genTestOutputArray, and compare the duplicate to the original (both full arrays) - both must equal element wise, because at this point, `af_approx1` should have already injected its output into the full original random array that we generated.

mark-poscablo:
As per the Skype call either between me, @9prady9, and @umar456, I've simplified my current code mainly by limiting the arrayfire calls to the C API. This also reduces the necessary member data, especially the `af::/af_arrays` needed. Thanks for the feedback guys, I actually like this version better than what I had - it's cleaner and it's still free of memory leaks (this whole PR actually gets rid of the existing memory leaks in the tests). Please let me know if there's anything else I can improve on this.

9prady9:
looks good.

test/testHelpers.hpp
```diff
@@ -979,4 +996,209 @@ ::testing::AssertionResult assertArrayNear(std::string hA_name, std::string aDim
...
+    FULL_ARRAY,
+    SUB_ARRAY,
+    REORDERED_ARRAY
+};
+
+class TestOutputArrayInfo {
+    af_array out_arr;
+    af_array out_arr_cpy;
+    af_array out_subarr;
+    std::vector<af_seq> out_subarr_idxs;
```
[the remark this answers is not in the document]
umar456: you are assigning to lhs and using lhs at the same time. I think that would leak. I don't see why you would get a heap use after free because of it because its on the stack.

PR #2358, opened 2018-11-21
Fix typo in matmul batch info table

9prady9:
@mlloreda docs need to be updated because of this fix. I don't think we need to update 3.6 branch yet. that's @umar456 call.

Sorry about the typo.

### DIFF
--- a/docs/details/blas.dox
+++ b/docs/details/blas.dox
@@ -29,14 +29,17 @@ data from memory. This results in no additional memory being used for temporary
 buffers.
 
 Batched matrix multiplications are supported. Given below are the supported
-formats for given matrices A and B.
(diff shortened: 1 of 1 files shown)


docs/details/blas.dox
```diff
@@ -29,14 +29,17 @@ data from memory. This results in no additional memory being used for temporary
 buffers.
 
 Batched matrix multiplications are supported. Given below are the supported
-formats for given matrices A and B.
+types of batch operations for any given set of two matrices A and B.
 
-| Input Matrix A             | Input Matrix B             | Output Matrix Size          |
+| Size of Input Matrix A     | Size of Input Matrix B     | Output Matrix Size          |
```
umar456: I feel like the original header name is fine.

PR #2359, opened 2018-11-21
Dilated convolve2 and backward gradients

syurkevi:
Adds dilation to forward convolve2 with cudnn integration.
Also adds functions to obtain convolve2 backwards gradients with respect to the filter or input data.


src/backend/cpu/convolve.cpp
```diff
@@ -107,5 +119,197 @@ INSTANTIATE(ushort ,   float)
...
+    s = {(double)(fDims[1] - 1), 0, -1};
+    flip_index[1] = s;
+    flip_index[2] = af_span;
+    flip_index[3] = af_span;
+
+    collapsed_filter = createSubArray(collapsed_filter, flip_index);
+    collapsed_filter.modDims(dim4(fDims[0] * fDims[1] * fDims[2], fDims[3]));
+
+    Array<accT> collapsed_gradient = cast<accT>(incoming_gradient);
+    collapsed_gradient = reorder(collapsed_gradient, dim4(0, 1, 3, 2));
```
umar456: I don't understand the reason for this reorder. you are only changing the 3rd and 4th dims. I feel that we can do something more efficient here and in the next operations. maybe request the right shape of these functions.

src/api/c/convolve.cpp
```diff
@@ -228,6 +229,51 @@ af_err af_convolve2(af_array *out, const af_array signal, const af_array filter,
...
+            case c32: output = convolve2Strided<cfloat ,  cfloat>(signal, filter, stride, padding, dilation); break;
+            case c64: output = convolve2Strided<cdouble, cdouble>(signal, filter, stride, padding, dilation); break;
+            case f32: output = convolve2Strided<float  ,   float>(signal, filter, stride, padding, dilation); break;
+            case f64: output = convolve2Strided<double ,  double>(signal, filter, stride, padding, dilation); break;
+            case u32: output = convolve2Strided<uint   ,   float>(signal, filter, stride, padding, dilation); break;
+            case s32: output = convolve2Strided<int    ,   float>(signal, filter, stride, padding, dilation); break;
+            case u16: output = convolve2Strided<ushort ,   float>(signal, filter, stride, padding, dilation); break;
+            case s16: output = convolve2Strided<short  ,   float>(signal, filter, stride, padding, dilation); break;
+            case u64: output = convolve2Strided<uintl  ,   float>(signal, filter, stride, padding, dilation); break;
+            case s64: output = convolve2Strided<intl   ,   float>(signal, filter, stride, padding, dilation); break;
```
9prady9: accumulation of `uintl` and `intl` should be `double`

src/api/c/convolve.cpp
```diff
@@ -254,3 +300,110 @@ af_err af_convolve2_sep(af_array *out, const af_array signal, const af_array col
...
 }
+
+template<typename T, typename accT>
+af_array conv2GradCall(const af_array incoming_gradient,
+                       const af_array original_signal,
+                       const af_array original_filter,
+                       const af_array convolved_output,
+                       af::dim4 stride, af::dim4 padding, af::dim4 dilation,
+                       af_conv_gradient_type gradType)
+{
```
9prady9: Since conv2GradCall doesn't handle complex inputs, we can remove `accT` template paramter and do the following inside this function.

```c++
using accT = typename std::conditional<std::is_same<T, double>::value, double, float>::type;
```

9prady9: This way, the switch case surrounding the `conv2GradCall`'s inside `af_convolve2Gradient` can be further trimmed down.

Actually, this removal of extra template parameter can trickle down all the way down to kernel implementation for `conv2FilterGradient`.

src/api/c/convolve.cpp
```diff
@@ -228,6 +229,51 @@ af_err af_convolve2(af_array *out, const af_array signal, const af_array filter,
...
+        const ArrayInfo& fInfo = getInfo(filter);
+
+        af::dim4 sDims  = sInfo.dims();
+        af::dim4 fDims  = fInfo.dims();
+
+        const af_dtype signalType  = sInfo.getType();
+
+        dim4 stride(stride_dims, strides);
+        dim4 padding(padding_dims, paddings);
+        dim4 dilation(dilation_dims, dilations);
```
9prady9: I am not sure if any values are allowed, already asked this in the comment I made earlier in the header file.

If all values for `strides`, `paddings` and `dilations` are not allowed, we should check for valid values.

src/backend/cpu/convolve.cpp
```diff
@@ -107,5 +119,197 @@ INSTANTIATE(ushort ,   float)
...
+#undef INSTANTIATE
+
+template<typename T, typename accT>
+Array<T> conv2DataGradient(const Array<T>& incoming_gradient,
+                           const Array<T>& original_signal,
+                           const Array<accT>& original_filter,
+                           const Array<T>& convolved_output,
+                           af::dim4 stride, af::dim4 padding, af::dim4 dilation) {
+    original_signal.eval();
+    original_filter.eval();
```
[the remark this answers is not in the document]
9prady9: Actually, as of now there is no `flip` backend level.

src/backend/opencl/convolve.cpp
```diff
@@ -85,5 +94,197 @@ INSTANTIATE(ushort ,   float)
...
+
+template<typename T, typename accT>
+Array<T> convolve2(Array<T> const& signal, Array<accT> const& filter,
+                   const dim4 stride, const dim4 padding, const dim4 dilation) {
+    signal.eval();
+    filter.eval();
+
+    Array<T> out  = createEmptyArray<T>(dim4());
+    out = convolve2_unwrap<T, accT>(signal, filter, stride, padding, dilation);
+
```
9prady9: A separate function `convolve2_unwrap` is not necessary if it is only used here. That way you can even avoid the extra call to `createEmptyArray`

src/backend/opencl/convolve.cpp
```diff
@@ -85,5 +94,197 @@ INSTANTIATE(ushort ,   float)
...
+    dim_t outputWidth  = 1 + (sDims[0] + 2 * padding[0] - (((fDims[0] - 1) * dilation[0]) + 1)) / stride[0];
+    dim_t outputHeight = 1 + (sDims[1] + 2 * padding[1] - (((fDims[1] - 1) * dilation[1]) + 1)) / stride[1];
+    Array<T> out = createValueArray<T>(dim4(outputWidth, outputHeight, fDims[3], sDims[3]), scalar<T>(0));
+
+    const bool retCols = false;
+    Array<accT> unwrapped = unwrap_dilated(cast<accT>(signal), fDims[0], fDims[1],
+                                                             stride[0], stride[1],
+                                                           padding[0], padding[1],
+                                                         dilation[0], dilation[1], retCols);
```
9prady9: the parameter indentations are off.

src/backend/opencl/convolve.cpp
```diff
@@ -85,5 +94,197 @@ INSTANTIATE(ushort ,   float)
...
+                          const dim4 stride,
+                          const dim4 padding,
+                          const dim4 dilation)
+{
+    dim4 sDims = signal.dims();
+    dim4 fDims = filter.dims();
+
+    dim_t outputWidth  = 1 + (sDims[0] + 2 * padding[0] - (((fDims[0] - 1) * dilation[0]) + 1)) / stride[0];
+    dim_t outputHeight = 1 + (sDims[1] + 2 * padding[1] - (((fDims[1] - 1) * dilation[1]) + 1)) / stride[1];
+    Array<T> out = createValueArray<T>(dim4(outputWidth, outputHeight, fDims[3], sDims[3]), scalar<T>(0));
```
9prady9: We don't need a `createValueArray` here when we are actually setting `out` in line 138 from the output of another operation.

src/backend/opencl/pool.cpp
```diff
@@ -0,0 +1,117 @@
...
+    udims[0] -= 1;
+    Array<T> max_unwrapped = reduce<af_max_t, T, T>(unwrapped, 0);
+
+
+    dim_t outputWidth  = 1 + (in.dims()[0] + 2 * padding_width - pool_width)   / stride_width;
+    dim_t outputHeight = 1 + (in.dims()[1] + 2 * padding_height - pool_height) / stride_height;
+
+    max_unwrapped.modDims(dim4(outputWidth, outputHeight, in.dims()[2], in.dims()[3]));
+    Array<T> out = createValueArray<T>(max_unwrapped.dims(), 0);
+    out = max_unwrapped;
```
9prady9: Can't we just return `max_unwrapped` ? Is it because of moddims ?

src/backend/opencl/pool.cpp
```diff
@@ -0,0 +1,117 @@
...
```
9prady9: same comment as on line line 93

src/backend/cuda/pool.cu
```diff
@@ -0,0 +1,394 @@
...
+                         const Array<T>& pooled_output,
+                         const dim_t pool_width, const dim_t pool_height,
+                         const dim_t padding_width, const dim_t padding_height,
+                         const int stride_width, const int stride_height,
+                         af_pooling_type pool_type)
+{
+    dim4 idims = incoming_gradient.dims();
+
+    cudnnHandle_t cudnn;
+    CUDNN(cudnnCreate(&cudnn));
```
9prady9: this is should move into platform.hpp/cpp and should be `thread_local` there if cudnn is similar to cublas or other libs from nvidia.

We shouldn't create cudnn context everytime this function is used.

src/backend/cuda/pool.cu
```diff
@@ -0,0 +1,394 @@
...
+    udims[0] -= 1;
+    Array<T> max_unwrapped = reduce<af_max_t, T, T>(unwrapped, 0);
+
+
+    dim_t outputWidth  = 1 + (in.dims()[0] + 2 * padding_width - pool_width)   / stride_width;
+    dim_t outputHeight = 1 + (in.dims()[1] + 2 * padding_height - pool_height) / stride_height;
+
+    max_unwrapped.modDims(dim4(outputWidth, outputHeight, in.dims()[2], in.dims()[3]));
+    Array<T> out = createValueArray<T>(max_unwrapped.dims(), 0);
+    out = max_unwrapped;
```
9prady9: same question from opencl backend

src/backend/cuda/pool.cu
```diff
@@ -0,0 +1,394 @@
...
```
9prady9: same as line 233

src/backend/cuda/pool.cu
9prady9: Also, this is may be causing a leak since there are no matching `cudnnDestroy` calls.

src/backend/cuda/pool.cu
```diff
@@ -0,0 +1,394 @@
...
+Array<T> pool2_cudnn(const Array<T>& in,
+                     const dim_t pool_width, const dim_t pool_height,
+                     const dim_t padding_width, const dim_t padding_height,
+                     const dim_t stride_width, const dim_t stride_height,
+                     af_pooling_type pool_type)
+{
+    dim4 idims = in.dims();
+
+    cudnnHandle_t cudnn;
+    CUDNN(cudnnCreate(&cudnn));
```
9prady9: Same issue as the other location where cudnnCreate is called.

src/backend/cuda/pool.cu
```diff
@@ -0,0 +1,394 @@
...
+            return CUDNN_DATA_UINT8; break;
+        default:
+            return CUDNN_DATA_FLOAT;
+    }
+}
+
+template<typename T>
+bool supported_cudnn_type() {
+    af_dtype dtype = ((af_dtype) dtype_traits<T>::af_type);
+    //is u8 == int8? s16 == half?
```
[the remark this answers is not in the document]
9prady9: to add to that, if we have half it will be `f16` following the convention of `f32` or `f64`

src/backend/cuda/convolve.cpp
```diff
@@ -98,5 +139,413 @@ INSTANTIATE(ushort ,   float)
...
+Array<T> convolve2_cudnn(const Array<T>& signal,
+                         const Array<accT>& filter,
+                         const dim4 stride,
+                         const dim4 padding,
+                         const dim4 dilation)
+{
+    dim4 sDims = signal.dims();
+    dim4 fDims = filter.dims();
+
+    cudnnHandle_t cudnn;
```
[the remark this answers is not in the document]
9prady9: @syurkevi note my comments about this below as well. You should see how cublas or cusparse handles are handled.

src/backend/cpu/pool.cpp
```diff
@@ -0,0 +1,130 @@
...
+    dim4 udims = unwrapped.dims();
+    udims[0] -= 1;
+    Array<T> max_unwrapped = reduce<af_max_t, T, T>(unwrapped, 0);
+
+    dim_t outputWidth  = 1 + (in.dims()[0] + 2 * padding_width - pool_width)   / stride_width;
+    dim_t outputHeight = 1 + (in.dims()[1] + 2 * padding_height - pool_height) / stride_height;
+
+    max_unwrapped.modDims(dim4(outputWidth, outputHeight, in.dims()[2], in.dims()[3]));
+    Array<T> out = createValueArray<T>(max_unwrapped.dims(), 0);
+    out = max_unwrapped;
```
9prady9: can't we return `max_unwrapped` itself ?

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,67 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
     }
 }
 
+template<typename T, int d>
```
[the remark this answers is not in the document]
9prady9: Yes, it won't help because d is indexing `dims` and the value of `dims[d]` is not a constant.

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -76,5 +76,67 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
     }
 }
 
+template<typename T, int d>
```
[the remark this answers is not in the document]
9prady9: I don't think it is even a little bit beneficial because d is not directly used in any looping condition. Only in the if check.

src/backend/cpu/convolve.cpp
```diff
@@ -107,5 +119,197 @@ INSTANTIATE(ushort ,   float)
...
+
+    const dim4 cDims = incoming_gradient.dims();
+    const dim4 sDims = original_signal.dims();
+    const dim4 fDims = original_filter.dims();
+
+    Array<accT> collapsed_filter = original_filter;
+
+    vector<af_seq> flip_index(4);
+    af_seq s = {(double)(fDims[0] - 1), 0, -1};
```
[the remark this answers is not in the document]
9prady9: no at the backend level.

src/backend/cuda/cudnn.hpp
```diff
@@ -0,0 +1,44 @@
...
+ ********************************************************/
+
+#pragma once
+#include <cudnn.h>
+#include <common/defines.hpp>
+#include <common/NNHandle.hpp>
+
+namespace cuda
+{
+typedef cudnnHandle_t NNHandle;
```
9prady9: NNHandle, isn't it little too unclear what it is. Probably expand it ? or `NeuralNetHandle`.

src/backend/cuda/platform.hpp
```diff
@@ -92,6 +93,8 @@ PlanCache& fftManager();
 
 BlasHandle blasHandle();
 
+NNHandle nnHandle();
```
9prady9: may be `neuralNetHandle` ? I personally think `nn` is too short and unclear.

9prady9:
also, rebase your branch.

src/.clang-format
```diff
@@ -1,11 +1,10 @@
----
 Language:        Cpp
```
mark-poscablo: Since #2407 has been closed, I think this file needs to be taken off this PR?

src/api/c/convolve.cpp
```diff
@@ -13,25 +13,31 @@
 #include <convolve.hpp>
 #include <fftconvolve.hpp>
 #include <handle.hpp>
+#include <handle.hpp>
```
mark-poscablo: Duplicate of same include on line 15; please remove this line

src/api/cpp/convolve.cpp
```diff
@@ -9,13 +9,16 @@
 
 #include <af/array.h>
 #include <af/compatible.h>
+#include <af/dim4.hpp>
+#include <af/ml.h>
 #include <af/signal.h>
 #include <algorithm>
 #include "error.hpp"
+#include "error.hpp"
```
mark-poscablo: Duplicate of line 16; please remove

src/backend/cpu/convolve.cpp
```diff
@@ -8,20 +8,31 @@
...
 #include <common/defines.hpp>
 #include <convolve.hpp>
+#include <handle.hpp>
 #include <kernel/convolve.hpp>
 #include <platform.hpp>
+#include <reorder.hpp>
+#include <transpose.hpp>
+#include <unwrap.hpp>
+#include <wrap.hpp>
+#include <iostream>
```
mark-poscablo: Please correct me if I'm wrong, but I don't think iostream is actually being used here. Please remove this include if so.

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -78,5 +78,68 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+                            optr[oloc] = scalar<T>(0.0);
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+}
+}
```
mark-poscablo: Minor thing, but can these brackets be commented on which namespaces they actually close? Like they previously did:

```
}  // namespace kernel
}  // namespace cpu
```

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,67 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+                            optr[oloc] += iptr[iloc];
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+}
+}
```
mark-poscablo: Here too, can these brackets be commented on which namespaces they actually close?

```
}  // namespace kernel
}  // namespace cpu
```

src/backend/cuda/convolve.cpp
```diff
@@ -8,17 +8,32 @@
  ********************************************************/
 
 #include <Array.hpp>
+#include <cast.hpp>
 #include <convolve.hpp>
+#include <cudnn.hpp>
 #include <err_cuda.hpp>
 #include <kernel/convolve.hpp>
+#include <kernel/convolve.hpp>
```
mark-poscablo: Duplicate include as line 15; please remove.

src/backend/cuda/convolve.cpp
```diff
@@ -8,17 +8,32 @@
...
+#include <platform.hpp>
+#include <af/dim4.hpp>
 #include <af/dim4.hpp>
```
mark-poscablo: Lines 18 and 19 are the same includes. Please remove either one.

src/backend/cuda/convolve.hpp
```diff
@@ -11,12 +11,29 @@
...
+                           const Array<T> &convolved_output, af::dim4 stride,
+                           af::dim4 padding, af::dim4 dilation);
+
+template <typename T, typename accT>
+Array<T> conv2FilterGradient(const Array<T> &incoming_gradient,
+                             const Array<T> &original_signal,
+                             const Array<accT> &original_filter,
+                             const Array<T> &convolved_output, af::dim4 stride,
+                             af::dim4 padding, af::dim4 dilation);
+}
```
mark-poscablo: Just for consistency, can we add back `// namespace cuda` here?

src/backend/cpu/convolve.cpp
```diff
@@ -124,5 +135,168 @@ INSTANTIATE(ushort, float)
...
+
+INSTANTIATE(double, double)
+INSTANTIATE(float, float)
+INSTANTIATE(uint, float)
+INSTANTIATE(int, float)
+INSTANTIATE(uchar, float)
+INSTANTIATE(ushort, float)
+INSTANTIATE(short, float)
+#undef INSTANTIATE
+}
```
mark-poscablo: Just for consistency, can we add back `// namespace cpu` here?

src/backend/cpu/convolve.hpp
```diff
@@ -12,12 +12,29 @@
...
+                           const Array<T> &convolved_output, af::dim4 stride,
+                           af::dim4 padding, af::dim4 dilation);
+
+template <typename T, typename accT>
+Array<T> conv2FilterGradient(const Array<T> &incoming_gradient,
+                             const Array<T> &original_signal,
+                             const Array<accT> &original_filter,
+                             const Array<T> &convolved_output, af::dim4 stride,
+                             af::dim4 padding, af::dim4 dilation);
+}
```
mark-poscablo: Minor thing, and just for consistency, but can we add back `// namespace cpu` here?

src/backend/cuda/convolve.cpp
```diff
@@ -112,5 +127,310 @@ INSTANTIATE(ushort, float)
...
+
+INSTANTIATE(double, double)
+INSTANTIATE(float, float)
+INSTANTIATE(uint, float)
+INSTANTIATE(int, float)
+INSTANTIATE(uchar, float)
+INSTANTIATE(ushort, float)
+INSTANTIATE(short, float)
+#undef INSTANTIATE
+}
```
mark-poscablo: Minor thing, but can we add back `// namespace cuda` here?

src/backend/cuda/platform.hpp
```diff
@@ -11,8 +11,10 @@
 
 #include <GraphicsResourceManager.hpp>
 #include <cublas.hpp>
+#include <cublas.hpp>
```
mark-poscablo: Duplicate include of line 13; please remove

src/backend/opencl/convolve.cpp
```diff
@@ -103,5 +112,184 @@ INSTANTIATE(ushort, float)
...
+Array<T> convolve2_unwrap(const Array<T> &signal, const Array<accT> &filter,
+                          const dim4 stride, const dim4 padding,
+                          const dim4 dilation) {
+    dim4 sDims = signal.dims();
+    dim4 fDims = filter.dims();
+
+    dim_t outputWidth =
+        1 +
+        (sDims[0] + 2 * padding[0] - (((fDims[0] - 1) * dilation[0]) + 1)) /
+            stride[0];
```
mark-poscablo: This seems like an odd kind of spacing/indentation. I suppose these were automatically done by clang-format?

src/backend/opencl/convolve.hpp
```diff
@@ -11,12 +11,29 @@
...
+                           const Array<T> &convolved_output, af::dim4 stride,
+                           af::dim4 padding, af::dim4 dilation);
+
+template <typename T, typename accT>
+Array<T> conv2FilterGradient(const Array<T> &incoming_gradient,
+                             const Array<T> &original_signal,
+                             const Array<accT> &original_filter,
+                             const Array<T> &convolved_output, af::dim4 stride,
+                             af::dim4 padding, af::dim4 dilation);
+}
```
mark-poscablo: Minor thing, but can we add back `// namespace opencl` here?

test/.clang-format
```diff
@@ -1,11 +1,10 @@
----
 Language:        Cpp
 # BasedOnStyle:  Google
```
mark-poscablo: We should probably take this file off from the PR because #2407 was closed.

test/convolve.cpp
```diff
@@ -7,8 +7,10 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
+#define GTEST_LINKED_AS_SHARED_LIBRARY 1
 #include <arrayfire.h>
 #include <gtest/gtest.h>
+#include <gtest/gtest.h>
```
mark-poscablo: Duplicate include as line 12; please remove.

src/backend/cuda/convolve.hpp
9prady9: File is too short to have that.

src/backend/opencl/convolve.cpp
9prady9: yup

test/.clang-format
9prady9: I am not sure about let umar respond on that.

src/.clang-format
9prady9: I am not sure about let umar respond on that.

9prady9:
@mark-poscablo Addressed your feedback but I didn't remove the clang format files because they are present in the master. We just have to figure if the changes @syurkevi made are valid. @umar456 ?

src/.clang-format
umar456: The changes can be reverted. Not the file

9prady9:
Removed changes to clang-format.

src/.clang-format
mark-poscablo: Thanks for reverting the changes - for some reason I was thinking that the clang format file wasn't in master yet.

test/.clang-format
mark-poscablo: Thanks for reverting the changes - for some reason I was thinking that the clang format file wasn't in master yet.

include/af/signal.h
```diff
@@ -590,6 +590,23 @@ AFAPI array convolve1(const array& signal, const array& filter, const convMode m
...
+   \param[in]  stride specifies the filter strides along each dimension
+   \param[in]  padding specifies the padding along each dimension
+   \param[in]  dilation specifies the amount to dilate the filter before convolution
+   \return     the convolved array
+
+   \ingroup signal_func_convolve2
+ */
+AFAPI array convolve2(const array& signal, const array& filter,
+                      const dim4 stride, const dim4 padding, const dim4 dilation);
+
```
9prady9: Strided convolution doesn't handle all the combinations that our old convolutions handle from what I saw in the code - I am referring to the combination table in the docs [here](http://arrayfire.org/docs/group__signal__func__convolve2.htm).

What I mean precisely is that this function has a narrow range of inputs compared to earlier convolve2 which should be mentioend in docs

src/backend/cpu/convolve.cpp
```diff
@@ -124,5 +134,169 @@ INSTANTIATE(ushort, float)
...
+    vector<af_seq> flip_index(4);
+    af_seq s      = {(double)(fDims[0] - 1), 0, -1};
+    flip_index[0] = s;
+    s             = {(double)(fDims[1] - 1), 0, -1};
+    flip_index[1] = s;
+    flip_index[2] = af_span;
+    flip_index[3] = af_span;
+
+    Array<accT> collapsedFilter = createSubArray(filter, flip_index);
+    collapsedFilter.modDims(dim4(fDims[0] * fDims[1] * fDims[2], fDims[3]));
```
9prady9: Shouldn't this be `collapsedFilter.modDims(dim4(fDims[0] * fDims[1], fDims[2], fDims[3]));` ?

src/backend/cpu/convolve.cpp
```diff
@@ -124,5 +134,169 @@ INSTANTIATE(ushort, float)
...
+    const dim4 sDims = original_signal.dims();
+    const dim4 fDims = original_filter.dims();
+
+    vector<af_seq> flip_index(4);
+    af_seq s      = {(double)(fDims[0] - 1), 0, -1};
+    flip_index[0] = s;
+    s             = {(double)(fDims[1] - 1), 0, -1};
+    flip_index[1] = s;
+    flip_index[2] = af_span;
+    flip_index[3] = af_span;
```
9prady9: the flipping operation can be extracted out into a function like `genFlipIndex(std::array<af_seq, 4> &out, dim4 dims)` since it is being used in multiple locations. Also, std::array should be used instead of std::vector here.

which would turn the above section into something like
```
std::array<af_seq, 4> flip_index;
genFlipIndex(flip_index, fDims);
```

src/backend/cpu/convolve.cpp
```diff
@@ -124,5 +134,169 @@ INSTANTIATE(ushort, float)
...
+
+    Array<accT> collapsed_filter = createSubArray(original_filter, flip_index);
+    collapsed_filter.modDims(dim4(fDims[0] * fDims[1] * fDims[2], fDims[3]));
```
9prady9: Isn't `fDims[2]` batch data dimension ?

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -78,5 +78,69 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+
+    af::dim4 idims    = in.dims();
+    af::dim4 odims    = out.dims();
+    af::dim4 istrides = in.strides();
+    af::dim4 ostrides = out.strides();
+
+    dim_t nx = 1 + (idims[0] + 2 * px - (((wx - 1) * dx) + 1)) / sx;
+
+    for (dim_t w = 0; w < odims[3]; w++) {
+        for (dim_t z = 0; z < odims[2]; z++) {
```
9prady9: These two loops can be combined into a single batch loop, isn't it ?

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -76,5 +76,67 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+                dim_t winx = (col % nx);
+
+                dim_t startx = winx * sx;
+                dim_t starty = winy * sy;
+
+                dim_t spx = startx - px;
+                dim_t spy = starty - py;
+
+                // Short cut condition ensuring all values within input dimensions
+                bool cond = (spx >= 0 && spx + (wx * dx) < idims[0] &&
```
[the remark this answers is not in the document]
9prady9: I doesn't seem like that because `cond` is based on loop index `col` but I believe it can pre-computed as a vector completely out of the batch loop since it seems to depend on `odims[d]`, `nx`, `px` & `py`, `sx` & `sy` which all are loop invariant in computing `cond`.

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -78,5 +78,69 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+
+                // Short cut condition ensuring all values within input
+                // dimensions
+                bool cond = (spx >= 0 && spx + (wx * dx) < idims[0] &&
+                             spy >= 0 && spy + (wy * dy) < idims[1]);
+
+                for (dim_t y = 0; y < wy; y++) {
+                    for (dim_t x = 0; x < wx; x++) {
+                        dim_t xpad = spx + x * dx;
+                        dim_t ypad = spy + y * dy;
```
9prady9: `ypad` and corresponding boolean condition calculation can be moved into outer loop, isn't it ?

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,68 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+                dim_t starty = winy * sy;
+
+                dim_t spx = startx - px;
+                dim_t spy = starty - py;
+
+                // Short cut condition ensuring all values within input
+                // dimensions
+                bool cond = (spx >= 0 && spx + (wx * dx) < odims[0] &&
+                             spy >= 0 && spy + (wy * dy) < odims[1]);
+
```
9prady9: same as unwrap_dilated, I believe this `cond` values can be pre-computed before the loops.

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,68 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+    af::dim4 idims    = in.dims();
+    af::dim4 odims    = out.dims();
+    af::dim4 istrides = in.strides();
+    af::dim4 ostrides = out.strides();
+
+    dim_t nx = 1 + (odims[0] + 2 * px - (((wx - 1) * dx) + 1)) / sx;
+
+    for (dim_t w = 0; w < idims[3]; w++) {
+        for (dim_t z = 0; z < idims[2]; z++) {
```
9prady9: merge into a single batch loop

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,68 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+
+                // Short cut condition ensuring all values within input
+                // dimensions
+                bool cond = (spx >= 0 && spx + (wx * dx) < odims[0] &&
+                             spy >= 0 && spy + (wy * dy) < odims[1]);
+
+                for (dim_t y = 0; y < wy; y++) {
+                    for (dim_t x = 0; x < wx; x++) {
+                        dim_t xpad = spx + x * dx;
+                        dim_t ypad = spy + y * dy;
```
9prady9: `ypad` and corresponding boolean condition calculation can be moved into outer loop, isn't it ?

src/backend/cpu/wrap.cpp
```diff
@@ -56,5 +56,45 @@ INSTANTIATE(uchar)
...
+    af::dim4 idims = in.dims();
+    af::dim4 odims(ox, oy, idims[2], idims[3]);
+
+    Array<T> out = createValueArray<T>(odims, scalar<T>(0));
+    out.eval();
+    in.eval();
+
+    const int d = is_column;
+    getQueue().enqueue(kernel::wrap_dim_dilated<T>, out, in, wx, wy, sx, sy, px,
+                       py, dx, dy, d);
```
9prady9: `is_column` can be directly passed to the enqueue for `d`.

src/backend/cpu/unwrap.cpp
```diff
@@ -60,5 +60,51 @@ INSTANTIATE(uchar)
...
+INSTANTIATE(cdouble)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(intl)
+INSTANTIATE(uintl)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
+#undef INSTANTIATE
```
9prady9: I think we can remove the instantiations for cfloat, cdouble , intl and uintl since I removed the those instantiations in strided convolution and convolve2Gradient @syurkevi Can you please confirm

src/backend/cpu/wrap.cpp
```diff
@@ -56,5 +56,45 @@ INSTANTIATE(uchar)
...
+INSTANTIATE(cdouble)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(intl)
+INSTANTIATE(uintl)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
+#undef INSTANTIATE
```
9prady9: I think we can remove the instantiations for cfloat, cdouble , intl and uintl since I removed the those instantiations in strided convolution and convolve2Gradient @syurkevi Can you please confirm

src/backend/opencl/convolve.cpp
```diff
@@ -103,5 +112,184 @@ INSTANTIATE(ushort, float)
...
+
+    Array<accT> collapsedFilter = filter;
+
+    vector<af_seq> flip_index(4);
+    af_seq s      = {(double)(fDims[0] - 1), 0, -1};
+    flip_index[0] = s;
+    s             = {(double)(fDims[1] - 1), 0, -1};
+    flip_index[1] = s;
+    flip_index[2] = af_span;
+    flip_index[3] = af_span;
```
9prady9: Perhaps, this `genFlipIndex` helper I suggested earlier should be moved somewhere into `src/backend/common`

src/backend/opencl/unwrap.cpp
```diff
@@ -51,4 +52,48 @@ INSTANTIATE(uchar)
 INSTANTIATE(char)
 INSTANTIATE(short)
 INSTANTIATE(ushort)
+#undef INSTANTIATE
+
+template <typename T>
+Array<T> unwrap_dilated(const Array<T> &in, const dim_t wx, const dim_t wy,
```
9prady9: I believe we follow camelCase convention for backend API or all normal C++ functions. Of course there are some exceptions from the past, but I think we should stick to the convention.

src/backend/opencl/unwrap.cpp
```diff
@@ -51,4 +52,48 @@ INSTANTIATE(uchar)
...
+INSTANTIATE(cdouble)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(intl)
+INSTANTIATE(uintl)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
+#undef INSTANTIATE
```
9prady9: Duplicate remark to remove cfloat, cdouble, intl and uintl instantiations.

src/backend/opencl/wrap.cpp
```diff
@@ -47,4 +47,39 @@ INSTANTIATE(uchar)
...
+INSTANTIATE(cdouble)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(intl)
+INSTANTIATE(uintl)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
+#undef INSTANTIATE
```
9prady9: Duplicate remark to remove cfloat, cdouble, intl and uintl instantiations.

test/convolve.cpp
```diff
@@ -804,3 +805,335 @@ TEST(DISABLED_ConvolveLargeDim3D, CPP) {
...
+    convolve2stridedTest<TypeParam>(
+        string(TEST_DIR "/convolve/sig81032_filt3334_s11_p11_d11.test"),
+        dim4(1, 1), dim4(1, 1), dim4(1, 1));
+}
+
+TYPED_TEST(ConvolveStrided, Gradient_sig81032_filt3334_s11_p11_d11) {
+    convolve2GradientTest<TypeParam>(
+        string(TEST_DIR "/convolve/sig81032_filt3334_s11_p11_d11.test"),
+        dim4(1, 1), dim4(1, 1), dim4(1, 1));
+}
```
9prady9: Is batch support available in all new convolution functions ? If yes, we should add tests for that also.

src/backend/opencl/kernel/wrap_dilated.cl
```diff
@@ -0,0 +1,79 @@
...
+            fx /= dx;
+
+            int win_end = win_end_y + fx;
+            int dim_end = dim_end_y + x;
+
+            if (is_column) {
+                idx = dim_end * in.strides[1] + win_end;
+            } else {
+                idx = dim_end + win_end * in.strides[1];
+            }
```
9prady9: I think we should replace `if` with something like the following.

```cpp
int offset = is_column * dim_end + (1 - is_column) * win_end;
int scalet = is_column *  win_end + (1 - is_column) * dim_end ;
idx = offset + scalet * in.strides[1];
```

src/backend/cuda/convolve.cpp
```diff
@@ -112,5 +125,311 @@ INSTANTIATE(ushort, float)
...
+        cudnn, &alpha, input_descriptor, cast<accT>(signal).device(),
+        filter_descriptor, filter.device(), convolution_descriptor,
+        convolution_algorithm, (void *)workspace_buffer.get(), workspace_bytes,
+        &beta, output_descriptor, out.device()));
+
+    // destroy all descriptors
+    CUDNN_CHECK(cudnnDestroyTensorDescriptor(input_descriptor));
+    CUDNN_CHECK(cudnnDestroyTensorDescriptor(output_descriptor));
+    CUDNN_CHECK(cudnnDestroyFilterDescriptor(filter_descriptor));
+    CUDNN_CHECK(cudnnDestroyConvolutionDescriptor(convolution_descriptor));
```
9prady9: @syurkevi @umar456 do you think there will be advantage if we cache these descriptors ?

Irrespective of that, we should wrap these cudnn resources to their respective RAII constructs. we already have cudnn.hpp/cpp , we should just place those definitions in there.

umar456: I think we looked into it and it didn't take a long time to create these. I think it may make a difference if add convolutions to the JIT tree. 

I agree these should be wrapped in a unique_ptr or something.

9prady9:
@syurkevi The cudnn cmake configure issue haven't seem to be resolved still.

src/backend/cuda/cudnn.cpp
```diff
@@ -0,0 +1,40 @@
...
+ ********************************************************/
+
+#include <common/err_common.hpp>
+#include <cudnn.hpp>
+
+namespace cuda {
+
+const char *errorString(cudnnStatus_t err) {
+    switch (err) {
+        case CUDNN_STATUS_SUCCESS: return "CUBLAS_STATUS_SUCCESS";
```
WilliamTambellini: CUDNN ?

include/af/signal.h
umar456: We have renamed the function convolve2_v2. This new function will batch using a different approach compared to the first version

src/backend/cpu/convolve.cpp
umar456: The filter can be 3+ dimensional in a 2d convolve.

src/backend/cpu/convolve.cpp
umar456: Not in this version of convolve.

src/backend/cpu/unwrap.cpp
umar456: I think its fine to have this function instantiated for those types. Its more generic than convolve.

src/backend/common/indexing_helpers.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace common {
+// will generate indexes to flip input array
+// of size original dims according to axes specified in flip
+template<typename T>
+detail::Array<T> flip(const detail::Array<T> &in, const std::array<bool, AF_MAX_DIMS> flip) {
+
+    std::vector<af_seq> index(4, af_span);
+    af::dim4 dims = in.dims();
+
```
9prady9: Nit: extra empty line.

src/backend/cpu/kernel/unwrap.hpp
```diff
@@ -33,12 +33,12 @@ void unwrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
         for (dim_t z = 0; z < odims[2]; z++) {
```
9prady9: batch loops can be merged.

src/backend/cpu/kernel/unwrap.hpp
9prady9: please address the above comment.

src/backend/cpu/kernel/unwrap.hpp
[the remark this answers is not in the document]
9prady9: It seems like `cond` calculation has changed since I made my comment. Is that correct ?

src/backend/cpu/kernel/wrap.hpp
9prady9: pls merge into a single batch loop

src/backend/cpu/kernel/wrap.hpp
```diff
@@ -75,5 +75,68 @@ void wrap_dim(Param<T> out, CParam<T> in, const dim_t wx, const dim_t wy,
...
+            dim_t cOut = w * ostrides[3] + z * ostrides[2];
+            const T* iptr_ = inPtr  + cIn;
+            T* optr= outPtr + cOut;
+
+            for(dim_t col = 0; col < idims[d]; col++) {
+                // Offset output ptr
+                const T* iptr = iptr_ + col * istrides[d];
+
+                // Calculate input window index
+                dim_t winy = (col / nx);
```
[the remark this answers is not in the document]
9prady9: that would bring in a branch statement based on value of `nx`. Do you think it would be faster ?

src/backend/cpu/unwrap.cpp
9prady9: We would be only increasing our compilation times without any added benefits since `unwrap_dilated` is not exposed to the user at the moment. `unwrap_dilated` is only used internally by the new convolution functions that were added in this PR. All the other instantiations except `float` & `double` are basically not used at all anywhere else in the code. These extra instantiations can be added later if and when required.

src/backend/cpu/wrap.cpp
```diff
@@ -56,5 +56,40 @@ INSTANTIATE(uchar)
...
+        const dim_t py, const dim_t dx, const dim_t dy, const bool is_column);
+
+INSTANTIATE(float)
+INSTANTIATE(double)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
```
9prady9: Same as my comment about `unwrap_dilated`.

src/backend/opencl/wrap.cpp
```diff
@@ -47,4 +47,35 @@ INSTANTIATE(uchar)
...
+        const dim_t py, const dim_t dx, const dim_t dy, const bool is_column);
+
+INSTANTIATE(float)
+INSTANTIATE(double)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(uchar)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
```
9prady9: Reminder remark to remove unnecessary instantiations.

src/backend/opencl/unwrap.cpp
```diff
@@ -51,4 +52,44 @@ INSTANTIATE(uchar)
...
+        const dim_t dy, const bool is_column);
+
+INSTANTIATE(float)
+INSTANTIATE(double)
+INSTANTIATE(int)
+INSTANTIATE(uint)
+INSTANTIATE(intl)
+INSTANTIATE(char)
+INSTANTIATE(short)
+INSTANTIATE(ushort)
```
9prady9: Reminder remark to remove unnecessary instantiations.

src/backend/opencl/kernel/wrap_dilated.cl
9prady9: Any objection to above suggestion ?

src/backend/cuda/platform.cpp
```diff
@@ -379,6 +379,24 @@ BlasHandle blasHandle() {
...
+    CUDNN_CHECK(
+        cudnnSetStream(cudnnHandles[deviceId], cuda::getStream(deviceId)));
+
+    return cudnnHandles[deviceId];
+}
+
+cudnnHandle_t nnHandle() {
+    return nnManager(cuda::getActiveDeviceId());
+}
+
```
9prady9: We should reset/release the cudnn resources inside `~DeviceManager` like we do for other cublas/cusolver/cusparse resource handlers .

https://github.com/arrayfire/arrayfire/blob/0d6fc70dab7d4882ef9d2d5ddf6ad2f71d18dda7/src/backend/cuda/platform.cpp#L144

test/convolve.cpp
9prady9: the numbers are confusing me a bit, can you please tell if there are batch tests among these.

WilliamTambellini:
Could you just confirm that after merging that one, we ll just need a local cudnn sdk (include+lib) in order to compile afcuda ?
https://developer.nvidia.com/rdp/cudnn-download

umar456:
Yes, you will need to have cuDNN installed on the system. Currently we are installing it in the cuda directly inline. If there are users who which to specify the location of cuDNN please make an issue and we can discuss.

WilliamTambellini:
On my side thats ok to install it in /usr/local/cuda.
Does it link with static or dyn cudnn ? 
Have you updated the packaging script(s) in order to embed libcudnn.so in the af installer ?

umar456:
Not with this commit. It should be done in the next couple of days.

PR #2362, opened 2018-11-29
Remove cudaMemcpys to the default streams in the CUDA backend

umar456:
There were some calls being made to the default CUDA stream. This is
problematic because the default cuda stream is blocks and blocks
all other streams when executing. We should only perform calls on
the current stream.

### DIFF
--- a/src/backend/cuda/copy.cu
+++ b/src/backend/cuda/copy.cu
@@ -18,47 +18,49 @@ namespace cuda
 {
 
     template<typename T>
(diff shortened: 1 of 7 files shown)


WilliamTambellini:
Cool. Could we expect any significant speedup ?

umar456:
I don't think there would be a speedup in your application here. We are still synchronizing after the call to cudaMemcpy because its interacting with user code. If you were using other kernels in your application you could see a small speedup then

PR #2363, opened 2018-11-30
Fix leaks in meanvar tests

umar456:
Fix several leaks in the meanvar tests.

### DIFF
--- a/test/meanvar.cpp
+++ b/test/meanvar.cpp
@@ -23,6 +23,7 @@ using af::cfloat;
 using af::dim4;
 using af::dtype_traits;
 using std::back_inserter;
+using std::move;
 using std::string;
 using std::vector;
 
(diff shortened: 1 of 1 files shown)


test/meanvar.cpp
```diff
@@ -178,20 +192,20 @@ meanvar_test_gen(string name, int in_index, int weight_index, af_var_bias bias,
   }
   if(weight_index == -1) {
       return meanvar_test<T> (name,
-                              inputs[in_index],
-                              empty,
+                              move(inputs[in_index]),
```
9prady9: I don't understand why inputs has to be a vector of af_array when there is only one af_array in use in all cases. Am I missing to see something ?

9prady9: Even `outputs` for that matter I think. I can see why `tests_` has to be vector of vectors but not why `outputs` has to be a vector of vectors.

test/meanvar.cpp
```diff
@@ -137,10 +155,8 @@ meanvar_test_gen(string name, int in_index, int weight_index, af_var_bias bias,
         readTests<double, typename varOutType<double>::type, double> (TEST_DIR"/meanvar/meanvar.data", numDims_, in_, tests_);
 
         inputs.resize(in_.size());
-        for(size_t i = 0; i < in_.size(); i++) {
-          af_create_array(&inputs[i], &in_[i].front(),
-                          numDims_[i].ndims(), numDims_[i].get(), f64);
-        }
```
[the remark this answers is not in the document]
umar456: Its unfortunate. The tests seem to be fast enough so speed is not an issue and if i tried to improve it; the tests would get more complicated.

test/meanvar.cpp
9prady9: It makes sense with the for loop back now.

Issue #2085, opened 2018-03-13
Incorrect result of scan in combination with OpenCL


meister87:
Hi,
The scan operator should calculate the exclusive and inclusive prefix sum if  AF_BINARY_ADD is used, right?

I noticed that the scan does not always correctly compute the prefix sum.

Using the following example
 
            afcl::addDevice(device_id,device_context, command_queue);

            afcl::setDevice(device_id,device_context);

            std::ifstream inFile;
            inFile.open("prefix_sum_array_fire_test_input.txt");
            std::vector<long> numbers;
            long S;
            while (inFile >> S) {
                numbers.emplace_back(S);
            }

            goo::util::opencl::OpenCLMemInput device_numbers(device_context,
                                                             CL_MEM_COPY_HOST_PTR,
                                                             numbers.size()*sizeof(long),
                                                             numbers.data());

            goo::util::opencl::OpenCLMemInput device_result(device_context, numbers.size()*sizeof(long));

            af::sync();

            af::array input = afcl::array(numbers.size(),
                                          device_numbers.getInput(),
                                          s64,
                                          true);

            af::array device_prefix_sum = af::scan(input,
                                            0,
                                            AF_BINARY_ADD,
                                            false);

            af::sync();

            clEnqueueCopyBuffer(command_queue->getValue(),
                                *device_prefix_sum.device<cl_mem>(),
                                device_result.getInput(),
                                0,
                                0,
                                numbers.size() * sizeof(long),
                                0,
                                nullptr,
                                nullptr
            );

            std::vector<long> result(numbers.size());
            device_result.read(command_queue, result.data(), numbers.size()* sizeof(long),true);

            size_t host_prefix_sum = 0;
            for(size_t element_id = 0; element_id < numbers.size();++element_id){
                ASSERT_EQ(host_prefix_sum, result[element_id]);
                host_prefix_sum += numbers[element_id];
            }`
with the following input file: 
[prefix_sum_array_fire_test_input.txt](https://github.com/arrayfire/arrayfire/files/1808161/prefix_sum_array_fire_test_input.txt)

There are some indices where the prefix sum has some strange outliers.
For example, at the positions the 8191-8193 the following sequence is provided: 4427, 8856, 4429 instead of 4427, 4428, 4429 

This repeats multiple times within the given input at the power of two (but not always).

I tested it on two different GPUs(mobile and Tesla) with the same result.
The example still includes some wrapper function of my framework, but the procedure should be clear.

[...]

PR #2366, opened 2018-12-06
Fix output indices on broadcast step of exclusive scan

mark-poscablo:
Addresses part of #2085. ~~This fixes only the 1-dimensional exclusive scan on CUDA. I still have to apply this fix to multiple dimensions (batched case) and on OpenCL.~~ This fixes exclusive scan along any dimension, on both CUDA and OpenCL

The problem lies in the last step of GPU scan where the scanned total sums of each block's data are broadcast to each block's scan results. The broadcast is performed one element early in the exclusive scan case, which means that the last element of each block data gets the cumulative sum that was supposed to be added on the next block's data. To illustrate for blocks of 5:

```
input           : 1 1 1 1 1 | 1 1 1 1 1  |  1  1  1  1  1
blockwise scan  : 0 1 2 3 4 | 0 1 2 3 4  |  0  1  2  3  4
block sums      : 5         | 5          |  5            
block sums scan : 0         | 5          | 10            
broadcast add*  : 0 0 0 0 5 | 5 5 5 5 10 | 10 10 10 10 15
global scan     : 0 1 2 3 9^| 5 6 7 8 14^| 10 12 13 14 19^

*where the bug lies. note however that the addition of zeros on the first four
elements is not really performed in the code itself.

^where the effects of the bug are seen

```

The solution then is to shift the broadcast addition (see [this line](https://github.com/arrayfire/arrayfire/blob/1b792bc3c12b6014079646d99c4f733c998d98af/src/backend/cuda/kernel/scan_first.hpp#L152)) one element to the right for exclusive scan. This pull request implements exactly that.

~~For multiple dimensions, I think I have to apply a similar change to `scan_dim.hpp`.~~ Both `scan_first.hpp` and `scan_dim.hpp` on CUDA and OpenCL have been modified for this fix.

### DIFF
--- a/src/backend/cuda/kernel/scan_dim.hpp
+++ b/src/backend/cuda/kernel/scan_dim.hpp
@@ -128,7 +128,7 @@ namespace kernel
             }
     }
 
-    template<typename To, af_op_t op, int dim>
+    template<typename To, af_op_t op, int dim, bool inclusive_scan>
     __global__
     static void bcast_dim_kernel(Param<To> out,
                                  CParam<To> tmp,
@@ -159,9 +159,13 @@ namespace kernel
         const int blockIdx_dim = ids[dim];
 
         ids[dim] = ids[dim] * blockDim.y * lim + tidy;
-        optr  += ids[3] * out.strides[3] + ids[2] * out.strides[2] + ids[1] * out.strides[1] + ids[0];
+        optr += ids[3] * out.strides[3] + ids[2] * out.strides[2] + ids[1] * out.strides[1] + ids[0];
         const int id_dim = ids[dim];
         const int out_dim = out.dims[dim];
+
+        // Shift broadcast one step to the right for exclusive scan (#2366)
+        int offset = inclusive_scan ? 0 : out.strides[dim];
+        optr += offset;
 
         bool is_valid =
             (ids[0] < out.dims[0]) &&
@@ -224,7 +228,7 @@ namespace kernel
 
 
 
-    template<typename To, af_op_t op, int dim>
+    template<typename To, af_op_t op, int dim, bool inclusive_scan>
     static void bcast_dim_launcher(Param<To> out,
                                    CParam<To> tmp,
                                    const uint threads_y,
@@ -242,7 +246,7 @@ namespace kernel
 
         uint lim = divup(out.dims[dim], (threads_y * blocks_all[dim]));
 
-        CUDA_LAUNCH((bcast_dim_kernel<To, op, dim>), blocks, threads,
+        CUDA_LAUNCH((bcast_dim_kernel<To, op, dim, inclusive_scan>), blocks, threads,
             out, tmp, blocks_all[0], blocks_all[1], blocks_all[dim], lim);
 
         POST_LAUNCH_CHECK();
@@ -296,7 +300,7 @@ namespace kernel
             }
 
             blocks_all[dim] = bdim;
-            bcast_dim_launcher<To, op, dim>(out, tmp, threads_y, blocks_all);
+            bcast_dim_launcher<To, op, dim, inclusive_scan>(out, tmp, threads_y, blocks_all);
         }
     }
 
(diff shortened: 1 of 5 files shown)


test/scan.cpp
```diff
@@ -262,3 +267,18 @@ TEST(Accum, DocSnippet) {
     array gold_accumB_dim1(3, 3, h_gold_accumB_dim1);
     ASSERT_ARRAYS_EQ(gold_accumB_dim1, accumB_dim1);
 }
+
+TEST(Scan, ExclusiveSum1D) {
+    const int block_work_size = 6912;
```
umar456: Can you make this 20000? I understand why you used 6912 but its an implementation detail and it may change in the future.

mark-poscablo: Yes, and we want 20000 to be the size of the whole input array, right? Currently, the input size this number * 4.

src/backend/cuda/kernel/scan_first.hpp
```diff
@@ -113,45 +113,47 @@ namespace kernel
...
-        tptr += wid * tmp.strides[3] + zid * tmp.strides[2] + yid * tmp.strides[1];
+        optr +=
+            wid * out.strides[3] + zid * out.strides[2] + yid * out.strides[1];
+        tptr +=
+            wid * tmp.strides[3] + zid * tmp.strides[2] + yid * tmp.strides[1];
 
         Binary<To, op> binop;
         To accum = tptr[blockIdx_x - 1];
 
+
```
pavanky: Do all the other lines before this need to be changed ? It makes it harder to review when bug fixes are coupled with formatting changes.

mark-poscablo:
Just updated the title and description to reflect what has been done so far. The fix has been applied to scan along any dimension, on both CUDA and OpenCL.

src/backend/cuda/kernel/scan_dim.hpp
```diff
@@ -128,7 +128,7 @@ namespace kernel
             }
     }
 
-    template<typename To, af_op_t op, int dim>
+    template<typename To, af_op_t op, int dim, bool inclusive_scan>
```
umar456: I think this argument can be a parameter instead of a template argument. There isn't anything that is going on that would significantly change based on this parameter.

src/backend/cuda/kernel/scan_first.hpp
```diff
@@ -113,21 +113,19 @@ namespace kernel
...
+                                              uint blocks_y,
+                                              uint lim) {
         const int tidx = threadIdx.x;
         const int tidy = threadIdx.y;
 
         const int zid = blockIdx.x / blocks_x;
         const int wid = (blockIdx.y + blockIdx.z * gridDim.y) / blocks_y;
-        const int blockIdx_x = blockIdx.x - (blocks_x) * zid;
-        const int blockIdx_y = (blockIdx.y + blockIdx.z * gridDim.y) - (blocks_y) * wid;
+        const int blockIdx_x = blockIdx.x - (blocks_x)*zid;
```
umar456: Spaces are good.

src/backend/opencl/kernel/scan_first.cl
```diff
@@ -118,11 +118,12 @@ void bcast_first_kernel(__global To *oData, KParam oInfo,
 
             To accum = tData[groupId_x - 1];
 
+            int offset = !inclusive_scan;
             for (int k = 0, id = xid;
-                 k < lim && id < oInfo.dims[0];
+                 k < lim && id + offset < oInfo.dims[0];
```
umar456: You could just increment id when initializing the loop instead of every time its used.

src/backend/cuda/kernel/scan_first.hpp
```diff
@@ -145,13 +143,12 @@ namespace kernel
         Binary<To, op> binop;
         To accum = tptr[blockIdx_x - 1];
 
+        int offset = !inclusive_scan;
         for (int k = 0, id = xid;
-             k < lim && id < out.dims[0];
+             k < lim && id + offset < out.dims[0];
```
umar456: Increment id during initialization instead of incrementing it every time its used.

PR #2368, opened 2018-12-12
Update matmul tolerances. Minor tweaks to CUDA builds

umar456:
Minor updates to the build and updated tolerances to the batched matmul tests


9prady9:
Why the change in order of some files in particular ?

umar456:
Those files take a significant amount of time to compile. These files will start compiling earlier in the build process so the compile times will be shorter.

PR #2398, opened 2019-01-03
Apply suggestions from clang-tidy tool on array.cpp

umar456:
Clang-tidy is a tool for static analysis and other quality checks on software.
These changes are some of the suggestions from clang-tidy on the array.cpp file.
This commit does not add the clang-tidy config file because that is still a
work in progress.

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -21,27 +21,35 @@
 
 #include <memory>
 #include <stdexcept>
+#include <vector>
 
-namespace af {
-static int gforDim(af_index_t *indices) {
+using af::calcDim;
+using af::dim4;
+using std::copy;
+using std::logic_error;
+using std::vector;
+
+namespace {
+int gforDim(af_index_t *indices) {
     for (int i = 0; i < AF_MAX_DIMS; i++) {
-        if (indices[i].isBatch) return i;
+        if (indices[i].isBatch) { return i; }
     }
     return -1;
 }
 
-static af_array gforReorder(const af_array in, unsigned dim) {
+af_array gforReorder(const af_array in, unsigned dim) {
     // This is here to stop gcc from complaining
-    if (dim > 3) AF_THROW_ERR("GFor: Dimension is invalid", AF_ERR_SIZE);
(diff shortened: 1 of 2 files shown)


src/api/cpp/array.cpp
```diff
@@ -100,93 +135,71 @@ struct array::array_proxy::array_proxy_impl {
...
+        : arr(0) {                                                             \
+        af::dtype ty = static_cast<af::dtype>(dtype_traits<T>::af_type);       \
+        initDataArray(&arr, ptr, ty, src, dim0, dim1, dim2);                   \
+    }                                                                          \
+    template<>                                                                 \
+    AFAPI array::array(dim_t dim0, dim_t dim1, dim_t dim2, dim_t dim3,         \
+                       const T *ptr, af::source src)                           \
+        : arr(0) {                                                             \
+        af::dtype ty = static_cast<af::dtype>(dtype_traits<T>::af_type);       \
+        initDataArray(&arr, ptr, ty, src, dim0, dim1, dim2, dim3);             \
```
9prady9: `arr(0)` -> `arr(nullptr)` in all above template instantiations.

src/api/cpp/array.cpp
```diff
@@ -100,93 +135,71 @@ struct array::array_proxy::array_proxy_impl {
...
 
-   private:
-    array_proxy_impl(const array_proxy_impl &);
-    array_proxy_impl(const array_proxy_impl &&);
-    array_proxy_impl operator=(const array_proxy_impl &);
-    array_proxy_impl operator=(const array_proxy_impl &&);
+    array_proxy_impl(const array_proxy_impl &)  = delete;
+    array_proxy_impl(const array_proxy_impl &&) = delete;
+    array_proxy_impl operator=(const array_proxy_impl &) = delete;
+    array_proxy_impl operator=(const array_proxy_impl &&) = delete;
```
9prady9: We should keep refactoring & style changes separate from the change like this. Perhaps a separate commit where the array_proxy_impl copy and move constructors are deleted.

umar456: This is equivalent to the previous code. Previously we made these constructors private which is the same as deleting them. Here we are using the `= delete` syntax.

umar456:
Addressed all feedback

PR #2399, opened 2019-01-03
Correct constness on createDeviceArray and Array<T> constructor

umar456:
The createDeviceArray pointer was const when it shouldn't be. This is
similiar to correcting the const in the api function but applies it
to the internal functions.

### DIFF
--- a/src/backend/common/SparseArray.cpp
+++ b/src/backend/common/SparseArray.cpp
@@ -48,11 +48,10 @@ SparseArrayBase::SparseArrayBase(af::dim4 _dims, dim_t _nNZ,
 #endif
 }
 
-SparseArrayBase::SparseArrayBase(af::dim4 _dims, dim_t _nNZ,
-                                 const int *const _rowIdx,
-                                 const int *const _colIdx,
-                                 const af::storage _storage, af_dtype _type,
-                                 bool _is_device, bool _copy_device)
+SparseArrayBase::SparseArrayBase(af::dim4 _dims, dim_t _nNZ, int *const _rowIdx,
+                                 int *const _colIdx, const af::storage _storage,
+                                 af_dtype _type, bool _is_device,
+                                 bool _copy_device)
     : info(getActiveDeviceId(), _dims, 0, calcStrides(_dims), _type, true)
     , stype(_storage)
     , rowIdx(_is_device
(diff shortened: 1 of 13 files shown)


src/backend/common/SparseArray.cpp
```diff
@@ -127,15 +126,18 @@ SparseArray<T> createHostDataSparseArray(const af::dim4 &_dims, const dim_t nNZ,
                                          const int *const _rowIdx,
                                          const int *const _colIdx,
                                          const af::storage _storage) {
-    return SparseArray<T>(_dims, nNZ, _values, _rowIdx, _colIdx, _storage,
-                          false);
+    return SparseArray<T>(_dims, nNZ, const_cast<T *>(_values),
+                          const_cast<int *>(_rowIdx),
+                          const_cast<int *>(_colIdx), _storage, false);
```
9prady9: What's the difference between earlier and now in this constructor ?

src/backend/cpu/Array.hpp
```diff
@@ -46,22 +46,24 @@ void evalMultiple(std::vector<Array<T> *> arrays);
...
+// Creates an array and copies from the \p data pointer located in host memory
+//
+// \param[in] dims The dimension of the array
+// \param[in] data The data that will be copied to the array
 template<typename T>
-Array<T> createHostDataArray(const af::dim4 &size, const T *const data);
+Array<T> createHostDataArray(const af::dim4 &dims, const T *const data);
 
 template<typename T>
-Array<T> createDeviceDataArray(const af::dim4 &size, const void *data);
```
9prady9: why is the `const` gone ?

src/backend/cuda/Array.cpp
```diff
@@ -303,23 +303,27 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
...
+    bool copy_device = false;
+    return Array<T>(dims, data, is_device, copy_device);
 }
 
 template<typename T>
-Array<T> createDeviceDataArray(const dim4 &size, const void *data) {
-    return Array<T>(size, (const T *const)data, true);
+Array<T> createDeviceDataArray(const dim4 &dims, void *data) {
+    bool is_device   = true;
+    bool copy_device = false;
```
9prady9: Nit: `is_device` and `copy_device` can be const too.

src/backend/common/SparseArray.cpp
umar456: The older SparseArray was const. this was a problem because we accept device functions and those can be modified by the internals. Moving them to non-const SparseArray requires this cast.

src/backend/cpu/Array.hpp
umar456: The DeviceDataArray uses the pointer from the data array. It may be modified in the future  by another operation so this is not a const parameter.

Issue #2310, opened 2018-10-10
Add env var to save generated kernels src to tmp dir


WilliamTambellini:
For easy debugging/profiling, create a new env var option in order to save the kernels sources to the tmp dir for example in getKernel() : 

```c++
string jit_ker = getKernelString(funcName, full_nodes, full_ids, output_ids, is_linear);
entry = compileKernel(funcName.c_str(), jit_ker);
kernelCaches[device][funcName] = entry;
if (getenv("AF_SAVE_KERNELS") {
    // Todo: check/create path /tmp/af/ker
    FILE* f = fopen(("/tmp/af/ker/"+funcName+".txt").c_str(), "w");
    if (f) { fputs(jit_ker.c_str(), f); fclose(f); }
}
```

WilliamTambellini:
@umar456  What about the user to set the env var
"AF_SAVE_JIT_KERNELS_DIR" to the desired folder for AF to write jit kernels into ?
if unset, then af not to write any kernels sources to disk.

PR #2404, opened 2019-01-07
Add env var to save generated kernels src to tmp dir

WilliamTambellini:
Add an env var option to save generated JIT kernels src to a folder.
https://github.com/arrayfire/arrayfire/issues/2310

### DIFF
--- a/src/backend/common/util.hpp
+++ b/src/backend/common/util.hpp
@@ -1,5 +1,5 @@
 /*******************************************************
- * Copyright (c) 2016, ArrayFire
+ * Copyright (c) 2019, ArrayFire
  * All rights reserved.
  *
  * This file is distributed under 3-clause BSD license.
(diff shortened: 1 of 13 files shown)


src/backend/common/util.hpp
```diff
@@ -15,3 +15,7 @@
 #pragma once
 
 std::string getEnvVar(const std::string &key);
+
+// Dump the kernel sources only if the environment variable is defined
+static const char* saveJitKernelsEnvVarName = "AF_DUMP_JIT_KERNELS_DIR";
```
umar456: Can you rename this to `AF_JIT_KERNEL_TRACE`?

It accepts `1` and writes it to stdout or "directory:..." and it will create a file for each kernel. In the future I will be creating a graphviz images so I want this to be slightly more flexible.

WilliamTambellini: Renamed. Now accepts stdout/stderr.

WilliamTambellini: I prefer to let you officially publish that option in the doc when the specs will be totally stable.

WilliamTambellini:
Tks Umar. @pavanky  and others : please review : it s today not possible to inspect/look at jit kernel sources without that change.

umar456:
Only one review is required. The branch has a couple of conflicts that I am going to address in a bit.

PR #2407, opened 2019-01-09
clang format compatability

syurkevi:
updates .clang-format file to be compatible with older versions of clang-format


src/.clang-format
```diff
@@ -141,4 +91,4 @@ SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        4
 UseTab:          Never
-
+...
```
mark-poscablo: I think this is a stray ellipsis? It gave me an error on clang 6 anyhow.

9prady9:
Also, this PR needs to be rebased of latest master to resolve conflicts

umar456:
I dont want to support older clang-formats. I will document the right development environment in another PR.

PR #2413, opened 2019-01-18
Initial f16 support

umar456:
Initial support for fp16:

- [x] JIT
- [x] Random Number Generation
- [x] matmul
- [x] reductions
- [x] reorder/transpose
- [x] printing
- [x] casts
- [x] tests
- ~[ ] convolution~
- ~[ ] anything else~


WilliamTambellini:
why is there 
"
Initial support for fp16:
CPU
"
in the description ?

umar456:
I was thinking of making a table but thought against it. Fixed the message.

CMakeModules/InternalUtils.cmake
```diff
@@ -104,7 +104,7 @@ macro(arrayfire_set_cmake_default_variables)
   set(CMAKE_PREFIX_PATH "${ArrayFire_BINARY_DIR};${CMAKE_PREFIX_PATH}")
   set(BUILD_SHARED_LIBS ON)
 
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 14)
```
9prady9: 👍

src/backend/cuda/blas.cpp
```diff
@@ -229,7 +236,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
                 getType<T>(), lStrides[1], rhs.get(), getType<T>(), rStrides[1],
                 &beta, out.get(), getType<T>(), out.strides()[1],
                 getType<T>(),  // Compute type
-                CUBLAS_GEMM_DEFAULT_TENSOR_OP));
+                selectGEMMAlgorithm()));
```
9prady9: This change should actually be a separate commit by itself when we merge this PR. Leaving a comment here so that we don't forget to reorganize the commits.

src/backend/cuda/CMakeLists.txt
```diff
@@ -53,7 +53,9 @@ cuda_include_directories(
 set(jit_kernel_headers
     "kernel_headers")
 
-file(GLOB jit_src "kernel/jit.cuh")
+set(jit_src "${CMAKE_CURRENT_SOURCE_DIR}/kernel/jit.cuh;${CUDA_INCLUDE_DIRS}/cuda_fp16.h;${CUDA_INCLUDE_DIRS}/cuda_fp16.hpp")
+
+message(STATUS ${jit_src})
```
9prady9: Debug message

src/backend/cuda/kernel/memcopy.hpp
```diff
@@ -86,11 +87,10 @@ void memcopy(T *out, const dim_t *ostrides, const T *in, const dim_t *idims,
     POST_LAUNCH_CHECK();
 }
 
-////////////////////////////// BEGIN - templated help functions for copy_kernel
-///////////////////////////////////
+///////////// BEGIN - templated help functions for copy_kernel /////////////////
```
9prady9: I thought we are removing these block comments whenever we are editing surrounding code.

src/backend/opencl/err_opencl.hpp
```diff
@@ -22,10 +23,13 @@
...
 template<typename T>
-void verifyDoubleSupport() {
+void verifyTypeSupport() {
     if ((std::is_same<T, double>::value || std::is_same<T, cdouble>::value) &&
         !isDoubleSupported(getActiveDeviceId())) {
         AF_ERROR("Double precision not supported", AF_ERR_NO_DBL);
+    } else if (std::is_same<T, common::half>::value &&
+               !isHalfSupported(getActiveDeviceId())) {
+        AF_ERROR("Half precision not supported", AF_ERR_NO_HALF);
     }
```
9prady9: My comment on #2415 , https://github.com/arrayfire/arrayfire/pull/2415#discussion_r249435163 , is more of concern  especially in OpenCL backend, don't you think so ? We can't assume if double support is there, there is half support.

Hm, is this okay for now because we only ammulate half computation using float for actual computation ?

src/backend/common/half.hpp
```diff
@@ -77,7 +76,7 @@ CONSTEXPR_DH uint16_t float2half(float value) {
     uint32_t bits = 0;  // = *reinterpret_cast<uint32*>(&value);
                         // //violating strict aliasing!
     std::memcpy(&bits, &value, sizeof(float));
-    constexpr uint16_t base_table[512] = {
+    uint16_t base_table[512] = {
```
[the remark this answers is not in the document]
umar456: Hmm. I should make that a static const array. The reason I removed constexpr is because of the memcpy later in the function. The memcpy is not a constexpr operation so the constexpr decorator would have been deceptive here. Also one of the compilers threw an error.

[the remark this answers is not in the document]
9prady9: Got it. Then `static const` should do the trick I guess.

src/backend/cuda/blas.cpp
umar456: We aren't using that function before this change. Additionally that function is not compatible with Kepler. Need to find another way to call it.

src/backend/cuda/CMakeLists.txt
umar456: Yes. I will need to remove it

src/api/c/blas.cpp
```diff
@@ -157,6 +162,11 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
             case c32: output = matmul<cfloat>(lhs, rhs, optLhs, optRhs); break;
             case f64: output = matmul<double>(lhs, rhs, optLhs, optRhs); break;
             case c64: output = matmul<cdouble>(lhs, rhs, optLhs, optRhs); break;
+#ifndef AF_CPU
+            case f16:
+                output = matmul<common::half>(lhs, rhs, optLhs, optRhs);
+                break;
+#endif
```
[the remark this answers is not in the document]
umar456: You get a type error. The only way to support this is by creating our own gemm implementation.

src/backend/cuda/blas.cpp
9prady9: Nevertheless, it is not fp16 related change. Hence, my comment.

src/backend/opencl/err_opencl.hpp
9prady9: Nevermind, I missed the type match validation. My bad.

include/af/half.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+typedef struct {
+    unsigned short data_:16;
+    struct {
+        unsigned short fraction:10;
+        unsigned short exponent:5;
+        unsigned short sign:1;
+    };
+} af_half;
```
pavanky: Where is this from ?

CMakeModules/InternalUtils.cmake
```diff
@@ -30,7 +30,7 @@ endfunction()
 
 function(arrayfire_get_cuda_cxx_flags cuda_flags)
   if(NOT MSVC)
-    set(flags "-std=c++11 -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
+    set(flags "-std=c++14 -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
```
WilliamTambellini: are you such a change/move should nt be split into a sub separate PullRequest "upgrade to std c++14" ?

include/af/device.h
```diff
@@ -75,10 +75,21 @@ namespace af
...
 
+    /// \brief Queries the current device for half precision floating point
+    ///        support
+    ///
+    /// \param[in] device the ID of the device to query
+    ///
+    /// \returns true if the \p device supports half precision operations.
+    ///          false otherwise
+    /// \ingroup device_func_half
+    AFAPI bool isHalfAvailable(const int device);
```
WilliamTambellini: Could you add more comments in order to give some examples of which device will return true which false ?

src/backend/cuda/platform.cpp
```diff
@@ -203,6 +203,12 @@ bool isDoubleSupported(int device) {
     return true;
 }
 
+bool isHalfSupported(int device) {
+    auto prop = getDeviceProp(device);
+    float compute = prop.major * 1000 + prop.minor * 10;
+    return compute >= 5030;
```
WilliamTambellini: hum does it mean that any device setup with CudaComp 5.3+ will really support half ?

src/backend/cuda/blas.cpp
```diff
@@ -179,21 +287,25 @@ void gemm(Array<T> &out, af_mat_prop optLhs, af_mat_prop optRhs,
     if (oDims.ndims() <= 2) {
         if (rDims[bColDim] == 1) {
             dim_t incr = (optRhs == AF_MAT_NONE) ? rStrides[0] : rStrides[1];
+            if (is_same<T, half>::value) {
+                AF_ERROR(
+                    "GEMV does not support half, Please create an issue on "
```
WilliamTambellini: is it impossible to gemv f16 arrays or just to be implemented ?

src/backend/cuda/CMakeLists.txt
```diff
@@ -450,7 +456,7 @@ cuda_add_library(afcuda
   )
 
 arrayfire_set_default_cxx_flags(afcuda)
-target_compile_definitions(afcuda PRIVATE AF_CUDA)
+target_compile_definitions(afcuda PRIVATE AF_CUDA CUDA_NO_HALF)
```
WilliamTambellini: Could you please add a comment what CUDA_NO_HALF does ?

CMakeModules/InternalUtils.cmake
umar456: Yeah, Thats a good idea. I can do that.

src/backend/cuda/platform.cpp
umar456: Yes, only those devices have support for half operations. I can try to incorporate older devices but that will take more time.

src/backend/cuda/blas.cpp
umar456: It will have to be implemented internally. Its not something that is part of cuBLAS

src/backend/cpu/cast.hpp
```diff
@@ -26,6 +26,112 @@ struct UnOp<To, Ti, af_cast_t> {
...
+        for (int i = 0; i < lim; i++) {
+            float val = in[i];
+            out[i]    = To(val);
+        }
+    }
+
+    void eval(jit::array<To> &out, const jit::array<float> &in, int lim) {
+        for (int i = 0; i < lim; i++) { out[i] = To(in[i]); }
+    }
+};
```
9prady9: The specialization for `std::complex<float>` output type for `struct UnOp<To, common::half, af_cast_t>` looks exactly same in code to the general implementation. Doesn't look like we need it from `half` to `To`.

src/backend/cpu/cast.hpp
```diff
@@ -26,6 +26,112 @@ struct UnOp<To, Ti, af_cast_t> {
...
```
9prady9: This specialization also looks identical to the general implementation of `struct UnOp<To, common::half, af_cast_t>`.

src/backend/cpu/kernel/transpose.hpp
```diff
@@ -12,6 +12,8 @@
 #include <err_cpu.hpp>
 #include <utility.hpp>
 
+using af::dim4;
```
9prady9: I thought we stopped adding `using` statements in headers ?

src/backend/cuda/blas.cpp
```diff
@@ -150,14 +156,116 @@ BLAS_FUNC(dot, cdouble, false, Z, u)
...
+    } else {
+        using Nt = typename common::kernel_type<T>::native;
+        return gemm_func<Nt>()(blasHandle(), lOpts, rOpts, M, N, K, (Nt *)alpha,
+                               (Nt *)lhs.get(), lStride, (Nt *)rhs.get(),
+                               rStride, (Nt *)beta, (Nt *)out.get(), oleading);
+    }
+}
+
+template<typename T>
+cublasStatus_t gemmBatched_dispatch(BlasHandle handle,
```
9prady9: I don't like this mixed naming convention. `gemmBatched_dispatch` should be `gemmBatchedDispatch` and the other one should be `gemmDispatch`.

src/backend/opencl/kernel/transpose.cl
```diff
@@ -15,6 +15,12 @@ T doOp(T in) {
 #define doOp(in) in
 #endif
 
+#pragma OPENCL EXTENSION all : enable
+#ifdef cl_khr_fp16
+#else
+#define half short
+#endif
```
9prady9: Interesting, why is it that we need these `#if`s only in `jit.cl` and `transpose.cl` ?

test/array.cpp
```diff
@@ -335,6 +335,16 @@ TYPED_TEST(Array, TypeAttributes) {
...
             break;
+        case f16:
+            EXPECT_TRUE(one.isfloating());
+            EXPECT_FALSE(one.isdouble());
+            EXPECT_FALSE(one.issingle());
+            EXPECT_TRUE(one.isrealfloating());
+            EXPECT_FALSE(one.isinteger());
+            EXPECT_TRUE(one.isreal());
+            EXPECT_FALSE(one.iscomplex());
+            EXPECT_FALSE(one.isbool());
```
9prady9: don't we have `isHalf` now ?

test/join.cpp
```diff
@@ -44,7 +44,7 @@ class Join : public ::testing::Test {
 
 // create a list of types to be tested
 typedef ::testing::Types<float, double, cfloat, cdouble, int, unsigned int,
-                         intl, uintl, char, unsigned char, short, ushort>
+                         intl, uintl, char, unsigned char, short, ushort, af_half>
```
9prady9: we should be C++ typedef `af::half` to be consistent with `cfloat` and `cboudle`. This applies to all the tests where `half` is enabled.

test/testHelpers.hpp
```diff
@@ -537,6 +576,10 @@ const af::cfloat &operator+(const af::cfloat &val) { return val; }
 
 const af::cdouble &operator+(const af::cdouble &val) { return val; }
 
+const af_half& operator+(const af_half& val) {
```
9prady9: `af::half` to be consistent with other `af::*` types.

test/testHelpers.hpp
```diff
@@ -34,6 +35,16 @@
 #include <mmio.h>
 #endif
 
+bool operator==(const af_half &lhs, const af_half &rhs) {
+    return lhs.data_ == rhs.data_;
+}
+
+std::ostream &operator<<(std::ostream &os, const af_half &val) {
```
9prady9: Do we use `half_float::half` from `extern/half/include/half.hpp` only in tests ?

src/backend/opencl/kernel/transpose.cl
umar456: We will probably need it in other places but those functions haven't been implemented yet.

test/testHelpers.hpp
umar456: Yes. Our half type is not fully functional. It would require us to reimplement the entire half library inorder to correctly implement it. I don't think we should directly include other libraries in our API. perhaps we can discuss it in a separate issue.

test/join.cpp
umar456: I am worried we will have conflicts with the half data type. I want to discourage the use of this type by users because of its inefficiency. I am not sure if this is the best way to handle this sort of thing.

include/af/device.h
WilliamTambellini: On my side I m more interested to know if the device can handle half natively with better perf than f32 (most of the time meaning Volta/Turing via TensorCores and apparently P100). 
Is nt that method going to return true on old GPUs (GTX10X0) where half is perhaps "supported" but perf is disastrous ?

umar456: The half support for the 10x0 cards are supported through floats.

Even on the Volta cards we need to improve the way we handle things to improve the compute performance. In order to fully utilize the GPU's half compute performance we need to use half2. This will require a major shift on some of the internal kernels so I pushed that change off to later. We can talk about what functions are important to you and improve those later down the line. All the library calls for half(i.e. matmul) will be using half2 so you will get full compute performance there. If there are kernels that are bandwidth bound moving to half2 will serve no purpose because it will still be a bandwidth bound kernel.

test/array.cpp
```diff
@@ -335,6 +335,17 @@ TYPED_TEST(Array, TypeAttributes) {
...
+            EXPECT_FALSE(one.ishalf());
```
syurkevi: EXPECT_TRUE?

PR #2415, opened 2019-01-20
Use SUPPORTED_TYPE_CHECK macro to test for supported type

umar456:
This is a more general macro which will check if the device surpports
the types being checked. This is done to support double types but
will be required for fp16 in the future

### DIFF
--- a/test/testHelpers.hpp
+++ b/test/testHelpers.hpp
@@ -439,6 +439,9 @@ bool noDoubleTests() {
     return ((isTypeDouble && !isDoubleSupported) ? true : false);
 }
 
+#define SUPPORTED_TYPE_CHECK(type) \
+    if (noDoubleTests<type>()) return;
+
 inline bool noImageIOTests() {
     bool ret = !af::isImageIOAvailable();
     if (ret) printf("Image IO Not Configured. Test will exit\n");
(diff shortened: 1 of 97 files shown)


test/testHelpers.hpp
```diff
@@ -439,6 +439,9 @@ bool noDoubleTests() {
     return ((isTypeDouble && !isDoubleSupported) ? true : false);
 }
 
+#define SUPPORTED_TYPE_CHECK(type) \
+    if (noDoubleTests<type>()) return;
+
```
9prady9: I see that we have the following
```cpp
#define SUPPORTED_TYPE_CHECK(type) \
  if (noDoubleTests<type>()) return; \
  if (noHalfTests<type>()) return
```
in PR #2413 , this will fail when the device has has no double support but half support - as in we will get false positives in test results, Unless I am missed seeing something in the changes.

I don't think reverse order will also work, because there could be devices with double support but no half support.

9prady9: Hm, is this okay for now because we only ammulate half computation using float for actual computation ?

9prady9: Nevermind, I missed the type match validation. My bad.

PR #2418, opened 2019-01-22
Added a basic feature summary to Cmake output

mark-poscablo:
Displays features (i.e. build options that were enabled) and packages found. The chosen packages that are  displayed are based on the required dependencies per backend as listed in the arrayfire build instructions.

Here's an example, as a snippet of the larger Cmake output:

```
-- Enabled Features: 
 * CPU Backend, for building the ArrayFire library that targets the CPU backend.
 * CUDA Backend, for building the ArrayFire library that targets the CUDA backend.
 * OpenCL Backend, for building the ArrayFire library that targets the OpenCL backend.
 * Unified Backend, for building a wrapper library which allows you to target all backends.
 * Graphics, which allows you to use the plotting capabilities of ArrayFire
 * ImageIO, which allows you to read and write images
 * Logging, which generates logs for debugging purposes

-- Required packages for CPU backend found: 
 * FFTW
 * CBLAS
 * LAPACKE

-- Required packages for CUDA backend found: 
 * CUDA (required version >= 7.0)

-- Required packages for OpenCL backend found: 
 * Boost
 * OpenCL

-- Required packages for graphics found: 
 * glfw3
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ include(build_cl2hpp)
 include(platform)
 include(GetPrerequisites)
 include(CheckCXXCompilerFlag)
+include(FeatureSummary)
 
 arrayfire_set_cmake_default_variables()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -334,3 +335,105 @@ conditional_directory(AF_BUILD_EXAMPLES examples)
 conditional_directory(AF_BUILD_DOCS docs)
 
 include(CPackConfig)
+
+set_property(GLOBAL APPEND PROPERTY FeatureSummary_PKG_TYPES CPU)
```
9prady9: I am not sure how `FeatureSummary_PKG_TYPES` affects the summary outlook, but I believe this is introduced in 3.8 CMake and that would bump up and minimum requirement to 3.8 from 3.5. This may not fail on CI systems as we don't have 3.5 cmake anywhere at the moment.

9prady9: Either that or CMake devs forget to document this variable in 3.5 docs :)

mark-poscablo: Yes, I just looked at the docs too and it looks like it was introduced in 3.8 instead. As umar suggested, I could disable the feature summary for older versions of cmake. For now though I'm going to close this PR to work on more urgent tasks, and I'll get back to this in the future. Thanks for reviewing!

PR #2420, opened 2019-01-22
Assert the nvidia drivers are recent enough for cuda

WilliamTambellini:
Assert the nvidia drivers are recent enough for cuda


9prady9:
I think the commit history is distorted. Please make sure to checkout the new branch from ArrayFire `remote`(git remote)

PR #2421, opened 2019-01-22
Assert the nvidia drivers are recent enough for cuda

WilliamTambellini:
Assert the nvidia drivers are recent enough for cuda

### DIFF
--- a/src/backend/cuda/platform.cpp
+++ b/src/backend/cuda/platform.cpp
@@ -11,12 +11,14 @@
 #include <windows.h>
 #endif
 
+#include <common/Logger.hpp>
 #include <common/defines.hpp>
 #include <common/host_memory.hpp>
 #include <common/util.hpp>
 #include <driver.h>
 #include <err_cuda.hpp>
 #include <platform.hpp>
+#include <spdlog/spdlog.h>
 #include <version.hpp>
 #include <af/cuda.h>
 #include <af/version.h>
@@ -36,8 +38,10 @@
 #include <vector>
 
 using namespace std;
+using std::to_string;
 
 namespace cuda {
+
 ///////////////////////////////////////////////////////////////////////////
 // HELPERS
 ///////////////////////////////////////////////////////////////////////////
@@ -129,13 +133,6 @@ static const std::string get_system(void) {
 #endif
 }
 
-template<typename T>
-static inline string toString(T val) {
-    stringstream s;
-    s << val;
-    return s.str();
-}
-
 static inline int getMinSupportedCompute(int cudaMajorVer) {
     // Vector of minimum supported compute versions
     // for CUDA toolkit (i+1).* where i is the index
@@ -160,14 +157,14 @@ string getDeviceInfo(int device) {
 
     bool show_braces = getActiveDeviceId() == device;
 
-    string id = (show_braces ? string("[") : "-") + toString(device) +
+    string id = (show_braces ? string("[") : "-") + to_string(device) +
                 (show_braces ? string("]") : "-");
     string name(dev.name);
-    string memory = toString((mem_gpu_total / (1024 * 1024)) +
-                             !!(mem_gpu_total % (1024 * 1024))) +
+    string memory = to_string((mem_gpu_total / (1024 * 1024)) +
+                              !!(mem_gpu_total % (1024 * 1024))) +
                     string(" MB");
-    string compute = string("CUDA Compute ") + toString(dev.major) +
-                     string(".") + toString(dev.minor);
+    string compute = string("CUDA Compute ") + to_string(dev.major) +
+                     string(".") + to_string(dev.minor);
 
     string info = id + string(" ") + name + string(", ") + memory +
                   string(", ") + compute + string("\n");
@@ -243,19 +240,21 @@ string getDriverVersion() {
 #endif
         int driver = 0;
         CUDA_CHECK(cudaDriverGetVersion(&driver));
-        return string("CUDA Driver Version: ") + toString(driver);
+        return string("CUDA Driver Version: ") + to_string(driver);
     } else {
         return string(driverVersion);
     }
 }
 
+string int_version_to_string(int version) {
+    return to_string(version / 1000) + "." +
+            to_string((int)((version % 1000) / 100.));
+}
+
 string getCUDARuntimeVersion() {
     int runtime = 0;
     CUDA_CHECK(cudaRuntimeGetVersion(&runtime));
-    if (runtime / 100.f > 0)
-        return toString((runtime / 1000) + (runtime % 1000) / 100.);
-    else
-        return toString(runtime / 1000) + string(".0");
+    return int_version_to_string(runtime);
 }
 
 unsigned getMaxJitSize() {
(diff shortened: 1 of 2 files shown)


src/backend/cuda/platform.cpp
```diff
@@ -37,7 +39,18 @@
 
 using namespace std;
 
+
 namespace cuda {
+
+// check with reviewers : perhaps to be moved to a static of DeviceManager
+std::shared_ptr<spdlog::logger> logger = common::loggerFactory("cuda");
```
9prady9: Your comment is right. This logger should be defined as member of CUDA backend's device manger class.

You can refer to `src/backend/common/MemoryManager.hpp` and `src/backend/common/MemoryManagerImpl.hpp` to see how the `mem` logger is tied to the memory manager's context.

Similarly, we should keep CUDA backend logger tied to the CUDA device manager instead of having a global variable.

src/backend/cuda/platform.cpp
```diff
@@ -37,7 +39,18 @@
...
+
+#if defined(AF_WITH_LOGGING)
+spdlog::logger* getLogger() {
+  return logger.get();
+}
+#endif
```
9prady9: `getLogger` should be member function of deviceManager.

You can refer to `src/backend/common/MemoryManager.hpp` and `src/backend/common/MemoryManagerImpl.hpp` to see how it is done.

src/backend/cuda/platform.cpp
9prady9: Forward declaring `spdlog::logger` inside `src/backend/cuda/platform.hpp` is sufficient to declare a shared_ptr of logger inside DeviceManager class. This forward declaration is actually already present in `src/backend/cuda/platform.hpp`, so you can just move the declaration to class interior and initialization to the class's constructor.

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +482,59 @@ SparseHandle sparseHandle() {
...
+    #endif
+    AF_TRACE("Minimum required driver version: {}", minimumDriverVersion);
+    if (driverVersion < minimumDriverVersion)
+	throw runtime_error("The driver " + driverVersionString + " are too old in order to run this verison of CUDA: " + cudaRuntimeVersionString 
+	+ ". Either upgrade the nvidia drivers on that machine (at least verison " + std::to_string(minimumDriverVersion) 
+	+ ") or use a build of ArrayFire linking with an older version of CUDA");
+}
+
 DeviceManager::DeviceManager()
     : cuDevices(0), nDevices(0), fgMngr(new graphics::ForgeManager()) {
```
9prady9: Hint on where I was referring to initialize logger
```
cuDevices(0), nDevices(0), fgMngr(new graphics::ForgeManager()), logger(common::loggerFactory("cuda"))
```

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +482,59 @@ SparseHandle sparseHandle() {
     return cusparseHandles[id].get()->get();
 }
 
+/*
+ * Map giving the minimum device driver needed in order to run a given version of CUDA
+ * https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html
+ * Map linking cuda version to the minimum driver versions for both Linux/Mac and Windows
+ */
+static const std::map<float, std::pair<double, double> > CudaToKernelVersionMap = {
```
9prady9: `static const` -> `constexpr`

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +482,59 @@ SparseHandle sparseHandle() {
...
+  { 10.0, { 410.48, 411.31 } },
+  { 9.2, { 396.37, 398.26 } },
+  { 9.1, { 390.46, 391.29 } },
+  { 9.0, { 384.81, 385.54 } },
+  { 8.0, { 375.26, 376.51 } },
+  { 7.5, { 352.31, 353.66 } }
```
9prady9: Nit: If doubles are not necessary, lets move to floats to store this info. Reduces the size of this map.

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +482,59 @@ SparseHandle sparseHandle() {
...
+    if (driverVersion < minimumDriverVersion)
+	throw runtime_error("The driver " + driverVersionString + " are too old in order to run this verison of CUDA: " + cudaRuntimeVersionString 
+	+ ". Either upgrade the nvidia drivers on that machine (at least verison " + std::to_string(minimumDriverVersion) 
+	+ ") or use a build of ArrayFire linking with an older version of CUDA");
+}
+
 DeviceManager::DeviceManager()
     : cuDevices(0), nDevices(0), fgMngr(new graphics::ForgeManager()) {
+
+    checkCudaVsDriverVersion();
```
9prady9: This can be made as private member of DeviceManager class after logger is made member variable of DeviceManager.

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +482,59 @@ SparseHandle sparseHandle() {
...
+  { 9.2, { 396.37, 398.26 } },
+  { 9.1, { 390.46, 391.29 } },
+  { 9.0, { 384.81, 385.54 } },
+  { 8.0, { 375.26, 376.51 } },
+  { 7.5, { 352.31, 353.66 } }
+};
+
+// check with reviewers: perhaps mutate this as a static of DeviceManager:
+// Check if the device driver version is recent enough to run the cuda libs linked with afcuda:
+void checkCudaVsDriverVersion() {
```
9prady9: Please add the `const` qualifier to this function as it is not modifying any state and also generating error using gcc.

src/backend/cuda/platform.cpp
9prady9: If you are going to keep doubles, then add `f` literal at the end of key floats of map so that it doesn't throw error on some compilers. (Windows CUDA job failed due to this error).

src/backend/cuda/platform.cpp
WilliamTambellini: Not fan as it would prevent to use the logger without the DeviceManager but anyway, done.

src/backend/cuda/platform.cpp
WilliamTambellini: done

src/backend/cuda/platform.cpp
WilliamTambellini: does nt compile under gcc 5

src/backend/cuda/platform.cpp
WilliamTambellini: converted to floats

src/backend/cuda/platform.cpp
WilliamTambellini: done

src/backend/cuda/platform.cpp
WilliamTambellini: done

src/backend/cuda/platform.cpp
WilliamTambellini: done

src/backend/cuda/platform.cpp
9prady9: The idea is to not have a single global logger with all the message clubbed into one group but instead have loggers tied to specific contexts like we have `mem` & `unified` for memory-manager and unified-api, respectively.

src/backend/cuda/platform.cpp
9prady9: sure ? because we have `constexpr` in various other locations in the code, in fact in CUDA backend itself. Check this one out.
https://github.com/arrayfire/arrayfire/blob/master/src/backend/cuda/Array.cpp

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +473,58 @@ SparseHandle sparseHandle() {
...
+    const float driverVersion = std::stof(driverVersionString);
+    if (driverVersion == 0)
+        throw runtime_error("Failed to parse driver version: " + driverVersionString);
+    const std::string cudaRuntimeVersionString = getCUDARuntimeVersion();
+    if (cudaRuntimeVersionString.empty())
+        throw runtime_error("Failed to get CUDA version (empty)");
+
+    AF_TRACE("AF built with cuda {}", cudaRuntimeVersionString);
+    if (CudaToKernelVersionMap.find(cudaRuntimeVersionString) == CudaToKernelVersionMap.end())
+        throw runtime_error("Failed to recognize this version of CUDA (" + cudaRuntimeVersionString + ") : it is either too old or too recent for this version of ArrayFire");
```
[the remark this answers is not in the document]
WilliamTambellini: I ve converted the throw to AF_ERROR.
Regarding moving the method, we agreed on the specs 2 weeks ago, so please move it later at your convenience.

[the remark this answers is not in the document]
WilliamTambellini: And no more error if cuda version unknown/not registered.

src/backend/cuda/platform.cpp
```diff
@@ -469,8 +473,58 @@ SparseHandle sparseHandle() {
...
+        throw runtime_error("Failed to retrieve nvidia driver version (empty string)");
+    // Nvidia driver versions are hopefully float based X.Y
+    const float driverVersion = std::stof(driverVersionString);
+    if (driverVersion == 0)
+        throw runtime_error("Failed to parse driver version: " + driverVersionString);
+    const std::string cudaRuntimeVersionString = getCUDARuntimeVersion();
+    if (cudaRuntimeVersionString.empty())
+        throw runtime_error("Failed to get CUDA version (empty)");
+
+    AF_TRACE("AF built with cuda {}", cudaRuntimeVersionString);
```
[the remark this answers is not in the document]
WilliamTambellini: done

9prady9:
@WilliamTambellini The commit history is still mixed up(see the commit list on Github) with merge commits - I believe there are actually only two/one commit(s) related to this change. Can you please fix that too. Thank you.

PR #2433, opened 2019-02-11
Add free eval functions for const arrays.

### LINKED ISSUE #2427: af::eval() does not accept const array
This code is cannot be compiled
```
const af::array myConstArray= af::constant(111, 10, f64);
af::eval(myConstArray);
```

But, this code can be compiled
```
const af::array myConstArray= af::constant(111, 10, f64);
myConstArray.eval();
```

Currently I wrote my own af::eval to get around this problem
```
af::eval(const af::array &a, const af::array &b, ...)
{
  a.eval();
  ...
}
```

But, why doesn't af::eval() accept const array? Is this just an oversight or this is intentional? What would be the consequences if I `const_cast<af::array *>(&a)` and feed that into your `af::eval(count, af::array*)` similar to how you wrote your `af::eval(const af::array &a, const af::array &b, ...)`?

Thank you

umar456:
Fixes #2427


include/af/array.h
```diff
@@ -1394,6 +1394,61 @@ namespace af
...
+#if AF_API_VERSION >= 37
+
+    /// Evaluate an expression (nonblocking).
+    /**
+       \ingroup method_mat
+       @{
+    */
+    inline const array &eval(const array &a) { a.eval(); return a; }
+
+#if __cplusplus > 199711L
```
9prady9: :+1:

test/jit.cpp
```diff
@@ -411,3 +411,19 @@ TEST(JIT, TransposeBuffers) {
...
+    const array d = constant(1, 1);
+    const array e = constant(1, 1);
+    const array f = constant(1, 1);
+    const array g = constant(1, 1);
+
+    eval(a, b, c, d, e, f, g);
+    af::sync();
+
+    // I can't think of a good test for this.
+    SUCCEED();
```
9prady9:
```
EXPECT_NO_THROW({
  eval(a, b, c, d, e, f, g);
  af::sync();
});
```
can you try if that works ? should :)

include/af/defines.h
```diff
@@ -36,7 +39,9 @@
 #else
     #define AF_DEPRECATED(msg) __attribute__((deprecated))
 #endif
-
+    #if __cpp_variadic_templates >= 200704
+        #define AF_HAS_VARIADIC_TEMPLATES
+    #endif
 #endif
```
9prady9: :+1:

PR #2437, opened 2019-02-15
Add CUDA runtime compilation support using nvrtc

9prady9:
- [x] Documetation for new functions (`src/backend/cuda/nvrtc/cache.hpp`)
- [x] Add checks for compute versions not supported in CUDA 10

Moved the following functions to use runtime compilation while refining the API defined in cache.hpp
* convolve
* scan
* scan_by_key
* separable convolution
* transpose
* where

All required headers(listed below) for runtime compilation will be embedded into the built library. Therefore, developers writing kernels just have to `#include` any required files as usually inside the cuh kernel file. Check `transpose` function to get an idea of how it is being done.
- backend.hpp
- complex.hpp
- jit.cuh    
- math.hpp   
- ops.hpp    
- optypes.hpp
- Param.hpp  
- shared.hpp 
- types.hpp  

The eventual goal is to use math.hpp even inside jit kernels and remove
the code-path controlled by isJIT parameter of buildKernel function. But this has been deferred for another PR.

Notes:
1. Just this change brought down afcuda.so file size by 200MB for single compute version 61. Hopefully, we will see drastic reduction in our final binary once all feasible functions are ported to runtime compilation.
2. CUB can't be included into this framework since it includes some system headers, there is an open [issue](https://github.com/NVlabs/cub/issues/131) regarding this on the corresponding repository.
3. Functions using thrust can only be ported if thrust calls and raw kernels are cleanly separated. Thrust API involving CUDA runtime constructs are basically the blockers.


9prady9:
Some other CUDA tests failed, may be which use JIT. I am looking into those failures.

pavanky:
This is nice. good to use this for full fledged half precision support.

9prady9:
@arrayfire/core-devel I am still debugging some failures on the linux-cuda ci job, but I think it is ready for reviews.

src/backend/cuda/nvrtc/kernel_map.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <map>
+#include <string>
+
+namespace cuda {
+
+/**
+    \brief A map of CUDA kernel source codes
+ */
+static const std::map<std::string, std::string> KernelMap = {
+    { "transpose", std::string(transpose_cuh, transpose_cuh_len) },
```
umar456: Thinking about this again. Maybe we should add the headers in the same source file. Like this:

```
KernelTemplate transposeTemplate = MakeKernel(transpose_cuh, transpose_cuh_len);
...
void transpose(...) {
    auto transpose = transposeTemplate(
                            {TemplateArgs(), ...},
                            {Defines(), ....}
                        );
}
```

src/backend/cuda/nvrtc/launch_args.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+        \brief EnqueueArgs constructor
+
+        \param[in] event is an event for which kernel execution has to wait for
+        \param[in] blks is number of blocks per grid
+        \param[in] thrds is number of threads per block
+        \param[in] sharedMemSize is number of bytes of shared memory allocation
+    */
+    EnqueueArgs(CUevent event, dim3 blks, dim3 thrds,
```
umar456: Blocks and threads should be first always. The optional stuff should be at the end.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,146 @@
...
+        essentially forwarded to cuLaunchKernel driver API call.
+
+        \param[in] qArgs is an object of struct \ref EnqueueArgs
+        \param[in] args is the placeholder for variadic arguments
+    */
+    template<typename... Args>
+    void operator()(const EnqueueArgs& qArgs, Args... args) {
+        void *params[] = {reinterpret_cast<void*>(&args)...};
+        for(auto& event: qArgs.mEvents)
+            CU_CHECK(cuEventSynchronize(event));
```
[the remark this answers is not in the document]
9prady9: I understood that API incorrectly.

9prady9:
thats an odd error on linux, that didn't happen prior to rebase I did today. trying a fresh build on my machine.

9prady9:
I have noticed `__int64` type failures on windows, will debug them soon.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -0,0 +1,383 @@
...
+                 const vector< string >& compileOpts) {
+    vector< string > args;
+    args.reserve(targs.size());
+
+    std::transform(targs.begin(), targs.end(), std::back_inserter(args),
+            [](const TemplateArg& arg) -> std::string {
+                 return arg._tparam;
+            });
+
+    std::string tInstance = nameExpr + "<" + args[0];
```
umar456: Nit: This code block could be a function.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -0,0 +1,383 @@
...
+template std::string toString<long long         >(long long         );
+template std::string toString<unsigned          >(unsigned          );
+template std::string toString<unsigned long     >(unsigned long     );
+template std::string toString<unsigned long long>(unsigned long long);
+template std::string toString<float             >(float             );
+template std::string toString<double            >(double            );
+template std::string toString<long double       >(long double       );
+
+
+template<> std::string toString(bool val) {
```
umar456: remove std:: from and use using statements at the top.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,166 @@
...
+    /// global(__constant__) variable declared inside the kernel.
+    ///
+    /// \param[in] name is the name of the global variable inside kernel
+    /// \param[in] src is the device pointer from which data will be copied
+    /// \param[in] bytes are the number of bytes of data to be copied
+    ///
+    void setConstant(const char* name, CUdeviceptr src, size_t bytes);
+
+    ///
+    /// \brief operator() overload to execute kernel
```
umar456: The "operator() overload" is redundant. A better brief would be

Calls a kernel given a set of Enqueue Args and kernel parameters.

Even that seems meh.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,166 @@
...
+        cuGetErrorString(res, &cu_err_string);                            \
+        snprintf(cu_err_msg, sizeof(cu_err_msg), "CU Error %s(%d): %s\n", \
+                 cu_err_name, (int)(res), cu_err_string);                 \
+        AF_ERROR(cu_err_msg, AF_ERR_INTERNAL);                            \
+    } while (0)
+
+namespace cuda {
+
+///
+/// \brief Kernel Functor that wraps CUDA nvrtc constructs
```
umar456: I think the interface can be expanded to OpenCL. I was thinking about using the Policy pattern. Maybe remove CUDA and nvrtc from the description.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,166 @@
...
+        return {std::string(dtype_traits<T>::getName())};
+    }
+};
+
+#define DefineKey(arg) "-D "#arg
+#define DefineValue(arg) "-D "#arg"=" + toString(arg)
+#define DefineKeyValue(key, arg) "-D "#key"=" + toString(arg)
+
+///
+/// \brief Fetch a Kernel that fits the given criteria
```
umar456: Nit: This function also builds the kernels if they are not available.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,166 @@
...
+///
+/// The paramter \p compileOpts is a list of strings that lets you add definitions
+/// such as `-D<NAME>` or `-D<NAME>=<VALUE>` to the compiler. To enable easy
+/// stringification of variables into their definition equation, three helper
+/// macros are provided: TemplateArg, DefineKey and DefineValue.
+///
+/// Example Usage: transpose
+///
+/// \code
+/// static const std::string src(transpose_cuh, transpose_cuh_len);
```
umar456: 👍

src/backend/cuda/nvrtc/launch_args.hpp
```diff
@@ -0,0 +1,55 @@
+/*******************************************************
```
umar456: You should name this file EnqueueArgs.hpp?

src/backend/cuda/nvrtc/launch_args.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#pragma once
+
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <nvrtc.h>
```
umar456: I don't think this header is needed here.

src/backend/cuda/scan.cpp
```diff
@@ -22,20 +22,10 @@ template<af_op_t op, typename Ti, typename To>
...
 
-    if (inclusive_scan) {
-        switch (dim) {
-            case 0: kernel::scan_first<Ti, To, op, true>(out, in); break;
-            case 1: kernel::scan_dim<Ti, To, op, 1, true>(out, in); break;
-            case 2: kernel::scan_dim<Ti, To, op, 2, true>(out, in); break;
-            case 3: kernel::scan_dim<Ti, To, op, 3, true>(out, in); break;
-        }
+    if (dim == 0) {
+        kernel::scan_first<Ti, To, op>(out, in, inclusive_scan);
```
umar456: Op can also be a function argument can't it?

src/backend/cuda/nvrtc/launch_args.hpp
```diff
@@ -0,0 +1,55 @@
...
+    /// \param[in] stream is CUDA steam on which kernel has to be enqueued
+    /// \param[in] sharedMemSize is number of bytes of shared memory allocation
+    /// \param[in] events is list of events to wait for kernel execution
+    ///
+    EnqueueArgs(dim3 blks, dim3 thrds, CUstream stream = 0,
+                const unsigned int sharedMemSize = 0,
+                const std::vector<CUevent> &events = {}) :
+      mBlocks(blks),
+      mThreads(thrds),
+      mSharedMemSize(sharedMemSize),
```
umar456: The variables should be initialized in the same order as they appear in the class.

src/backend/cuda/nvrtc/cache.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+namespace cuda {
+
+///
+/// \brief Kernel Functor that wraps CUDA nvrtc constructs
+///
+/// This struct encapsulates CUmodule and CUfunction pointers that are required
+/// to execution of CUDA C++ kernels compiled at runtime.
+///
+struct Kernel {
```
umar456: We should probably put classes/structs like this in their own file.

umar456:
We need to do this but I am worried that there are going to be a few combinations of template parameters that are valid but will not be tested. We need be vigilant  to test all types are parameters.

src/backend/cuda/scan.cpp
9prady9: It can, but the problem is larger than it seems to be. Lot of these template parameters,especially the ones in `src/backend/[cuda|opencl|cpu]/*.cpp` level, trickle down from `src/api/c` level. Even though we can now have `op` as normal argument to reduce template instantiations in CUDA backend alone, that will effect the function signatures at `src/backend/[cuda|opencl|cpu]/*.hpp` level. Unless we change all the headers in every other backend, it will become head ache.

Having said that, since it is already a template paramater when it reaches cuda backend implementation level, we might as well use it untill we change the backend API too reduce number of template parameters.

src/backend/cuda/nvrtc/launch_args.hpp
9prady9: We have `CUStream` and `CUEvent` usage in that header.

src/backend/cuda/nvrtc/cache.hpp
9prady9: I can work on making a skeleton and then moving transpose/convolve to use new framework in OpenCL too.

src/backend/cuda/nvrtc/cache.hpp
9prady9: It will probably be moved anyway the way you say if we make it common to cuda/opencl backend.

src/backend/cuda/nvrtc/cache.cpp
9prady9: It isn't being used anywhere else even in the very same source file.

src/backend/cuda/scan.cpp
umar456: Its not something you need to do in this PR. We can change that later.

9prady9:
I have rebased/squashed all CUDA work. Will push once OpenCL changes are ready.

src/backend/cuda/nvrtc/cache.cpp
umar456: I was saying this to better document what is going on without reading the code. Its not a big deal.

umar456:
You should run [Bloaty McBloatface](https://github.com/google/bloaty) before and after on this if its not too difficult. It would be interesting to see.

9prady9:
bloaty output
```
nvrtc ./src/backend/cuda/libafcuda.so.3.7.0
     VM SIZE                         FILE SIZE
 --------------                   --------------
 100.0%   119Mi TOTAL               529Mi 100.0%
  

master ./src/backend/cuda/libafcuda.so
   VM SIZE             FILE SIZE
 --------------          --------------
 100.0%  180Mi TOTAL        640Mi 100.0%
```

PR #2448, opened 2019-03-03
Improve CUDA version checking code

umar456:
* Use cudaRuntimeGetVersion and cudaDriverGetVersion to check the
  compatibility of the toolkit versions vs driver version
* Prints warning messages if the driver version is not included in the
  CudaToDriverVersion in Debug builds.
* Fix int_version_to_string function

### DIFF
--- a/src/backend/cuda/platform.cpp
+++ b/src/backend/cuda/platform.cpp
@@ -137,7 +137,7 @@ static inline int getMinSupportedCompute(int cudaMajorVer) {
     // Vector of minimum supported compute versions
     // for CUDA toolkit (i+1).* where i is the index
     // of the vector
-    static const std::array<int, 10> minSV{1, 1, 1, 1, 1, 1, 2, 2, 3, 3};
+    static const std::array<int, 10> minSV{{1, 1, 1, 1, 1, 1, 2, 2, 3, 3}};
 
     int CVSize = static_cast<int>(minSV.size());
     return (cudaMajorVer > CVSize ? minSV[CVSize - 1]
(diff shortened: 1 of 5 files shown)


src/backend/cuda/platform.cpp
```diff
@@ -239,15 +239,15 @@ string getDriverVersion() {
...
+        return to_string(driver);
     } else {
         return string(driverVersion);
     }
 }
 
 string int_version_to_string(int version) {
     return to_string(version / 1000) + "." +
-           to_string((int)((version % 1000) / 100.));
+           to_string((int)((version % 1000) / 10.));
```
9prady9: Was `100` a bug ?

src/backend/cuda/platform.cpp
```diff
@@ -460,82 +460,146 @@ SparseHandle sparseHandle() {
...
         string msg =
             "ArrayFire was built with CUDA %s which requires GPU driver "
-            "version %.2f or later. Please download the latest drivers from "
-            "https://www.nvidia.com/drivers. Alternatively, you could rebuild "
-            "ArrayFire with CUDA Toolkit version %s to use the current "
-            "drivers.";
+            "version %.2f or later. Please download and install the latest "
+            "drivers from https://www.nvidia.com/drivers for your GPU. "
+            "Alternatively, you could rebuild ArrayFire with CUDA Toolkit "
+            "version %s to use the current drivers.";
```
9prady9: `msg` can be moved down after if block with return statement.

src/backend/cuda/platform.cpp
umar456: Yes

PR #2453, opened 2019-03-04
Clarified and added illustration to rotate docs

mark-poscablo:
The rotate description can be confusing. This is an attempt to clarify it, especially by adding an illustration for the effect of the `crop` parameter. I also added the interpolation method options based on [`src/api/c/rotate.cpp`](https://github.com/arrayfire/arrayfire/blob/f38feccccf04a594c066c6ae6f81225d06d716c2/src/api/c/rotate.cpp#L4) and[`defines.h`](https://github.com/arrayfire/arrayfire/blob/f38feccccf04a594c066c6ae6f81225d06d716c2/include/af/defines.h#L23). Here's what the detailed description looks like from this PR:

rotate_docs

The illustration can be found in this PR: https://github.com/arrayfire/assets/pull/9

### DIFF
--- a/docs/details/image.dox
+++ b/docs/details/image.dox
@@ -626,9 +626,13 @@ dimensions will stay the same as the original image's, but the rotated image's
 portions outside the dashed orange lines will be cropped, and the rest of the
 output image (the area between the solid black and solid orange lines) will be
 filled with zeros. However, if \p crop is false, then the output image's
-dimensions will be bigger (outermost box with dashed black lines) in order to
-accommodate all of the rotated image's data, and the remainder will be filled
-with zeros (the area between the solid orange lines and dashed black lines).
+dimensions will be bigger (at least in this illustration), as represented by
+the outermost box with dashed black lines. This change is necessary to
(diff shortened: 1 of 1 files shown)


9prady9:
@mark-poscablo I have merged the assets PR. Please update the assets submodule in this PR now.
The docs look good.

docs/details/image.dox
```diff
@@ -626,9 +626,13 @@ dimensions will stay the same as the original image's, but the rotated image's
 portions outside the dashed orange lines will be cropped, and the rest of the
 output image (the area between the solid black and solid orange lines) will be
 filled with zeros. However, if \p crop is false, then the output image's
-dimensions will be bigger (outermost box with dashed black lines) in order to
-accommodate all of the rotated image's data, and the remainder will be filled
-with zeros (the area between the solid orange lines and dashed black lines).
+dimensions will be bigger (at least in this illustration), as represented by
```
9prady9: I think `will be bigger (at least in this illustration)` should be something like the following.
```
...
dimensions might get bigger for some cases (as shown in the above illustration)
...
```

PR #2454, opened 2019-03-11
Add support for multigpu training in the neural net example

WilliamTambellini:
Add support for multigpu parallel training in the neural net example


9prady9:
The updated example looks good. I have made some cmake changes.

WilliamTambellini:
Cool tks. Feel free to merge/modify your way.

9prady9:
I ran it on multiple(two: GTX 1060, AMD Spectre R7) devices, works without any issues. This change demonstrates how to use multiple devices but not so much on how to distribute training data across multiple devices. How difficult would be such an example to implement ?

WilliamTambellini:
> how to distribute training data across multiple devices. How difficult would be such an example to implement ?
That will be worth it to do as soon as 
https://github.com/arrayfire/arrayfire/issues/2463
done. 
That example at least shows how to do multigpu multimodels training.

WilliamTambellini:
That item would allow to also compare speed of mutiple vs single gpus execution.  
Distruibuted single model training is out of scope here.
Review please ?

Issue #2436, opened 2019-02-13
CUDA_ERROR_INVALID_IMAGE error when running arrayfire on GPU


gonzaloiglesiasiglesias:
I am running into the following error:

terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Internal error:998):
In function cuda::kc_entry_t cuda::compileKernel(const char*, std::string)
In file src/backend/cuda/jit.cpp:342
CU Error CUDA_ERROR_INVALID_IMAGE(200): device kernel image is invalid

In function af::array af::operator+(const af::array&, const af::array&)
In file src/api/cpp/array.cpp:920

Running with AF_CUDA_MAX_JIT_LEN=10 makes the error go away. Also, running on cpu is fine.
Any ideas?

9prady9:
Can you please give more details such as the following:
- The version of ArrayFire are you using
- The GPU on which ArrayFire is running
- CUDA driver version
- Operating System

WilliamTambellini:
- AF 3.6.2
- GPU : reproduced with both TitanV and TitanX
- NVIDIA-SMI 396.37 
- OS: CentOS Linux release 7.5.1804 (Core)

WilliamTambellini:
Probably similar to 
https://github.com/arrayfire/arrayfire/issues/2084
which has been closed without any fix ?

umar456:
Thats my guess. I have a fix that is in the works. It will take me a bit to get it working. For now you may need to add an eval manually.

umar456:
But this is a different error so I am curious about the kernel generated.

WilliamTambellini:
Also reproduced on GTX1060 using default AF_CUDA_MAX_JIT_LEN=100 : 
terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Internal error:998):
In function cuda::kc_entry_t cuda::compileKernel(const char*, std::string)
In file src/backend/cuda/jit.cpp:342
CU Error CUDA_ERROR_INVALID_IMAGE(200): device kernel image is invalid
In function af::array af::operator+(const af::array&, const af::array&)
In file src/api/cpp/array.cpp:920
Aborted (core dumped)

I have to reduce AF_CUDA_MAX_JIT_LEN down to 20 to make it run.

WilliamTambellini:
The very first step would be to log which kernel is failing to load : 
example : in cache.cpp line 188 : 

[...]

Issue #2389, opened 2018-12-24
Address binary JIT operations where combined node is too large


umar456:
JIT operations that take two or more operands can fail due to a large parameter size in case of non-linear JIT kernels. This sceanario appeared in issue https://github.com/arrayfire/arrayfire-python/issues/199. You can recreate this error using the following C++ code:

```
TEST(JIT, TwoLargeNonLinear) {
  array a(10, 10);
  array b(10, 10);

  for(int i = 0; i < 23; i++) {
    a += tile(randu(10), 1, 10) + randu(10, 10);
  }

  for(int i = 0; i < 23; i++) {
    b += tile(randu(10), 1, 10) + randu(10, 10);
  }


  array c = a + b;
  c.eval();
  af::sync();
}
```

9prady9:
Fixed in https://github.com/arrayfire/arrayfire/pull/2484

PR #2458, opened 2019-03-16
Fix errors caused by large JIT kernels

umar456:
Fixes errors caused by JIT kernels that were too large. The errors occurred because the size of the arguments passed to the JIT kernel was greater than 4kb. This is the limit on NVIDIA devices. This PR fixes this by passing the shape information via global memory. The kernels have been modified so that the parameter array is read into the shared memory and then used in the rest of the kernel.

Here is an example of a non-linear kernel with the new changes. Notice that I am passing the pointer and the index. This index is the index in the `dims` Param array that stores the dimensions and strides of that buffer.
```
extern "C" __global__ void
KER6346349204703026994(
int *in0_ptr, int in_index0,
int *in1_ptr, int in_index1,
Param out2, 
Param out3, 
Param* dims, int num_params, uint blocks_x, uint blocks_y, uint blocks_x_total, uint num_odims)
{

const Param &outref = out2;

        extern __shared__ Param params[];
        long long offsets[8];

        if (threadIdx.x < num_params) { params[threadIdx.x] = dims[threadIdx.x]; }
        __syncthreads();
    
    for (int blockIdx_x = blockIdx.x; blockIdx_x < blocks_x_total; blockIdx_x += gridDim.x) {
    
        long long id0 = 0, id1 = 0, id2 = 0, id3 = 0;
        long blockIdx_y = blockIdx.z * gridDim.y + blockIdx.y;
        if (num_odims > 2) {
            id2 = blockIdx_x / blocks_x;
            id0 = blockIdx_x - id2 * blocks_x;
            id0 = threadIdx.x + id0 * blockDim.x;
            if (num_odims > 3) {
                id3 = blockIdx_y / blocks_y;
                id1 = blockIdx_y - id3 * blocks_y;
                id1 = threadIdx.y + id1 * blockDim.y;
            } else {
                id1 = threadIdx.y + blockDim.y * blockIdx_y;
            }
        } else {
            id3 = 0;
            id2 = 0;
            id1 = threadIdx.y + blockDim.y * blockIdx_y;
            id0 = threadIdx.x + blockDim.x * blockIdx_x;
        }

        bool cond = id0 < outref.dims[0] &&
                    id1 < outref.dims[1] &&
                    id2 < outref.dims[2] &&
                    id3 < outref.dims[3];

        if (!cond) { continue; }

        long long idx = outref.strides[3] * id3 +
                        outref.strides[2] * id2 +
                        outref.strides[1] * id1 + id0;

        for(int i = 0; i < num_params; i++) {
            offsets[i] = (id3 < params[i].dims[3]) * params[i].strides[3] * id3 +
                         (id2 < params[i].dims[2]) * params[i].strides[2] * id2 +
                         (id1 < params[i].dims[1]) * params[i].strides[1] * id1 +
                         (id0 < params[i].dims[0]) * id0;
        }
        int val0 = in0_ptr[offsets[in_index0]];
int val1 = in1_ptr[offsets[in_index1]];
int val2 = __add(val0, val1);
int val3 = __sub(val0, val1);
((int*)(out2.ptr))[idx] = val2;
((int*)(out3.ptr))[idx] = val3;
}
```

Fixes: #2436 #2389

### DIFF
--- a/src/backend/common/jit/Node.hpp
+++ b/src/backend/common/jit/Node.hpp
@@ -49,28 +49,46 @@ class Node {
     int getNodesMap(Node_map_t &node_map, std::vector<const Node *> &full_nodes,
                     std::vector<Node_ids> &full_ids) const;
 
+    /// Generates the string that will be used to hash the kernel
     virtual void genKerName(std::stringstream &kerStream,
-                            const Node_ids &ids) const {
-        UNUSED(kerStream);
-        UNUSED(ids);
-    }
+                            const Node_ids &ids) const = 0;
+
+    /// Generates the function parameters for the node.
+    ///
+    /// \param[in/out] kerStream  The string will be written to this stream
+    /// \param[in]     ids        The integer id of the node and its children
+    /// \param[in]     is_linear  True if the kernel is a linear kernel
     virtual void genParams(std::stringstream &kerStream, int id,
                            bool is_linear) const {
         UNUSED(kerStream);
         UNUSED(id);
         UNUSED(is_linear);
     }
+
+    /// Generates the variable that stores the thread's/work-item's offset into
+    /// the memory.
+    ///
+    /// \param[in/out] kerStream  The string will be written to this stream
+    /// \param[in]     ids        The integer id of the node and its children
+    /// \param[in]     is_linear  True if the kernel is a linear kernel
     virtual void genOffsets(std::stringstream &kerStream, int id,
                             bool is_linear) const {
         UNUSED(kerStream);
         UNUSED(id);
         UNUSED(is_linear);
     }
-    virtual void genFuncs(std::stringstream &kerStream,
-                          const Node_ids &ids) const {
-        UNUSED(kerStream);
-        UNUSED(ids);
-    }
+
+    /// Generates the code for the operation of the node.
+    ///
+    /// Generates the soruce code of the operation that the node needs to
+    /// perform. For example this function will create the string
+    /// "val2 = __add(val1, val2);" for the addition node.
+    ///
+    /// \param[in/out] kerStream  The string will be written to this stream
+    /// \param[in]     ids        The integer id of the node and its children
+    /// \param[in]     is_linear  True if the kernel is a linear kernel
+    virtual void genFuncs(std::stringstream &kerStream, const Node_ids &ids,
+                          bool is_linear) const = 0;
 
     /// Calls the setArg function on each of the arguments passed into the
     /// kernel
@@ -90,6 +108,11 @@ class Node {
         return start_id;
     }
 
+    // Sets the index of the Param object stored in global memory. (CUDA ONLY)
+    virtual void setParamIndex(int index) {}
+    // Gets the index of the Param object stored in global memory. (CUDA ONLY)
+    virtual int getParamIndex() const { return -1; }
+
     virtual void getInfo(unsigned &len, unsigned &buf_count,
                          unsigned &bytes) const {
         UNUSED(buf_count);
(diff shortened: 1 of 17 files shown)


9prady9:
The failures are not random I think, same set failed in two consecutive runs on `linux-cuda-k80`.

src/backend/cuda/jit.cpp
```diff
@@ -155,21 +179,31 @@ struct Param
...
+                       << ".ptr))"
+                       << "[idx] = val" << id << ";\n";
     }
 
+    // TODO(umar): find the ideal size for the offset array. This
+    // size shouldn't be too large I would think. I am sure someone
+    // will find a way to hit this bug waiting to happen.
+    paramreadstream << R"JIT(
+        extern __shared__ Param params[];
+        long long offsets[8];
```
9prady9: We should be able to do the following and avoid the hard coded declaration of offsets array. Assuming, we already included "Shared.hpp" thats available via buildKernel function.
```
SharedMemory<unsigned char> shared;
unsigned char* shrdMem = shared.getPointer();

Param* params = (Param*)shrdMem;
long long* offsets = (long long*)(&shrdMem[offsets_offset]);
```
`offsets_offset` could be passed along with `num_params`. Once we have `offsets` pointer, we would only access offsets array for only `num_params` values.

src/backend/cuda/jit.cpp
```diff
@@ -319,14 +383,27 @@ void evalNodes(vector<Param<T>> &outputs, vector<Node *> output_nodes) {
...
+
     args.push_back((void *)&blocks_x_);
     args.push_back((void *)&blocks_y_);
     args.push_back((void *)&blocks_x_total);
     args.push_back((void *)&num_odims);
 
     CU_CHECK(cuLaunchKernel(ker, blocks_x, blocks_y, blocks_z, threads_x,
-                            threads_y, 1, 0, getActiveStream(), args.data(),
-                            NULL));
+                            threads_y, 1, sizeof(Param<T>) * params.size(),
```
9prady9: `params.size() * sizeof(Param<T>)` would become our `offsets_offset` from my earlier comment.
`params.size() * (sizeof(Param<T>) + sizeof(long long))` would become our total shared memory size.

src/backend/cuda/jit.cpp
umar456: offset is a local array not a shared array

src/backend/cuda/jit.cpp
```diff
@@ -72,12 +85,11 @@ static string getKernelString(const string funcName,
...
     const std::string paramTStr = R"JIT(
-template<typename T>
 struct Param
 {
-  T *ptr;
   dim_t dims[4];
   dim_t strides[4];
+  void *ptr;
 };
 )JIT";
```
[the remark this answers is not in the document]
umar456: I am doing this so that I can avoid adding types to Params since these param objects are really just used to store the stride and the dimensions of the inputs. I should probably make another type for this but it doesn't really add much.

src/backend/common/jit/Node.hpp
```diff
@@ -49,28 +49,46 @@ class Node {
     int getNodesMap(Node_map_t &node_map, std::vector<const Node *> &full_nodes,
                     std::vector<Node_ids> &full_ids) const;
 
+    /// Generates the string that will be used to hash the kernel
     virtual void genKerName(std::stringstream &kerStream,
-                            const Node_ids &ids) const {
-        UNUSED(kerStream);
-        UNUSED(ids);
-    }
+                            const Node_ids &ids) const = 0;
```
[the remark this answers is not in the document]
umar456: Its just used to create the hash string. There are better ways to do this. I will eventually replace this in place of a better hash function

src/backend/cuda/jit.cpp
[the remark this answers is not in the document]
9prady9: It would be just more consistent with how regular kernels compiled by nvrtc will also look like when traced. And one less type to manager in multiple locations.

src/backend/cuda/jit.cpp
9prady9: May be I wasn't clear, I am asking you to make it shared along with `Param`s. Making it a fixed size `8` and then waiting for the bug to happen(as described by you in the TODO comment) is not correct.

9prady9: After our little chat, I do see that offsets are based on global thread ids, hence have to be stored for each thread in the block. We can still do it by setting the shared memory size to the following
```
params.size() * (sizeof(Param<T>) + sizeof(long long)*total_num_of_threads_in_a_block)
```

9prady9:
@umar456 fyi in case you haven't noticed, jit opencl tests are failing.

umar456:
superseded by #2487

PR #2459, opened 2019-03-16
Remove void* memAlloc instantiation. Replace with uchar

umar456:


9prady9:
@umar456 I force pushed to trigger ci jobs.

PR #2460, opened 2019-03-18
Add framework for extensible ArrayFire memory managers

jacobkahn:
## Motivation
Many different use cases require performance across many different memory allocation patterns. Even different devices/backends have different costs associated with memory allocations/manipulations. Having the flexibility to implement different memory management schemes can help optimize performance for the use case and backend.

## Framework
- The basic interface lives in a new header: `include/af/memory.h` and includes two interfaces:
  - A C-style interface defined in the `af_memory_manager` struct, which includes function pointers to which custom memory manager implementations should be defined along with device/backend-specific functions that can be called by the implementation (e.g. `nativeAlloc`) and will be dynamically set. Typesafe C-style struct inheritance should be used.
  - A C++ style interface using `MemoryManagerBase`, which defines pure-virtual methods for the API along with device/backend-specific functions as above.

A C++ implementation is simple, and requires only:
```
#include <af/memory.h>
...
class MyCustomManager : public af::MemoryManagerBase {
...
  void* alloc(const size_t size, bool user_lock) override {
    ...
    void* ptr = this->nativeAlloc(...);
    ...
  }
};

// In some code run at startup:
af::MemoryManagerBase* p = new MyCustomMemoryManager();
af::setMemoryManager(p);
```

For the C API:
```
#include <af/memory.h>
...
af_memory_manager_impl_alloc(const size_t size, bool user_lock) {
  ...
}

typedef struct af_memory_manager_impl {
  af_memory_manager manager; // inherit base methods
  // define custom implementation
} af_memory_manager_impl;

// In some code run at startup:
af_memory_manager* p = malloc(sizeof(af_memory_manager_impl));
p->af_memory_manager_alloc = &af_memory_manager_impl_alloc;
af_set_memory_manager(p);

```

### Details
- The F-bound polymorphism pattern present in the existing `MemoryManager` implementation is removed; removing this was required as it precludes dynamically dispatching to a derived implementation.
- New interfaces are defined for C/C++ (see below)
- `MemoryManagerCWrapper` wraps a C struct implementation of a memory manager and facilitates using the same backend and `DeviceManager` APIs to manipulate a manager implemented in C.

### API Design Decisions
  - If a custom memory manager is not defined or set, the default memory manager will be used. While the default memory manager implements the new existing interface, behavior is completely identical to existing behavior by default (as verified by tests)
  - Memory managers should be stored on the existing `DeviceManager` framework so as to preserve the integrity of existing backend APIs. As such, memory managers can exist on a per-backend basis and work with the unified backend.
  - Existing ArrayFire APIs expect garbage collection and memory step sizing to be implemented in a memory manager. As such, these and a few other slightly opinionated methods are included in the overall API.
    - That said, these methods can be noops or throw exceptions (e.g. garbage collection) if the style of custom memory manager implementation doesn't implement those facilities.
  - Setting a memory manager should use one API in the same C/C++ fashion so as to be compatible with the unified backend via dynamic invocation of symbols in a shared object. As such, the C and C++ APIs should have a polymorphic relationship such that either can be passed to the public API (`af::MemoryManagerBase` is this a subtype of `af_memory_manager`, a C struct)

<hr/>

- Adds tests defining custom memory manager implementations in both the C++ and C API and testing end-to-end Array allocations and public AF API calls (e.g. garbage collection, step size).


jacobkahn:
[still need to rebase this properly]

PR #2472, opened 2019-03-30
Remove the use of addition and scalars in tile jit nodes

umar456:
Tiles used to be performed using the add jit node with a zero scalar. This
performed extra operations and created an extra parameter which was
unnecessary.


umar456:
> I haven't been able to understand why we need additional parameter `outDim` for `unaryOp`. Can't we use `op == af_noop_t` comparison which is compile time check?

umar456:
@9prady9 We need to resize the output after the noop. I wanted to do this in a generic  way. You could potentially do the same thing with any other operation.

PR #2476, opened 2019-03-31
Add Apple platform to jit parameter size check

umar456:
The apple platform can target NVIDIA GPUs so the checks for parameter sizes needs to be performed on there as well.

### DIFF
--- a/src/backend/opencl/Array.cpp
+++ b/src/backend/opencl/Array.cpp
@@ -247,8 +247,8 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
 
             bool isBufferLimit =
                 lock_bytes > getMaxBytes() || lock_buffers > getMaxBuffers();
-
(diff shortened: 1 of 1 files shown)


src/backend/opencl/Array.cpp
```diff
@@ -247,8 +247,8 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
 
             bool isBufferLimit =
                 lock_bytes > getMaxBytes() || lock_buffers > getMaxBuffers();
-
-            bool isNvidia = getActivePlatform() == AFCL_PLATFORM_NVIDIA;
+            bool isNvidia = getActivePlatform() == AFCL_PLATFORM_NVIDIA ||
```
9prady9: is this parameter named incorrectly or something ? or `AFCL_PLATFORM_APPLE` can be any OpenCL device , can't it be ?

9prady9: After some offline discussion, this is considered okay OSX.

PR #2477, opened 2019-04-01
Use compute for nvrtc based on CUDA toolkit version

### LINKED ISSUE #2471: Titan RTX Cuda backend crashing
When running benchmarks/blas_cuda I get
```
ArrayFire v3.7.0 (CUDA, 64-bit Linux, build 6da99359)
Platform: CUDA Toolkit 10.0, Driver: 418.43
[0] TITAN RTX, 24191 MB, CUDA Compute 7.5
Benchmark N-by-N matrix multiply
ArrayFire Exception (Internal error:998):
In function cuda::Kernel cuda::buildKernel(const string&, const string&, const std::vector<std::__cxx11::basic_string<char> >&, bool)
In file src/backend/cuda/nvrtc/cache.cpp:157
NVRTC ERROR

In function af::array af::matmul(const af::array&, const af::array&, af::matProp, af::matProp)
In file src/api/cpp/blas.cpp:18
In function cuda::Kernel cuda::buildKernel(const string&, const string&, const std::vector<std::__cxx11::basic_string<char> >&, bool)
In file src/backend/cuda/nvrtc/cache.cpp:157
NVRTC ERROR

In function af::array af::matmul(const af::array&, const af::array&, af::matProp, af::matProp)
In file src/api/cpp/blas.cpp:18
Aborted (core dumped)
```
I also tried the installer version. The Open CL backend also gives different results and sometimes crashes on my code (comparing against a quadro p2000 mobile). 

Running Linux 4.19.30-1

I checked issue #2343, CMake seems to have found the correct cuda compute (7.5, it is in the output above as well).

9prady9:
Fixes #2471

### DIFF
--- a/src/backend/cuda/nvrtc/cache.cpp
+++ b/src/backend/cuda/nvrtc/cache.cpp
@@ -95,49 +95,6 @@ void Kernel::setConstant(const char* name, CUdeviceptr src, size_t bytes) {
     CU_CHECK(cuMemcpyDtoDAsync(dst, src, bytes, getActiveStream()));
 }
 
-pair<int, int> getComputeFlag() {
-    struct cuNVRTCcompute {
-        /// The CUDA Toolkit version returned by cudaRuntimeGetVersion
-        int cuda_version;
-
-        /// Maximum major compute flag supported by cuda_version
-        int major;
-
-        /// Maximum minor compute flag supported by cuda_version
-        int minor;
-    };
-    static const cuNVRTCcompute Toolkit2Compute[] = {
-        {10010, 7, 5},
-        {10000, 7, 2},
-        {9020,  7, 2},
-        {9010,  7, 2},
-        {9000,  7, 2},
-        {8000,  5, 3},
-        {7050,  5, 3},
-        {7000,  5, 3}
-    };
-    int runtime_cuda_ver = 0;
-    CUDA_CHECK(cudaRuntimeGetVersion(&runtime_cuda_ver));
-    auto tkit_max_compute =
-        find_if(begin(Toolkit2Compute), end(Toolkit2Compute),
-                [runtime_cuda_ver](cuNVRTCcompute v) {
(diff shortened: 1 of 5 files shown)


src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -68,6 +74,7 @@ using kc_t = map<string, Kernel>;
         char* logptr = log.get();                      \
         nvrtcGetProgramLog(prog, logptr);              \
         logptr[logSize] = '\x0';                       \
+        printf("Compile error log: \n %s \n", logptr); \
```
WilliamTambellini: any way to print to stderr ?

9prady9: In fact, we will remove this soon once we add logging to nvrtc compilation pipeline. At which point, it would would use `AF_JIT_KERNEL_TRACE` environment variable I think.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -89,6 +96,49 @@ void Kernel::setConstant(const char* name, CUdeviceptr src, size_t bytes) {
...
+        int cuda_version;
+
+        /// Maximum major compute flag supported by cuda_version
+        int major;
+
+        /// Maximum minor compute flag supported by cuda_version
+        int minor;
+    };
+    static const cuNVRTCcompute Toolkit2Compute[] = {
+        {10010, 7, 5},
```
umar456: I think you should convert the compute versions to int format on the right. There is not maximum minor version. For example if the maximum version was 7.0, then the condition below will fail for version 5.3. You can use the https://github.com/arrayfire/arrayfire/blob/bce559f881006932de7c2a6787217dacc4c52937/src/backend/cuda/platform.cpp#L246 to convert it to string.

9prady9:
I still have to debug the `vision` function failures after this change. Not sure how what I changed is causing them to fail. Will push a fix soon.

PR #2324, opened 2018-10-23
Allow af_approx1 to use an existing af_array for output

mark-poscablo:
@umar456 had an idea to allow arrayfire functions to be able to use existing arrays to place their output on, instead of the functions creating new output arrays all the time. This saves overhead time, memory, and extraneous code especially if the user already has an existing memory block allocated for results to be placed in. `af_approx1` is the first function that we tried this idea on, and here's the implementation.

Note that this can only be done on the C API for now.

Issue #2446, opened 2019-02-26
matmul and reduce over huge tensors


gonzaloiglesiasiglesias:
Hi,
Let's say that we have these two tensors
A: 64000 x 10 x 100
B: 10 x 512 x 100

C= af::matmul(A,B) has size 64000 x 512 x 100.
In my application, this matmul is in many cases followed by summing all slices together, which of course returns a matrix D of dimensions 64000 x 512.
To avoid OOM issues due to matmul I'm currently looping through the slices of A, B to run
matmul operations and accumulate into D on the fly.   Is there a more efficient way?

Thanks!

umar456:
Instead of doing the batches individually. You can work on a larger batch size. For example instead of performing 100 matmuls you can do smaller batch operations over 20 or so.

We have some ideas to perform this operation on pre-allocated memory so you would place the results directly into the memory space. It's not implemented yet and I don't think it will be in the upcoming release but should be in the one after.

WilliamTambellini:
Hi @umar456 
Outputing to a preallocated space would nt resolve that issue : the issue here is that a giant tensor is allocated in order to store the full matmul output (100 slices) before reducing it (sum on the 3rd dim). 
The title of that item should consequently better be "matmul and reduce" : in that case, there is no need to store the full matmul output tensor, reducing/accumulating on the fly slice after slice should do a better job and save tons of memory.    
I m wondering if simply using cublasSgemmStridedBatched() and setting "strideC" to 0 would nt do the job :
https://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-gemmstridedbatched
?

WilliamTambellini:
Ok some good news :  using the beta param as documented there 
https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemm

 C = α op ( A ) op ( B ) + β C 

does allow to accumulate the output to a single 2d output matrix :   

A=
0 1 
1 2 
2 3 
 0  1 
 1  2 
 2  3 
B=
0 1 2 
1 2 3 
 0  1  2 
 1  2  3 

SerializedGemm: A=3 2 2  B=2 3 2  C=3 3  beta=0
C=
1 2 3 
2 5 8 
3 8 13 

SerializedGemm: A=3 2 2  B=2 3 2  C=3 3  beta=1
C=
2 4 6 
4 10 16 
6 16 26 


So if I m correct all we need is to be able to set the value of beta.
[...]

PR #2481, opened 2019-04-02
Allow af_matmul to use an existing af_array for output

mark-poscablo:
Similar to #2324, but for matmul. Still WIP as the OpenCL BLAS tests are failing. Made a PR for now so that my changes will be visible to you guys while I'm working on this.

### DIFF
--- a/src/api/c/blas.cpp
+++ b/src/api/c/blas.cpp
@@ -17,7 +17,10 @@
 #include <sparse_handle.hpp>
 #include <af/array.h>
 #include <af/blas.h>
+#include <af/data.h>
 #include <af/defines.h>
+#include <af/dim4.hpp>
+#include <type_util.hpp>
 
 template<typename T>
 static inline af_array sparseMatmul(const af_array lhs, const af_array rhs,
@@ -27,10 +30,14 @@ static inline af_array sparseMatmul(const af_array lhs, const af_array rhs,
 }
 
 template<typename T>
-static inline af_array matmul(const af_array lhs, const af_array rhs,
-                              af_mat_prop optLhs, af_mat_prop optRhs) {
-    return getHandle(
-        detail::matmul<T>(getArray<T>(lhs), getArray<T>(rhs), optLhs, optRhs));
+static inline void gemm(af_array *out, af_mat_prop optLhs, af_mat_prop optRhs,
+                        const T* alpha,
+                        const af_array lhs, const af_array rhs,
+                        const T* betas) {
+    detail::gemm<T>(getArray<T>(*out), optLhs, optRhs,
+                    alpha,
+                    getArray<T>(lhs), getArray<T>(rhs),
+                    betas);
 }
 
 template<typename T>
@@ -105,16 +112,15 @@ af_err af_sparse_matmul(af_array *out, const af_array lhs, const af_array rhs,
     return AF_SUCCESS;
 }
 
-af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
-                 const af_mat_prop optLhs, const af_mat_prop optRhs) {
-    using namespace detail;
+af_err af_gemm(af_array *out,
+               const af_mat_prop optLhs, const af_mat_prop optRhs,
+               const void* alpha, const af_array lhs, const af_array rhs,
+               const void* beta) {
+    using namespace detail; // needed for cfloat and cdouble
 
     try {
-        const ArrayInfo &lhsInfo = getInfo(lhs, false, true);
-        const ArrayInfo &rhsInfo = getInfo(rhs, true, true);
-
-        if (lhsInfo.isSparse())
-            return af_sparse_matmul(out, lhs, rhs, optLhs, optRhs);
+        const ArrayInfo &lhsInfo    = getInfo(lhs, false, true);
+        const ArrayInfo &rhsInfo    = getInfo(rhs, true, true);
 
         af_dtype lhs_type = lhsInfo.getType();
         af_dtype rhs_type = rhsInfo.getType();
@@ -131,11 +137,11 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
                      AF_ERR_NOT_SUPPORTED);
         }
 
-        dim4 lDims = lhsInfo.dims();
-        dim4 rDims = rhsInfo.dims();
+        af::dim4 lDims = lhsInfo.dims();
+        af::dim4 rDims = rhsInfo.dims();
 
         if (lDims.ndims() > 2 && rDims.ndims() > 2) {
-            DIM_ASSERT(1, lDims.ndims() == rDims.ndims());
+            DIM_ASSERT(3, lDims.ndims() == rDims.ndims());
             if (lDims[2] != rDims[2] && lDims[2] != 1 && rDims[2] != 1) {
                 AF_ERROR("Batch size mismatch along dimension 2", AF_ERR_BATCH);
             }
@@ -145,23 +151,113 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
         }
 
         TYPE_ASSERT(lhs_type == rhs_type);
-        af_array output = 0;
 
         int aColDim = (optLhs == AF_MAT_NONE) ? 1 : 0;
         int bRowDim = (optRhs == AF_MAT_NONE) ? 0 : 1;
 
         DIM_ASSERT(1, lhsInfo.dims()[aColDim] == rhsInfo.dims()[bRowDim]);
 
-        switch (lhs_type) {
-            case f32: output = matmul<float>(lhs, rhs, optLhs, optRhs); break;
-            case c32: output = matmul<cfloat>(lhs, rhs, optLhs, optRhs); break;
-            case f64: output = matmul<double>(lhs, rhs, optLhs, optRhs); break;
-            case c64: output = matmul<cdouble>(lhs, rhs, optLhs, optRhs); break;
-            default: TYPE_ERROR(1, lhs_type);
+        // Assume that *out is either initialized to null or an actual af_array
+        // Otherwise, this function has undefined behavior
+        af_array output = 0;
+        if (*out) {
+            output = *out;
         }
+        else {
+            const int aRowDim = (optLhs == AF_MAT_NONE) ? 0 : 1;
+            const int bColDim = (optRhs == AF_MAT_NONE) ? 1 : 0;
+            const int M = lDims[aRowDim];
+            const int N = rDims[bColDim];
+            const dim_t d2 = std::max(lDims[2], rDims[2]);
+            const dim_t d3 = std::max(lDims[3], rDims[3]);
+            const af::dim4 oDims = af::dim4(M, N, d2, d3);
+            AF_CHECK(af_create_handle(&output, lhsInfo.ndims(),
+                                      oDims.get(), lhs_type));
+        }
+
+        switch (lhs_type) {
+            case f32: gemm<float>  (&output, optLhs, optRhs,
+                                    static_cast<const float*  >(alpha), lhs, rhs,
+                                    static_cast<const float*  >(beta)); break;
+            case c32: gemm<cfloat> (&output, optLhs, optRhs,
+                                    static_cast<const cfloat* >(alpha), lhs, rhs,
+                                    static_cast<const cfloat* >(beta)); break;
+            case f64: gemm<double> (&output, optLhs, optRhs,
+                                    static_cast<const double* >(alpha), lhs, rhs,
+                                    static_cast<const double* >(beta)); break;
+            case c64: gemm<cdouble>(&output, optLhs, optRhs,
+                                    static_cast<const cdouble*>(alpha), lhs, rhs,
+                                    static_cast<const cdouble*>(beta)); break;
+            default: TYPE_ERROR(3, lhs_type);
+        }
+
         std::swap(*out, output);
     }
     CATCHALL
+    return AF_SUCCESS;
+}
+
+af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
+                 const af_mat_prop optLhs, const af_mat_prop optRhs) {
+    using namespace detail; // needed for cfloat and cdouble
+
+    try {
+
+        const ArrayInfo &lhsInfo = getInfo(lhs, false, true);
+        const ArrayInfo &rhsInfo = getInfo(rhs, true, true);
+
+        if (lhsInfo.isSparse())
+            return af_sparse_matmul(out, lhs, rhs, optLhs, optRhs);
+
+        const int aRowDim = (optLhs == AF_MAT_NONE) ? 0 : 1;
+        const int bColDim = (optRhs == AF_MAT_NONE) ? 1 : 0;
+
+        const af::dim4 lDims = lhsInfo.dims();
+        const af::dim4 rDims = rhsInfo.dims();
+        const int M = lDims[aRowDim];
+        const int N = rDims[bColDim];
+
+        const dim_t d2 = std::max(lDims[2], rDims[2]);
+        const dim_t d3 = std::max(lDims[3], rDims[3]);
+        const af::dim4 oDims = af::dim4(M, N, d2, d3);
+        const int num_batch = oDims[2] * oDims[3];
+
+        af_array gemm_out = 0;
+        AF_CHECK(af_create_handle(&gemm_out, oDims.ndims(), oDims.get(), lhsInfo.getType()));
+
+        af_dtype lhs_type = lhsInfo.getType();
+        switch (lhs_type) {
+            case f32: {
+                    float alpha = 1.f;
+                    float beta  = 0.f;
+                    AF_CHECK(af_gemm(&gemm_out, optLhs, optRhs, &alpha, lhs, rhs, &beta));
+                    break;
+            }
+            case c32: {
+                    cfloat alpha = {1.f, 0.f};
+                    cfloat beta  = {0.f, 0.f};
+
+                    AF_CHECK(af_gemm(&gemm_out, optLhs, optRhs, &alpha, lhs, rhs, &beta));
+                    break;
+            }
+            case f64: {
+                    double alpha = 1.0;
+                    double beta  = 0.0;
+                    AF_CHECK(af_gemm(&gemm_out, optLhs, optRhs, &alpha, lhs, rhs, &beta));
+                    break;
+            }
+            case c64: {
+                    cdouble alpha = {1.0, 0.0};
+                    cdouble beta  = {0.0, 0.0};
+                    AF_CHECK(af_gemm(&gemm_out, optLhs, optRhs, &alpha, lhs, rhs, &beta));
+                    break;
+            }
+            default: TYPE_ERROR(1, lhs_type);
+        }
+
+        std::swap(*out, gemm_out);
+    }
+    CATCHALL;
     return AF_SUCCESS;
 }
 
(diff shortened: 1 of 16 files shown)


WilliamTambellini:
Cool. Would that work also for batched matmul (per slice matmul) ?

mark-poscablo:
@WilliamTambellini I think that must be ensured, though I haven't added this specific test yet in the published commits. The goal is that all of the existing behavior of `af_matmul` will not change (which includes its current batching behavior), but we will have the added ability to reuse an existing af_array. Thanks for bringing it up!

9prady9:
@WilliamTambellini This change is to the C-API, we will be adding a new `gemm` API that we have been [discussing](https://github.com/arrayfire/arrayfire/issues/2446#issuecomment-476691353) in the issue #2446 . It will look different from regular `matmul`. Something like the following

```cpp
void af::gemm(af::array& C, const array &lhs, const array &rhs,
const matProp optLhs = AF_MAT_NONE, const matProp optRhs = AF_MAT_NONE,
const float alpha = 1, const float beta = 1);
```
C would be your pre-allocated output array. After each call to this new API, C will have `aplha*(lhs x rhs) + beta*C`. What happens when `lhs` and `rhs` are batched inputs is a design choice we haven't made yet. Once this PR is merged, we will raise the new function PR and we can continue this discussion on that.

src/api/c/pinverse.cpp
```diff
@@ -129,8 +129,12 @@ Array<T> pinverseSvd(const Array<T> &in, const double tol) {
                          0, uT.dims()[2] - 1, 0, uT.dims()[3] - 1);
     }
 
-    Array<T> out = matmul<T>(matmul<T>(v, sPinv, AF_MAT_NONE, AF_MAT_NONE), uT,
-                             AF_MAT_NONE, AF_MAT_NONE);
+    // Array<T> out = matmul<T>(matmul<T>(v, sPinv, AF_MAT_NONE, AF_MAT_NONE), uT,
```
9prady9: nit: remove when PR is ready

src/api/c/blas.cpp
```diff
@@ -145,23 +145,32 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
...
         TYPE_ASSERT(lhs_type == rhs_type);
-        af_array output = 0;
+        // af_array output = 0;
+        int aRowDim = (optLhs == AF_MAT_NONE) ? 0 : 1;
+        int bColDim = (optRhs == AF_MAT_NONE) ? 1 : 0;
+        int M       = lDims[aRowDim];
+        int N       = rDims[bColDim];
+        dim_t d2    = std::max(lDims[2], rDims[2]);
+        dim_t d3    = std::max(lDims[3], rDims[3]);
+        const dim4 outDims(M, N, d2, d3);
```
9prady9: I am thinking if this output size calculation should be abstracted out into respective backends. If not, at least put in `src/backend/common` instead of repeating this code every where.

We should use something like the following
```cpp
dim4 blasOutputDims(dim4 lhsDims, dim4 rhsDims, matProp optLhs, matProp optRhs);
```

src/backend/cpu/blas.cpp
```diff
@@ -226,7 +226,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
     dim_t d2 = std::max(lDims[2], rDims[2]);
     dim_t d3 = std::max(lDims[3], rDims[3]);
     const dim4 oDims(M, N, d2, d3);
-    Array<T> out = createEmptyArray<T>(oDims);
+    // Array<T> out = createEmptyArray<T>(oDims);
```
9prady9: nit: remove once PR is ready.

src/api/c/blas.cpp
```diff
@@ -145,23 +145,32 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
         }
 
         TYPE_ASSERT(lhs_type == rhs_type);
-        af_array output = 0;
+        // af_array output = 0;
```
9prady9: nit: remove when PR is ready

src/backend/cpu/blas.cpp
```diff
@@ -308,7 +307,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
     };
     getQueue().enqueue(func, out, lhs, rhs);
 
-    return out;
+    // return out;
```
9prady9: nit: remove when PR is ready

src/backend/cuda/blas.cpp
```diff
@@ -171,7 +171,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
     dim_t d2     = std::max(lDims[2], rDims[2]);
     dim_t d3     = std::max(lDims[3], rDims[3]);
     dim4 oDims   = dim4(M, N, d2, d3);
-    Array<T> out = createEmptyArray<T>(oDims);
+    // Array<T> out = createEmptyArray<T>(oDims);
```
9prady9: nit: remove when PR is ready

src/backend/cuda/blas.cpp
```diff
@@ -241,7 +241,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
             rStrides[1], &beta, (T **)d_optrs.get(), oStrides[1], batchSize));
     }
 
-    return out;
+    // return out;
```
9prady9: nit: remove when PR is ready

src/backend/opencl/blas.cpp
```diff
@@ -68,7 +69,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
     dim_t d2     = std::max(lDims[2], rDims[2]);
     dim_t d3     = std::max(lDims[3], rDims[3]);
     dim4 oDims   = af::dim4(M, N, d2, d3);
-    Array<T> out = createEmptyArray<T>(oDims);
+    // Array<T> out = createEmptyArray<T>(oDims);
```
9prady9: nit: remove when PR is ready

src/backend/opencl/blas.cpp
```diff
@@ -104,19 +105,19 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
...
                                    lOffset, lStrides[1], (*rhs.get())(),
                                    rOffset, rStrides[1], beta, (*out.get())(),
-                                   oOffset, out.dims()[0], 1, &getQueue()(), 0,
+                                   oOffset, oStrides[1], 1, &getQueue()(), 0,
                                    nullptr, &event()));
         }
     }
 
-    return out;
+    // return out;
```
9prady9: nit: remove when PR is ready

src/backend/opencl/cpu/cpu_blas.cpp
```diff
@@ -172,7 +173,7 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
     dim4 oDims = af::dim4(M, N, d2, d3);
 
     // FIXME: Leaks on errors.
-    Array<T> out = createValueArray<T>(oDims, scalar<T>(0));
+    // Array<T> out = createValueArray<T>(oDims, scalar<T>(0));
```
9prady9: nit: remove when PR is ready

src/backend/opencl/solve.cpp
```diff
@@ -160,7 +160,7 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
                            A.getOffset(), A.strides()[1], &h_tau[0], (*dT)(),
                            tmp.getOffset(), NB, queue, &info);
 
-        B = matmul(A, B, AF_MAT_NONE, AF_MAT_NONE);
+        matmul(B, A, B, AF_MAT_NONE, AF_MAT_NONE);
```
9prady9: solve_dense test in opencl backend is failing, can you please check it.

WilliamTambellini:
ok tks @9prady9 
@mark-poscablo  Could you consequently add a unitest for accumulated batched gemm : 
Example:
A=
0 1
1 2
2 3
0 1
1 2
2 3

B=
0 1 2
1 2 3
0 1 2
1 2 3

A=3 2 2 B=2 3 2 C=3 3 beta=1
C=
2 4 6
4 10 16
6 16 26

src/api/c/blas.cpp
mark-poscablo: Thanks, this was actually one thing that I was iffy about too, but I decided that time to move on temporarily in order to address the failing tests. In any case, coming back to this issue now - my first attempt to resolve this is e9fada9 - please let me know your thoughts.

src/api/c/transform_coordinates.cpp
```diff
@@ -25,7 +25,15 @@ using namespace detail;
...
-    return matmul(lhs, createSubArray(rhs, idx), AF_MAT_NONE, AF_MAT_NONE);
+    Array<T> rhs_subarr = createSubArray(rhs, idx);
+    int M    = lhs.dims()[0];
+    int N    = rhs_subarr.dims()[1];
+    dim_t d2 = std::max(lhs.dims()[2], rhs_subarr.dims()[2]);
+    dim_t d3 = std::max(lhs.dims()[3], rhs_subarr.dims()[3]);
+    const dim4 out_dims(M, N, d2, d3);
+    Array<T> out = createEmptyArray<T>(out_dims);
+    matmul(out, lhs, rhs_subarr, AF_MAT_NONE, AF_MAT_NONE);
+    return out;
```
9prady9: Using the latest change, you can even avoid calculating output dimensions here and let matmul do the allocation.

src/backend/opencl/solve.cpp
```diff
@@ -160,7 +160,12 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
                            A.getOffset(), A.strides()[1], &h_tau[0], (*dT)(),
                            tmp.getOffset(), NB, queue, &info);
 
-        B = matmul(A, B, AF_MAT_NONE, AF_MAT_NONE);
+        int B_new_dim2 = A.dims()[2] >= B.dims()[2] ? A.dims()[2] : B.dims()[2];
+        int B_new_dim3 = A.dims()[3] >= B.dims()[3] ? A.dims()[3] : B.dims()[3];
+        Array<T> B_new = createEmptyArray<T>(dim4(A.dims()[0], B.dims()[1],
+                                                  B_new_dim2, B_new_dim3));
+        matmul(B_new, A, B, AF_MAT_NONE, AF_MAT_NONE);
```
9prady9: Using the latest change, you can even avoid calculating output dimensions here and let matmul do the allocation.

pavanky:
Why just matmul ? None of the other functions support using an existing af_array.

src/api/c/blas.cpp
```diff
@@ -27,10 +27,12 @@ static inline af_array sparseMatmul(const af_array lhs, const af_array rhs,
...
 
 template<typename T>
-static inline af_array matmul(const af_array lhs, const af_array rhs,
-                              af_mat_prop optLhs, af_mat_prop optRhs) {
-    return getHandle(
-        detail::matmul<T>(getArray<T>(lhs), getArray<T>(rhs), optLhs, optRhs));
+static inline void matmul(af_array *out,
+                          const af_array lhs, const af_array rhs,
+                          af_mat_prop optLhs, af_mat_prop optRhs) {
+    detail::matmul<T>(getArray<T>(*out),
```
pavanky: ...
This assumes `out` is already created. What happens if someone wants to create out instead of reusing an existing one ?

src/api/c/blas.cpp
```diff
@@ -145,23 +147,25 @@ af_err af_matmul(af_array *out, const af_array lhs, const af_array rhs,
         }
 
         TYPE_ASSERT(lhs_type == rhs_type);
-        af_array output = 0;
+        if (*out == 0) {
+            const dim4 outDims(0, 0, 0, 0);
+            *out = createHandle(outDims, lhs_type);
+        }
```
pavanky:
```
af_array out; // This is not zero.
af_matmul(&out, lhs, rhs, opts);
```
This is now broken. This could break python and other language wrappers because the pointers are not explicitly set to NULL.

If you want to support a way to reuse existing array for matmul, write a new function. Do not have hidden behavior this way.

mark-poscablo:
> Why just matmul ? None of the other functions support using an existing af_array.

matmul specifically because this is a part of the approach we're taking regarding #2446 (see @9prady9's comment above). We have done this also with a few other functions such as `af_approx1` (#2324)

src/api/c/blas.cpp
mark-poscablo: I think your comment below (quoting `src/api/c/blas.cpp:150`) answers this question. I'll respond to that comment shortly

src/api/c/blas.cpp
mark-poscablo: Thank you for pointing this out. This may be beside the fact, but I remember @umar456 has also said that this needs an assertion that `out` is null indeed.

In any case, @9prady9 and @umar456, thoughts on this? It's a valid point that pointers do need to be always initialized after all, and the question then now becomes: does the responsibility of initializing the af_array mainly to the user? Or to the library? Based on the [previous code](https://github.com/arrayfire/arrayfire/blob/6c28d07218045f8ab997a6b455db24a8786a3f5b/src/api/c/blas.cpp#L85) before this PR, it seems like the responsibility has been with the library for a long time now, and thus this is the expected behavior among arrayfire users (as it is still currently with other functions). Is there any other way to determine if `out` is not a pre-allocated/existing array (that covers both null and uninitialized pointers)?

WilliamTambellini: I vote for AF to init/create/allocate if the out is null/empty.

mark-poscablo:
Coming back to this after some time. Working on leaving the old behavior intact and separating the new behavior into new functions.

mark-poscablo:
@9prady9 Yes I have more coming in. Pushed some changes for now to see how they fare in different systems though.

src/backend/cpu/blas.cpp
```diff
@@ -286,29 +307,36 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
...
+                                   lDims[0], lDims[1],
+                                   alpha_.getScale(),
+                                   lptrs[n], lStrides[1], rptrs[n], incr,
+                                   beta_.getScale(),
+                                   optrs[n], oStrides[0]);
                 } else {
-                    gemm_func<T>()(CblasColMajor, lOpts, rOpts, M, N, K, alpha,
+                    gemm_func<T>()(CblasColMajor, lOpts, rOpts,
+                                   M, N, K,
+                                   alpha_.getScale(),
```
9prady9: Shouldn't the `alpha_` and `beta_` change for each iteration ? or does the `gemm_func` correctly picks ith batch set's alpha and beta accordingly ?

src/backend/opencl/blas.cpp
```diff
@@ -101,22 +98,20 @@ Array<T> matmul(const Array<T> &lhs, const Array<T> &rhs, af_mat_prop optLhs,
...
+                                   (*out.get())(), oOffset, oStrides[0], 1, &getQueue()(),
                                    0, nullptr, &event()));
         } else {
             gpu_blas_gemm_func<T> gemm;
-            OPENCL_BLAS_CHECK(gemm(lOpts, rOpts, M, N, K, alpha, (*lhs.get())(),
+            OPENCL_BLAS_CHECK(gemm(lOpts, rOpts, M, N, K, *alpha, (*lhs.get())(),
                                    lOffset, lStrides[1], (*rhs.get())(),
-                                   rOffset, rStrides[1], beta, (*out.get())(),
-                                   oOffset, out.dims()[0], 1, &getQueue()(), 0,
+                                   rOffset, rStrides[1], *beta, (*out.get())(),
```
9prady9: Same question as CPU: Shouldn't the alpha and beta change for each iteration ? or does the gemm call correctly picks i^th batch set's alpha and beta accordingly ?

test/blas.cpp
```diff
@@ -319,3 +322,246 @@ TEST(MatrixMultiply, RhsBroadcastBatched) {
         }
     }
 }
+
+////////////// Test C API using special/preallocated output arrays //////////////
+
+float alpha = 1.f;
+int num_alpha  = 1;
```
9prady9: I think `num_alpha` & `num_beta` being `1` in the tests is the reason why we didn't catch the problem of non-changing alpha & beta in CPU and OpenCL backend. See my comments in respective backend blas.cpp file.

src/backend/cpu/blas.cpp
```diff
@@ -202,55 +220,58 @@ toCblasTranspose(af_mat_prop opt) {
...
     using CBT = const typename blas_base<T>::type;
 
-    dim_t d2 = std::max(lDims[2], rDims[2]);
-    dim_t d3 = std::max(lDims[3], rDims[3]);
-    const dim4 oDims(M, N, d2, d3);
-    Array<T> out = createEmptyArray<T>(oDims);
+    auto alpha_ = scale_type<T, false>(alpha);
+    auto beta_ = scale_type<T, false>(beta);
+    auto alpha_batched = scale_type<T, true>(alpha);
+    auto beta_batched = scale_type<T, true>(beta);
```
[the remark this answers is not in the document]
mark-poscablo: I actually tried that before, but it runs the risk of the original alpha and beta that the caller passes in being undefined by the time the async lambda is executed. The results get wrong randomly if I do that. However, if I leave them outside the lambda like this, the lambda is able to capture them and it consistently gives the correct results.

src/backend/cpu/blas.cpp
mark-poscablo: Umar and I originally planned to be like that, where alpha and beta could be different for each slice in batching. However, cuBLAS's batched gemm couldn't handle that case (see https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemmbatched). Thus, to maintain behavior parity across all backends, we decided that we should just have all backends use a single alpha and beta for all batched slices.

src/backend/opencl/blas.cpp
mark-poscablo: Please see https://github.com/arrayfire/arrayfire/pull/2481#discussion_r295248645

test/blas.cpp
mark-poscablo: This was intentional because of https://github.com/arrayfire/arrayfire/pull/2481#discussion_r295248645.

src/backend/cpu/blas.cpp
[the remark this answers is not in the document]
9prady9: thats unfortunate!

src/api/unified/blas.cpp
```diff
@@ -10,6 +10,14 @@
 #include <af/blas.h>
 #include "symbol_manager.hpp"
 
+AFAPI af_err af_gemm(af_array *out,
+                     const af_mat_prop optLhs, const af_mat_prop optRhs,
+                     const void* alpha, const af_array lhs, const af_array rhs,
+                     const void* beta) {
+    CHECK_ARRAYS(out, lhs, rhs);
```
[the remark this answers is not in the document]
mark-poscablo: Addressed this in cc43610. It failed before since `out` is an `af_array*`. Now we pass in `*out`, which doesn't make `CHECK_ARRAYS` fail because the `checkArray` function it eventually calls [returns true if the array is null](https://github.com/arrayfire/arrayfire/blob/ac9119d895c6fdb21640a0cd8e7cba4bf8fac923/src/api/unified/symbol_manager.cpp#L242).

PR #2483, opened 2019-04-04
Update install tutorial with min required CUDA driver info

### LINKED ISSUE #2480: Info user on minimum drivers for CUDA toolkits used by ArrayFire binaries

9prady9:
Resolves #2480

### DIFF
--- a/docs/pages/install.md
+++ b/docs/pages/install.md
@@ -26,6 +26,27 @@ vendor’s website.
 * [Linux](#Linux)
 * [macOS](#macOS)
 
+Tabulated below are the CUDA toolkits used to generate the binaries
+available for download on our website. The last column are the
+respective minimum required NVIDIA GPU driver versions.
+
+| ArrayFire | CUDA Toolkit | Linux  | Windows   |
+|:---------:|:------------:|:------:|:---------:|
+| 3.0.x     |     7.0      | 346.46 |   347.62  |
+| 3.1-3.x   |     7.5      | 352.31 |   353.66  |
+| 3.4.0     |     7.5      | 352.31 |   353.66  |
+| 3.4.1-2   |     8.0      | 367.48 |   369.30  |
+| 3.5.0     |     8.0      | 375.26 |   376.51  |
+| 3.5.1     |     9.1      | 390.46 |   391.29  |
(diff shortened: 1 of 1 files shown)


umar456:
I don't want to have this in the documentation. ArrayFire can target multiple computes and that should be something we offer.

docs/pages/install.md
```diff
@@ -26,6 +26,27 @@ vendor’s website.
...
+| 3.1-3.x   |     7.5      | 352.31 |   353.66  |
+| 3.4.0     |     7.5      | 352.31 |   353.66  |
+| 3.4.1-2   |     8.0      | 367.48 |   369.30  |
+| 3.5.0     |     8.0      | 375.26 |   376.51  |
+| 3.5.1     |     9.1      | 390.46 |   391.29  |
+| 3.6.0     |     9.1      | 390.46 |   391.29  |
+| 3.6.1     |     9.2      | 396.37 |   398.26  |
+| 3.6.2     |    10.0      | 410.48 |   411.31  |
+
+Prior to ArrayFire 3.6.0, the user had to install CUDA toolkit on their
```
WilliamTambellini: That part is good but not the table above : I use 3.6.2 with cuda 9.2 and it works well.

9prady9:
This can be removed when we start offering multiple-computes. This helps until then.

docs/pages/install.md
9prady9: The table doesn't claim requirement of any particular CUDA toolkit versions. In fact, no where in the change it is claimed that a particular toolkit is required by a particular ArrayFire version.

More over, the line above the table explicitly says the table is about cuda toolkits used for generating binaries that we made available on our website (or archive - I didn't mention archive though).

Since, you pointed out. We can add another statement that says - all ArrayFire versions can be used against any CUDA toolkit so long as the respective cuda toolkits are hot-swappable.

WilliamTambellini: That could still easily fool the user that AF version X is to be used with cuda version Y.
Perhaps could you enhance the table with 2 columns : min cuda version and max cuda version ?

9prady9:
After some discussions, closing this in favor of adding similar information on either our binaries download page or installer.

PR #2484, opened 2019-04-07
Treat all JIT buffer parameters as non-linear parameters

### LINKED ISSUE #2436: CUDA_ERROR_INVALID_IMAGE error when running arrayfire on GPU

umar456:
This commit addresses very large kernels that were generated by the
JIT for smaller buffers. This was occuring because of how we were
estimating the JIT parameter sizes.

When targeting linear kernels we only pass the pointer to the kernel
instead of the entire dims and stride arrays. This generates smaller
parameter sets so we calculated kernels based on the pointer size
instead of the Param<T> size.

The issue occured when this kernel came in below the threshold for
evaluation but when it was used as a node to another operation, the
generated kernel became a non-linear kernel requiring us to pass
the Param<T> object for each buffer. This causes the size to go past
the limit of the parameters supported by CUDA and compilation fails.

This commit treats all kernels as non-linear kernel when calculating
the parameter sizes.

Fixes #2436 #2389

### DIFF
--- a/src/backend/cuda/Array.cpp
+++ b/src/backend/cuda/Array.cpp
@@ -241,7 +241,7 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
             // TODO: Find better solution than the following emperical solution.
             if (node->getHeight() > 25 || isBufferLimit) {
                 // This is the size of the params that are passed by default
-                constexpr int param_base_size =
+                constexpr size_t param_base_size =
                     sizeof(Param<T>) + (4 * sizeof(uint));
 
                 // This is the maximum size of the params that can be allowed by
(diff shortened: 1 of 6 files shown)


src/backend/cuda/Array.cpp
```diff
@@ -249,30 +249,25 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
                 // some_buffer_size) BUT kernels who's kernel sizes come close
                 // to this value are not passing and cuModuleLoadDataEx is
                 // failing with CUDA_ERROR_INVALID_IMAGE(200). 35*sizeof(int)
-                // seems to be the magic number that passes all tests. I have no
-                // idea why this is the case.
+                // seems to be the magic number that passes all tests.
                 constexpr int max_param_size =
```
9prady9: make this also `size_t` to be consistent ?

test/jit.cpp
```diff
@@ -421,9 +421,24 @@ TEST(JIT, ConstEval7) {
...
 
-    // I can't think of a good test for this.
     EXPECT_NO_THROW({
         eval(a, b, c, d, e, f, g);
         af::sync();
-      });
+    });
+}
+
+TEST(JIT, IncrementingBinary) {
```
9prady9: Perhaps `VeryLargeTreeEvaluation` instead of `IncrementingBinary` makes more sense as the test name.

src/backend/cuda/Array.cpp
```diff
@@ -249,30 +249,25 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
...
                     (4096 - (sizeof(Param<T>) + 35 * sizeof(uint)));
                 Node *n = node.get();
 
                 struct tree_info {
                     size_t buffer_size;
```
9prady9: This looks like is the size of all buffers under the current jit node. If that is correct, can we rename this appropriately.

umar456:
Addressed all feedback

Issue #2416, opened 2019-01-21
Use pinned memory in batched matmul


umar456:
The batched version of `matmul` requires several `cudaMemcpyAsync`'s to pass the device pointers to the GPU. This is currently being performed from a std::vector so that the operations are not asynchronous. Use pinned memory instead.

PR #2492, opened 2019-04-19
Used pinned memory to store pointers in batched gemm

umar456:
We were using vectors to store the pointers to matrices in the
blas function when performing batched operations. This caused
them to be synchronous because the memory wasn't pinned.

Fixes:  #2416


umar456:
This is failing because pinned memory is not being reallocated and used before the previous memcpy is has performed. Normally the regular memory allocator works because the GPU operations are synchronized within the same stream. This is not the case for host side pinned memory. We need to create a set of events that are tracked before the memory is reallocated to avoid this sort of thing.

PR #2495, opened 2019-04-25
New Window method: setAxesLabelFormat

### LINKED ISSUE #2006: Add Window method to change axes label format
Originally reported in upstream graphics library, Forge. https://github.com/arrayfire/forge/issues/152

Code used to reproduce the issue.
```
#include <arrayfire.h>
{
    try {
        af::info();
        af::array X = af::seq(0., 0.4, 0.1);
        af::array Y = af::seq(0., 0.9, 0.1);
        af::array Z = af::constant(1., 5, 10);
        af::Window window(1500, 800, "The door.");
        do {
            window.surface(X, Y, Z, "The weird surface");
        } while(!window.close());
    } catch (af::exception& e) {
        throw;
    }
    return 0;
}
```

Current work around to for the issue is to explicitly set the limits as follows.
` window.setAxesLimits(0.0f, 0.4f, 0.0f, 0.9f, -1.0f, 1.0f, true);`

9prady9:
This member function of af::Window class enables the user
to set the format in which chart axes labels are generated.

The formats for each axis are specified as a printf-style
format specifier.

Resolves #2006

### DIFF
--- a/include/af/graphics.h
+++ b/include/af/graphics.h
@@ -496,9 +496,9 @@ class AFAPI Window {
 
            \ingroup gfx_func_window
         */
-        void setAxesLabelFormat(const char *const xtitle = "4.1%f",
-                                const char *const ytitle = "4.1%f",
-                                const char *const ztitle = NULL);
+        void setAxesLabelFormat(const char *const xformat = "4.1%f",
+                                const char *const yformat = "4.1%f",
+                                const char *const zformat = NULL);
 #endif
 
         /**
--- a/include/af/graphics.h
+++ b/include/af/graphics.h
@@ -485,6 +485,22 @@ class AFAPI Window {
                            const char * const ytitle = "Y-Axis",
                            const char * const ztitle = NULL);
 #endif
+
+#if AF_API_VERSION >= 37
+        /**
+           Setup the axes label formats for charts
+
+           \param[in] xformat is a printf-style format specifier for x-axis
+           \param[in] yformat is a printf-style format specifier for y-axis
+           \param[in] zformat is a printf-style format specifier for z-axis
+
+           \ingroup gfx_func_window
+        */
+        void setAxesLabelFormat(const char *const xtitle = "4.1%f",
+                                const char *const ytitle = "4.1%f",
+                                const char *const ztitle = NULL);
+#endif
+
         /**
            Setup grid layout for multiview mode in a window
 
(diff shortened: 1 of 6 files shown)


include/af/graphics.h
```diff
@@ -485,6 +485,23 @@ class AFAPI Window {
                            const char * const ytitle = "Y-Axis",
                            const char * const ztitle = NULL);
 #endif
+
+#if AF_API_VERSION >= 37
+        /**
+           Setup the axes label formats for charts
+
+           \param[in] xformat is a printf-style format specifier string for
```
umar456: I think the formatting was messed up by clang format.

include/af/graphics.h
```diff
@@ -1088,6 +1105,33 @@ AFAPI af_err af_set_axes_titles(const af_window wind,
                                 const af_cell* const props);
 #endif
 
+#if AF_API_VERSION >= 37
+/**
+   C Interface wrapper for setting axes labels formats for 2d charts
+
+   Passing correct value to \p zformat dictates the right behavior when it comes
```
umar456: I think you should have a better description and then add that note at the bottom with the `\note` tag.

src/api/c/window.cpp
```diff
@@ -221,6 +221,37 @@ af_err af_set_axes_titles(const af_window window, const char* const xtitle,
     return AF_SUCCESS;
 }
 
+af_err af_set_axes_label_format(const af_window window,
+                                const char* const xformat,
+                                const char* const yformat,
+                                const char* const zformat,
+                                const af_cell* const props) {
+    try {
+        if (window == 0) { AF_ERROR("Not a valid window", AF_ERR_INTERNAL); }
```
umar456: Check that the arguments are not null before using them.

src/api/c/window.cpp
```diff
@@ -221,6 +221,37 @@ af_err af_set_axes_titles(const af_window window, const char* const xtitle,
...
+
+        ForgeManager& fgMngr = forgeManager();
+
+        fg_chart chart = NULL;
```
umar456: nit: nullptr

include/af/graphics.h
```diff
@@ -485,6 +485,22 @@ class AFAPI Window {
...
+
+           \param[in] xformat is a printf-style format specifier for x-axis
+           \param[in] yformat is a printf-style format specifier for y-axis
+           \param[in] zformat is a printf-style format specifier for z-axis
+
+           \ingroup gfx_func_window
+        */
+        void setAxesLabelFormat(const char *const xtitle = "4.1%f",
+                                const char *const ytitle = "4.1%f",
+                                const char *const ztitle = NULL);
```
mark-poscablo: The argument names here doesn't match with the definition in src/api/cpp/graphics.cpp:192, and also those in the docstring. The CI's docs check also complains about it. I think you meant `x/y/zformat` here?

PR #2499, opened 2019-04-26
Cleanup platform headers in all backends

9prady9:


src/backend/opencl/platform.cpp
```diff
@@ -37,16 +37,12 @@
...
 #include <string>
 #include <thread>
 #include <utility>
 #include <vector>
 
-using std::ostringstream;
-using std::runtime_error;
-using std::string;
-using std::vector;
+using namespace std;
```
umar456: Never use `using namespace std`. Always define each symbol you want to introduce into the file. Even if it is a cpp file.

src/backend/opencl/device_manager.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <common/MemoryManager.hpp>
+#include <platform.hpp>
+
+// Forward declaration from clFFT.h
+struct clfftSetupData_;
+using clfftSetupData = clfftSetupData_;
```
umar456: Are these supposed to be defined in the global namespace?

src/backend/opencl/device_manager.cpp
```diff
@@ -0,0 +1,442 @@
...
+#include <boost/compute/context.hpp>
+#include <boost/compute/utility/program_cache.hpp>
+
+#include <algorithm>
+#include <iterator>
+#include <sstream>
+#include <string>
+#include <vector>
+
+using namespace std;
```
umar456: Please remove `using namespace std;`

src/backend/cuda/device_manager.cpp
```diff
@@ -0,0 +1,506 @@
...
+#include <cstdio>
+#include <memory>
+#include <mutex>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <thread>
+#include <vector>
+
+using namespace std;
```
umar456: Please remove `using namespace std;`

src/backend/cuda/cu_handles.hpp
```diff
@@ -0,0 +1,20 @@
...
+ ********************************************************/
+
+#pragma once
+
+#include <cublas_v2.h>
+#include <cusolverDn.h>
+#include <cusparse_v2.h>
+
+namespace cuda {
+using BlasHandle = cublasHandle_t;
```
umar456: Why not just define them in the cublas.hpp file like earlier? These are handles from distinct libraries why mix them if you don't have to?

src/backend/cpu/device_manager.cpp
```diff
@@ -0,0 +1,137 @@
...
+ ********************************************************/
+
+#include <common/graphics_common.hpp>
+#include <device_manager.hpp>
+#include <af/version.h>
+
+#include <cctype>
+#include <sstream>
+
+using namespace std;
```
umar456: Please remove `using namespace std;`

src/backend/cpu/device_manager.cpp
```diff
@@ -0,0 +1,137 @@
...
+#else
+
+CPUInfo::CPUInfo()
+    : mVendorId("")
+    , mModelName("")
+    , mNumSMT(0)
+    , mNumCores(0)
+    , mNumLogCpus(0)
+    , mIsHTT(false) {
+    mVendorId   = "Unknown";
```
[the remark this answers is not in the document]
9prady9: I didn't notice it, was just moving things around. Fixed it.

src/backend/opencl/platform.hpp
```diff
@@ -114,102 +113,23 @@ GraphicsResourceManager& interopManager();
...
-    std::vector<cl::Context*> mContexts;
-    std::vector<cl::CommandQueue*> mQueues;
-    std::vector<bool> mIsGLSharingOn;
-    std::vector<int> mDeviceTypes;
-    std::vector<int> mPlatforms;
-    unsigned mUserDeviceOffset;
+static afcl::platform getPlatformEnum(cl::Device dev) {
+    std::string pname = getPlatformName(dev);
+    if (verify_present(pname, "AMD")) return AFCL_PLATFORM_AMD;
+    if (verify_present(pname, "NVIDIA")) return AFCL_PLATFORM_NVIDIA;
```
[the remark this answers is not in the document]
9prady9: Added elses. this function is required in DeviceManager constructor as well as another backend API from platform.hpp. Hence, I have it here.

src/backend/opencl/device_manager.hpp
9prady9: May be the `using` can be moved inside `DeviceManager` class. But the the other forward declaration is from clfft. We can't move it.

src/backend/cuda/cu_handles.hpp
9prady9: These handles are `cu*` libraries. I can declare them like I did `clfft` stuff, let me check.

src/backend/cuda/device_manager.cpp
```diff
@@ -0,0 +1,512 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
umar456: 2019

src/backend/opencl/device_manager.cpp
```diff
@@ -0,0 +1,446 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
umar456: 2019

PR #2501, opened 2019-04-27
Revert  "Move parameters to shared memory commit #2487"

umar456:
Revert the Move parameters to shared memory commit #2487 and reduce the threshold to for max parameter size.


9prady9:
@umar456 Can you please fix the conflict, I think this just happened after my last two PR merges.

test/jit.cpp
```diff
@@ -124,10 +125,15 @@ TEST(JIT, CPP_Multi_linear) {
     x.host(&hx[0]);
     y.host(&hy[0]);
```
mark-poscablo: Looks like you don't need the `hx` and `hy` vectors anymore

src/backend/cpu/Array.cpp
```diff
@@ -224,29 +250,7 @@ Array<T> createNodeArray(const dim4 &dims, Node_ptr node) {
...
-                               [=](const size_t prev, const Node &n) {
-                                   // getBytes returns the size of the data
-                                   // Array. Sub arrays will be represented by
-                                   // their parent size.
-                                   return prev + n.getBytes();
-                               });
-
-                if (2 * bytes > lock_bytes) { out.eval(); }
-            }
+            if (!passesJitHeuristics<T>(node.get())) { out.eval(); }
```
9prady9: Is there a specific reason, this if block isn't condensed further ?

src/backend/cuda/Array.hpp
```diff
@@ -36,6 +36,10 @@ void evalNodes(std::vector<Param<T>> &out, std::vector<common::Node *> nodes);
 template<typename T>
 void evalMultiple(std::vector<Array<T> *> arrays);
 
+template<typename T, typename Ti, typename FUNC>
+Array<T> createNodeArray(const dim4 &odims, FUNC createNode,
+                         const Array<Ti> *children, int num_children);
+
```
9prady9: As per our slack discussion, this is just a left over. Just leaving a comment as a reminder to remove this.

PR #2510, opened 2019-05-11
Add support for subprojects in CDash with labels

umar456:


umar456:
ArrayFire doesn't work with CMake Labels because our project makes use of the same source files amongst multiple targets. This causes some problems assigning the warnings and errors in CDash.

PR #2519, opened 2019-05-22
Fix convolve3 launch configuration in CUDA backend

9prady9:
Batched input of Convolve3 was incorrectly folding batch size
into CUDA launch grid's y & z dimensions. Since the batch for 3d
inputs can happen along 4th dimension only, folding it onto x dimesion
of CUDA grid is sufficient.

Have to merge https://github.com/arrayfire/arrayfire-data/pull/14 before this is merged.

### DIFF
--- a/docs/details/signal.dox
+++ b/docs/details/signal.dox
@@ -19,21 +19,8 @@ batch mode convolutions take place.
 - **Identical Batches** - A set of filters applied onto to a set of inputs in one-to-one correspondence.
 - **Non overlapping Batches** - All batched filters are applied to all batched signals. The batch dimension of Signal and Filter **should not** be the same.
 
-
-
-\page signal_func_conv2_batch_desc convolve2
-
-For example, if the signal is two dimensional with m & n as sizes along the 0th & 1st dimensions
-respectively, then the possible batch operations are as follows.
-
-| Input Signal Dimensions | Filter Dimensions | Output Dimensions | Batch Mode | Explanation |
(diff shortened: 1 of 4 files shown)


syurkevi:
Would it be possible to update the convolve3 documentation to describe the expected batching behavior?

9prady9:
@syurkevi We are already explain the batching behavior for all convolve functions in our [documentation](http://arrayfire.org/docs/group__signal__func__convolve3.htm). This change is an internal fix for CUDA kernel launch configuration and doesn't change the external batching behavior.

syurkevi:
> @syurkevi We are already explain the batching behavior for all convolve functions in our [documentation](http://arrayfire.org/docs/group__signal__func__convolve3.htm). This change is an internal fix for CUDA kernel launch configuration and doesn't change the external batching behavior.

The documentation you linked is copy-pasted from convolve2 and doesn't apply to convolve3. It would be nice to update the correct batching behavior to the documentation as it is difficult to understand what it should be without reading the source code.

9prady9:
Screenshot from 2019-05-30 20-37-33
Screenshot from 2019-05-30 20-37-38
Screenshot from 2019-05-30 20-37-59
Screenshot from 2019-05-30 20-38-12
Screenshot from 2019-05-30 20-38-21
Screenshot from 2019-05-30 20-38-27

9prady9:
@arrayfire/core-devel Alright, here's the new updated documentation for convolution. have a look and share feedback pls.

PR #2520, opened 2019-05-22
Address issues with modified inputs for fftC2R functions.

### LINKED ISSUE #2518: in v3.6.4 c2r FFT changes the value of the input
While testing `ArrayFire.jl` errors show up with FFT tests for 3.6.4, for some reason it changes the values of inputs, is it expected? Last tested 3.6.2 and everything was fine.
```
julia> a = rand(AFArray{Complex{Float32}}, 2, 3)
AFArray: 2×3 Array{Complex{Float32},2}:
 0.314803+0.368697im  0.376258+0.269053im  0.228255+0.379484im
 0.961944+0.929568im  0.389558+0.223507im   0.64615+0.835269im

julia> irfft2(a, 1/2, true); # af_fft2_c2r

julia> a # note that the values have changed
AFArray: 2×3 Array{Complex{Float32},2}:
 0.919316+1.01723im  0.108182+0.172603im  -0.0830895-0.0837456im
  1.99765+1.98834im   0.97389+0.177964im  -0.0857113+0.622396im 
```

The same problem happens with af_fft3_c2r

umar456:
The fftC2R function was modifying the input array when the fftw library
was used. This was happening because the fftw library does not have
and algorithm that can compute the fft without modifying the input array.

This problem does not seem to be happening on the MKL based fft.

Fixes #2518 For the CPU backend. I haven't been able to reproduce this with
the cuda backend.

### DIFF
--- a/src/backend/cpu/kernel/fft.hpp
+++ b/src/backend/cpu/kernel/fft.hpp
@@ -121,7 +121,7 @@ void fft_r2c(Param<Tc> out, const af::dim4 oDataDims, CParam<Tr> in,
     plan = transform.create(rank, t_dims, (int)batch, (Tr *)in.get(), in_embed,
                             (int)istrides[0], (int)istrides[rank],
                             (ctype_t *)out.get(), out_embed, (int)ostrides[0],
-                            (int)ostrides[rank], FFTW_ESTIMATE);
+                            (int)ostrides[rank], FFTW_ESTIMATE | FFTW_PRESERVE_INPUT);
 
     transform.execute(plan);
     transform.destroy(plan);
--- a/test/fft.cpp
+++ b/test/fft.cpp
@@ -6,12 +6,15 @@
  * The complete license agreement can be obtained at:
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
+#define GTEST_LINKED_AS_SHARED_LIBRARY 1
 
 #include <arrayfire.h>
 #include <gtest/gtest.h>
 #include <testHelpers.hpp>
 #include <af/dim4.hpp>
 #include <af/traits.hpp>
+
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -41,6 +44,7 @@ using af::span;
 using std::abs;
 using std::endl;
 using std::string;
+using std::stringstream;
 using std::vector;
 
 TEST(fft, Invalid_Type) {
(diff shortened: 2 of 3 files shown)


src/backend/cpu/kernel/fft.hpp
```diff
@@ -121,7 +121,7 @@ void fft_r2c(Param<Tc> out, const af::dim4 oDataDims, CParam<Tr> in,
     plan = transform.create(rank, t_dims, (int)batch, (Tr *)in.get(), in_embed,
                             (int)istrides[0], (int)istrides[rank],
                             (ctype_t *)out.get(), out_embed, (int)ostrides[0],
-                            (int)ostrides[rank], FFTW_ESTIMATE);
+                            (int)ostrides[rank], FFTW_ESTIMATE | FFTW_PRESERVE_INPUT);
```
gaika: Is this flag alone enough to fix the problem? Why copy the array yet again?

umar456: This only works in real to complex ffts. The complex 2 real function will return a NULL plan because FFTW does not have an algorithm for this.

> FFTW_PRESERVE_INPUT specifies that an out-of-place transform must not change its input array. This is ordinarily the default, except for c2r and hc2r (i.e. complex-to-real) transforms for which FFTW_DESTROY_INPUT is the default. In the latter cases, passing FFTW_PRESERVE_INPUT will attempt to use algorithms that do not destroy the input, at the expense of worse performance; for multi-dimensional c2r transforms, however, no input-preserving algorithms are implemented and the planner will return NULL if one is requested.

gaika: Yikes! I'm so glad ArrayFire has separate functions for in-place transforms and it is clear from the name of the function of what to expect.

umar456:
Removed tests that were failing on CUDA 10.1. Should be added back once the bug is resolved.

test/fft.cpp
```diff
@@ -682,82 +686,232 @@ TEST(fft3, GFOR) {
...
+INSTANTIATE_TEST_CASE_P(InputsND, FFTC2R,
+                        ::testing::Values(
+                            fft_params(dim4(513), false, 0.5),
+                            fft_params(dim4(1025), false, 0.5),
+                            fft_params(dim4(1025, 1024), false, 0.5),
+                            fft_params(dim4(1025, 1024, 3), false, 0.5)),
+                        to_test_params);
+
+// Does not work well with CUDA 10.1
+// TEST_P(FFTC2R2D, Complex32ToRealInputsPreserved) {
```
9prady9: Not a major requirement, we should lean towards using `DISABLED_` prefix for the test so that it is reported as disabled in ctest suite runs.

PR #2521, opened 2019-05-23
Update k-means example to show graphical result

mark-poscablo:
The k-means example has a non-working code stub that looks like it was supposed to be updated once arrayfire had the functionality to show windows. This change simply converts that code stub into a working implementation.

Here is a screenshot of the result (for k = 16, which is the example's default setting):
image

### DIFF
--- a/examples/machine_learning/kmeans.cpp
+++ b/examples/machine_learning/kmeans.cpp
@@ -113,7 +113,7 @@ int kmeans_demo(int k, bool console) {
     printf("** ArrayFire K-Means Demo (k = %d) **\n\n", k);
 
     array img =
-        loadImage(ASSETS_DIR "/examples/images/vegetable-woman.jpg", true) /
+        loadImage(ASSETS_DIR "/examples/images/spider.jpg", true) /
         255;  // [0-255]
 
     int w = img.dims(0), h = img.dims(1), c = img.dims(2);
@@ -133,16 +133,16 @@ int kmeans_demo(int k, bool console) {
         array out_half = moddims(means_half(span, clusters_half, span), img.dims());
         array out_dbl  = moddims(means_dbl (span, clusters_dbl , span), img.dims());
 
-        af::Window wnd("ArrayFire K-Means Demo");
-        wnd.grid(1, 4);
+        af::Window wnd(800, 800, "ArrayFire K-Means Demo");
+        wnd.grid(2, 2);
         std::string out_full_caption = "k = " + std::to_string(k);
         std::string out_half_caption = "k = " + std::to_string(k / 2);
         std::string out_dbl_caption = "k = " + std::to_string(k * 2);
         while (!wnd.close()) {
             wnd(0, 0).image(img, "Input Image");
             wnd(0, 1).image(out_full, out_full_caption.c_str());
-            wnd(0, 2).image(out_half, out_half_caption.c_str());
-            wnd(0, 3).image(out_dbl, out_dbl_caption.c_str());
+            wnd(1, 0).image(out_half, out_half_caption.c_str());
+            wnd(1, 1).image(out_dbl, out_dbl_caption.c_str());
             wnd.show();
         }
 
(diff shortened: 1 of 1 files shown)


9prady9:
Thanks for reviving this good example :+1: 

Perhaps, we should use a smaller image size so that the example runs relatively fast on lower level NVIDIA/AMD GPUs.

I have tried few images from our examples-images-assets, `spider.jpg`(512x512) does perform faster and the banding that appears in the visual display after clustering is more evident with this image. I have changed the layout to 2x2 grid with 1024x1024 as window size to have better image display. Checkout the spider output with following changes and let me know.

```diff
diff --git a/examples/machine_learning/kmeans.cpp b/examples/machine_learning/kmeans.cpp
index 81f1503a..e3a80018 100644
--- a/examples/machine_learning/kmeans.cpp
+++ b/examples/machine_learning/kmeans.cpp
@@ -113,7 +113,7 @@ int kmeans_demo(int k, bool console) {
     printf("** ArrayFire K-Means Demo (k = %d) **\n\n", k);
 
     array img =
-        loadImage(ASSETS_DIR "/examples/images/vegetable-woman.jpg", true) /
+        loadImage(ASSETS_DIR "/examples/images/spider.jpg", false) /
         255;  // [0-255]
 
     int w = img.dims(0), h = img.dims(1), c = img.dims(2);
@@ -133,16 +133,16 @@ int kmeans_demo(int k, bool console) {
         array out_half = moddims(means_half(span, clusters_half, span), img.dims());
         array out_dbl  = moddims(means_dbl (span, clusters_dbl , span), img.dims());
 
-        af::Window wnd("ArrayFire K-Means Demo");
-        wnd.grid(1, 4);
+        af::Window wnd(1024, 1024, "ArrayFire K-Means Demo");
+        wnd.grid(2, 2);
         std::string out_full_caption = "k = " + std::to_string(k);
         std::string out_half_caption = "k = " + std::to_string(k / 2);
         std::string out_dbl_caption = "k = " + std::to_string(k * 2);
         while (!wnd.close()) {
             wnd(0, 0).image(img, "Input Image");
             wnd(0, 1).image(out_full, out_full_caption.c_str());
-            wnd(0, 2).image(out_half, out_half_caption.c_str());
-            wnd(0, 3).image(out_dbl, out_dbl_caption.c_str());
+            wnd(1, 0).image(out_half, out_half_caption.c_str());
+            wnd(1, 1).image(out_dbl, out_dbl_caption.c_str());
             wnd.show();
         }
 
 ``

**Comment by mark-poscablo** [comment]
@9prady9 Thanks. I tried the spider example, and it looks like it's being processed faster indeed on my laptop, I assume largely because it is monochrome. I also think changing the grid layout to 2x2 is good, but I think it's better to have the window auto-sized because 1024 x 1024 fills pretty much the whole vertical space on common full HD screens (1920 x 1080). Lastly I think reducing k from 16 to 8 is good in terms of processing time and just being able to make the banding more obvious.

image

**Comment by 9prady9** [comment]
Yes, it is three times less data in a monochrome image. Sure, you can reduce it to 800x800 and cluster parameter to 8. The reason I suggested smaller image is since CUDA/OpenCL takes so long, on CPU backend the example will seem like it is running forever.

File: examples/machine_learning/kmeans.cpp [hunk 1]
```diff
@@ -112,7 +113,7 @@ int kmeans_demo(int k, bool console) {
     printf("** ArrayFire K-Means Demo (k = %d) **\n\n", k);
 
     array img =
-        loadImage(ASSETS_DIR "/examples/images/vegetable-woman.jpg", true) /
+        loadImage(ASSETS_DIR "/examples/images/spider.jpg", true) /
```
9prady9: the `isColor` is still true, this would result in an image with three channels with identical values in each channel. Can you make this `false`.

mark-poscablo: Oops yes, thanks for catching that.

PR #2526, opened 2019-05-31
Enforce dependencies between allocs and frees using events

umar456:
* Creates an event class which is based on cudaEvent_t and OpenCL
  Events
* Use these events to enforce dependencies between frees and allocs
  to make sure that the data is not overwritten in later operations.
  This is not really an issue now because we are only running ops on
  one stream, but this will be an issue when we move to multiple streams
* Track MemoryEvent pairs in the memory manager.

### DIFF
--- a/src/api/c/random.cpp
+++ b/src/api/c/random.cpp
@@ -27,9 +27,7 @@ using namespace common;
 using af::dim4;
 
 Array<uint> emptyArray() {
-    static const Array<uint> EMPTY_ARRAY = createEmptyArray<uint>(af::dim4(0));
-
-    return EMPTY_ARRAY;
+    return createEmptyArray<uint>(af::dim4(0));
 }
 
 struct RandomEngine {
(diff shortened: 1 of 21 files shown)


9prady9:
Depends on https://github.com/arrayfire/threads/pull/3

src/api/c/random.cpp
```diff
@@ -110,13 +106,14 @@ static void validateRandomType(const af_random_engine_type type) {
...
+}
 
 af_err af_get_default_random_engine(af_random_engine *r) {
     try {
         AF_CHECK(af_init());
 
-        thread_local RandomEngine re;
-        *r = static_cast<af_random_engine>(&re);
+        thread_local RandomEngine *re = new RandomEngine;
+        *r = static_cast<af_random_engine>(re);
```
9prady9: Why are we creating a intentional leak here now ?

umar456: Yeah, its causing problems on Windows. Sometimes the destructor is called after the CUDA driver is unloaded and it causes segfaults on exit.

PR #2528, opened 2019-06-01
Adjust max CUDA kernel parameter size, fix #2527

### LINKED ISSUE #2527: CUDA_ERROR_INVALID_IMAGE(200): device kernel image is invalid
I'm running into:

```
ERROR: ArrayFire Error (998) : Internal error
In function cuda::Kernel cuda::buildKernel(int, const string&, const string&, const std::vector<std::__cxx11::basic_string<char> >&, bool)
In file src/backend/cuda/nvrtc/cache.cpp:191
CU Error CUDA_ERROR_INVALID_IMAGE(200): device kernel image is invalid

ArrayFire v3.7.0 (CUDA, 64-bit Linux, build 093efe58)
Platform: CUDA Toolkit 10.1, Driver: 418.67
[0] GeForce RTX 2080 Ti, 10987 MB, CUDA Compute 7.5
```

Related - https://github.com/arrayfire/arrayfire/issues/2436 - setting AF_CUDA_MAX_JIT_LEN=10 also fixes the issue for me.

How do I debug it further?

gaika:


gaika:
I can tweak parameters in my code and it is failing even with 96 byte adjustment - just takes a lot longer to trigger.

umar456:
Is there any way I can run your code?

gaika:
It is in Julia, let me get a minimal example that triggers this.

PR #2535, opened 2019-06-13
Dump debug symbols into separate files from binaries

9prady9:
Dump debug symbols into separate files from binaries
    
    * Linux systems will created separate `lib<target>`.debug
      files with debug symbols of respective backend binaries.
    * Windows by default splits debug symbols into pdb files.
    * OSX debug symbol files are not generated - DISABLED for now
    
    OSX support will be added later.
    
    Component based installers will have *_debug_symbols component(s)
    disabled(not-installed) by default. The user has to explicitly select
    them during installation. The idea is to skip installation on these
    components on deployment systems and have them installed on development
    systems.

### DIFF
--- a/CMakeModules/SplitDebugInfo.cmake
+++ b/CMakeModules/SplitDebugInfo.cmake
@@ -1,0 +1,94 @@
+# Tailored after https://github.com/GerbilSoft/mcrecover/blob/master/cmake/macros/SplitDebugInformation.cmake
+# Minor modifications to original
+
+if (NOT WIN32)
+  include(CMakeFindBinUtils)
+  if (NOT APPLE AND NOT CMAKE_OBJCOPY)
+    message(WARNING "'objcopy' tool not found; debug information will not be split.")
+  elseif (NOT CMAKE_STRIP)
+    message(WARNING "'strip' tool not found; debug information will not be split.")
+  elseif (APPLE)
+    # TODO(pradeep) debug info splits on OSX are disabled
+    # this section of elseif will be removed when Apple support is added
+    message(WARNING "Debug information is not split on OSX")
+  endif ()
+endif (NOT WIN32)
+
+function(af_split_debug_info _target)
+  set(SPLIT_TOOL_EXISTS ON)
+  if (WIN32)
+    set(SPLIT_TOOL_EXISTS OFF)
+    if (MSVC)
+      install(FILES
+        $<TARGET_PDB_FILE:${_target}>
+        DESTINATION ${AF_INSTALL_LIB_DIR}
+        COMPONENT "${_target}_debug_symbols"
+        )
+    endif()
+  elseif (NOT APPLE AND NOT CMAKE_OBJCOPY)
+    set(SPLIT_TOOL_EXISTS OFF)
+  elseif (NOT CMAKE_STRIP)
+    set(SPLIT_TOOL_EXISTS OFF)
+  elseif (APPLE)
+    # TODO(pradeep) debug info splits on OSX are disabled
+    # this section of elseif will be removed when Apple support is added
+    set(SPLIT_TOOL_EXISTS OFF)
+  endif ()
+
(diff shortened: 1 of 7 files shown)


CMakeModules/SplitDebugInfo.cmake
```diff
@@ -0,0 +1,88 @@
...
+  elseif (NOT APPLE AND NOT CMAKE_OBJCOPY)
+    set(SPLIT_TOOL_EXISTS OFF)
+  elseif (NOT CMAKE_STRIP)
+    set(SPLIT_TOOL_EXISTS OFF)
+  endif ()
+
+  if (SPLIT_TOOL_EXISTS)
+    get_target_property(TARGET_TYPE ${_target} TYPE)
+    set(PREFIX_EXPR_1
+      "$<$<STREQUAL:$<TARGET_PROPERTY:${_target},PREFIX>,>:${CMAKE_${TARGET_TYPE}_PREFIX}>")
```
umar456: I think it would be cleaner if this wasn't a generator expression. Generator expressions are usually only useful for IDEs such as visual studio and xcode where you need to change behavior based on the configuration. It doesn't looks like you need that here.

CMakeModules/SplitDebugInfo.cmake
```diff
@@ -0,0 +1,88 @@
+# Tailored after https://github.com/GerbilSoft/mcrecover/blob/master/cmake/macros/SplitDebugInformation.cmake
+# Minor modifications to original
+
+if (NOT MSVC)
+  include(CMakeFindBinUtils)
+  if (NOT APPLE AND NOT CMAKE_OBJCOPY)
```
umar456: You can probably use find_program to find a different program on apple. I think there are g versions of equivalent programs available on brew.

CMakeModules/SplitDebugInfo.cmake
```diff
@@ -0,0 +1,88 @@
...
+    set(SPLIT_TOOL_EXISTS OFF)
+    if(AF_PACK_DEBUG_SYMBOLS AND WIN32)
+      install(FILES
+        $<TARGET_PDB_FILE:${_target}>
+        DESTINATION ${AF_INSTALL_LIB_DIR}
+        COMPONENT "${_target}_debug_symbols"
+        )
+    endif()
+  elseif (NOT APPLE AND NOT CMAKE_OBJCOPY)
+    set(SPLIT_TOOL_EXISTS OFF)
```
umar456: These two else statements do the same thing. is that intended?

umar456:
I will test this on OSX sometime later.

CMakeModules/SplitDebugInfo.cmake
9prady9: gnu binutils installed by brew can end in places where find_program doesn't look for by default. This happened on jesse(OSX system) too. Hence, I am using dsymutil on APPLE, which is guaranteed to be available from what I read so far.

src/backend/cpu/CMakeLists.txt
```diff
@@ -352,6 +352,8 @@ if(LAPACK_FOUND OR MKL_FOUND)
       WITH_LINEAR_ALGEBRA)
 endif()
 
+af_split_debug_info(afcpu)
```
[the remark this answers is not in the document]
9prady9: Sure

CMakeModules/SplitDebugInfo.cmake
9prady9: I have checked these and this procedure is what works reliably consistently across Windows, OSX and Linux.

CMakeModules/SplitDebugInfo.cmake
9prady9: Yes, we are looking for OBJCOPY only on Linux style distributions.

9prady9:
I have updated few things and remove unnecessary stuff. Will push soon.

PR #2540, opened 2019-06-15
Sets flags to enforce c++ compliance and correct cplusplus definition

umar456:
* Sets the "permissive-" flag which instructs the compiler to enforce
  strict compliance to the standard
* Sets the Zc:__cplusplus flag which corrects the __cplusplus defintion
  so that move constructors are enabled in MSVC

Required for #2537


umar456:
Merged with #2537

PR #2542, opened 2019-06-19
Fix queue on the CPU backend. Remove evals

umar456:


src/backend/cpu/queue.hpp
```diff
@@ -56,7 +56,7 @@ class queue {
                      getEnvVar("AF_SYNCHRONOUS_CALLS") == "1") {}
 
     template<typename F, typename... Args>
-    void enqueue(const F func, Args... args) {
+    void enqueue(const F func, Args&&... args) {
         count++;
         if (sync_calls) {
             func(toParam(args)...);
```
9prady9: Why aren't we forwarding args using `std::forward<Args>(toParam(args))...` ? Is it because `toParam` is creating the rvalue out of args for us ?

umar456: it should be
toParam(std::forward<Args>(args))

umar456: Addressed all feedback

PR #2544, opened 2019-06-19
Use correct reduce op type for boolean product

### LINKED ISSUE #2543: Product of b8 uses sum
It seems taking the product of a bit vector (`b8` array) is summing the elements. To reproduce:

```c
void test_bool () {
  bool data[5] = { 1,1,1,1,1 };                      
  dim_t * dims = malloc(sizeof(dim_t) * 4);
  dims[0] = 5;                                               
  dims[1] = 1;                                               
  dims[2] = 1;                                               
  dims[3] = 1;                                               
  af_array arrin;                                            
  af_create_array(&arrin, &data, 4, dims, b8);
  af_print_array(arrin);                                      
  af_array arrout;                                              
  af_product(&arrout, arrin, 0);                        
  printf("printing output array\n");                  
  af_print_array(arrout);                                   
} 
```
Expected:
```
printing output array
No Name Array
[1 1 1 1]
         1
```
Actual:
```
printing output array
No Name Array
[1 1 1 1]
         5 
```

9prady9:
Fixes #2543

### DIFF
--- a/test/reduce.cpp
+++ b/test/reduce.cpp
@@ -685,3 +685,22 @@ TEST(Reduce, AllSmallIndexed) {
     array b       = a(seq(len / 2), span);
     ASSERT_EQ(max<float>(b), len / 2 - 1);
 }
+
+TEST(ProductAll, BoolIn_ISSUE2543_All_Ones) {
+    ASSERT_EQ(true, product<int>(constant(1, 5, 5, b8)) > 0);
+}
+
+TEST(ProductAll, BoolIn_ISSUE2543_Random_Values) {
+    array in = randu(5, 5, b8);
+    vector<char> hostData(25);
+    in.host(hostData.data());
+    unsigned int gold = 1;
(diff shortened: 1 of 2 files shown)


test/reduce.cpp
```diff
@@ -685,3 +685,22 @@ TEST(Reduce, AllSmallIndexed) {
     array b       = a(seq(len / 2), span);
     ASSERT_EQ(max<float>(b), len / 2 - 1);
 }
+
+TEST(ProductAll, BoolIn_ISSUE2543_All_Ones) {
+    ASSERT_EQ(true, product<int>(constant(1, 5, 5, b8)) > 0);
```
umar456: Shouldn't product return char?

Wouldn't the Result be one? In that case use

```

```
```suggestion
    ASSERT_EQ(1, product<char>(constant(1, 5, 5, b8));
```

PR #2549, opened 2019-06-20
Build Release configs for OCL upstream dependencies

9prady9:


9prady9:
Closing in favor of keeping them as is. In fact, forge build type should be switched back to match these upstreams.

PR #2558, opened 2019-06-28
Improve the indexing tutorial

umar456:
Adding more details to the indexing tutorials.

image

### DIFF
--- a/docs/pages/indexing.md
+++ b/docs/pages/indexing.md
@@ -33,7 +33,7 @@ matrix by passing `0,0` into the parenthesis operator of the af::array.
 
 \snippet test/index.cpp index_tutorial_first_element
 
-\f[ A(0, 0) = [ 0 ] \f]
+\f[ A(2, 3) = [ 14 ] \f]
 
 We can also access the array using linear indexing by passing in one value. Here
 we are accessing the fifth element of the array.
@@ -42,16 +42,23 @@ we are accessing the fifth element of the array.
 
 \f[ A(5) = [ 5 ] \f]
 
-Normally you want to avoid accessing individual elements of the array like this
+\note Normally you want to avoid accessing individual elements of the array like this
 for performance reasons.
 
-## af::span and af::seq
+Negative indexing values will index from the end of the array. For example, the
+value negative one and negative two(-2) will return the last and second to last
+element of the array. ArrayFire provides the end alias for this which also
+allows you to index the last element of the array.
 
-You can access regions regions of the array via the af::seq and af::span
-objects. The span objects allows you to select the entire set of elements across
-a particular dimension/axis of an array. For example, we can select the third
-column of the array by passing span as the first agument and 2 as the second
-argument to the parenthesis operator.
+\snippet test/index.cpp index_tutorial_negative_indexing
+
+## Indexing slices and subarrays
+
+You can access regions of the array via the af::seq and af::span objects. The
+span objects allows you to select the entire set of elements across a particular
+dimension/axis of an array. For example, we can select the third column of the
+array by passing span as the first agument and 2 as the second argument to the
+parenthesis operator.
 
 \snippet test/index.cpp index_tutorial_third_column
 
@@ -98,13 +105,13 @@ three constructors for af::seq.
 * af::seq(begin, end) Defines a range between begin and end inclusive
 * af::seq(begin, end, step) defines a range between begin and end striding by step values
 
-You can select the second and forth rows by passing (seq(1, 4, 2), span) to the
-indexing operator.
+You can select the second and forth(last) rows by passing (seq(1, end, 2), span)
+to the indexing operator.
 
 \snippet test/index.cpp index_tutorial_second_and_fourth_rows
 
 \f[
-A(seq(1, 3, 2), span) =
+A(seq(1, end, 2), span) =
 \begin{bmatrix}
      1 & 5 &  9 & 13 \\
      3 & 7 & 11 & 15
@@ -179,10 +186,24 @@ approx2(A,
 \end{bmatrix}
 \f]
 
+Boolean(b8) arrays can be used to index into another array. In this type of
+indexing the non-zero values will be selected by the boolean operation. If we
+want to select all values less than 5, we can pass a boolean expression into
+the parenthesis operator.
+
+\snippet test/index.cpp index_tutorial_boolean
+
+\f[
+out =
+\begin{bmatrix}
+0 \\ 1 \\ 2 \\ 3 \\ 4
+\end{bmatrix}
+\f]
+
 ## References and copies
 
 All ArrayFire indexing functions return af::array(technically its an array_proxy
-class but we will discuss that later** objects. These objects may be new arrays
+class but we will discuss that later) objects. These objects may be new arrays
 or they may reference the original array depending on the type of indexing that
 was performed on them. If an array was indexed using another af::array or it
 was indexed using the af::approx functions, then a new array is created. It does
@@ -190,7 +211,7 @@ not reference the original data.
 
 If an array was indexed using a scalar, af::seq or af::span, then the resulting
 array will reference the original data IF the first dimension is continuous. The
-following indexing approaches will not allocate additional memory.
+following lines will not allocate additional memory.
 
 \snippet test/index.cpp index_tutorial_references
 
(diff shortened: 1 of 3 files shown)


docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+
+Notice that even though the copy3 array is refrencing continuous memory in the
+original array, a new array is created because we used an array to index into
+the af::array.
+
+## Assignment
+
+An assignment on an af::array, will update the data of the array with the result
+of the experession on the right hand side of the equal(=) opeartor. Assignments
+will not updated the array was previously refrencing through an indexing
```
9prady9: typo

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+
+This will update only the array being modified. If there are arrays that
+are referring to this array because of an indexing operation, those values
+will remain unchanged.
+
+Allocation will only be performed if there are other arrays referencing the data
+at the point of assignment. In the previous example, an allocation will be
+performed when assigning to the `A` array because the `ref` array is pointing
+to the original data. Here is another example demonstrating when an allocation
+will occure:
```
ShadyBoukhary: typo
```suggestion
will occur:
```

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+
+There are several member functions which allow you to index into an af::array. These
+functions have similar functionallity but may be easier to parse for some.
+
 * [row(i)](\ref af::array::row) or [col(i)](\ref af::array::col) specifying a single row/column
 * [rows(first,last)](\ref af::array::rows) or [cols(first,last)](\ref af::array::cols)
- specifying a span of rows or columns
+ specifying multiple rows or columns
+
+# Additional examples
```
ShadyBoukhary: Maybe also show the result of the `af_print()`s?

umar456:
Addressed Feedback:

image

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+    9 \\
+    10 \\
+    11
+\end{bmatrix}
+\f]
+
+You can read that as saying that you want all values across the first dimension,
+but only from index 2 of the second dimension.
+
+You can access the second rows by passing (1, span) to the array
```
9prady9: typo, `rows` -> `row`

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+
+\snippet test/index.cpp index_tutorial_second_row
+
+\f[ A(1, span) = [ 1, 5, 9, 13 ] \f]
+
+You can use the af::seq (short for sequence) object defines a range of values
```
9prady9: `... af::seq object defines a range of values ... ` 
->
 `... af::seq object to define a range ... ` ?

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+when indexing. For example if you wanted to get the first two columns, you would
```
9prady9: - Comma after `For example`.
- `wanted` -> `want` I think .
- `you would` -> `you can` .

It should look like the following

```
For example, if you want to fetch the first two columns, you can access the array by passing span for the first argument and seq(2) as the second argument.
```

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+\begin{bmatrix}
+     0 & 4 \\
+     1 & 5 \\
+     2 & 6 \\
+     3 & 7
+\end{bmatrix}
+\f]
+
+The af::seq object can also be used to define ranges that are not continuous. There are
+three constructors for af::seq.
```
9prady9: Second sentence stops abruptly. I think it should end like the following
```
There are three constructors for af::seq that can help create non-continuous ranges(af::seq objects)`

File: docs/pages/indexing.md [hunk 8]
```diff
@@ -1,18 +1,325 @@
...
+## References and copies
+
+All ArrayFire indexing functions return af::array(technically its an array_proxy
+class but we will discuss that later) objects. These objects may be new arrays
+or they may reference the original array depending on the type of indexing that
+was performed on them. If an array was indexed using another af::array or it
+was indexed using the af::approx functions, then a new array is created. It does
+not reference the original data.
+
+If an array was indexed using a scalar, af::seq or af::span, then the resulting
```
9prady9: This new paragraph is not necessary as we still continuing topic from first paragraph. We should either merge this into above paragraph or list the two types of outcome in a bullet list fashion after we end the sentence `that was performed on them.`.

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+
+Notice that even though the copy3 array is referencing continuous memory in the
+original array, a new array is created because we used an array to index into
+the af::array.
+
+## Assignment
+
+An assignment on an af::array, will update the data of the array with the result
+of the expression on the right hand side of the equal(=) operator. Assignments
+will not updated the array was previously referencing through an indexing
```
9prady9: typo, `will not update` and something else is missing, perhaps a conjunction `that`.

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+\end{bmatrix}
+\f]
+
+## Member functions
+
+There are several member functions which allow you to index into an af::array. These
+functions have similar functionality but may be easier to parse for some.
+
 * [row(i)](\ref af::array::row) or [col(i)](\ref af::array::col) specifying a single row/column
 * [rows(first,last)](\ref af::array::rows) or [cols(first,last)](\ref af::array::cols)
```
9prady9: We have `slices` also right ?

docs/pages/indexing.md
```diff
@@ -1,18 +1,325 @@
...
+we are accessing the fifth element of the array.
+
+\snippet test/index.cpp index_tutorial_fifth_element
+
+\f[ A(5) = [ 5 ] \f]
+
+\note Normally you want to avoid accessing individual elements of the array like this
+for performance reasons.
+
+Negative indexing values will index from the end of the array. For example, the
```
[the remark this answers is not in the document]
umar456: Reworded it to:

```
Indexing with negative values will index from the end of the array.
```

docs/pages/indexing.md
umar456: Updated that section to:

```
There are three constructors for af::seq. 

* af::seq(N): Defines a range between 0 and N-1
* af::seq(begin, end) Defines a range between begin and end inclusive
* af::seq(begin, end, step) defines a range between begin and end striding by step values

The last constructor that can help create non-continuous ranges. For example,
you can select the second and forth(last) rows by passing (seq(1, end, 2), span)
to the indexing operator.
```

docs/pages/indexing.md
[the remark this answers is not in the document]
9prady9: the word `index` again after the start `Indexing with negative ... ` feels recursive to me. Just `will fetch/access/get` is more simple in my opinion.

docs/pages/indexing.md
9prady9: `The af::seq object can also be used to define ranges that are not continuous.` Is this sentence gone now ?

docs/pages/indexing.md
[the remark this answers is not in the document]
umar456:
```
Indexing with negative values will access from the end of the array. For example,
the value negative one and negative two(-2) will return the last and second to
last element of the array, respectivly. ArrayFire provides the `end` alias for
this which also allows you to index the last element of the array.
```

docs/pages/indexing.md
umar456: Yeah, I moved that message to the paragraph after the list.

docs/pages/indexing.md
umar456:
```
All ArrayFire indexing functions return af::array(technically its an array_proxy
class) objects. These objects may be new arrays or they may reference the
original array depending on the type of indexing that was performed on them. 

- If an array was indexed using another af::array or it was indexed using the
af::approx functions, then a new array is created. It does not reference the
original data.
- If an array was indexed using a scalar, af::seq or af::span, then
the resulting array will reference the original data IF the first dimension is
continuous. The following lines will not allocate additional memory.

\note The new arrays wither references or newly allocated arrays, are
independent of the original data. Meaning that any changes to the original array
will not propagate to the references. Likewise, any changes to the reference
arrays will not modify the original data.

\snippet test/index.cpp index_tutorial_references

The following code snippet shows some examples of indexing that will allocate
new memory.

\snippet test/index.cpp index_tutorial_copies

Notice that even though the copy3 array is referencing continuous memory in the
original array, a new array is created because we used an array to index into
the af::array.
```

docs/pages/indexing.md
umar456:
```
An assignment on an af::array, will replace the array with the result of the
expression on the right hand side of the equal(=) operator. This means that the
type and shape of the result can be different from the array on the left had
side of the equal operator. Assignments will not update the array that was
previously referenced through an indexing operation. Here is an example:
```

PR #2571, opened 2019-07-08
Fix bugs in SIFT and blas

umar456:
* Fixes a bug in blas after the type was changed from common::half to compute<T>
* Fix bug in SIFT caused by the change in the resize internal backend API
* Fix several leaks in the blas, gloh, and gen_index tests.

### DIFF
--- a/src/backend/cuda/kernel/sift_nonfree.hpp
+++ b/src/backend/cuda/kernel/sift_nonfree.hpp
@@ -78,6 +78,7 @@
 #include <debug_thrust.hpp>
 #include <memory.hpp>
 #include "shared.hpp"
+#include <af/defines.h>
 
 #include "convolve.hpp"
 #include "resize.hpp"
@@ -1061,7 +1062,7 @@ Array<T> createInitialImage(CParam<T> img, const float init_sigma,
     Array<convAccT> filter = gauss_filter<convAccT>(s);
 
     if (double_input) {
-        resize<T, InterpolationType::Bilinear>(init_img, img);
+        resize<T>(init_img, img, AF_INTERP_BILINEAR);
         convolve2<T, convAccT>(init_tmp, init_img, filter, 0, false);
     } else
         convolve2<T, convAccT>(init_tmp, img, filter, 0, false);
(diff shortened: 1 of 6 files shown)


src/backend/cuda/types.hpp
```diff
@@ -149,16 +149,13 @@ struct kernel_type<common::half> {
...
     // These are the types within a kernel
-
-#if __CUDA_ARCH__ > 530 && __CUDA_ARCH__ != 610
+#if __CUDA_ARCH__ >= 530 && __CUDA_ARCH__ != 610
     using compute = __half;
 #else
     using compute = float;
 #endif
 #else
-
```
9prady9: `__CUDA_ARCH__` is defined for nvrtc compilation also. The only time `#else` of the `#ifdef __CUDA_ARCH__` is taken is when this header is included by `.cpp` in CUDA backend.

The following(not in this PR's diff)
```
#if defined(NVCC) || defined(__CUDACC_RTC__)
    using native  = __half;
#else
    using native = common::half;
#endif
```
can be just
```
#if defined(NVCC)
    using native  = __half;
#else
    using native = common::half;
#endif
```
Don't you think so ?

src/backend/cuda/kernel/sift_nonfree.hpp
```diff
@@ -1108,8 +1109,7 @@ std::vector<Array<T>> buildGaussPyr(Param<T> init_img, const unsigned n_octaves,
                 tmp_pyr.push_back(
                     createEmptyArray<T>({tmp_pyr[src_idx].dims()[0] / 2,
                                          tmp_pyr[src_idx].dims()[1] / 2}));
-                resize<T, InterpolationType::Bilinear>(tmp_pyr[idx],
-                                                       tmp_pyr[src_idx]);
+                resize<T>(tmp_pyr[idx], tmp_pyr[src_idx], AF_INTERP_BILINEAR);
```
9prady9: thanks for this  fix. I must have forgotten to turn on build-non-free when I switched computers while fixing these in my PR.

src/backend/cuda/types.hpp
9prady9: This is an obsolete check and doesn't hurt the builds at the moment. Deferring to fix in later PR.

PR #2572, opened 2019-07-08
Add half support for several data functions

umar456:
Added half support for

- logical binary operations
- diagonal
- identity
- iota
- ireduce
- range
- triangle
- transpose
- transpose_inplace

Refactored half usage to be explicit.

The OpenCL backend compiles but it is probably broken for f16. I will
be testing that in another PR.

### DIFF
--- a/CMakeModules/InternalUtils.cmake
+++ b/CMakeModules/InternalUtils.cmake
@@ -30,7 +30,7 @@ endfunction()
 
 function(arrayfire_get_cuda_cxx_flags cuda_flags)
   if(NOT MSVC)
-    set(flags "-std=c++14 -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
+    set(flags "-std=c++14 --expt-relaxed-constexpr -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
   else()
     set(flags "-Xcompiler /wd4251 -Xcompiler /wd4068 -Xcompiler /wd4275 -Xcompiler /bigobj -Xcompiler /EHsc")
     if(CMAKE_GENERATOR MATCHES "Ninja")
--- a/src/api/c/ops.hpp
+++ b/src/api/c/ops.hpp
@@ -25,82 +25,51 @@ using namespace detail;
 
 template<typename T, af_op_t op>
 struct Binary {
-    static __DH__ detail::compute_t<T> init();
+    static __DH__ T init();
 
-    __DH__ detail::compute_t<T> operator()(detail::compute_t<T> lhs,
-                                           detail::compute_t<T> rhs);
+    __DH__ T operator()(T lhs, T rhs);
 };
 
 template<typename T>
 struct Binary<T, af_add_t> {
-    static __DH__ detail::compute_t<T> init() {
-        return detail::scalar<detail::compute_t<T>>(0);
-    }
+    static __DH__ T init() { return detail::scalar<T>(0); }
 
-    __DH__ detail::compute_t<T> operator()(detail::compute_t<T> lhs,
-                                           detail::compute_t<T> rhs) {
-        return lhs + rhs;
-    }
+    __DH__ T operator()(T lhs, T rhs) { return lhs + rhs; }
 };
 
 template<typename T>
 struct Binary<T, af_mul_t> {
-    static __DH__ detail::compute_t<T> init() {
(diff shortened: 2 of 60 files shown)


src/api/cpp/array.cpp
```diff
@@ -949,6 +949,8 @@ INSTANTIATE(long long)
 INSTANTIATE(unsigned long long)
 INSTANTIATE(short)
 INSTANTIATE(unsigned short)
+INSTANTIATE(af_half)
+INSTANTIATE(half_float::half)
```
9prady9: Can we create arrays of type `half_float::half` type as well ? I was under the impression that `half_float::float` is used only in our unit tests.

src/backend/cpu/cast.hpp
```diff
@@ -38,7 +38,7 @@ struct UnOp<To, common::half, af_cast_t> {
 
     void eval(jit::array<To> &out, const jit::array<Ti> &in, int lim) {
         for (int i = 0; i < lim; i++) {
-            float val = in[i];
+            float val = static_cast<float>(in[i]);
```
9prady9: :+1:

src/backend/cpu/kernel/iota.hpp
```diff
@@ -16,18 +16,18 @@ namespace kernel {
...
 void iota(Param<T> output, const af::dim4& sdims) {
     const af::dim4 dims    = output.dims();
-    T* out                 = output.get();
+    data_t<T>* out      = output.get();
     const af::dim4 strides = output.strides();
 
     for (dim_t w = 0; w < dims[3]; w++) {
         dim_t offW = w * strides[3];
-        T valW     = (w % sdims[3]) * sdims[0] * sdims[1] * sdims[2];
+        dim_t valW     = (w % sdims[3]) * sdims[0] * sdims[1] * sdims[2];
```
9prady9: This looks like a bug and should go into a separate PR if true.

src/backend/cuda/kernel/iota.hpp
```diff
@@ -41,15 +41,15 @@ __global__ void iota_kernel(Param<T> out, const int s0, const int s1,
...
+    dim_t val = (ow % s3) * s2 * s1 * s0;
     val += (oz % s2) * s1 * s0;
 
     const int incy = blocksPerMatY * blockDim.y;
     const int incx = blocksPerMatX * blockDim.x;
 
     for (int oy = yy; oy < out.dims[1]; oy += incy) {
         int oyzw = ozw + oy * out.strides[1];
-        T valY   = val + (oy % s1) * s0;
+        dim_t valY   = val + (oy % s1) * s0;
```
9prady9: Same as CPU, it is odd that we used to stored indexing type in `T` which is most likely template parameter.

src/backend/common/half.hpp
```diff
@@ -609,36 +815,25 @@ class alignas(2) half {
...
+        memcpy(&data_, &bits, sizeof(uint16_t));
     }
 
-#ifndef __CUDACC_RTC__
-    template<typename T,
-        typename std::enable_if_t<std::is_unsigned<T>::value>* = nullptr>
-    CONSTEXPR_DH explicit half(T value) noexcept
-        : data_(int2half<std::round_to_nearest, false>(value)) {}
+    CONSTEXPR_DH explicit half(double value) noexcept
+        : data_(float2half<double>(value)) {}
```
9prady9: I assume, we don't have a doubl2half and hence we have float2half here. Shouldn't we just add `static_cast<float>(value)` too ?

src/api/c/ops.hpp
```diff
@@ -25,82 +25,51 @@ using namespace detail;
 
 template<typename T, af_op_t op>
 struct Binary {
-    static __DH__ detail::compute_t<T> init();
```
9prady9: We moved to this `compute` change not too long ago. What change resulted in this inversion ?

CMakeModules/InternalUtils.cmake
```diff
@@ -30,7 +30,7 @@ endfunction()
 
 function(arrayfire_get_cuda_cxx_flags cuda_flags)
   if(NOT MSVC)
-    set(flags "-std=c++14 -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
+    set(flags "-std=c++14 --expt-relaxed-constexpr -Xcompiler -fPIC -Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY}hidden")
```
9prady9: Why do we need this on non-MSVC platforms ?

PR #2575, opened 2019-07-08
Fixed unnecessary  memcpy during array assignment

### LINKED ISSUE #2425: assign operations are performing unnecessary memcpys
The following code performs a device to device memcpy of the large array if the input array is a 2D matrix. This behavior is unnecessary in this example.

```
#include <arrayfire.h>

using namespace af;

int main() {
  info();

  array a = constant(0.0f, 1000, 1000);
  float hb[] = { 3, 5, 6, 8, 9 };
  array b(5, hb);
  array c = constant(1.0f, 5);
  array tmp = a(b);
  a(b) = c;
  a.eval();
  af_print(a(seq(10)));
}
```

ShadyBoukhary:
Fixes #2425

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -471,25 +471,37 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
         }
     }
 
-    af_array par_arr = nullptr;
+    af_array par_arr = 0;
 
+    dim4 parent_dims = impl->parent_->dims();
     if (impl->is_linear_) {
         AF_THROW(af_flat(&par_arr, impl->parent_->get()));
+        // The set call will dereference the impl->parent_ array. We are doing
+        // this because the af_flat call above increases the reference count of the
+        // parent array which triggers a copy operation. This triggers a copy operation
+        // inside the af_assign_gen function below. The parent array will be reverted
+        // to the original array and shape later in the code.
+        af_array empty = 0;
+        impl->parent_->set(empty);
         nd = 1;
     } else {
         par_arr = impl->parent_->get();
     }
 
-    af_array tmp = nullptr;
-    AF_THROW(af_assign_gen(&tmp, par_arr, nd, impl->indices_, other_arr));
+    af_array flat_res = 0;
+    AF_THROW(af_assign_gen(&flat_res, par_arr, nd, impl->indices_, other_arr));
 
-    af_array res = nullptr;
+    af_array res = 0;
+    af_array unflattened = 0;
     if (impl->is_linear_) {
-        AF_THROW(af_moddims(&res, tmp, this_dims.ndims(), this_dims.get()));
+        AF_THROW(af_moddims(&res, flat_res, this_dims.ndims(), this_dims.get()));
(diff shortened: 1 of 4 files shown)


src/api/cpp/array.cpp
```diff
@@ -471,21 +471,31 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
         }
     }
 
-    af_array par_arr = nullptr;
+    af_array par_arr = 0;
 
+    dim4 dims = impl->parent_->dims();
```
umar456: Can you name this parent_dim or original_dim?

src/api/cpp/array.cpp
```diff
@@ -471,21 +471,31 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
...
     if (impl->is_linear_) {
         AF_THROW(af_flat(&par_arr, impl->parent_->get()));
+        // af_flat creates a new reference, remove the original reference
+        // in order to prevent triggering a memcpy in af_assign_gen when refCount > 1
```
umar456: I think a better comment would be:

```
The set call will dereference the impl->parent_ array. We are doing
this because the af_flat call above increases the reference count of the
parent array which triggers a copy operation. This triggers a copy operation
inside the af_assign_gen function below. The parent array will be reverted
to the original array and shape later in the code.
```

src/api/cpp/array.cpp
```diff
@@ -471,21 +471,31 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
...
 
-    af_array res = nullptr;
+    af_array res = 0;
+    af_array unflattened = 0;
     if (impl->is_linear_) {
         AF_THROW(af_moddims(&res, tmp, this_dims.ndims(), this_dims.get()));
+        // Unflatten the af_array and reset the original reference
+        AF_THROW(af_moddims(&unflattened, par_arr, dims.ndims(), dims.get()));
+        impl->parent_->set(unflattened);
+        unflattened = 0;
```
umar456: This line is not necessary.
```suggestion
```

src/api/cpp/array.cpp
```diff
@@ -471,21 +471,31 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
...
+        // in order to prevent triggering a memcpy in af_assign_gen when refCount > 1
+        af_array empty = 0;
+        impl->parent_->set(empty);
         nd = 1;
     } else {
         par_arr = impl->parent_->get();
     }
 
-    af_array tmp = nullptr;
+    af_array tmp = 0;
```
umar456: I know we have this all over the place in this code but can you call this `flat_result` or something more descriptive?

ShadyBoukhary: Done

src/api/cpp/array.cpp
ShadyBoukhary: Changed to `parent_dim`

src/api/cpp/array.cpp
ShadyBoukhary: Done

PR #2585, opened 2019-07-15
Add f16 support for unary types

umar456:


src/api/c/implicit.cpp
```diff
@@ -42,6 +42,8 @@ af_dtype implicit(const af_dtype lty, const af_dtype rty) {
 
     if ((lty == s16) || (rty == s16)) return s16;
 
+    if ((lty == f16) || (rty == f16)) return f16;
+
```
9prady9: I am aware this is right change. Just saying it aloud that the `implicit` function is dangerous one, a simple line swap will result in completely different set of results. I wonder if we can rewrite it in a better way.

src/backend/cuda/math.hpp
```diff
@@ -54,6 +54,10 @@ static inline __DH__ size_t max(size_t lhs, size_t rhs) {
 }
 
 #ifdef __CUDA_ARCH__
+static inline __device__ __half abs(__half val) {
+    return __short_as_half(__half_as_short(val) & 0x7FFF);
```
9prady9: It's unfortunate that they have lot half `h*` math functions but not `habs`.

test/math.cpp
```diff
@@ -30,57 +34,47 @@ typedef std::complex<double> complex_double;
...
-                    << "for value: " << h_a[i] << endl;          \
-            }                                                    \
-        } catch (exception & ex) { FAIL() << ex.what(); }        \
+#define TEST_REAL(T, func, err, lo, hi)                                     \
+    TEST(MathTests, Test_##func##_##T) {                                    \
+        try {                                                               \
+            SUPPORTED_TYPE_CHECK(T);                                        \
+            af_dtype ty = (af_dtype)dtype_traits<T>::af_type;               \
+            array a     = (hi - lo) * randu(num, ty) + lo + err;            \
+            a           = a.as(ty);                                         \
```
9prady9: Is this explicit necessary ? The reason I ask that is if this line isn't there, then `TEAT_REAL` and `TEST_CPLX` are identical.

src/api/c/implicit.cpp
gaika: Shouldn't it be `double > float > half > uintl > intl > uint > int > uchar > char`
and not `double > float > uintl > intl > uint > int > half > uchar > char` as it currently implemented?

9prady9: What we have now is correct I think since the ordering is based on the range of values a given type can store.

gaika: When I have a float32 and I multiply it by 2 (Int64) I do not expect the result to be Int64. Why would it be different for a float16?

9prady9: There are couple of troubling things with putting half ahead of it's bit-size peers. Any value from the other array(lets say `rhs`) whose type is of higher bit size will be lost - all higher order bits will be truncated. This will result in totally incorrect results not just less accuracy. Therefore, I would think 
`double > float > uintl > intl > uint > int > half > ushort > short > uchar > char` is the right order.

@umar456 What do you think ?

gaika: By that logic half(0.3) + 1 == 1 while float(1.3) + 1 == 1.3

9prady9: true, that example negates my order. But what I said is also true - https://en.wikipedia.org/wiki/Half-precision_floating-point_format#Precision_limitations_on_integer_values. I don't think it is a good idea to use `half` when one of the type is `half` and other is either 32-bit or more.

I think ArrayFire should error out if one of operands in half and other operand of higher bit-size than half.

gaika: At least in `julia> Float16(1) + 1 == Float16(2.0)` , sorry, don't know any other language with a native `half` type.

9prady9: Right, but if we do
```
julia> Float16(1) + 657890
Inf16
```
the result is infinity as half precision can't represent it.  I wonder if thats we should do or figure out a better way if possible.

umar456: I think @gaika is right. For things like float16 + 1 the result should be float16. If the values fall out of the range then that will need to be handled by the user by casting the float16 array to another type.

src/backend/cuda/math.hpp
umar456: Yeah, I doubt this is an expensive operation. Those are reinterpret casts so basically noops.

test/math.cpp
umar456: No, this is not necessary. It was required earlier because we were using the new ASSERT_VEC_ARRAY_NEAR macro.

PR #2587, opened 2019-07-18
Fix random number generator for f16 and short on the CPU backend

umar456:
The CPU random number generator was incorrectly calculating the shifts
for the transform function. This logic seemed to be copied from the
uchar transform function. The incorrect behavior only appeared on
Windows and for arrays larger than 256 elements. In this commit I
mask the shifted value before the float factor is calculated.


9prady9:
This PR addressed the math_cpu failures on windows for https://github.com/arrayfire/arrayfire/pull/2585

PR #2589, opened 2019-07-19
Fix several OpenCL fp16 kernels.

umar456:
Fixes the following kernels on the AMD Vega card with fp16 kernels.

- rotate
- triangle
- reduce
- JIT
- identity
- lookup
- random
- range
- transpose
- iota

Not addressed in this PR:

- matmul(half test is failing on clblast)
- JIT LargeTree kernel is still failing
- rotate is still failing for the short datatype
- iota is failing because of rounding issues between the two fp16 types

### DIFF
--- a/src/backend/opencl/kernel/reduce.hpp
+++ b/src/backend/opencl/kernel/reduce.hpp
@@ -28,6 +28,12 @@
 #include "config.hpp"
 #include "names.hpp"
 
+#include <iostream>
+
+namespace opencl {
+
+namespace kernel {
+
 using cl::Buffer;
 using cl::EnqueueArgs;
 using cl::Kernel;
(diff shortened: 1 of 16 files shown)


src/backend/opencl/kernel/reduce.hpp
```diff
@@ -28,6 +28,12 @@
 #include "config.hpp"
 #include "names.hpp"
 
+#include <iostream>
```
9prady9: Forgot it

src/backend/opencl/kernel/reduce.hpp
```diff
@@ -28,6 +28,12 @@
...
+
+namespace opencl {
+
```
9prady9: Nit: Unnecessary new line, can be removed

PR #2590, opened 2019-07-20
Adjust JIT param size for AMD ROCm/Vega GPUs

umar456:


src/backend/opencl/Array.cpp
```diff
@@ -280,7 +280,7 @@ bool passesJitHeuristics(Node *root_node) {
         // This is the maximum size of the params that can be allowed by the
         // CUDA platform.
         constexpr size_t max_nvidia_param_size = (4096 - base_param_size);
-        constexpr size_t max_amd_param_size    = (3670 - base_param_size);
+        constexpr size_t max_amd_param_size    = (3520 - base_param_size);
```
9prady9: Doesn't this change based on device ?

umar456: We don't know what these numbers are. They don't seem to be published.

9prady9: thats unfortunate.

PR #2595, opened 2019-07-21
Move constructor and assignment operator

### LINKED ISSUE #2371: Move ctor + move assignment for af::array
Currently `af::array` does not implement a move constructor, so any attempt at move construction will default to copying.

    af::array a1;
    auto a2 = std::move(a1); // invokes copy ctor
    // mutating a2 will make a deep copy

umar456:
Fixes #2371

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -169,7 +169,7 @@ namespace af
             @{
         */
         /**
-            Create undimensioned array (no data, undefined size)
+            Create an uninitialized array (no data, undefined size)
 
             \code
             array A, B, C;   // creates three arrays called A, B and C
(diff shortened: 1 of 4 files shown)


include/af/array.h
```diff
@@ -169,14 +169,39 @@ namespace af
             @{
         */
         /**
-            Create undimensioned array (no data, undefined size)
+            Create an uninitialized array (no data, undefined size)
```
9prady9: haha, `undimensioned` ?

src/api/cpp/array.cpp
```diff
@@ -491,12 +501,14 @@ array::array_proxy &af::array::array_proxy::operator=(const array &other) {
     af_array flat_res = 0;
     AF_THROW(af_assign_gen(&flat_res, par_arr, nd, impl->indices_, other_arr));
 
-    af_array res = 0;
+    af_array res         = 0;
     af_array unflattened = 0;
```
9prady9: Nit: unaligned equals it seems.

PR #2596, opened 2019-07-21
Fix coverage exclude regular expressions

umar456:


9prady9:
Is coverage even working now ?

umar456:
It is but its including external dependencies so it throws off the result.

PR #2597, opened 2019-07-22
Fix CMP0074 policy warnings. add set_policies macro

umar456:


CMakeLists.txt
```diff
@@ -24,6 +20,10 @@ include(platform)
 include(GetPrerequisites)
 include(CheckCXXCompilerFlag)
 
+set_policies(
+  TYPE NEW
+  POLICIES CMP0073
+           CMP0074)
```
9prady9: Can we set NEW or OLD individually for each policy ? It would be nice if this generic in that manner, otherwise looks fine.

umar456: You can add another set_policies call for OLD behaviours. We should only stick with NEW in most cases.

umar456:
It looks like its an issue with the vcpkg script and the warning is only appearing on windows. The scope of these policies do not seem to extend to those scripts.

PR #2598, opened 2019-07-22
Add matmul helper function in internal API

umar456:


umar456:
Closing because this is already done in #2359

PR #2599, opened 2019-07-22
Let af_approx1 allocate the output array, have af_approx1_v2 accept preallocated outputs

mark-poscablo:
This extends #2324 in order to address the concern raised in https://github.com/arrayfire/arrayfire/pull/2481#discussion_r273752946. Here, `af_approx1`'s behavior towards allocating output arrays is reverted to the original (before #2324), where it simply creates a new `af_array` each time. However, a new function is added: `af_approx1_v2`, which implements the same behavior as in #2324 (uses the existing array passed as the `out` argument and only allocate if it is null).

Here are some screenshots of the added documentation as well:
image
image

### DIFF
--- a/include/af/signal.h
+++ b/include/af/signal.h
@@ -776,6 +776,9 @@ extern "C" {
 AFAPI af_err af_approx1(af_array *out, const af_array in, const af_array pos,
                         const af_interp_type method, const float off_grid);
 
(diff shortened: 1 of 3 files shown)


mark-poscablo:
> Looks good. I have requested a couple of changes. Can you also run clang-format on this?

I ran clang-format on the modified files but left out the specially formatted lines already there, which I assume were written that way to make them easier to read.

PR #2603, opened 2019-07-30
Add approx1_*_v2 functions to the unified backend

umar456:


src/api/unified/symbol_manager.cpp
```diff
@@ -246,6 +246,14 @@ bool checkArray(af_backend activeBackend, af_array a) {
     return backend == activeBackend;
 }
 
+bool checkArray(af_backend activeBackend, const af_array* a) {
```
mark-poscablo: I like this, it makes checking the output array (whether it exists yet or not) cleaner.

src/api/unified/signal.cpp
```diff
@@ -13,22 +13,36 @@
...
+                     const af_interp_type method, const float offGrid) {
+    CHECK_ARRAYS(yo, yi, xo);
     return CALL(yo, yi, xo, method, offGrid);
 }
 
 af_err af_approx2(af_array *zo, const af_array zi, const af_array xo,
                   const af_array yo, const af_interp_type method,
                   const float offGrid) {
-    CHECK_ARRAYS(zi, xo, yo);
+    CHECK_ARRAYS(zo, zi, xo, yo);
```
mark-poscablo: My changes for adding the _v2 versions of `af_approx2`/`af_approx2_uniform` are still underway (no PR yet). Should we include this change in this PR now? Alternately, I can add them in the changes that I'm working on right now.

umar456: This is working on the original function. you don't have to update that in your PR. You are going to be adding the v2 functions.

WilliamTambellini:
tks guys

PR #2604, opened 2019-08-01
Add af_approx2*_v2 functions

mark-poscablo:
Similar to #2324 / #2599, but for `af_approx2*_v2` functions.

This also modifies `af_approx1*_v2` tests so that they use test fixtures instead of repeating defining the host C arrays over and over again.

### DIFF
--- a/test/approx2.cpp
+++ b/test/approx2.cpp
@@ -760,7 +760,7 @@ class Approx2V2 : public ::testing::Test {
     typedef typename dtype_traits<T>::base_type BT;
 
     static const int h_gold_size = 4;
-    static const int h_in_size = 9;
+    static const int h_in_size   = 9;
     static const int h_pos1_size = 4;
     static const int h_pos2_size = 4;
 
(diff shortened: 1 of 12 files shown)


test/testHelpers.hpp
```diff
@@ -1417,4 +1417,40 @@ ::testing::AssertionResult assertArrayEq(std::string aName, std::string bName,
     return testWriteToOutputArray(aName, bName, a, b, metadata);
 }
 
+class TestData {
```
umar456: Can you put this class in another file. I think this class is very specific to the approx functions because it is only working on floats. I can see the benefit of having a class like this because our current approach to reading test data is tedious. Perhaps we can come up with a better design later.

test/approx2.cpp
```diff
@@ -753,3 +753,352 @@ TEST(Approx2, CPPEmptyPosAndInput) {
...
+
+TEST_F(Approx2NullArgs, V2NullOutputPtr) {
+    af_array* out_ptr = 0;
+    ASSERT_EQ(af_approx2_v2(out_ptr, this->in, this->pos1, this->pos2,
+                            AF_INTERP_LINEAR, 0.f),
+              AF_ERR_ARG);
+}
+
+TEST_F(Approx2NullArgs, V2NullInputArray) {
+    ASSERT_EQ(af_approx2_v2(&this->out, 0, this->pos1, this->pos2,
```
umar456: Nit: Expected is the first argument.

Issue #2478, opened 2019-04-01
Memory leak 


WilliamTambellini:
I would try to reproduce via a simple main.cpp with direct AF but these few lines make AFCUDA  totally leaking at least with AF 3.6.2 (AF_MEM_DEBUG=1, no garbage collecting): 

    // left and right are just 3d arrays
    const unsigned ns = left.dims(2); // usually 20 slices
    for (unsigned z = 0; z < ns; ++z) {
      o += af::matmul(left.slice(z), right.slice(z), transposeLeft ? AF_MAT_TRANS : AF_MAT_NONE, transposeRight ? AF_MAT_TRANS : AF_MAT_NONE);
    }

when calling getMemInfo() before and after : 
BEFORE : AllocBytes=5083372752	AllocBuffers=1391	LockBytes=5083372752	LockBuffers=1391
...
AFTER : AllocBytes=5104344272	AllocBuffers=1411	LockBytes=5104344272	LockBuffers=1411

AF has clearly created 20 buffers for the 20 slices and these will never been freed leading to OOM.

WilliamTambellini:
It seems to come from the inplace addition (+=), not the matmul : 

    for (unsigned z = 0; z < ns; ++z) {
              af::matmul(left.slice(z), right.slice(z), transposeLeft ? AF_MAT_TRANS : AF_MAT_NONE, transposeRight ? AF_MAT_TRANS : AF_MAT_NONE);
    }

does nt leak (but ofcourse does nt do the job).

9prady9:
Sure, a reproducible code snippet will help in exactly replicating the problem. 

I shall also try to replicate the issue on my end.

WilliamTambellini:
ok, so it s probably a bad (undesired ?) effect of the jit approach : 
the inplace addition is just adding ops to the tree without evaluating anything BUT allocating (potentially big) memory/buffers, easily leading to oom.
Forcing .eval() in the loop is probably avoiding that but to be double checked and potentially hurting the speed :

// left and right are just const 3d arrays
const unsigned ns = left.dims(2); // usually 20 slices
for (unsigned z = 0; z < ns; ++z) {
  // todo : print af::getMemInfo();
  o += af::matmul(left.slice(z), right.slice(z), transposeLeft ? AF_MAT_TRANS : AF_MAT_NONE, transposeRight ? AF_MAT_TRANS : AF_MAT_NONE);
  // todo : print af::getMemInfo();
  o.eval();
[...]

PR #2608, opened 2019-08-14
Fixes infinite loop when eval-ing under memory pressure

syurkevi:
Adds different return codes to jit heuristic allowing
different behaviour depending on why eval is triggered


WilliamTambellini:
Sounds very interesting.
Would it kind of implement the idea mentioned there :  
https://github.com/arrayfire/arrayfire/issues/2478
?

umar456:
Its not quite the same thing. Here we addressed a bug where we didn't handle the case were we were under memory pressure but the JIT tree wasn't very large. I think that issue is trying to suggest something else.

src/backend/common/defines.hpp
```diff
@@ -49,6 +49,13 @@ typedef enum {
     AF_BATCH_DIFF,             /* signal and filter have different batch size */
 } AF_BATCH_KIND;
 
+enum class kJITHeuristics {
+    PASS               = 0, /* no eval necessary */
+    TREE_HEIGHT        = 1, /* eval due to jit tree height */
+    KERNEL_PARAM_SIZE  = 2, /* eval due to many kernel parameters */
+    MEM_PRESSURE       = 3  /* eval due to memory pressure */
+};
```
9prady9: I think enum class values should be `CamelCase` since they need `name::` resolution anyway while being used. It results in less characters and easy to read names like the following, hence better in my opinion.

```cpp
enum class kJITHeuristics : int {
    Pass = 0,
    TreeHeight,
    KernelParameterSize,
    MemoryPressure
};
```

syurkevi: Done! I actually like the CamelCase for these, it does look better.

9prady9:
I have reschulded the failed jobs since they failed to pull from GitHub.

PR #2615, opened 2019-08-20
A few small clarifications

r-barnes:


src/backend/common/MemoryManagerImpl.hpp
```diff
@@ -253,7 +253,7 @@ void MemoryManager<T>::printInfo(const char *msg, const int device) {
     printf("%s\n", msg);
     printf(
         "---------------------------------------------------------\n"
-        "|     POINTER      |    SIZE    |  AF LOCK  | USER LOCK |\n"
+        "|  DEVICE POINTER  |    SIZE    |  AF LOCK  | USER LOCK |\n"
```
umar456: This is not strictly true for a memory manager that handles pinned memory.

r-barnes: @umar456: But it's also the case that the pointer in question is not, say, the `af_array` - which I thought for a while it might be. If not here, I wonder if there's some other place to clarify what pointer this message refers to?

9prady9: Perhaps, something like `DEVICE/PINNED-MEMORY POINTER` I know it is long but it caters to both the situations.

If not this, we can add a note to after horizontal line in line 296. something like the following.
```c
printf("---------------------------------------------------------\n"); // This is already there at end of table
printf("POINTER here is either device or pinned memory pointer\n");
printf("---------------------------------------------------------\n"); // new horizontal line
```
something along those lines.

I like the second approach personally.

umar456: @9prady9 I think that sort of information would be in the documentation rather than printed every time.

I understand your concern but this would also be misleading.

9prady9: Some times the proximity of this info helps a round trip to the docs or code, given that it is just few words.

r-barnes: I don't think the table itself is actually documented anywhere and, as a new user, I did not know how to interpret any of the headers.
I'll amend the pull request to document `AF LOCK  | USER LOCK` somewhere and can document pointer at the same time.

Is `DEVICE/PINNED-MEM POINTER` or `DEVICE/PINNED-MEM PTR` still legible while carrying more information?

umar456: How about DATA or BUFFER POINTER instead of DEVICE POINTER?

r-barnes: Maybe? It's very clear to me what "device pointer" means because it's used all the time in CUDA.

r-barnes: I've added some documentation explaining the table (see c90fa9f).

include/af/device.h
```diff
@@ -363,17 +363,28 @@ extern "C" {
...
-    ///
-    /// \ingroup device_func_mem
-    ///
+    /**
+       Prints buffer details from the ArrayFire Device Manager.
+
+       The result is a table with several columns:
+
+        * POINTER:   The hex address of the array's device or pinned-memory
+                     pointer (depending on the backend)
```
9prady9: We can remove the "depending on backend" is not backend dependent per say

include/af/array.h
```diff
@@ -1580,7 +1580,7 @@ extern "C" {
        \ingroup method_mat
        @{
 
-       Get the use count of `af_array`
+       Get the reference count of \ref af_array
```
9prady9: :+1:

src/backend/common/MemoryManagerImpl.hpp
9prady9: Although you have added the table header description in the function documentation, you haven't changed the first column header name to just 'POINTER'

r-barnes: Fixed.

include/af/device.h
r-barnes: Fixed.

include/af/array.h
r-barnes: I can't find this in my current commit, so I think it is fixed.

PR #2617, opened 2019-08-21
Clarify array reference counting

r-barnes:


include/af/array.h
```diff
@@ -203,7 +203,9 @@ namespace af
 #endif
 #endif
         /**
-            Creates an array from an \ref af_array handle
+            Creates an array from an \ref af_array handle. Does not increment 
+            a reference counter: the array assumes it has ownership of the
```
9prady9: Nit: the array assumes ownership of the handle.

include/af/array.h
```diff
@@ -1573,7 +1574,9 @@ extern "C" {
     AFAPI af_err af_release_array(af_array arr);
 
     /**
-       Increments an \ref af_array reference count
+       Increments an \ref af_array reference count. The `out` pointer
+       need not be retained in situations where the reference count
+       is the only effect needed.
```
9prady9: I am confused here, what is out pointer retain ?

r-barnes: The function has two side effects. One is to set the value of the `out` pointer, which is the function's first argument. The other is to increment the `af_array`'s reference count. In situations where the only desired effect is the reference count, the `out` pointer can be tossed without further side-effects occurring.

include/af/array.h
r-barnes: Can fix.

r-barnes:
I'm not sure why `ci/win-opencl-k5000` is failing to build since these changes affect only comments.

umar456:
Thats a known bug in clBlast. We can ignore that failure for this PR.

include/af/array.h
r-barnes: I don't think this was strictly true, so I've removed this text.

include/af/array.h
r-barnes: Fixed.

PR #2621, opened 2019-08-22
Use a comma instead of a colon to match documentation

r-barnes:
Simple syntax fix


9prady9:
For future visitors: The delimiter in this case doesn't actually doesn't play an active role. As long as the required values are a substring of the entire string set to the environment variable, the tracing will work fine. Hence, even though the documentation had conflicting information, the functionality wasn't affected.

PR #2622, opened 2019-08-22
Clarify what elements() does

r-barnes:


include/af/array.h
```diff
@@ -572,7 +572,7 @@ namespace af
         af_array get() const;
 
         /**
-           get the number of elements in array
+           Get the number of elements in array. This is the product of the array's dimensions.
```
9prady9: perhaps "Get the total number of elements across all dimensions of the array" is more succinct ?

And move the next statement as a note `\note This is the product of the array's dimensions` - Although I think this is not required given that is very basic thing with multidimensional arrays, especially if we rephrase it to the above.

r-barnes: Fixed.

include/af/array.h
```diff
@@ -1631,7 +1631,8 @@ extern "C" {
         @{
     */
     /**
-        \brief Gets the number of elements in an array.
+        \brief Gets the number of elements in an array. This is the product
+        of the array's dimensions.
```
9prady9: @r-barnes You forgot to change this one.

r-barnes: Fixed.

9prady9:
Thank you!

PR #2626, opened 2019-08-27
Add several workarounds for older CMake versions

umar456:
* Fixes errors with cublas_devices in some versions of CMake
* Remove compiler check header creation to support older CMake
* Workaround for compile definitions in the CUDA backend

### DIFF
--- a/src/backend/opencl/blas.cpp
+++ b/src/backend/opencl/blas.cpp
@@ -24,10 +24,7 @@
 
 // Includes one of the supported OpenCL BLAS back-ends (e.g. clBLAS, CLBlast)
 #include <magma/magma_blas.h>
-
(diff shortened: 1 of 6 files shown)


src/backend/opencl/blas.cpp
```diff
@@ -24,10 +24,7 @@
 
 // Includes one of the supported OpenCL BLAS back-ends (e.g. clBLAS, CLBlast)
 #include <magma/magma_blas.h>
-
-#if defined(WITH_LINEAR_ALGEBRA)
 #include <cpu/cpu_blas.hpp>
-#endif
```
9prady9: Shouldn't `gemm_fallback` definition and specilization be placed within these guards instead of removing this guard here. We would be compiling with those headers even when we won't use build with linear algebra.

Issue #1994, opened 2017-11-21
Build error: OpenCL with installed Boost greater or equal to 1.61


meister87:
Hi,
In the following opencl-based CMakeFiles the target Boost::boost is used:
- src/backend/opencl/kernel/sort_by_key/CMakeLists.txt
- src/backend/opencl/kernel/scan_by_key/CMakeLists.txt
- src/backend/opencl/CMakeLists.txt

The target Boost::boost is only defined in the following file:
- CMakeModules/build_boost_compute.cmake

This file is only included in the following file:
- src/backend/opencl/CMakeLists.txt

```
if("${Boost_VERSION}" VERSION_LESS 106100)
  dependency_check(Boost_FOUND "Boost not found.")
  include(build_boost_compute)
endif()
```
Hence, if already a boost version greater than 1.61 is installed, the target will never be defined and the build fails.

9prady9:
This has been addressed in 3.6 release.

PR #2641, opened 2019-09-08
Use Boost_VERSION_STRING to detect v1.71.0 correctly

### LINKED ISSUE #2640: Boost v1.71.0 not recognized as greater than v1.70.0
Output from `cmake`:

```
WARN: Found Boost v1.71.  Required 1.70.0.  Build will download Boost
```

The error is due to int version missing trailing 00. Technically likely a problem with `cmake`, but the easy fix may to be to always use the string-style version compare. Issue #1994 was closed saying that this would be addressed in the v3.6 release.

A pull request will be inbound shortly.

kriswuollett:
- Fixes #2640, earlier similar reference is #1994.

### DIFF
--- a/CMakeModules/boost_package.cmake
+++ b/CMakeModules/boost_package.cmake
@@ -7,7 +7,7 @@
 
 find_package(Boost)
 
-if("${Boost_VERSION}" VERSION_LESS 107000)
+if("${Boost_VERSION_STRING}" VERSION_LESS 1.70.0)
   set(VER 1.70.0)
   set(MD5 e160ec0ff825fc2850ea4614323b1fb5)
   include(ExternalProject)


umar456:
Thanks for creating a PR regarding this issue. Unfortunately that variable was only added in CMake version 3.14. As much as I want to update the CMake requirements to the latest version, this will exclude several recent distros including Ubuntu 19.04(3.13), Debian 10(3.13), and Visual Studio 2017(3.12). Currently we are targeting 3.5 which is included with Ubuntu 16.04.

Sounds like CMake is transitioning away from the FindBoost module starting from version 1.70 (https://gitlab.kitware.com/cmake/cmake/issues/19402). Perhaps a better solution would be to transition to the new CMake configuration file provided by boost as we are targeting a fairly new version of boost anyway.

umar456:
Closing. Superseded by #2735

PR #2653, opened 2019-10-15
Add support for f16 mean var and meanvar

WilliamTambellini:
WIP : not ready

### DIFF
--- a/src/api/cpp/mean.cpp
+++ b/src/api/cpp/mean.cpp
@@ -15,7 +15,6 @@
 #include "error.hpp"
 #include "half.hpp"
 #ifdef AF_CUDA
-// NOTE: Adding ifdef here to avoid copying code constructor in the cuda backend
 #include <cuda_fp16.h>
 #include <traits.hpp>
 #endif
@@ -35,25 +34,12 @@ array mean(const array& in, const array& weights, const dim_t dim) {
     return array(temp);
 }
 
-template<typename To, typename T>
-To cast(T in) {
-    return static_cast<To>(in);
-}
-
-template<>
-af_half cast<af_half, double>(double in) {
-    half_float::half tmp = static_cast<half_float::half>(in);
-    af_half out;
-    memcpy(&out, &tmp, sizeof(af_half));
-    return out;
-}
-
 #define INSTANTIATE_MEAN(T)                                                  \
     template<>                                                               \
     AFAPI T mean(const array& in) {                                          \
         double ret_val;                                                      \
         AF_THROW(af_mean_all(&ret_val, NULL, in.get()));                     \
-        return cast<T>(ret_val);                                      \
+        return cast<T>(ret_val);                                             \
     }                                                                        \
     template<>                                                               \
     AFAPI T mean(const array& in, const array& wts) {                        \
@@ -101,7 +87,7 @@ INSTANTIATE_MEAN(unsigned long long);
 INSTANTIATE_MEAN(short);
 INSTANTIATE_MEAN(unsigned short);
 INSTANTIATE_MEAN(af_half);
-INSTANTIATE_MEAN(half_float::half); // Add support for public API
+INSTANTIATE_MEAN(half_float::half);  // Add support for public API
 #ifdef AF_CUDA
 INSTANTIATE_MEAN(__half);
 #endif
--- a/src/api/cpp/mean.cpp
+++ b/src/api/cpp/mean.cpp
@@ -13,6 +13,12 @@
 #include <af/statistics.h>
 #include "common.hpp"
 #include "error.hpp"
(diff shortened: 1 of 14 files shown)


src/backend/cuda/mean.cu
```diff
@@ -53,6 +55,8 @@ Array<T> mean(const Array<T>& in, const Array<Tw>& wts, const int dim) {
 
 INSTANTIATE(double, double, double);
 INSTANTIATE(float, float, float);
+INSTANTIATE(half, half, half);
+INSTANTIATE(half, float, half);
```
WilliamTambellini: strange but really needed for some other collateral algos : if not then linking issues: cannot find mean<half, float, half>

WilliamTambellini:
Could anyone check why "All checks have failed" ? (I cannot access https://ci.arrayfire.org:8010)

umar456:
@9prady9 This is ready for review.

src/api/cpp/mean.cpp
```diff
@@ -13,6 +13,12 @@
 #include <af/statistics.h>
 #include "common.hpp"
 #include "error.hpp"
+#include "half.hpp"
+#ifdef AF_CUDA
+// NOTE: Adding ifdef here to avoid copying code constructor in the cuda backend
```
9prady9: `code` constructor ?

src/api/cpp/mean.cpp
```diff
@@ -34,13 +40,13 @@ array mean(const array& in, const array& weights, const dim_t dim) {
     AFAPI T mean(const array& in) {                                          \
         double ret_val;                                                      \
         AF_THROW(af_mean_all(&ret_val, NULL, in.get()));                     \
-        return (T)ret_val;                                                   \
+        return cast<T>(ret_val);                                      \
```
9prady9: Identation seems to be different from other lines on this one.

src/api/cpp/var.cpp
```diff
@@ -12,6 +12,12 @@
 #include <af/statistics.h>
 #include "common.hpp"
 #include "error.hpp"
+#include "half.hpp"
+#ifdef AF_CUDA
+// NOTE: Adding ifdef here to avoid copying code constructor in the cuda backend
```
9prady9: same as earlier, I didn't understand what constructor ?

src/backend/cpu/kernel/mean.hpp
```diff
@@ -108,9 +109,10 @@ struct mean_dim<Ti, Tw, To, 0> {
         To* out            = output.get();
 
         dim_t istride = istrides[dim];
-        MeanOp<Ti, To, Tw> Op(0, 0);
-        for (dim_t i = 0; i < idims[dim]; i++) {
-            Op(in[inOffset + i * istride], 1);
+        dim_t end     = inOffset + idims[dim] * istride;
+        MeanOp<compute_t<Ti>, compute_t<To>, compute_t<Tw>> Op(0, 0);
```
9prady9: Shouldn't this start with the first element since you are starting with non-zero offset in the for loop below ?

src/backend/cpu/mean.cpp
```diff
@@ -12,10 +12,14 @@
 #include <mean.hpp>
 #include <platform.hpp>
 #include <queue.hpp>
+#include <types.hpp>
 #include <af/dim4.hpp>
 #include <complex>
 
+#include <common/half.hpp>
```
9prady9: this is our internal header, we can move it into above header club.

test/mean.cpp
```diff
@@ -196,10 +195,34 @@ void meanAllTest(T const_value, dim4 dims) {
...
+
+    vector<float> hundred(dims.elements(), const_value);
+
+    float gold = float(0);
+    for (int i = 0; i < (int)hundred.size(); i++) { gold = gold + hundred[i]; }
+    gold = gold / dims.elements();
+
+    array a = array(dims, &(hundred.front())).as(f16);
+    half output = mean<half>(a);
+    af_half output2 = mean<af_half>(a);
```
9prady9: Is this supposed to be checked against `output` ?

test/mean.cpp
```diff
@@ -221,6 +244,14 @@ T random() {
     return T(std::rand() % 10);
 }
 
+template<>
+half random<half>() {
+    // create values from -0.5 to 0.5 to ensure sum does not deviate
+    // too far out of half's useful range
```
9prady9: 👍

src/backend/cpu/kernel/mean.hpp
umar456: The offset determines the start of the axis where you want to perform the mean. You still want the mean to start anew from that point.

PR #2661, opened 2019-11-04
Locate cudnn header for cmake include command

### LINKED ISSUE #2660: cmake not searching/checking the cudnn header file
Looks like the afcuda cmake is not searching/checking that the required cudnn.h file is available: 

cmake
success
make afcuda
...
In file included from /local1/wtambellini/arrayfire/src/backend/cuda/convolve.cpp:15:0:
/local1/wtambellini/arrayfire/src/backend/cuda/cudnn.hpp:13:19: fatal error: cudnn.h: No such file or directory
 #include <cudnn.h>

Expected behavior : cmake to at least fail if cudnn.h not findable in usual suspect dirs (/usr/local/cuda/include, ...)

Ref:
https://github.com/soumith/cudnn.torch/blob/master/cmake/FindCUDNN.cmake

9prady9:
Resolves #2660

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -13,6 +13,13 @@ dependency_check(CUDA_FOUND "CUDA not found.")
 find_cuda_helper_libs(nvrtc)
 find_cuda_helper_libs(nvrtc-builtins)
 find_cuda_helper_libs(cudnn)
+get_filename_component(__found_cudnn_root ${CUDA_cudnn_LIBRARY} PATH)
+find_path(CUDA_cudnn_INCLUDE_DIR
+    NAMES cudnn.h
+    HINTS ${__found_cudnn_root} ${CUDNN_ROOT_DIR} ${CUDA_TOOLKIT_INCLUDE}
+    PATH_SUFFIXES include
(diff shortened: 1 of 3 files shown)


src/backend/cuda/CMakeLists.txt
```diff
@@ -13,6 +13,13 @@ dependency_check(CUDA_FOUND "CUDA not found.")
 find_cuda_helper_libs(nvrtc)
 find_cuda_helper_libs(nvrtc-builtins)
 find_cuda_helper_libs(cudnn)
+get_filename_component(__found_cudnn_root ${CUDA_cudnn_LIBRARY} PATH)
+find_path(CUDA_cudnn_INCLUDE_DIR
```
WilliamTambellini: cool. Should nt we add 'required' somewhere ?

9prady9: The `REQUIRED` keyword works for only `find_package`. There is not official Find module for CUDNN from what I found from cmake docs. I am adding a check based on outcome of `find_path` , that should do the job. Have you tried the change on your system ?

WilliamTambellini: okok tks

PR #2668, opened 2019-11-12
Forward generator platform to external projects on windows

### LINKED ISSUE #2655: building failure with clFFT-ext missing
I was trying to build the arrayfire source code by strictly following the instructions offered in the wiki except that I used VS2019 Build Tools instead of VS2017. It has been said that clfft is automatically downloaded by Cmake.
    When the building was nearly finished, however, I saw 

> CUSTOMBUILD : CMake error : The following variables are used in this project, but they are set to NOTFOUND.[D:\Code\arrayfire\build\src\backend\opencl\clFFT-ext.vcxproj]

9prady9:
Resolves #2655 

On windows for multi-platform generators such as Visual Studio, without forwarding the platform string to external projects, cmake is choosing x86 default platform. Hence, the error some times.

I have observed the error especially with latest cmake using Visual Studio 16 2019 latest editions.

### DIFF
--- a/CMakeModules/build_clBLAS.cmake
+++ b/CMakeModules/build_clBLAS.cmake
@@ -12,6 +12,12 @@ set(clBLAS_location ${prefix}/lib/import/${CMAKE_STATIC_LIBRARY_PREFIX}clBLAS${C
 
 find_package(OpenCL)
 
+if(WIN32)
+  set(extproj_gen_opts "-G${CMAKE_GENERATOR}" "-A${CMAKE_GENERATOR_PLATFORM}")
+else(WIN32)
+  set(extproj_gen_opts "-G${CMAKE_GENERATOR}")
+endif(WIN32)
+
 ExternalProject_Add(
     clBLAS-ext
     GIT_REPOSITORY https://github.com/arrayfire/clBLAS.git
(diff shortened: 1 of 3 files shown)


CMakeModules/build_clBLAS.cmake
```diff
@@ -21,7 +27,8 @@ ExternalProject_Add(
     INSTALL_DIR "${prefix}"
     UPDATE_COMMAND ""
     DOWNLOAD_NO_PROGRESS 1
-    CONFIGURE_COMMAND ${CMAKE_COMMAND} "-G${CMAKE_GENERATOR}" -Wno-dev <SOURCE_DIR>/src
+    CONFIGURE_COMMAND ${CMAKE_COMMAND} ${extproj_gen_opts}
+      -Wno-dev <SOURCE_DIR>/src
```
umar456: Can't you just set `-DCMAKE_GENERATOR`? Setting the configure command seems like a hack.

9prady9: `-A` is valid option for `cmake` command and it can't take empty string on non-Windows platform (verified that on Linux). Without forwarding this generator platform for `Visual Studio 16 2019`, cmake is choosing `x86` as platform target even if ArrayFire's platform target is x64. Hence, causing compilation failures due to mismatch in target platforms between third party projects and ArrayFire project. I have already explained this briefly in PR description.

We are conditionally forwarding value of `CMAKE_GENERATOR_PLATFORM` variable to external projects, I don't understand how that is a hack!

9prady9:
Awaiting test with Ninja generator on Windows
TODO(pradeep).

PR #2677, opened 2019-11-21
Disable events by default. Add enable events flag

### LINKED ISSUE #2673: AF unified master up to 30% slower than 3.6.4
Hello,
I have to call that as today with 
c30d5455f3d60582cb81968551b69fadcfc7c880
AF master is up to 30% slower than 3.6.4 in some cases.
For some (non-batched) NN inference : 
3.6.4  2.81 secs
3.7.0  3.99 secs

One of the root causes is definitely the new event creation introduced in
https://github.com/arrayfire/arrayfire/pull/2526
The fix in 
https://github.com/arrayfire/arrayfire/pull/2574
just reduced the speed drop but did nt totally fix it.
Commenting out 
 auto err = cuEventCreate(e, CU_EVENT_DISABLE_TIMING); 
in 
https://github.com/arrayfire/arrayfire/blob/c30d5455f3d60582cb81968551b69fadcfc7c880/src/backend/cuda/Event.hpp#L26
reduced further the speed drop : 
3.7.0 (no cuda event) : 3.467 secs
but still not as fast as 3.6.4.

TBContinued
W.

umar456:
This commit disables the use of events but does not remove the
events based code from the memory manager. This is required because
it is not necessary as of now and there is a performance penalty
for using events. This may be enabled in the future once we begin
using streams where events will be necessary.

Fixes #2673

### DIFF
--- a/include/af/internal.h
+++ b/include/af/internal.h
@@ -187,6 +187,23 @@ extern "C"
     AFAPI af_err af_get_allocated_bytes(size_t *bytes, const af_array arr);
 #endif
 
+#if AF_API_VERSION >= 37
+    /**
+       Enables and disables(default) the use of events in the memory manager
+       for synchronization. Enabling events incurrs a performance penilty
+       but allows you to use streams(not currently supported).
+
+       \note The default for this variable may change in future versions of
+             ArrayFire
+
+       \param[in] enabled If true, events will be used in the memory manager
+       \returns AF_SUCCESS
+
+       \ingroup internal_func_eventsbasedmemorymanager
+    */
+    AFAPI af_err af_set_use_events_based_memory_manager(int enable);
+#endif
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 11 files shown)


pavanky:
why not have `make_event` check for `getEventsFlag`  implicitly ? If disabled, just return `Event()` else create one based on the Queue? 

This makes it easier to add events in other locations in the future.

umar456:
I thought that would be misleading. If we decide to use events in another part of the code base it could cause issues. Its just a problem in the memory manager because it is such a fundamental part of the library and it is called so frequently.

9prady9:
@WilliamTambellini Can you please verify if the performance slow down you notice in #2673 is resolved with this change. Thank you.

WilliamTambellini:
Tested the PR and good for me.
Are there some CI failures though ?

9prady9:
Yeah, we are looking into the ci related stuff. It is definitely not related to this PR by itself.

WilliamTambellini:
this one could be droped, is nt it ?

PR #2678, opened 2019-11-24
Ensure query and info functions do not fail if there is no device

umar456:
Some of the info and query functions failed when called on a system
with no OpenCL runtime or graphics drivers installed. These functions
should be resilient to these situations and provide feed back to the
user about possible next steps. The following functions are able to
function even if there is not device or driver installed.

af_info
af_info_string
af_get_device_count
af_get_size_of
af_get_backend_count
af_get_version
af_get_revision

This also includes their respective C++ functions.

Added unit tests.

### DIFF
--- a/src/backend/cuda/device_manager.cpp
+++ b/src/backend/cuda/device_manager.cpp
@@ -473,35 +473,52 @@ DeviceManager::DeviceManager()
     , cuDevices(0)
     , nDevices(0)
     , fgMngr(new graphics::ForgeManager()) {
-    checkCudaVsDriverVersion();
+    try {
+        checkCudaVsDriverVersion();
 
-    CUDA_CHECK(cudaGetDeviceCount(&nDevices));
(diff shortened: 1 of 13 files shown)


src/backend/cuda/device_manager.cpp
```diff
@@ -352,37 +352,53 @@ DeviceManager::DeviceManager()
...
-            continue;
+    , fgMngr(new graphics::ForgeManager()) {
+    try {
+        checkCudaVsDriverVersion();
+
+        CUDA_CHECK(cudaGetDeviceCount(&nDevices));
+        AF_TRACE("Found {} CUDA devices", nDevices);
+        if (nDevices == 0) {
+            AF_ERROR("No CUDA capable devices found", AF_ERR_DRIVER);
+            return;
```
9prady9: nit: `return` valid and won't hurt, but we can remove it.

Issue #2676, opened 2019-11-21
How to use borderType AF_PAD_SYM for convolve operations?


krish021195:
Hi,

The convolve operations in ArrayFire seems to use the AF_PAD_ZERO borderType. How to change this borderType to AF_PAD_SYM?

Thanks

9prady9:
Currently, border type can be set only for median, min and max filters only. Convolutions only do zero padding as of now.

krish021195:
Can it be expected in the next release?

For a workaround , Trying to provide a border replicated array to the convolve function. 
My method seems to take more time for just replicating the border.

Is the process of replicating the borders done directly in cuda kernel of median or min or max filters? Is there any simple way of replicating the borders in ArrayFire?

9prady9:
Unfortunately, we don't have it in the line for the next release.

http://arrayfire.org/docs/indexing.htm You can use indexing feature to create padding with the values of your choice.

Can you please share the code that you used. It would help to see whats exactly is going from code.

krish021195:
I am trying to use indexing feature for padding aswell. But it seems to take more time , the time taken for padding nearly equals the time for convolution operation with the padded array.

I have used the below code for padding,

`
array ar_A = constant(0,1924,1084,3,2,u8);
array ar_B = randu(1920,1080,3,2,u8);

ar_A(seq(2,ar_A.dims(0)-3,1),seq(2,ar_A.dims(1)-3,1)) = ar_B;
ar_A(0,seq(2,ar_A.dims(1)-3,1)) += ar_B(2,span);
ar_A(1,seq(2,ar_A.dims(1)-3,1)) += ar_B(1,span);
ar_A(ar_A.dims(0)-1,seq(2,ar_A.dims(1)-3,1)) += ar_B(ar_B.dims(0)-3,span);
ar_A(ar_A.dims(0)-2,seq(2,ar_A.dims(1)-3,1)) += ar_B(ar_B.dims(0)-2,span);
ar_A(span,0) += ar_A(span,4);
ar_A(span,1) += ar_A(span,3);
ar_A(span,ar_A.dims(1)-1) += ar_A(span,ar_A.dims(1)-5);
ar_A(span,ar_A.dims(1)-2) += ar_A(span,ar_A.dims(1)-4);
`

krish021195:
@9prady9 , Any suggestions on how to improve the timings of the above code?

9prady9:
[...]

PR #2682, opened 2019-12-03
FEAT: function to pad array borders

9prady9:
Addresses #2676 partially - allows the user to pad inputs and pass to convolve.

Performance comparison of indexing based padding vs kernel in this PR for a HD image is posted 
https://github.com/arrayfire/arrayfire/issues/2676#issuecomment-561073244

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -264,25 +264,25 @@ typedef enum {
 } af_interp_type;
 
 typedef enum {
-  ///
-  /// Out of bound values are 0
-  ///
-  AF_PAD_ZERO = 0,
+    ///
+    /// Out of bound values are 0
+    ///
+    AF_PAD_ZERO = 0,
 
-  ///
-  /// Out of bound values are symmetric over the edge
-  ///
-  AF_PAD_SYM,
+    ///
+    /// Out of bound values are symmetric over the edge
+    ///
+    AF_PAD_SYM,
 
-  ///
-  /// Out of bound values are clamped to the edge
-  ///
-  AF_PAD_CLAMP_TO_EDGE,
+    ///
+    /// Out of bound values are clamped to the edge
+    ///
+    AF_PAD_CLAMP_TO_EDGE,
 
-  ///
-  /// Out of bound values are mapped to range of the dimension in cyclic fashion
-  ///
-  AF_PAD_PERIODIC
+    ///
+    /// Out of bound values are mapped to range of the dimension in cyclic fashion
+    ///
+    AF_PAD_PERIODIC
 } af_border_type;
 
 typedef enum {
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -280,7 +280,7 @@ typedef enum {
   AF_PAD_CLAMP_TO_EDGE,
 
   ///
-  /// Out of bound values are clamped to the edge
+  /// Out of bound values are mapped to range of the dimension in cyclic fashion
   ///
   AF_PAD_PERIODIC
 } af_border_type;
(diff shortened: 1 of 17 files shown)


sebastienleclaire:
If possible, it would be nice to support periodic padding as I always manually pad periodic image before convolve. It It is slow and memory consuming. I think my  application would benefit from this kind of ability.

9prady9:
@sebastienleclaire That shouldn't be hard I think. I wouldn't be able to say much about performance unless we run nvprof on it.

sebastienleclaire:
For information, this is the function I use for periodic (circularly) padding (it could be use to write and debug new code):

array paddingArrayCircularly(const array& arrayToPad,
		bool isPaddingDim0, bool isPaddingDim1, bool isPaddingDim2, bool isPaddingDim3)
	{
		int dim0 = arrayToPad.dims(0);
		int dim1 = arrayToPad.dims(1);
		int dim2 = arrayToPad.dims(2);
		int dim3 = arrayToPad.dims(3);

		array id0 = isPaddingDim0 ? join(0, seq(dim0 - 1, dim0 - 1), seq(0, dim0 - 1), seq(0, 0)) : seq(0, dim0 - 1);
		array id1 = isPaddingDim1 ? join(0, seq(dim1 - 1, dim1 - 1), seq(0, dim1 - 1), seq(0, 0)) : seq(0, dim1 - 1);
		array id2 = isPaddingDim2 ? join(0, seq(dim2 - 1, dim2 - 1), seq(0, dim2 - 1), seq(0, 0)) : seq(0, dim2 - 1);
		array id3 = isPaddingDim3 ? join(0, seq(dim3 - 1, dim3 - 1), seq(0, dim3 - 1), seq(0, 0)) : seq(0, dim3 - 1);

		return  arrayToPad(id0, id1, id2, id3);
	}

I'm pretty sure that whatever you come up with will be faster and will consume less memory. As it stand right now, This function use approximately twice the memory inside the array "arrayToPad" to work.

9prady9:
@sebastienleclaire For the same sizes as you tried, periodic padding has the following global read/write efficiency.

```
    Kernel: void cuda::padBorders<float, af_border_type=3>(cuda::Param<float>, cuda::CParam<cuda::Param>, int, int, int, int, unsigned int, unsigned int)
         25                            gld_efficiency             Global Memory Load Efficiency      79.64%      79.64%      79.64%
         25                            gst_efficiency            Global Memory Store Efficiency      88.75%      88.75%      88.75%
```

sebastienleclaire:
I am not sure what the above mean (gld_efficiency and gst_efficiency), but thank you for the periodic padding!

9prady9:
That is global memory read and write efficiency metric names.

include/af/defines.h
```diff
@@ -264,20 +264,25 @@ typedef enum {
...
+  ///
+  AF_PAD_SYM,
+
+  ///
+  /// Out of bound values are clamped to the edge
+  ///
+  AF_PAD_CLAMP_TO_EDGE,
+
+  ///
+  /// Out of bound values are clamped to the edge
```
syurkevi: Wrong docstring

syurkevi: also, 4 spaces in the rest of the file

src/backend/cpu/kernel/pad_array_borders.hpp
```diff
@@ -8,9 +8,12 @@
  ********************************************************/
 
 #pragma once
+
 #include <Param.hpp>
+#include <utility.hpp>
 
 #include <algorithm>
+#include <cstdlib>
```
syurkevi: What is this header used for?

syurkevi:
👍 👍 
Need this for a client project. Adding 3.7 milestone.

PR #2686, opened 2019-12-04
Update dev props and driver arrays for cuda 10.2

9prady9:


src/backend/cuda/device_manager.cpp
```diff
@@ -52,7 +52,7 @@ using std::stringstream;
 
 namespace cuda {
 
-void findJitDevCompute(pair<int, int>& prop) {
+void checkAndSetDevMaxCompute(pair<int, int>& prop) {
```
9prady9: I changed this to reflect what the function is doing in a better way. It was misleading earlier.

src/backend/cuda/device_manager.cpp
```diff
@@ -396,7 +398,7 @@ DeviceManager::DeviceManager()
         if (i < nDevices) {
             auto prop = make_pair(cuDevices[i].prop.major,
                     cuDevices[i].prop.minor);
-            findJitDevCompute(prop);
+            checkAndSetDevMaxCompute(prop);
```
umar456: Can you make sure its more resilient to future toolkit updates?

9prady9: Please check now

9prady9:
I would prefer to keep the two commits separately. However, I wouldn't mind if they are squashed.

src/backend/cuda/device_manager.cpp
```diff
@@ -54,29 +54,62 @@ using std::stringstream;
...
+            "Runtime cuda version not found in toolkit info array."
+            "Current device compute version (%d.%d) exceeds supported maximum "
+            "runtime cuda compute version (%d.%d) of latest known cuda toolkit."
+            "Using %d.%d.";
+        fprintf(stderr, errMsg, originalCompute.first, originalCompute.second,
+                prop.first, prop.second, prop.first, prop.second);
+#endif
+    } else if (prop.first < 3) {
+        // all compute versions prior to Kepler
+        prop = make_pair(3, 0);
```
umar456: I think its fine if we not handle this case and not change the prop values. As I said in my other commit. this will be different in different versions of the toolkits.

PR #2689, opened 2019-12-10
Refactor memory manager. Remove buffer_info. Simplify allocs. Disable Events

umar456:
* Removed buffer_info from the allocation calls

The buffer_info object was created so that the event synchonization was
performed in the memAlloc call instead of in the memory manager. It did not have
significant benefit over performing the synchronization within the memory
manager and increased complexity of the API. Now that the events API is
exposed, it shouldn't be necessary to include this.

* Moved memory manager classes into separate files in the common directory
* Use memAlloc instead of bufferAlloc in OpenCL Array<T> initializtion to
  provide type context to the memory manager. Avoid inconsistencies between
  backends.
* Use objects instead of handles within the codebase


src/backend/common/AllocatorInterface.hpp
```diff
@@ -0,0 +1,43 @@
...
+class logger;
+}
+namespace common {
+namespace memory {
+
+/**
+ * An interface that provides backend-specific memory management functions,
+ * typically calling a dedicated backend-specific native API. Stored, wrapped,
+ * and called by a MemoryManagerBase, from which calls to its interface are
+ * delegated.
```
9prady9: :+1:

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -37,20 +42,21 @@ void DefaultMemoryManager::cleanDeviceMemoryManager(int device) {
     // This vector is used to store the pointers which will be deleted by
     // the memory manager. We are using this to avoid calling free while
     // the lock is being held because the CPU backend calls sync.
-    vector<af_buffer_info> free_ptrs;
+    vector<void*> free_ptrs;
     size_t bytes_freed           = 0;
-    memory::memory_info &current = memory[device];
+    DefaultMemoryManager::memory_info &current = memory[device];
```
9prady9: nit: A lot of these refactors in this file could have been avoided via auto + using statement at the top.

src/backend/common/MemoryManagerBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+   private:
+    // A threshold at or above which JIT evaluations will be triggered due to
+    // memory pressure. Settable via a call to setMemoryPressureThreshold
+    float memoryPressureThreshold_{1.0};
+    // A backend-specific memory manager, containing backend-specific
+    // methods that call native memory manipulation functions in a device
+    // API. We need to wrap these since they are opaquely called by the
+    // memory manager.
+    std::unique_ptr<AllocatorInterface> nmi_;
```
9prady9: nit: its easier to understand something like `nativeMemInterface` and makes code more readable.

src/backend/opencl/Array.cpp
```diff
@@ -51,7 +51,7 @@ template<typename T>
 Array<T>::Array(dim4 dims)
     : info(getActiveDeviceId(), dims, 0, calcStrides(dims),
            (af_dtype)dtype_traits<T>::af_type)
-    , data(bufferAlloc(info.elements() * sizeof(T)), bufferFree)
+    , data(memAlloc<T>(info.elements()).release(), bufferFree)
```
9prady9: This has been bothering me too for a while now - two different ways to allocate buffers in OpenCL.

WilliamTambellini:
For one of my NN, the speed seems a little better than the master branch (as today):
master: 34s
this change: 32s 
but still far from 3.6.4: 23s

umar456:
Yes. This PR also disables events.

test/memory.cpp
```diff
@@ -988,18 +865,8 @@ TEST(MemoryManagerApi, E2ETest) {
...
-            ASSERT_EQ(payload->lastDims, af::dim4(bDim * b.numdims()));
-            ASSERT_EQ(payload->lastElementSize, sizeof(float));
-        } else {
-            // alloced as bytes
-            ASSERT_EQ(payload->lastDims,
-                      af::dim4(bDim * b.numdims() * sizeof(float)));
-            ASSERT_EQ(payload->lastElementSize, 1);
-        }
+        ASSERT_EQ(payload->lastDims, af::dim4(bDim * b.numdims()));
+        ASSERT_EQ(payload->lastElementSize, sizeof(float));
```
jacobkahn: 👍

src/backend/common/DefaultMemoryManager.hpp
```diff
@@ -0,0 +1,139 @@
...
+    bool jitTreeExceedsMemoryPressure(size_t bytes) override;
+
+   protected:
+    DefaultMemoryManager()                                  = delete;
+    ~DefaultMemoryManager()                                 = default;
+    DefaultMemoryManager(const DefaultMemoryManager &other) = delete;
+    // DefaultMemoryManager(const DefaultMemoryManager &&other) = default;
+    DefaultMemoryManager &operator=(const DefaultMemoryManager &other) = delete;
+    // DefaultMemoryManager &operator=(const DefaultMemoryManager &&other) =
+    //     default;
```
jacobkahn: remove or uncomment the move ctor/assignment op? Although this should be movable.

umar456:
I think the events could be handled in the memory manager without exposing it to ArrayFire. Earlier, I was thinking we could add the returned events and make it part of the Array<T> object. I don't think this is necessary because the memory manager can enqueue those waits before it returns the pointer. The end result would be the same.

With the events API exposed the memory manager can handle this sort of thing externally.

umar456:
@9prady9 converting from bufferAlloc to memAlloc would be a pretty large change. I think we should do that on another PR.

PR #2691, opened 2019-12-13
Fix missing headers

9prady9:
- Add missing header inclusions that are causing compilation issues
- Fix external dependencies cmake build scripts for Windows+ninja combo (MSVC generators take -A option)


9prady9:
I fake-forced to trigger ci.

PR #2695, opened 2019-12-19
Optimize unified backend. Remove exceptions in af_get_backend_id.

umar456:
* The unified backend calls af_get_backend_id which throw an exception
  if the af_array passed to it is not an initialized array. This occures
  when you create an af::array with a default constructor or if a move
  is performed. In those cases the af_get_backend_id was throwing an
  exception and catching it then returning the error code. This was
  causing the slowdown in the unified backend.
* Remove ARG_ASSERT calls on simple functions to avoid overhead
* Remove the unordered map from the call function.
* Optimize the CALL macro
* Move the backend handling code from the C API to the C++ API. The
  C API function call will behave like a regular function call.

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -29,6 +29,11 @@
 #include <traits.hpp>
 #endif
 
+#ifdef AF_UNIFIED
+#include <symbol_manager.hpp>
+#include <af/backend.h>
+#endif
+
 #include <memory>
 #include <stdexcept>
 #include <vector>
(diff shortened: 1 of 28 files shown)


src/api/cpp/array.cpp
```diff
@@ -250,9 +242,50 @@ INSTANTIATE(__half);
...
+                    static af_release_array_ptr func =
+                        reinterpret_cast<af_release_array_ptr>(
+                            common::getFunctionPointer(opencl_handle,
+                                                       "af_release_array"));
+                    func(get());
+                    break;
+                }
+                case AF_BACKEND_CUDA: {
+                    static auto cuda_handle = instance.getHandle();
+                    static af_release_array_ptr func =
```
9prady9: Are these statics helping that much ? Is `getFunctionPointer` that costly even now ?

src/api/unified/symbol_manager.hpp
```diff
@@ -133,16 +135,21 @@ bool checkArrays(af_backend activeBackend, T a, Args... arg) {
...
-#endif
+#define CALL(FUNCTION, ...)                                                   \
+    using af_func = std::add_pointer<decltype(FUNCTION)>::type;               \
+    static af_backend index_ =                                                \
+        unified::AFSymbolManager::getInstance().getActiveBackend();           \
+    static af_func func = (af_func)common::getFunctionPointer(                \
+        unified::AFSymbolManager::getInstance().getHandle(), __func__);       \
+    if (index_ !=                                                             \
+        unified::AFSymbolManager::getInstance().getActiveBackend()) {         \
+        index_ = unified::AFSymbolManager::getInstance().getActiveBackend();  \
```
9prady9: I understand initialization of `index_` is not an issue, but what happens when more than one thread call the same line of code ? This feels more like a `thread_local` instead of just static.

src/api/unified/symbol_manager.hpp
```diff
@@ -105,10 +80,37 @@ class AFSymbolManager {
...
+    } else {
+        return true;
+    }
+}
+
+bool checkArrays(af_backend activeBackend) {
+    UNUSED(activeBackend);
+    // Dummy
+    return true;
+}
```
9prady9: Is there a specific reason you moved implementations from cpp to header ? This would later cause issue if and when this header is included in more than on translation unit.

umar456: It's used in a template and unless if the implementation is in the header, the function will not be inlined.

src/api/cpp/array.cpp
umar456: Yes, I experimented with removing the static variables and it helps the runtime.

WilliamTambellini:
Hi @umar456 This PR seems cool. Is it ready to be tested on my side ?

umar456:
Yes. This is pretty much complete. I am working on optimizing it for multiple threads but I don't think you will see a benefit from that.

umar456:
My optimizations didn't offer any improvements. This PR is ready to review and merge.

WilliamTambellini:
sounds good. Bravo.

PR #2702, opened 2020-01-02
Add f16 support for topk

umar456:


src/backend/cpu/topk.cpp
```diff
@@ -60,13 +62,13 @@ void topk(Array<T>& vals, Array<unsigned>& idxs, const Array<T>& in,
                 partial_sort_copy(
                     idx_itr, idx_itr + in.strides()[1], kiptr, kiptr + k,
                     [ptr](const uint lhs, const uint rhs) -> bool {
-                        return ptr[lhs] < ptr[rhs];
+                        return compute_t<T>(ptr[lhs]) < compute_t<T>(ptr[rhs]);
```
9prady9: Probably some other backend header is including `types.hpp`. I don't see the header included in this source file but yet we still use `compute_t` alias in here. It may fail compilation in one or the other compiler, most  probably MSVC.

src/backend/opencl/kernel/reduce.hpp
```diff
@@ -295,15 +295,15 @@ To reduce_all(Param in, int change_nan, double nanval) {
...
+        Binary<compute_t<To>, op> reduce;
+        compute_t<To> out       = Binary<compute_t<To>, op>::init();
+        compute_t<To> nanval_to = scalar<compute_t<To>>(nanval);
 
         for (int i = 0; i < (int)in_elements; i++) {
-            To in_val = transform(h_ptr[i]);
+            compute_t<To> in_val = transform(h_ptr[i]);
             if (change_nan) in_val = IS_NAN(in_val) ? nanval_to : in_val;
-            out = reduce(out, in_val);
+            out = reduce(out, compute_t<To>(in_val));
```
9prady9: I do see that sort_index is used for implementing topk in OpenCL backend, so changes to that makes sense.

However, I don't see how these are related to this PR ? Are they a fix and accidentally clubbed together ? We need to have a unit test if it is a bug in reduce unit tests ?

umar456: If you want to compile the topk changes then these changes are also required. I am not able to compile without it.

src/backend/cuda/kernel/topk.hpp
```diff
@@ -56,15 +57,16 @@ static __global__ void kerTopkDim0(Param<T> ovals, Param<uint> oidxs,
     uint* ires = oidxs.ptr + by * oidxs.strides[1] + bz * oidxs.strides[2] +
                  bw * oidxs.strides[3];
 
-    T keys[TOPK_IDX_THRD_LOAD];
+    compute_t<T> keys[TOPK_IDX_THRD_LOAD];
     ValueType vals[TOPK_IDX_THRD_LOAD];
 
     for (uint li = 0, i = gx; li < TOPK_IDX_THRD_LOAD; i += gxStride, li++) {
         if (i < elements) {
             keys[li] = kdata[i];
```
WilliamTambellini: keys[li] = (compute_t<T>)kdata[i];
is a must have to compile under old gcc and/or cuda10.1

WilliamTambellini:
ok my unittests seem to pass with that branch. Dont know yet for the speed for worst case it s still possible to convert input to f32. Approving. Tks.

PR #2710, opened 2020-01-08
Add missing af::meanvar function. Fix bias parameter

umar456:
* Fixes the bias parameter from the API that accepted the bias parameter as a boolean.
* Adds missing meanvar C++ API function implementation.

### DIFF
--- a/src/api/c/var.cpp
+++ b/src/api/c/var.cpp
@@ -177,7 +177,7 @@ af_err af_var(af_array* out, const af_array in, const bool isbiased,
 
         af_array no_weights = 0;
         af_var_bias bias =
-            (isbiased) ? AF_VARIANCE_POPULATION : AF_VARIANCE_SAMPLE;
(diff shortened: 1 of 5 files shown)


src/api/c/var.cpp
```diff
@@ -177,7 +177,7 @@ af_err af_var(af_array* out, const af_array in, const bool isbiased,
 
         af_array no_weights = 0;
         af_var_bias bias =
-            (isbiased) ? AF_VARIANCE_POPULATION : AF_VARIANCE_SAMPLE;
+            (isbiased) ? AF_VARIANCE_SAMPLE: AF_VARIANCE_POPULATION;
```
WilliamTambellini: add any warning/note in the af.h header ?

9prady9: I think this sort of info should go into https://github.com/arrayfire/arrayfire/blob/master/docs/pages/release_notes.md and it will when we make a release.

PR #2720, opened 2020-01-14
Fix unified segfault if no backends are found

umar456:
Fix segfault in the unified backend caused by an infinite recursive call if no backends were found.

Refactored some error code logic.

### DIFF
--- a/src/backend/common/err_common.cpp
+++ b/src/backend/common/err_common.cpp
@@ -93,12 +93,13 @@ const string &DimensionError::getExpectedCondition() const { return expected; }
 
 int DimensionError::getArgIndex() const { return argIndex; }
 
-void print_error(const string &msg) {
+af_err set_global_error_string(const string &msg, af_err err) {
     std::string perr = getEnvVar("AF_PRINT_ERRORS");
     if (!perr.empty()) {
         if (perr != "0") fprintf(stderr, "%s\n", msg.c_str());
     }
     get_global_error_string() = msg;
+    return err;
 }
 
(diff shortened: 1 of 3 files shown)


src/backend/common/err_common.hpp
```diff
@@ -139,16 +139,14 @@ void print_error(const std::string& msg);
...
-        print_error(s.str());                                                \
-        return ERR_TYPE;                                                     \
+#define AF_RETURN_ERROR(MSG, ERR_TYPE)                                \
+    do {                                                              \
+        std::stringstream s;                                          \
+        s << "Error in " << __PRETTY_FUNCTION__ << "\n"               \
+          << "In file " << __AF_FILENAME__ << ":" << __LINE__ << ": " \
+          << MSG;                                                     \
+        set_error(s.str(), ERR_TYPE);                                 \
+        return ERR_TYPE;                                              \
```
9prady9: Last two lines could be merged to avoid double editing in future.
`return set_error(s.str(), ERR_TYPE)`

src/backend/common/err_common.cpp
```diff
@@ -93,12 +93,13 @@ const string &DimensionError::getExpectedCondition() const { return expected; }
 
 int DimensionError::getArgIndex() const { return argIndex; }
 
-void print_error(const string &msg) {
+af_err set_error(const string &msg, af_err err) {
```
9prady9: something like `set_global_error_string` is more understandable and less misleading.

PR #2726, opened 2020-01-22
Delegate all non-batch matmul to cuBLAS GEMM API

### LINKED ISSUE #2701: f16 vector-vector and matrix-vector afcuda matmul bug
Hello
vector-vector (kind of dotprod) does nt seem to work at half prec but does at f32 :
auto left = af::array(1, 1024, f16); 
auto right = af::array(1024, 1, f16);
af::matmul(left, right) : 

ArrayFire Exception (Function not supported:301):
In function af_err af_matmul(void**, af_array, af_array, af_mat_prop, af_mat_prop)
In file src/api/c/blas.cpp:241
In function af::array af::matmul(const af::array&, const af::array&, af::matProp, af::matProp)
In file src/api/cpp/blas.cpp:18

Note: gevm seems to work (1x1024 matmul 1024x1024).
Note: af::dot() only works for non batched 2d vectors. Would be better for matmul to be consistent between f32 and f16 anyway.

9prady9:
This also handles "Add gemv f16 support in matmul impl"

Resolves #2701

### DIFF
--- a/src/backend/cuda/blas.cpp
+++ b/src/backend/cuda/blas.cpp
@@ -14,13 +14,17 @@
 #include <platform.hpp>
 
 #include <arith.hpp>
+#include <cast.hpp>
 #include <common/err_common.hpp>
 #include <common/half.hpp>
 #include <complex.hpp>
+#include <copy.hpp>
 #include <cublas.hpp>
 #include <err_cuda.hpp>
 #include <math.hpp>
 #include <reduce.hpp>
+#include <tile.hpp>
+#include <transpose.hpp>
 
 #include <cassert>
 #include <functional>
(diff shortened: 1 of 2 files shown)


WilliamTambellini:
@9prady9  Cool. Have you looked the speed vs f32 inputs ?

src/backend/cuda/blas.cpp
```diff
@@ -312,16 +340,26 @@ void gemm(Array<T> &out, af_mat_prop optLhs, af_mat_prop optRhs, const T *alpha,
...
+            auto t1 = (lOpts == CUBLAS_OP_C ?
+                       transpose(lhs, true) :
+                       (lOpts == CUBLAS_OP_T ? transpose(lhs, false) : lhs));
+            auto t2 = (rOpts == CUBLAS_OP_C ?
+                       conj<T>(tRhs) :
+                       (rOpts == CUBLAS_OP_T ? tRhs : transpose(tRhs, false)));
+
+            auto product = arithOp<T, af_mul_t>(t1, t2, t1.dims());
+
+            out = reduce<af_add_t, T, T>(product, 1, false, 0);
```
umar456: The out array for this function is already allocated and the results from the reduced need to be copied into it. For this to work we will need to modify reduce which is not a great idea. For now lets allocate a new array and copy into it.

src/backend/cuda/blas.cpp
```diff
@@ -312,16 +340,26 @@ void gemm(Array<T> &out, af_mat_prop optLhs, af_mat_prop optRhs, const T *alpha,
...
+
+            //TODO Remove gemvDispatch implementations once
+            //     above code is verified and tested
+            //dim_t incr = (optRhs == AF_MAT_NONE) ? rStrides[0] : rStrides[1];
+            //gemvDispatch(out, lOpts, alpha, lhs, beta, rhs, incr);
```
umar456: Lets move this logic into gemvDispatch.

PR #2730, opened 2020-01-25
Set ctest dashboard based on ci build typer:push/pr

9prady9:


9prady9:
That's a valid change and that's the way to set environment variable from a job step to subsequent step

PR #2733, opened 2020-01-26
Make cuDNN a runtime dependency rather than a link time dependency

umar456:
* Load cuDNN at runtime
* Take into account different versions of cuDNN
* Add support for older cuDNN versions (Tested with 4, 6 and 7)

### DIFF
--- a/src/backend/common/DependencyModule.cpp
+++ b/src/backend/common/DependencyModule.cpp
@@ -47,7 +47,7 @@ namespace common {
 
 DependencyModule::DependencyModule(const char* plugin_file_name,
                                    const char** paths)
-    : handle(nullptr), logger(loggerFactory("platform")) {
+    : handle(nullptr), logger(common::loggerFactory("platform")) {
     // TODO(umar): Implement handling of non-standard paths
     UNUSED(paths);
     if (plugin_file_name) {
@@ -66,7 +66,6 @@ DependencyModule::DependencyModule(const vector<string> plugin_base_file_name,
                                    const vector<string> suffixes,
                                    const vector<string> paths)
     : handle(nullptr), logger(common::loggerFactory("platform")) {
-    UNUSED(paths);
     for (const string& base_name : plugin_base_file_name) {
         for (const string& path : paths) {
             for (const string& suffix : suffixes) {
--- a/src/backend/common/DependencyModule.cpp
+++ b/src/backend/common/DependencyModule.cpp
@@ -8,6 +8,7 @@
  ********************************************************/
 
 #include <common/DependencyModule.hpp>
+#include <common/Logger.hpp>
 #include <common/module_loading.hpp>
 #include <algorithm>
 #include <string>
@@ -33,6 +34,7 @@ static const char* libraryPrefix = "lib";
 #endif
 
 using std::string;
+using std::vector;
 
 namespace {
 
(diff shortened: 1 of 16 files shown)


src/backend/common/DependencyModule.hpp
```diff
@@ -42,7 +42,7 @@ class DependencyModule {
     /// Returns true if the module was successfully loaded
     bool isLoaded();
 
-    /// Returns true if the module was successfully loaded
+    /// Returns true if all of the symbols for the module were loaded
```
9prady9: :+1:

src/backend/common/util.hpp
```diff
@@ -17,5 +17,13 @@
 std::string getEnvVar(const std::string &key);
 
 // Dump the kernel sources only if the environment variable is defined
-static const char* saveJitKernelsEnvVarName = "AF_JIT_KERNEL_TRACE";
 void saveKernel(const std::string& funcName, const std::string& jit_ker, const std::string& ext);
+namespace {
+const char* saveJitKernelsEnvVarName = "AF_JIT_KERNEL_TRACE";
+
+std::string int_version_to_string(int version) {
```
9prady9: A comment on intended usage is useful. Are the values in the formulation (1000 , 10) always fixed ?

src/backend/cuda/CMakeLists.txt
```diff
@@ -525,6 +524,7 @@ target_include_directories (afcuda
     ${CMAKE_CURRENT_SOURCE_DIR}/kernel
     ${CMAKE_CURRENT_SOURCE_DIR}/jit
     ${CMAKE_CURRENT_BINARY_DIR}
+    #${CUDNN_INCLUDE_DIR}
```
9prady9: Interesting, why is this include commented instead of removal ?

src/backend/cuda/convolve.cpp
```diff
@@ -39,14 +39,20 @@ template<>
...
     return CUDNN_DATA_DOUBLE;
 }
+
+#if CUDNN_MAJOR >= 6
 template<>
 cudnnDataType_t getCudnnDataType<int>() {
     return CUDNN_DATA_INT32;
 }
+
+#elif CUDNN_MAJOR >= 6
```
9prady9: Why are the else and if conditionals same ?

src/backend/cuda/cudnn.hpp
```diff
@@ -10,22 +10,136 @@
...
+
+
+// cuDNN Wrappers
+//
+// cuDNN deprecates and releases function names often between releases. in order
+// to prevent locking arrayfire versions to specific cuDNN versions, we wrap all
+// cuDNN calls so that the main codebase is not full of ifdefs. The Following
+// functions are wrappers around cuDNN functions that abstract out the version
+// differences between older versions of cuDNN.
+//
```
9prady9: How do we handle this inside the wrapper call or the cuDNN module loader ? There doesn't seem to be any guards inside cuDNN module loader based on the cuDNN versions.

umar456:
Addressed all feedback

src/backend/cuda/convolve.cpp
```diff
@@ -39,14 +39,20 @@ template<>
...
     return CUDNN_DATA_DOUBLE;
 }
+
+#if CUDNN_MAJOR >= 6
 template<>
 cudnnDataType_t getCudnnDataType<int>() {
     return CUDNN_DATA_INT32;
 }
+
+#elif CUDNN_MAJOR > 7 || (CUDNN_MAJOR >= 7 && CUDNN_MINOR >= 1)
```
9prady9: We will never reach the code-path under this condition. I think you were aiming for a separate `#if` here.

src/backend/cuda/cudnn.cpp
```diff
@@ -27,14 +27,169 @@ const char *errorString(cudnnStatus_t err) {
...
+            convDesc, pad_h, pad_w, u, v, upscalex, upscaley, mode);
+#endif
 }
+
+cudnnStatus_t cudnnSetFilter4dDescriptor(cudnnFilterDescriptor_t filterDesc,
+                                         cudnnDataType_t dataType,
+                                         cudnnTensorFormat_t format, int k,
+                                         int c, int h, int w) {
+    int major;
+    tie(major, std::ignore, std::ignore) = getCudnnPlugin().getVersion();
```
9prady9: these two lines can be moved into the `#if CUDNN_MAJOR >= 6` arm.

src/backend/cuda/cudnn.cpp
```diff
@@ -27,14 +27,172 @@ const char *errorString(cudnnStatus_t err) {
...
+#if CUDNN_VERSION >= 6000
+    int version = getCudnnPlugin().cudnnGetVersion();
+    if (version >= 6000) {
+        return getCudnnPlugin().cudnnSetFilter4dDescriptor(filterDesc, dataType,
+                                                           format, k, c, h, w);
+    } else if (version == 4000) {
+        return getCudnnPlugin().cudnnSetFilter4dDescriptor_v4(
+            filterDesc, dataType, format, k, c, h, w);
+    }
+    CUDA_NOT_SUPPORTED(
```
9prady9: :+1:

src/backend/cuda/convolve.cpp
```diff
@@ -39,14 +39,21 @@ template<>
...
     return CUDNN_DATA_DOUBLE;
 }
+
+#if CUDNN_VERSION >= 6000
 template<>
 cudnnDataType_t getCudnnDataType<int>() {
     return CUDNN_DATA_INT32;
 }
+
+#if CUDNN_VERSION >= 7100
```
9prady9: :+1:

9prady9:
Strided convolution tests are failing on k5000.

src/backend/common/DependencyModule.cpp
```diff
@@ -45,14 +47,42 @@ namespace common {
...
+            AF_TRACE("Unable to open {}", plugin_file_name);
+        }
     }
 }
 
+DependencyModule::DependencyModule(const vector<string> plugin_base_file_name,
+                                   const vector<string> suffixes,
+                                   const vector<string> paths)
+    : handle(nullptr), logger(common::loggerFactory("platform")) {
+    UNUSED(paths);
```
9prady9: Must be a left over from copy-paste.

umar456:
The ASAN failure is because we are now dynamically loading cuDNN and we close it before we exit the executable. The close operation unloads all of the symbols and the leak sanitizer cannot find the symbols. You can get around this issue by LD_PRELOADing the library and address sanitizer and the leaks go away.

PR #2743, opened 2020-02-06
Several CMake fixes on Windows

### LINKED ISSUE #2623: cmake build issue
When I build arrayfire like so:
```
CXX=g++-6 CC=gcc-6 cmake -GNinja -DBUILD_TESTING=OFF -DAF_BUILD_EXAMPLES=OFF -DCMAKE_BUILD_TYPE=Release ..
sudo ninja install
```
I later try to use it in a project with CMAKE:

    find_package(ArrayFire REQUIRED)
    ...
    target_link_libraries(mytarget PRIVATE ArrayFire::afcuda)

And disaster strikes:
```
CMake Error at /usr/local/share/ArrayFire/cmake/ArrayFireConfig.cmake:131 (get_property):
  get_property given invalid argument "RELEASE".
Call Stack (most recent call first):
  src/CMakeLists.txt:6 (find_package)
```
So, I rebuild without specifying the build type:
```
CXX=g++-6 CC=gcc-6 cmake -GNinja -DBUILD_TESTING=OFF -DAF_BUILD_EXAMPLES=OFF  ..
sudo ninja install
```
And now things work as expected.

It seems as though using `CMAKE_BUILD_TYPE` creates a problem, tough it is recommended [here](https://github.com/arrayfire/arrayfire/wiki/Build-Instructions-for-Linux).

umar456:
Fixed several issues when building on Windows.

* For certain generators the CMAKE_GENERATOR_PLATFORM was not set and caused an error due to changes in #2736 
* Release builds do not generate PDB files but the install script failed because it couldn't find them
* MKL sets the LIB environment variable instead of the LIBRARY_PATH environment variable on linux
* Fixed a ArrayFireConfig.cmake issue caused by multiple configuration files in the build/install folder. This fixes #2623


CMakeModules/build_CLBlast.cmake
```diff
@@ -37,7 +37,7 @@ index 9446499..786f7db 100644
   set(CLBLAST_PATCH_COMMAND ${GIT} apply ${ArrayFire_BINARY_DIR}/clblast.patch)
 endif()
 
-if(WIN32 AND NOT CMAKE_GENERATOR MATCHES "Ninja")
+if(WIN32 AND CMAKE_GENERATOR_PLATFORM AND NOT CMAKE_GENERATOR MATCHES "Ninja")
```
9prady9: Interestingly, [CMAKE_GENERATOR_PLATFORM](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html) is available only when user sets it. I didn't realize that earlier, I was under the impression this is auto-populated (when not explicitly provided for relevant generators) with project call/during compiler check.

Starting with CMake 3.14, there is a new variable [CMAKE_VS_PLATFORM_NAME_DEFAULT](https://cmake.org/cmake/help/v3.16/variable/CMAKE_VS_PLATFORM_NAME_DEFAULT.html#variable:CMAKE_VS_PLATFORM_NAME_DEFAULT) that contains default platform for Visual Studio generators. This variable defaults to Win32 for VS 2017 or below.

PR #2749, opened 2020-02-09
Support f16 for af_isinf, af_isnan

gaika:


umar456:
```
@@ -96,7 +96,8 @@ Array<T> unaryOp(const Array<T> &in, dim4 outDim = dim4(-1, -1, -1, -1)) {
 #define CHECK_FN(name, op)                                                   \
     template<typename T>                                                     \
     struct UnOp<char, T, af_##name##_t> {                                    \
-        void eval(jit::array<char> &out, const jit::array<T> &in, int lim) { \
+        void eval(jit::array<char> &out, const jit::array<compute_t<T>> &in, \
+                  int lim) {                                                 \
             for (int i = 0; i < lim; i++) { out[i] = op(in[i]); }            \
         }                                                                    \
     };
```

umar456:
Thanks for pointing this out. If you want to reopen this PR with the change above please do. The CPU backend requires slightly more work to make it work because x86 does not support half operations. If you want I can also fix this issue if you feel uncomfortable with the change.

gaika:
You will do it better, and then there are tests.

umar456:
Please let me know if you need any other function. Some are easier than others.

PR #2751, opened 2020-02-09
Add support for isInf, isNan, and iszero

### LINKED ISSUE #2750: af_isinf, af_isnan, af_iszero for f16 arrays convert inputs to f32
https://github.com/arrayfire/arrayfire/blob/c1e141691187c4a56f17fffbc52b832d3fa3d2c9/src/api/c/unary.cpp#L671-L674

umar456:
Fixes #2750

### DIFF
--- a/src/api/c/unary.cpp
+++ b/src/api/c/unary.cpp
@@ -673,6 +673,7 @@ static af_err af_check(af_array *out, const af_array in) {
         switch (type) {
             case f32: res = checkOp<float, op>(in); break;
             case f64: res = checkOp<double, op>(in); break;
+            case f16: res = checkOp<half, op>(in); break;
             case c32: res = checkOpCplx<cfloat, float, op>(in); break;
             case c64: res = checkOpCplx<cdouble, double, op>(in); break;
             default: TYPE_ERROR(1, in_type); break;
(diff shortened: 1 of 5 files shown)


src/api/c/unary.cpp
```diff
@@ -673,6 +673,7 @@ static af_err af_check(af_array *out, const af_array in) {
         switch (type) {
```
gaika: `type` is  f32 when the input is f16.

gaika: need `        if (in_type == f16) { type = f16; }` before that line

umar456: ahh good catch.

src/api/c/unary.cpp
```diff
@@ -669,10 +669,12 @@ static af_err af_check(af_array *out, const af_array in) {
 
         // Convert all inputs to floats / doubles / complex
         af_dtype type = implicit(in_type, f32);
+        if(in_type == f16) type = f16;
```
9prady9: In this case, you can just call `af_dtype type = implicit(in_type, f16);`

umar456: The implicit function will be used to convert integer types to floating point types. If we use f16 on the rhs, the integer types will be converted to f16 which has a pretty limited range compared to float.

9prady9: I have checked that, Integer types are not supported in this function

9prady9: We will probably never incude integer types for these three functions. Hence my comment

PR #2753, opened 2020-02-09
Fix cast from f16 to u8 on CUDA backend

umar456:


src/backend/cuda/cast.hpp
```diff
@@ -74,6 +74,13 @@ struct CastOp<cdouble, cdouble> {
     const char *name() { return "__convert_z2z"; }
 };
 
+// Casting from half to unsigned char causes compilation issues. First convert
+// to short then to half
+template<>
+struct CastOp<unsigned char, common::half> {
+    const char *name() { return "(short)"; }
```
9prady9: Interesting, so after being cast to short. The value is implicitly casted to f16 because of identical bit size ?

umar456: No. NVRTC was throwing compilation errors when converting from half to unsigned char. Instead I am first converting to short to avoid this issue. I chose short because it would effectively convert half to an integer type and then short to unsigned char conversion would cover the range of the more limited type.

PR #2761, opened 2020-02-12
Fix forge build dependency when standalone package is generated

9prady9:


CMakeModules/AFconfigure_forge_submodule.cmake
```diff
@@ -34,6 +33,8 @@ if(AF_BUILD_FORGE)
       $<TARGET_FILE:forge>
       $<$<PLATFORM_ID:Linux>:$<TARGET_SONAME_FILE:forge>>
       $<$<PLATFORM_ID:Darwin>:$<TARGET_SONAME_FILE:forge>>
+      $<$<PLATFORM_ID:Linux>:$<TARGET_LINKER_FILE:forge>>
```
9prady9: This is needed since runtime dependency module is not taking into account the soname extensions yet.

src/backend/common/CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ target_link_libraries(afcommon_interface
     ${CMAKE_DL_LIBS}
 )
 
+if(AF_BUILD_FORGE)
+  add_dependencies(afcommon_interface forge)
```
9prady9: This doesn't add link dependency in the linker, I have verified this on Linux and Windows. This approach works on both the OSes unlike the earlier one which weirdly doesn't work on Windows.

PR #2784, opened 2020-03-04
Fp16 print fix

syurkevi:
Adds missing print<half> in array_to_string.
Also checks if half type is supported during array creation in cuda backend.

### DIFF
--- a/src/backend/cuda/Array.cpp
+++ b/src/backend/cuda/Array.cpp
@@ -34,6 +34,18 @@ using std::shared_ptr;
 using std::vector;
 
 namespace cuda {
+
+template<typename T>
+void verifyTypeSupport() {
+    if ((std::is_same<T, double>::value || std::is_same<T, cdouble>::value) &&
+        !isDoubleSupported(getActiveDeviceId())) {
+        AF_ERROR("Double precision not supported", AF_ERR_NO_DBL);
+    } else if (std::is_same<T, common::half>::value &&
+               !isHalfSupported(getActiveDeviceId())) {
+        AF_ERROR("Half precision not supported", AF_ERR_NO_HALF);
+    }
+}
+
 template<typename T>
 Node_ptr bufferNodePtr() {
     return Node_ptr(new BufferNode<T>(getFullName<T>(), shortname<T>(true)));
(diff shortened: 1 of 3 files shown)


src/backend/cuda/Array.cpp
```diff
@@ -34,6 +34,18 @@ using std::shared_ptr;
...
+template<typename T>
+void verifyTypeSupport() {
+    if ((std::is_same<T, double>::value || std::is_same<T, cdouble>::value) &&
+        !isDoubleSupported(getActiveDeviceId())) {
+        AF_ERROR("Double precision not supported", AF_ERR_NO_DBL);
+    } else if (std::is_same<T, common::half>::value &&
+               !isHalfSupported(getActiveDeviceId())) {
+        AF_ERROR("Half precision not supported", AF_ERR_NO_HALF);
+    }
+}
```
9prady9: I think we should move this to `src/backend/common` or `src/api/c` since this can be used from other locations too.

Also, Don't we need similar checks in `create*Array` helpers in others backends too ?

9prady9:
> Warning: this could easily kill perf for users creating thousand of small arrays per secs : have you checked the impact of checkTypeSupported on perf ?
> Example:
> 
>     * in a for loop, just create arrays of random shapes (to bypass the GC buffer, or turn off GC)
> 
>     * track the time to do all these creations
> 
>     * compare with and without this checkTypeSupported
>       ?

I am not sure if these checks are avoided for fp16 for the reasons @WilliamTambellini mentioned. Having said that, some sort of check is required at some location I would think. @umar456 Can you please confirm if this is intentional or if check is done else where ?

src/backend/cuda/Array.cpp
umar456: This is already done in the OpenCL backend. It was not done in this backend because I intended to support half on older hardware and forgot to put it back in once I abandoned that idea.

9prady9: In that case, lets try to move this check into `src/api/c/` a more common location is perhaps `af_create_array` ?

umar456:
I don't think this check is going to be very expensive but it may be a good idea to memoize the result of the call for each device.

src/backend/cuda/Array.cpp
umar456: lots of functions create arrays. for example randu. I think implementing it in the common namespace is sufficient.

9prady9: Whichever location makes sense most.

src/backend/cuda/platform.cpp
```diff
@@ -236,9 +236,17 @@ bool isDoubleSupported(int device) {
 }
 
 bool isHalfSupported(int device) {
-    auto prop     = getDeviceProp(device);
```
9prady9: Isn't getDeviceProp inexpensive ?

umar456: Yes but this will be slightly faster.

PR #2790, opened 2020-03-11
Cleanup cuda sources

9prady9:
- Rename cu to cpp where possible and cleanup CUDA fast/orb
- move CUDA kernels to runtime(nvrtc) compilation


src/backend/cuda/select.cpp
```diff
@@ -32,7 +35,7 @@ void select(Array<T> &out, const Array<char> &cond, const Array<T> &a,
 template<typename T, bool flip>
```
umar456: would be nice to remove this template parameter as well.

9prady9: True, these are changes to backend API level and have to be done along with other backends. We can do this in later change where other backends also change the API to be consistent.

PR #2791, opened 2020-03-12
Move Lookup tables in CUDA backend to texture memory to reduce global constant memory usage

9prady9:
There is negligible to no difference between texture based look-up table and constant memory look-up table. Hence, shifted to texture memory look-up table to reduce global constant memory usage.

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -375,6 +375,7 @@ cuda_add_library(afcuda
 
     Array.cpp
     Array.hpp
+    LookupTable1D.hpp
     Param.hpp
     anisotropic_diffusion.hpp
     approx.hpp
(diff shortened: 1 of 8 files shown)


src/backend/cuda/texture.hpp
```diff
@@ -0,0 +1,66 @@
+/*******************************************************
```
umar456: Can you rename this file to LookupTable1D.hpp?

src/backend/cuda/texture.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <type_traits>
+
+namespace cuda {
+
+template<typename T>
+class LookupTable1D {
+   public:
+    LookupTable1D()                          = delete;
+    LookupTable1D(const LookupTable1D& arg)  = delete;
```
umar456: It makes sense to have a copy constructor for this right? You can copy a texture if you need it.

src/backend/cuda/texture.hpp
9prady9: Okay

src/backend/cuda/texture.hpp
9prady9: Then we have to take care of how texture object is copied also. Just handle copy won't work. Move operation wont have issues though. We don't need neither for these use cases.

PR #2793, opened 2020-03-12
Do not check out af_array of gemm since it can be non-existant

9prady9:


9prady9:
Seems like it is handled here https://github.com/arrayfire/arrayfire/blob/master/src/api/unified/symbol_manager.hpp#L104

PR #2801, opened 2020-03-20
Fix memory pressure in DefaultMemoryManager

padentomasello:
Summary: This fixes a bug that caused memory to not be garbage collected until we ran out of memory in v3.7.0. 

When we switched from MemoryManagerImpl to DefaultMemoryManager, we switched the memory pressure check to check for `lock_buffers` instead of `total_buffers` which I'm assuming was unintentional:

https://github.com/arrayfire/arrayfire/blob/v3.6.4/src/backend/common/MemoryManagerImpl.hpp#L379

https://github.com/arrayfire/arrayfire/blob/v3.7.0/src/backend/common/DefaultMemoryManager.cpp#L132

Also, we now check against `getMemoryPressureThreshold()` which default value is 1.0. Since we were returning 1.0 from `getMemoryPressure()`, it would never trigger. 
(https://github.com/padentomasello/arrayfire/blob/v3.7.0/src/backend/common/DefaultMemoryManager.cpp#L165)

I changed it to return to ratio now instead. This should have the intended behavior, plus allow users to adjust the memory_threshold for meaningful changes.

### DIFF
--- a/src/backend/common/DefaultMemoryManager.cpp
+++ b/src/backend/common/DefaultMemoryManager.cpp
@@ -162,7 +162,8 @@ void* DefaultMemoryManager::alloc(bool user_lock, const unsigned ndims,
         if (!this->debug_mode) {
             // FIXME: Add better checks for garbage collection
             // Perhaps look at total memory available as a metric
-            if (getMemoryPressure() > getMemoryPressureThreshold()) {
+            if (current.lock_bytes >= current.max_bytes ||
+                current.total_buffers >= this->max_buffers) {
                 this->signalMemoryCleanup();
             }
 
--- a/src/backend/cpu/queue.hpp
+++ b/src/backend/cpu/queue.hpp
@@ -69,7 +69,7 @@ class queue {
 #ifndef NDEBUG
         sync();
 #else
-        if (getMemoryPressure() > getMemoryPressureThreshold() || count >= 25) {
+        if (getMemoryPressure() >= getMemoryPressureThreshold() || count >= 25) {
             sync();
         }
 #endif
--- a/src/backend/cpu/queue.hpp
+++ b/src/backend/cpu/queue.hpp
@@ -69,7 +69,8 @@ class queue {
 #ifndef NDEBUG
         sync();
 #else
-        if (getMemoryPressure() >= getMemoryPressureThreshold() || count >= 25) {
+        if (getMemoryPressure() >= getMemoryPressureThreshold() ||
+            count >= 25) {
             sync();
         }
 #endif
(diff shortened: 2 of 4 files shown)


jacobkahn:
So after talking to @padentomasello offline, we've discovered the issue is a bit simpler. There are two distinct issues that affect the default memory manager:
1. A JIT evaluation/memory pressure bug that was introduced when moving JIT heuristics into the memory manager. The [new memory pressure measure](https://github.com/arrayfire/arrayfire/blob/master/src/backend/common/DefaultMemoryManager.cpp#L128) returns `1.0` if the default JIT eval heuristic is met, and `0.0` otherwise. In the [CPU](https://github.com/arrayfire/arrayfire/blob/master/src/backend/cpu/Array.cpp#L242) backend, the `memoryPressureThreshold` (returned by `getMemoryPressureThreshold()`) defaults to 1.0, and the comparison checks if `getMemoryPressure() >= getMemoryPressureThreshold()`. In the [OpenCL](https://github.com/arrayfire/arrayfire/blob/master/src/backend/opencl/Array.cpp#L299) and [CUDA](https://github.com/arrayfire/arrayfire/blob/master/src/backend/cuda/Array.cpp#L251) backends, however, this check is only `getMemoryPressure() > getMemoryPressureThreshold()` (to be clear, this is `>`, not `>=`, so the JIT eval condition will never be met even if `getMemoryPressure()`, which means CUDA and OpenCL backends won't call kernel launches, leading to OOMs and odd behavior.
2. The behavior of garbage collection in the default memory manager has changed slightly in 3.7.0. The [check for garbage collection in 3.6.4](https://github.com/arrayfire/arrayfire/blob/v3.6.4/src/backend/common/MemoryManagerImpl.hpp#L144) during allocations was based on, among other things, [a comparison between `total_buffers` and `max_buffers`](https://github.com/arrayfire/arrayfire/blob/v3.6.4/src/backend/common/MemoryManagerImpl.hpp#L379). In 3.7.0, this is [implicitly calling `getMemoryPressure()`](https://github.com/arrayfire/arrayfire/blob/master/src/backend/common/DefaultMemoryManager.cpp#L163), which is [comparing `lock_buffers` and `max_buffers`](https://github.com/arrayfire/arrayfire/blob/master/src/backend/common/DefaultMemoryManager.cpp#L132).

**tl;dr** — we don't need the ratio idea presented in this PR and can use the existing `getMemoryPressure()` implementation, but we have to make sure that JIT eval checks are always checking that `getMemoryPressure() >= getMemoryPressureThreshold()`, and also that garbage collection checks at `alloc` time in the memory manager are consistent with 3.6.4 behavior.

cc @umar456 @9prady9

jacobkahn:
[This gist](https://gist.github.com/jacobkahn/487cdb58a007ef63c48a6c6869779b6a) should cover everything that needs changing for now.

padentomasello:
@jacobkahn @umar456 Thank you for the feedback. I did not initially realize that getMemoryPressure was used outside of the memory manager. I have updated the PR to match @jacobkahn's gist.

Issue #2800, opened 2020-03-20
OpenCL kernel failing to compile on Clover 


glavaux2:
I have an AMD Radeon running on Fedora 31, its device signature is:

Clover: Radeon RX 580 Series (POLARIS10, DRM 3.36.0, 5.5.8-200.fc31.x86_64, LLVM 9.0.0), 8192 MB

However running a number of examples like "surface_opencl" (master branch as of now) produces this kind of error:

ArrayFire v3.8.0 (OpenCL, 64-bit Linux, build bc37e8a9)
[0] Clover: Radeon RX 580 Series (POLARIS10, DRM 3.36.0, 5.5.8-200.fc31.x86_64, LLVM 9.0.0), 8192 MB
```
Cannot open file 1/KER17252040479444784050.cl
Cannot open file 1/KER7820264272521714025.cl
Cannot open file 1/KER5090553150507457729.cl
Cannot open file 1/KER15803858198028938426.cl
Cannot open file 1/KER6403184010839580536.cl
Cannot open file 1/KER5651956347504768133.cl
In file included from <built-in>:1:
In file included from /usr/include/clc/clc.h:31:
/usr/include/clc/workitem/get_global_size.h:1:39: error: expected ')'
<command line>:4:13: note: expanded from here
/usr/include/clc/workitem/get_global_size.h:1:33: note: to match this '('
In file included from <built-in>:1:
In file included from /usr/include/clc/clc.h:32:
/usr/include/clc/workitem/get_global_id.h:1:37: error: expected ')'
<command line>:4:13: note: expanded from here
/usr/include/clc/workitem/get_global_id.h:1:31: note: to match this '('
In file included from <built-in>:1:
In file included from /usr/include/clc/clc.h:33:
/usr/include/clc/workitem/get_local_size.h:1:38: error: expected ')'
<command line>:4:13: note: expanded from here
/usr/include/clc/workitem/get_local_size.h:1:32: note: to match this '('
In file included from <built-in>:1:
In file included from /usr/include/clc/clc.h:34:
/usr/include/clc/workitem/get_local_id.h:1:36: error: expected ')'
<command line>:4:13: note: expanded from here
/usr/include/clc/workitem/get_local_id.h:1:30: note: to match this '('
In file included from <built-in>:1:
In file included from /usr/include/clc/clc.h:35:
/usr/include/clc/workitem/get_num_groups.h:1:38: error: expected ')'
<command line>:4:13: note: expanded from here
[...]

PR #2802, opened 2020-03-21
Use kDim instead of dim to avoid name collision

glavaux2:
This is related to https://github.com/arrayfire/arrayfire/issues/2800


glavaux2:
I have renamed to kDim which looks like working on Mesa AMD Clover

src/backend/opencl/kernel/join.hpp
```diff
@@ -48,7 +48,7 @@ void join(Param out, const Param in, const af::dim4 offset) {
         std::ostringstream options;
         options << " -D To=" << dtype_traits<To>::getName()
                 << " -D Ti=" << dtype_traits<Ti>::getName()
-                << " -D dim=" << dim;
+                << " -D kDim=" << dim;
```
9prady9: `join.cl` OpenCL kernel doesn't use this one at all. I wonder why it's being passed to this at all.

PR #2821, opened 2020-04-02
Fix documentation to mem step size and clean up memory manager test

jacobkahn:
- Add documentation to set and get mem step size functions in the C/C++ API to properly reflect behavior if a custom memory manager is set
- Remove af_event use from unlock in memory management test
- Ensure memory manager test memory pressure functions and gc threshold more closely matches default memory manager behavior

### DIFF
--- a/include/af/device.h
+++ b/include/af/device.h
@@ -236,12 +236,14 @@ namespace af
     AFAPI void deviceGC();
     /// @}
 
-    /// \brief Set the resolution of memory chunks
+    /// \brief Set the resolution of memory chunks. Works only with the default
+    /// memory manager - throws if a custom memory manager is set.
     ///
     /// \ingroup device_func_mem
     AFAPI void setMemStepSize(const size_t size);
 
-    /// \brief Get the resolution of memory chunks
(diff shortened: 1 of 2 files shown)


include/af/device.h
```diff
@@ -236,12 +236,14 @@ namespace af
     AFAPI void deviceGC();
     /// @}
 
-    /// \brief Set the resolution of memory chunks
+    /// \brief Set the resolution of memory chunks. Works only with the default
+    /// memory manager - throws if a custom memory manager is set.
```
9prady9, line 240: 👍

test/memory.cpp
```diff
@@ -773,7 +769,7 @@ af_err alloc_fn(af_memory_manager manager, void **ptr,
         get_memory_pressure_fn(manager, &pressure);
         float threshold;
         af_memory_manager_get_memory_pressure_threshold(manager, &threshold);
-        if (pressure > threshold) { signal_memory_cleanup_fn(manager); }
+        if (pressure >= threshold) { signal_memory_cleanup_fn(manager); }
```
9prady9, line 773: The changes in this file basically reflect the change in #2801 Is that correct ?

PR #2838, opened 2020-04-11
Fix empty array handling in lu, rank, and qr. Other minor refactoring

umar456:
* some linear algebra functions could have leaked memory in case the input was null. This PR addresses that potential leak and handles the empty or null array case.
* Minor refactor in median
* Minor refactor in DefaultMemoryManager. Just changed names of the variables.

### DIFF
--- a/src/api/c/lu.cpp
+++ b/src/api/c/lu.cpp
@@ -49,6 +49,9 @@ af_err af_lu(af_array *lower, af_array *upper, af_array *pivot,
 
         af_dtype type = i_info.getType();
 
+        ARG_ASSERT(0, lower != nullptr);
+        ARG_ASSERT(1, upper != nullptr);
+        ARG_ASSERT(2, pivot != nullptr);
         ARG_ASSERT(3, i_info.isFloating());  // Only floating and complex types
 
         if (i_info.ndims() == 0) {
@@ -81,13 +84,13 @@ af_err af_lu_inplace(af_array *pivot, af_array in, const bool is_lapack_piv) {
         }
 
         ARG_ASSERT(1, i_info.isFloating());  // Only floating and complex types
+        ARG_ASSERT(0, pivot != nullptr);
 
         if (i_info.ndims() == 0) {
             return af_create_handle(pivot, 0, nullptr, type);
         }
 
         af_array out;
-
         switch (type) {
             case f32: out = lu_inplace<float>(in, is_lapack_piv); break;
             case f64: out = lu_inplace<double>(in, is_lapack_piv); break;
@@ -95,7 +98,7 @@ af_err af_lu_inplace(af_array *pivot, af_array in, const bool is_lapack_piv) {
             case c64: out = lu_inplace<cdouble>(in, is_lapack_piv); break;
             default: TYPE_ERROR(1, type);
         }
-        if (pivot != NULL) std::swap(*pivot, out);
+        std::swap(*pivot, out);
     }
     CATCHALL;
 
--- a/src/api/c/median.cpp
+++ b/src/api/c/median.cpp
@@ -41,14 +41,12 @@ static double median(const af_array& in) {
     } else if (nElems == 2) {
         T result[2];
         AF_CHECK(af_get_data_ptr((void*)&result, in));
-        if (input.isFloating()) {
-            return division(result[0] + result[1], 2.0);
-        } else {
-            return division((float)result[0] + (float)result[1], 2.0);
-        }
+        return division(
+            (static_cast<double>(result[0]) + static_cast<double>(result[1])),
+            2.0);
     }
 
-    double mid       = (nElems + 1) / 2;
+    double mid       = static_cast<double>(nElems + 1) / 2.0;
     af_seq mdSpan[1] = {af_make_seq(mid - 1, mid, 1)};
 
     Array<T> sortedArr = sort<T>(input, 0, true);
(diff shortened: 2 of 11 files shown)


src/api/c/lu.cpp
```diff
@@ -49,6 +49,9 @@ af_err af_lu(af_array *lower, af_array *upper, af_array *pivot,
 
         af_dtype type = i_info.getType();
 
+        ARG_ASSERT(0, lower != nullptr);
```
9prady9, line 52: Are we going to add such `nullptr` checks for all functions going forward ?

test/lu_dense.cpp
```diff
@@ -235,3 +235,42 @@ TYPED_TEST(LU, RectangularLarge1) {
...
+    if (noLAPACKTests()) return;
+    dim4 dims(3, 3);
+    af_array in = 0;
+    ASSERT_SUCCESS(af_randu(&in, dims.ndims(), dims.get(), f32));
+
+    af_array lower, upper;
+    ASSERT_EQ(AF_ERR_ARG, af_lu(&lower, &upper, NULL, in));
+}
+
+TEST(LU, NullOutput) {
```
9prady9, line 269: `InplaceNullOutput` as name sounds more better.

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -166,13 +166,15 @@ void *DefaultMemoryManager::alloc(bool user_lock, const unsigned ndims,
             }
 
             lock_guard_t lock(this->memory_mutex);
-            free_iter iter = current.free_map.find(alloc_bytes);
```
9prady9, line 169: This type alias should be removed from here
https://github.com/arrayfire/arrayfire/blob/3693e6a0ea5ec89a39f4699102db1967de3b923d/src/backend/common/DefaultMemoryManager.hpp#L42

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -206,8 +208,8 @@ void *DefaultMemoryManager::alloc(bool user_lock, const unsigned ndims,
 size_t DefaultMemoryManager::allocated(void *ptr) {
     if (!ptr) return 0;
     memory_info &current = this->getCurrentMemoryInfo();
-    locked_iter iter     = current.locked_map.find((void *)ptr);
```
9prady9, line 209: This type alias should be removed from header
https://github.com/arrayfire/arrayfire/blob/3693e6a0ea5ec89a39f4699102db1967de3b923d/src/backend/common/DefaultMemoryManager.hpp#L39

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -206,8 +208,8 @@ void *DefaultMemoryManager::alloc(bool user_lock, const unsigned ndims,
...
-    if (iter == current.locked_map.end()) return 0;
+    auto iter            = current.locked_map.find(ptr);
```
9prady9, line 211: nit: I am aware this variable is not used in a very large scope, but it could also use a better name to help while reading the code.

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -325,7 +332,7 @@ void DefaultMemoryManager::userLock(const void *ptr) {
 
     lock_guard_t lock(this->memory_mutex);
 
-    locked_iter iter = current.locked_map.find(const_cast<void *>(ptr));
+    auto iter = current.locked_map.find(const_cast<void *>(ptr));
```
9prady9, line 335: nit: I am aware this variable is not used in a very large scope, but it could also use a better name to help while reading the code.

src/backend/common/DefaultMemoryManager.cpp
```diff
@@ -342,12 +349,9 @@ void DefaultMemoryManager::userUnlock(const void *ptr) {
 bool DefaultMemoryManager::isUserLocked(const void *ptr) {
     memory_info &current = this->getCurrentMemoryInfo();
     lock_guard_t lock(this->memory_mutex);
-    locked_iter iter = current.locked_map.find(const_cast<void *>(ptr));
-    if (iter != current.locked_map.end()) {
-        return iter->second.user_lock;
-    } else {
-        return false;
-    }
+    auto iter = current.locked_map.find(const_cast<void *>(ptr));
```
9prady9, line 352: nit: I am aware this variable is not used in a very large scope, but it could also use a better name to help while reading the code.

src/api/c/median.cpp
```diff
@@ -90,9 +86,9 @@ static af_array median(const af_array& in, const dim_t dim) {
 
     Array<T> sortedIn = sort<T>(input, dim, true);
 
-    int dimLength = input.dims()[dim];
-    double mid    = (dimLength + 1) / 2;
-    af_array left = 0;
+    size_t dimLength = input.dims()[dim];
```
9prady9, line 89: :+1:

test/median.cpp
```diff
@@ -150,3 +150,19 @@ MEDIAN(float, uchar)
...
+    ASSERT_ARRAYS_EQ(in, out);
+}
+
+TEST(Median, TwoElements) {
+    af::array in = randu(2, f32);
+
+    af::array out = median(in);
+    af::array gold = mean(in);
+    ASSERT_ARRAYS_EQ(gold, out);
+}
```
9prady9, line 168: :+1:

src/api/c/lu.cpp
umar456: We should but not now.

umar456:
Addressed all feedback

PR #2840, opened 2020-04-14
Fix pinnedMemManager check. Fix MKL Sequential layer

umar456:
* The setMemorymanagerPinned function was incorrectly checking the function pointer instead of the pinnedMemManager pointer.
* MKL with the Sequential layer was broken because the ThreadingLibrary variable was required. The sequential layer does not have a threading library so MKL was set to not found because that variable was not set.

### DIFF
--- a/CMakeModules/FindMKL.cmake
+++ b/CMakeModules/FindMKL.cmake
@@ -357,8 +357,7 @@ find_package_handle_standard_args(MKL_Shared
   REQUIRED_VARS MKL_INCLUDE_DIR
                 MKL_Core_LINK_LIBRARY
                 MKL_Interface_LINK_LIBRARY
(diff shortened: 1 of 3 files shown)


CMakeModules/FindMKL.cmake
```diff
@@ -357,17 +357,15 @@ find_package_handle_standard_args(MKL_Shared
   REQUIRED_VARS MKL_INCLUDE_DIR
                 MKL_Core_LINK_LIBRARY
                 MKL_Interface_LINK_LIBRARY
-                MKL_ThreadLayer_LINK_LIBRARY
-                MKL_ThreadingLibrary_LINK_LIBRARY)
+                MKL_ThreadLayer_LINK_LIBRARY)
```
9prady9, line 360: The commit message for this commit is incorrect.

Issue #2845, opened 2020-04-16
Compilation fails with Visual Studio 2015


cschreib-ibex:
The current master branch of ArrayFire does not compile with Visual Studio 2015:
```
src\backend\common/half.hpp(701): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(702): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(705): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(706): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(707): error : a constexpr function must contain exactly one return statement
src\backend\common/half.hpp(708): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(709): error : a constexpr function must contain exactly one return statement
src\backend\common/half.hpp(710): error : a constexpr function must contain exactly one return statement
src\backend\common/half.hpp(712): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(713): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(714): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(715): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(716): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(717): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(718): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(719): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(720): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(721): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(723): error : statement may not appear in a constexpr function
src\backend\common/half.hpp(724): error : a constexpr function must contain exactly one return statement

src\backend\common/unique_handle.hpp(77): error : statement may not appear in a constexpr constructor
            detected during:
              instantiation of "common::unique_handle<T>::unique_handle(common::unique_handle<T> &&) [with T=cusparseMatDescr_t]"
  (111): here
              instantiation of "common::unique_handle<T> common::make_handle<T>() [with T=cusparseMatDescr_t]"
  src/backend/cuda/sparse_arith.cu(146): here
              instantiation of "common::SparseArray<T> cuda::arithOp<T,op>(const common::SparseArray<T> &, const common::SparseArray<T> &) [with T=float, op=af_add_t]"
  src/backend/cuda/sparse_arith.cu(224): here

[...]

Issue #2503, opened 2019-05-01
Cache compiled kernels to disk


umar456:
With the latest nvrtc changes incorporated into ArrayFire, the initialization times for the initial launch will go up significantly. This is because we are compiling these kernels at runtime instead of embedding them into the library. We did this to reduce the binary size of the binary which has steadily gone up with every release.

An approach to reduce the overhead would be to write the compiled binaries to disk and load them on the first call. This avoids compiling to code to the intermediate representation and then to binary by the driver. Loading these binaries should be faster than compiling them from source.

We could also do this with JIT kernels and further reduce the initialization overhead.

One of the things to consider is where are we going to store these binaries. On Linux we can store them in the tmp folder or a hidden folder in the user's directory. On Windows we need to find a standard location for these files. Systems may be loading from a network drive which could hurt the performance so this location needs to be configurable.

9prady9:
#### For Windows
The best way to store the kernels is on per-user-account basis. We can can query `AppData` path of the user running the program using [SHGetKnownFolderPath](https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath) using the flag `FOLDERID_AppDataProgramData`.

#### For Linux
I don't like the idea of `tmp`, there is no guarantee of persistent storage there. `tmp` potentially gets cleaned up on reboot or startup. We should store it under `$HOME/.arrayfire` . This per-user location can also be used to store other configuration data we potentially might want to save to disk in future. One such avenue is environment variables: right now can be set either in command line or in bashrc; we can have a configuration file(whichever may be the format) inside this per-user folder and use those settings on library load.

cschreib-ibex:
This seems like a very desirable feature for improving runtime performances. Is there any plan to include this in ArrayFire in a coming release?

9prady9:
Unfortunately, this won't make the next release, v3.7. We shall update this issue for future discussions and/or progress related to this feature.

cschreib-ibex:
Ok, thank you for the update!

cschreib-ibex:
This has become an important issue for us. Because of the kernel compilation overhead, our processing pipeline takes about 20 seconds when run "cold" (first time run), and just 2 seconds when "hot" (second run within the same process). 

[...]

PR #2848, opened 2020-04-17
Automatically cache compiled CUDA kernels on disk to speed up kernel compilation

cschreib-ibex:
This PR addresses #2503 ~~and #2845~~ (addressed in separate PR: #2850). 

Kernels are saved to disk in CUBIN format straight after being compiled in ``buildKernel()``. The folder they are saved into is platform-dependent:
 - linux: ``/var/lib/arrayfire``, or ``~/.arrayfire``, or ``/tmp/arrayfire`` (in order of decreasing priority)
 - windows: ``%APPDATA%\Temp\ArrayFire``

Each kernel is saved in a separate file, and the file name is built from a hash of the kernel function name (that name is already used internally for caching, so hopefully won't collide). The device compute capabilities and the AF API version are also encoded in the file name, so if these ever change, kernels will get recompiled automatically.

Disk usage looks very reasonable; on average 5 kB per kernel on the tests I ran.

Then, whenever a kernel is requested, we first look if the kernel is in the memory cache (as was done before), and if not, try the disk cache (that's new), and if that fails too, build the kernel from scratch (as was done before).

I have tested the implementation on Windows by running our own software with the ArrayFire DLLs built from that branch, and all worked fine. The linux implementation, however, is untested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ option(AF_BUILD_FORGE
 option(AF_WITH_NONFREE  "Build ArrayFire nonfree algorithms"   OFF)
 option(AF_WITH_LOGGING  "Build ArrayFire with logging support" ON)
 option(AF_WITH_STACKTRACE  "Add stacktraces to the error messages." ON)
+option(AF_CACHE_KERNELS "Enable caching kernels on disk"       ON)
 
 if(WIN32)
   set(AF_STACKTRACE_TYPE "Windbg" CACHE STRING "The type of backtrace features. Windbg(simple), None")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,10 @@ foreach(backend ${built_backends})
     target_compile_definitions(${backend}
       PRIVATE AF_WITH_LOGGING)
   endif()
+  if(AF_CACHE_KERNELS_TO_DISK)
+    target_compile_definitions(${backend}
+      PRIVATE AF_CACHE_KERNELS_TO_DISK)
+  endif()
 endforeach()
 
 if(AF_BUILD_FRAMEWORK)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ option(AF_BUILD_FORGE
 option(AF_WITH_NONFREE  "Build ArrayFire nonfree algorithms"   OFF)
 option(AF_WITH_LOGGING  "Build ArrayFire with logging support" ON)
 option(AF_WITH_STACKTRACE  "Add stacktraces to the error messages." ON)
-option(AF_CACHE_KERNELS "Enable caching kernels on disk"       ON)
+option(AF_CACHE_KERNELS_TO_DISK "Enable caching kernels to disk" ON)
 
 if(WIN32)
   set(AF_STACKTRACE_TYPE "Windbg" CACHE STRING "The type of backtrace features. Windbg(simple), None")
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -180,7 +180,7 @@ const string& getCacheDirectory() {
         auto iterDir = std::find_if(pathList.begin(), pathList.end(), 
             isDirectoryWritable);
 
-        return iterDir != pathList.end() ? *iterDir : "";
+        cacheDirectory = iterDir != pathList.end() ? *iterDir : "";
     });
 
     return cacheDirectory;
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -27,6 +27,7 @@
 #include <fstream>
 #include <numeric>
 #include <string>
+#include <thread>
 #include <vector>
 
 using std::string;
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -20,6 +20,7 @@
 #include <af/defines.h>
 
 #include <sys/stat.h>
+#include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
@@ -146,6 +147,10 @@ bool removeFile(const string& path) {
 #endif
 }
 
+bool renameFile(const string& sourcePath, const string& destPath) {
+    return std::rename(sourcePath.c_str(), destPath.c_str()) == 0;
+}
+
 bool isDirectoryWritable(const string& path) {
     if (!directoryExists(path) && !createDirectory(path)) return false;
 
@@ -180,4 +185,15 @@ const string& getCacheDirectory() {
     });
 
     return cacheDirectory;
+}
+
+string makeTempFilename() {
+    thread_local std::size_t fileCount = 0u;
+
+    ++fileCount;
+    const std::size_t threadID =
+        std::hash<std::thread::id>{}(std::this_thread::get_id());
+
+    return std::to_string(std::hash<string>{}(std::to_string(threadID) + "_" +
+                                              std::to_string(fileCount)));
 }
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -120,7 +120,7 @@ string getTemporaryDirectory() {
 }
 #endif
 
-bool folderExists(const string& path) {
+bool directoryExists(const string& path) {
 #if !defined(OS_WIN)
     struct stat status;
     return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
@@ -130,7 +130,7 @@ bool folderExists(const string& path) {
 #endif
 }
 
-bool createFolder(const string& path) {
+bool createDirectory(const string& path) {
 #if !defined(OS_WIN)
     return mkdir(path.c_str(), 0777) == 0;
 #else
@@ -147,7 +147,7 @@ bool removeFile(const string& path) {
 }
 
 bool isDirectoryWritable(const string& path) {
-    if (!folderExists(path) && !createFolder(path)) return false;
+    if (!directoryExists(path) && !createDirectory(path)) return false;
 
     const string testPath = path + AF_PATH_SEPARATOR + "test";
     if (!std::ofstream(testPath).is_open()) return false;
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -120,7 +120,7 @@ string getTemporaryDirectory() {
 }
 #endif
 
-bool folderExists(const string &path) {
+bool folderExists(const string& path) {
 #if !defined(OS_WIN)
     struct stat status;
     return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
@@ -130,7 +130,7 @@ bool folderExists(const string &path) {
 #endif
 }
 
-bool createFolder(const string &path) {
+bool createFolder(const string& path) {
 #if !defined(OS_WIN)
     return mkdir(path.c_str(), 0777) == 0;
 #else
@@ -138,7 +138,7 @@ bool createFolder(const string &path) {
 #endif
 }
 
-bool removeFile(const string &path) {
+bool removeFile(const string& path) {
 #if !defined(OS_WIN)
     return unlink(path.c_str()) == 0;
 #else
@@ -146,7 +146,7 @@ bool removeFile(const string &path) {
 #endif
 }
 
-bool isDirectoryWritable(const string &path) {
+bool isDirectoryWritable(const string& path) {
     if (!folderExists(path) && !createFolder(path)) return false;
 
     const string testPath = path + AF_PATH_SEPARATOR + "test";
@@ -156,7 +156,7 @@ bool isDirectoryWritable(const string &path) {
     return true;
 }
 
-const string &getCacheDirectory() {
+const string& getCacheDirectory() {
     thread_local std::once_flag flag;
     thread_local string cacheDirectory;
 
@@ -171,7 +171,7 @@ const string &getCacheDirectory() {
 #endif
         };
 
-        for (const string &path : pathList) {
+        for (const string& path : pathList) {
             if (isDirectoryWritable(path)) {
                 cacheDirectory = path;
                 break;
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -178,8 +178,8 @@ const string& getCacheDirectory() {
 #endif
         };
 
-        auto iterDir = std::find_if(pathList.begin(), pathList.end(), 
-            isDirectoryWritable);
+        auto iterDir =
+            std::find_if(pathList.begin(), pathList.end(), isDirectoryWritable);
 
         cacheDirectory = iterDir != pathList.end() ? *iterDir : "";
     });
@@ -201,8 +201,8 @@ string makeTempFilename() {
 std::size_t deterministicHash(const void* data, std::size_t byteSize) {
     // Fowler-Noll-Vo "1a" 32 bit hash
     // https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
-    constexpr std::size_t seed = 0x811C9DC5;
-    constexpr std::size_t prime = 0x01000193;
+    constexpr std::size_t seed   = 0x811C9DC5;
+    constexpr std::size_t prime  = 0x01000193;
     const std::uint8_t* byteData = static_cast<const std::uint8_t*>(data);
     return std::accumulate(byteData, byteData + byteSize, seed,
                            [&](std::size_t hash, std::uint8_t data) {
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -20,6 +20,7 @@
 #include <af/defines.h>
 
 #include <sys/stat.h>
+#include <algorithm>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -176,12 +177,10 @@ const string& getCacheDirectory() {
 #endif
         };
 
-        for (const string& path : pathList) {
-            if (isDirectoryWritable(path)) {
-                cacheDirectory = path;
-                break;
-            }
-        }
+        auto iterDir = std::find_if(pathList.begin(), pathList.end(), 
+            isDirectoryWritable);
+
+        return iterDir != pathList.end() ? *iterDir : "";
     });
 
     return cacheDirectory;
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -101,7 +101,16 @@ std::string int_version_to_string(int version) {
            std::to_string((int)((version % 1000) / 10.));
 }
 
-#if !defined(OS_WIN)
+#if defined(OS_WIN)
+string getTemporaryDirectory() {
+    DWORD bufSize = 261;  // limit according to GetTempPath documentation
+    string retVal;
+    retVal.resize(bufSize);
+    bufSize = GetTempPathA(bufSize, &retVal[0]);
+    retVal.resize(bufSize);
+    return retVal;
+}
+#else
 string getHomeDirectory() {
     string home = getEnvVar("XDG_CACHE_HOME");
     if (!home.empty()) return home;
@@ -111,40 +120,31 @@ string getHomeDirectory() {
 
     home = getpwuid(getuid())->pw_dir;
 }
-#else
-string getTemporaryDirectory() {
-    DWORD bufSize = 261;  // limit according to GetTempPath documentation
-    string retVal;
-    retVal.resize(bufSize);
-    bufSize = GetTempPathA(bufSize, &retVal[0]);
-    retVal.resize(bufSize);
-    return retVal;
-}
 #endif
 
 bool directoryExists(const string& path) {
-#if !defined(OS_WIN)
-    struct stat status;
-    return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
-#else
+#if defined(OS_WIN)
     struct _stat status;
     return _stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
+#else
+    struct stat status;
+    return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
 #endif
 }
 
 bool createDirectory(const string& path) {
-#if !defined(OS_WIN)
-    return mkdir(path.c_str(), 0777) == 0;
-#else
+#if defined(OS_WIN)
     return CreateDirectoryA(path.c_str(), NULL) != 0;
+#else
+    return mkdir(path.c_str(), 0777) == 0;
 #endif
 }
 
 bool removeFile(const string& path) {
-#if !defined(OS_WIN)
-    return unlink(path.c_str()) == 0;
-#else
+#if defined(OS_WIN)
     return DeleteFileA(path.c_str()) != 0;
+#else
+    return unlink(path.c_str()) == 0;
 #endif
 }
 
@@ -168,12 +168,12 @@ const string& getCacheDirectory() {
 
     std::call_once(flag, []() {
         const vector<string> pathList = {
-#if !defined(OS_WIN)
+#if defined(OS_WIN)
+            getTemporaryDirectory() + "\\ArrayFire"
+#else
             "/var/lib/arrayfire",
             getHomeDirectory() + "/.arrayfire",
             "/tmp/arrayfire"
-#else
-            getTemporaryDirectory() + "\\ArrayFire"
 #endif
         };
 
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -25,6 +25,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
+#include <numeric>
 #include <string>
 #include <vector>
 
@@ -195,4 +196,20 @@ string makeTempFilename() {
 
     return std::to_string(std::hash<string>{}(std::to_string(threadID) + "_" +
                                               std::to_string(fileCount)));
+}
+
+std::size_t deterministicHash(const void* data, std::size_t byteSize) {
+    // Fowler-Noll-Vo "1a" 32 bit hash
+    // https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
+    constexpr std::size_t seed = 0x811C9DC5;
+    constexpr std::size_t prime = 0x01000193;
+    const std::uint8_t* byteData = static_cast<const std::uint8_t*>(data);
+    return std::accumulate(byteData, byteData + byteSize, seed,
+                           [&](std::size_t hash, std::uint8_t data) {
+                               return (hash ^ data) * prime;
+                           });
+}
+
+std::size_t deterministicHash(const std::string& data) {
+    return deterministicHash(data.data(), data.size());
 }
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -162,8 +162,8 @@ bool isDirectoryWritable(const string& path) {
 }
 
 const string& getCacheDirectory() {
-    thread_local std::once_flag flag;
-    thread_local string cacheDirectory;
+    static std::once_flag flag;
+    static string cacheDirectory;
 
     std::call_once(flag, []() {
         const vector<string> pathList = {
--- a/src/backend/common/util.cpp
+++ b/src/backend/common/util.cpp
@@ -10,17 +10,24 @@
 /// This file contains platform independent utility functions
 #if defined(OS_WIN)
 #include <Windows.h>
+#else
+#include <pwd.h>
+#include <unistd.h>
 #endif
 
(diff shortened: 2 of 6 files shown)


src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
 template void Kernel::setScalar<int>(const char *, int);
 template void Kernel::getScalar<int>(int &, const char *);
 
+#if !defined(OS_WIN)
+string getHomeDirectory() {
+    string home = getEnvVar("XDG_CONFIG_HOME");
```
9prady9, line 162: Shouldn't this be `XDG_CACHE_HOME` for kernel cache ?

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+    if (!home.empty()) return home;
+
+    home = getEnvVar("HOME");
+    if (!home.empty()) return home;
+
+    home = getpwuid(getuid())->pw_dir;
+}
+#else
+string getTemporaryDirectory() {
+    DWORD bufSize = 261;  // limit according to GetTempPath documentation
```
9prady9, line 172: We can do what I have mentioned [here](https://github.com/arrayfire/arrayfire/issues/2503#issuecomment-488546754) for windows. Cache will be persistent for at least a given session or even reboot depending on settings.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+const string &getKernelCacheDirectory() {
+    thread_local std::once_flag flag;
+    thread_local string cacheDirectory;
+
+    std::call_once(flag, []() { 
+        const vector<string> pathList = {
+#if !defined(OS_WIN)
+            "/var/lib/arrayfire",
+            getHomeDirectory() + "/.arrayfire",
+            "/tmp/arrayfire"
```
9prady9, line 226: `tmp` folder caching will not help for reboot cases however there is no guarantee on how much time cache will be there in tmp while the program is running ? I never thought about this before. @umar456 @cschreib-ibex Will it persistent until the user log-outs ?

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+bool isDirectoryWritable(const string &path) {
+    if (!folderExists(path) && !createFolder(path)) return false;
+
+    const string testPath = path + AF_PATH_SEPARATOR + "test";
+    if (!std::ofstream(testPath).is_open()) return false;
+    removeFile(testPath);
+    
+    return true;
+}
+
```
9prady9, line 216: All the helper functions can go into `src/backend/common` since they are not specific to compute backends.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -457,21 +457,73 @@ Kernel buildKernel(const int device, const string &nameExpr,
...
+        CU_CHECK(cuModuleGetFunction(&kernel, module, name.c_str()));
+
+        AF_TRACE("{{{:<30} : loaded from {} for {} }}",
+            nameExpr, cacheFile, getDeviceProp(device).name);
+    } catch (...) {
+        if (module != nullptr) {
+            CU_CHECK(cuModuleUnload(module));
+        }
+        removeFile(cacheFile);
+        throw;
```
9prady9, line 499: It should be `AF_ERROR("message", AF_ERR_RUNTIME)` instead of an empty throw.

9prady9:
@cschreib-ibex If would be great if you can separate [Removed constexpr not supported by VS2015 ](https://github.com/arrayfire/arrayfire/pull/2848/commits/57b49d0d6c2d5704642c4e468229c0ad22a12ff1) into it's own PR.

Thanks for working on this! Nice job. I have left some feedback, we can continue discussion about the caching in this PR while the VS2015 doesn't need to wait for this change to go into.

9prady9:
We can actually move all directory related functions to `src/backend/common` so that they can be used by other backends if and when necessary.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+}
+
+string hashKernelName(const string &nameExpr) {
+    return to_string(std::hash<string>{}(nameExpr));
+}
+
+string getKernelCacheFilename(const int device, const string &nameExpr) {
+    string mangledName = nameExpr;
+    if (mangledName.substr(0, 3) != "KER") {
+        mangledName = "KERF" + hashKernelName(nameExpr);
```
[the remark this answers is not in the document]
cschreib-ibex: I wanted to reduce the risk of collisions between JIT and non-JIT kernels, and to achieve this I thought to add a letter to the prefix. I chose "F" for "fixed" kernels. But it's true we could just hash everything as you suggested. I will do that.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+    }
+
+    auto computeFlag = getComputeCapability(device);
+
+    return mangledName
+        + "_CU" + to_string(computeFlag.first) + to_string(computeFlag.second) 
+        + "_AF" + to_string(AF_API_VERSION_CURRENT)
```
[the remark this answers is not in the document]
cschreib-ibex: Done

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+                cacheDirectory = path;
+                break;
+            }
+        } 
+    });
+
+    return cacheDirectory;
+}
+
+string hashKernelName(const string &nameExpr) {
```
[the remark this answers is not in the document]
cschreib-ibex: Removed.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -149,6 +157,107 @@ void Kernel::getScalar(T &out, const char *name) {
...
+    const string testPath = path + AF_PATH_SEPARATOR + "test";
+    if (!std::ofstream(testPath).is_open()) return false;
+    removeFile(testPath);
+    
+    return true;
+}
+
+const string &getKernelCacheDirectory() {
+    thread_local std::once_flag flag;
+    thread_local string cacheDirectory;
```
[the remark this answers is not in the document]
cschreib-ibex: My understanding is that a file cannot be opened twice with write permission. This means that two threads cannot write the same kernel file at the same time; one of the threads will fail to open the file for writing. By default, ``std`` streams fail silently so this will simply do nothing, both threads will continue normally, and the file will be written on disk correctly.

The other risk is if one thread is writing the file while another thread tries to read that same file. This can happen, although because the byte size of the CUBIN is included in the file itself, this acts as some crude form of checksum, and there will be a read error (not enough bytes to read from file). Currently, this will make ``loadKernel()`` throw an exception. We can make it so any exception is caught and an empty ``Kernel{nullptr, nullptr}`` is returned instead, so the thread will fallback to building the kernel again.

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: Didn't know about it! Will change

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: On my machine ``GetTempPath`` returned ``AppData\Local\Temp``, which I think is the default unless the user has overridden the ``TMP`` or ``TEMP`` environment variables. 

``FOLDERID_AppDataProgramData``, which you suggest as an alternative, looks like it was introduced in Windows 10, I don't know if this is an issue for supporting older Windows OSes.

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: This folder order was suggested by @umar456, I think his reasoning was that, even if ``/tmp`` is not persistent across reboots/logouts, this is still better than nothing. Looks like ``/var/tmp/`` is persistent though, so we could try that before ``/tmp``.

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: Will do

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: See comment above; maybe we don't want to throw from this function and just return an empty kernel, so that we fallback to building the kernel again if loading fails.

cschreib-ibex:
@9prady9 Thank you for your review! I'll implement your suggestions and split the VS2015 stuff in a separate branch / PR.

cschreib-ibex:
I've renamed the function ``getKernelCacheDirectory()`` into ``getCacheDirectory()`` when moving it to ``util.cpp``. I figure this might get used later for other purposes than caching kernels.

9prady9:
@cschreib-ibex You can drop `Removed constexpr not supported by VS2015` commit from this PR now. You will be able to rebase this from latest master once I merge #2850 (waiting for the ci jobs to go green).

cschreib-ibex:
@9prady9 Done!

src/backend/cuda/nvrtc/cache.cpp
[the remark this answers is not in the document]
9prady9: I didn't know that std stream are thread safe. From what I found in the following links
- [Only iostream is thread safe, even then the order of characters/words in the output cannot be deterministically identified. ](https://docs.microsoft.com/en-us/cpp/standard-library/thread-safety-in-the-cpp-standard-library?redirectedfrom=MSDN&view=vs-2019)
- [So, if your platform's C library is threadsafe, then your fstream I/O operations will be threadsafe at the lowest level. For higher-level operations, such as manipulating the data contained in the stream formatting classes (e.g., setting up callbacks inside an std::ofstream), you need to guard such accesses like any other critical shared resource. ](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_concurrency.html)

There is no guarantee that the behavior is well defined with file streams in C++. They have to be guarded to be safe in a multi-threaded context. A similar discussion on stackoverflow can be found here https://stackoverflow.com/a/15513639/1985476

@umar456 what are you thoughts ?

src/backend/cuda/nvrtc/cache.cpp
9prady9: That is unfortunate. There must be a way to handle it for older windows versions too I think.

9prady9:
> I've renamed the function `getKernelCacheDirectory()` into `getCacheDirectory()` when moving it to `util.cpp`. I figure this might get used later for other purposes than caching kernels.
> 
> I have also created a separate PR for the VS2015 compilation fix, #2850. I can revert the relevant commit from the current PR if you wish, although that will make it more painful for me to make further adjustments.

We can have the VS PR merged soon, it is a minor change. I am waiting for the Windows jobs to finish.

src/backend/cuda/nvrtc/cache.cpp
cschreib-ibex: ``FOLDERID_AppDataProgramData`` is nothing more than ``%LOCALAPPDATA%\ProgramData`` by default, although the Windows docs say:

> This FOLDERID is used internally by .NET applications to enable cross-platform app functionality. It is not intended to be used directly from an application.

We could instead just ask for ``FOLDERID_LocalAppData`` (which should just be ``%LOCALAPPDATA%``), and build the rest ourselves.

src/backend/cuda/nvrtc/cache.cpp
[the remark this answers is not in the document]
cschreib-ibex: I think the links above are mostly about accessing the same stream object from multiple threads, which is not what we are doing here. The same file is accessed by different streams (which will have a different underlying file handles too), so the thread-safety is down to the file system and whether it will allow a file being open for concurrent write.

I just did a test on my Linux and Windows machines with running two instances of the following program 
```c++
#include <fstream>
#include <iostream>
#include <string>

int main() {
    // Open the file
    std::ofstream out("test.txt");
    out.exceptions(std::ios::failbit | std::ios::badbit);

    // Wait for user signal to write to it
    int c;
    std::cin >> c;
    std::cout << out.is_open() << std::endl;

    // Write and close
    out << c;
    out.close();

    return 0;
}
```

On both OSes, the two program instances were able to open ``test.txt`` and write to it concurrently. So not thread-safe. I expected this wouldn't be possible...

That being said, the two threads would be writing exactly the same content to the file, so I don't expect this would be an issue in practice. If we want to make this more thread-safe, we can wrap any operation on cached kernels files by a mutex. That would solve multi-threaded access within one process, but not data races across multiple processes.

9prady9:
@cschreib-ibex #2850 is merged, you can rebase this branch for future changes/testing.

src/backend/cuda/nvrtc/cache.cpp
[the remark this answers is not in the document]
cschreib-ibex: C++ streams use ``fopen()`` [under the hood](https://en.cppreference.com/w/cpp/io/basic_filebuf/open).

According to the [C standard](https://wiki.sei.cmu.edu/confluence/display/c/FIO24-C.+Do+not+open+a+file+that+is+already+open), whether calling ``fopen()`` twice on the same file is allowed or not is implementation-defined. For those OSes that don't allow it, we're fine, because the file will just refuse to open twice and only one write will occur. From the above however, it looks like common OSes do allow it.

According to [this SO question](https://stackoverflow.com/questions/11723397/opening-a-file-using-fopen-with-same-flag-in-c), if the files are small and if we don't open the files in "append" mode, there is no risk of interleaving data in the output file, as the last thread that calls ``close()`` will flush its buffer and overwrite the content written by other threads. Although here I'm sure we're also in implementation-defined land (up to which file size will the OS use such buffering?).

I've open an [SO question](https://stackoverflow.com/questions/61339944/is-it-safe-for-two-threads-to-write-identical-content-to-the-same-file) to ask for further inputs.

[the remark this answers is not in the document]
cschreib-ibex: It's been proposed on SO that we should write the kernel to a temporary file, and rename it to the final cached file once the writing is complete. This prevents concurrent writes to the same file, and incidentally, it also ensures that whenever a thread finds a cached kernel file, the file is complete. If you agree, I will implement this.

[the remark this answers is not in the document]
9prady9: I find it reasonable, let's wait for @umar456 thoughts on the idea ? I will ping him tonight.

Thanks for checking out the potential scenarios. This is going to be a great asset for kernel caching.

src/backend/cuda/nvrtc/cache.cpp
9prady9: I will check about windows and your suggestions, will keep you posted.

cschreib-ibex:
@umar Thanks for the comments, I'll implement your suggested changes. 

As for the threaded caching, using a separate thread for writing would still not safeguard us against two separate ArrayFire executables writing the same kernel file at the same time. The solution of writing to a temporary file with a thread-unique name (which guarantees no concurrent writes) and trying to move the file at the end is really easy to implement and provides full safety. I've implemented and tested it locally, but haven't pushed the changes here yet, waiting on your decision.

Another solution, which has been suggested on SO, is to use OS-specific functions to lock the file prior to reading/writing into it, which will then prevent two threads (even from different processes) from accessing it simultaneously. I like this solution less because it requires platform-dependent code (which can get harder to maintain in the long run) while the other solution above is standard C/C++.

cschreib-ibex:
Hmm small issue: looks like ``std::hash`` is not guaranteed to always return the same value for identical inputs if the program is stopped and restarted. [Source](https://en.cppreference.com/w/cpp/utility/hash).
> Hash functions are only required to produce the same result for the same input within a single execution of a program; this allows salted hashes that prevent collision denial-of-service attacks.

This means we should be using something else for setting the cached kernel filename, but also for setting the hashed "name" of JIT kernels. There's plenty of simple deterministic string hashing functions out there that could do the job. Any recommendation?

src/backend/common/util.cpp
```diff
@@ -84,3 +91,86 @@ void saveKernel(const std::string& funcName, const std::string& jit_ker,
...
+    return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
+#else
+    struct _stat status;
+    return _stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
+#endif
+}
+
+bool createFolder(const string& path) {
+#if !defined(OS_WIN)
+    return mkdir(path.c_str(), 0777) == 0;
```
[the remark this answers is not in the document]
cschreib-ibex: I am not too versed in these permission strings. Isn't the "execute" bit required for simple things like "ls"? Also I think we can't make this directory only owned by the current user, as if the folder is somewhere on the root (like "/var/tmp/arrayfire/") this could potentially be accessed by multiple users.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -334,21 +362,70 @@ Kernel buildKernel(const int device, const string &nameExpr,
...
+        name.resize(nameSize);
+        in.read(&name[0], nameSize);
+
+        size_t cubinSize;
+        in >> cubinSize;
+        vector<char> cubin(cubinSize);
+        in.read(cubin.data(), cubinSize);
+        in.close();
+
+        CU_CHECK(cuModuleLoadDataEx(&module, cubin.data(), 0, 0, 0));
```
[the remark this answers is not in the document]
cschreib-ibex: I've added the ``#ifdef`` and set it as default ``ON`` in the CMake file.

src/backend/common/util.cpp
```diff
@@ -84,3 +91,86 @@ void saveKernel(const std::string& funcName, const std::string& jit_ker,
...
+    const string testPath = path + AF_PATH_SEPARATOR + "test";
+    if (!std::ofstream(testPath).is_open()) return false;
+    removeFile(testPath);
+
+    return true;
+}
+
+const string& getCacheDirectory() {
+    thread_local std::once_flag flag;
+    thread_local string cacheDirectory;
```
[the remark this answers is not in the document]
cschreib-ibex: Done

src/backend/common/util.cpp
```diff
@@ -84,3 +91,86 @@ void saveKernel(const std::string& funcName, const std::string& jit_ker,
...
+#if !defined(OS_WIN)
+            "/var/lib/arrayfire",
+            getHomeDirectory() + "/.arrayfire",
+            "/tmp/arrayfire"
+#else
+            getTemporaryDirectory() + "\\ArrayFire"
+#endif
+        };
+
+        for (const string& path : pathList) {
```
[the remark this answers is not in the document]
cschreib-ibex: Done

src/backend/common/util.cpp
```diff
@@ -84,3 +91,86 @@ void saveKernel(const std::string& funcName, const std::string& jit_ker,
...
+    struct stat status;
+    return stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
+#else
+    struct _stat status;
+    return _stat(path.c_str(), &status) == 0 && (status.st_mode & S_IFDIR) != 0;
+#endif
+}
+
+bool createFolder(const string& path) {
+#if !defined(OS_WIN)
```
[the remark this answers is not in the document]
cschreib-ibex: Done

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -313,6 +327,20 @@ Kernel buildKernel(const int device, const string &nameExpr,
...
+    if (!cacheDirectory.empty()) {
+        const string cacheFile = cacheDirectory + AF_PATH_SEPARATOR +
+                                 getKernelCacheFilename(device, nameExpr);
+
+        std::ofstream out(cacheFile, std::ios::binary);
+        const size_t nameSize = strlen(name);
+        out << nameSize;
+        out.write(name, nameSize);
+        out << cubinSize;
+        out.write(static_cast<const char *>(cubin), cubinSize);
```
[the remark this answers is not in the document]
cschreib-ibex: I added this feature and did not measure a noticeable impact on performances.

src/backend/cuda/nvrtc/cache.cpp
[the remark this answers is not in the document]
cschreib-ibex: FYI I've implemented the temporary file + rename and pushed the changes.

src/backend/common/util.cpp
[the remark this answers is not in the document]
umar456: the execute bit is needed on binary executable but since we will only be writing once and reading multiple times 644 would be a conservative option. Other users and groups can still read the file but they don't necessarily need to write there.

umar456:
FNV1a is a fine algorithm for now. If we run into any issues we can look into another algorithm although that would be a hell of a debug session. 

> @umar Thanks for the comments, I'll implement your suggested changes.
> 
> As for the threaded caching, using a separate thread for writing would still not safeguard us against two separate ArrayFire executables writing the same kernel file at the same time. The solution of writing to a temporary file with a thread-unique name (which guarantees no concurrent writes) and trying to move the file at the end is really easy to implement and provides full safety. I've implemented and tested it locally, but haven't pushed the changes here yet, waiting on your decision.
> 
> Another solution, which has been suggested on SO, is to use OS-specific functions to lock the file prior to reading/writing into it, which will then prevent two threads (even from different processes) from accessing it simultaneously. I like this solution less because it requires platform-dependent code (which can get harder to maintain in the long run) while the other solution above is standard C/C++.

Good point about the multiple process issue. I agree that could be a problem with multiple application running at the same time. Lets go with the move approach you suggested. I think as long as we don't hit this situation often we will not run into performance issues where you have multiple threads writing to disk.

src/backend/common/util.cpp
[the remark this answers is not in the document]
cschreib-ibex: Note this is the permission for the directory, not for the files that get written in it. For directories, the execute bit must be set at least for the current user, [otherwise you cannot create / remove files](https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work). Other users might need write access to that directory if they want to cache their own kernels there too (if that directory is not in the users's home, that is).

CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ option(AF_BUILD_FORGE
 option(AF_WITH_NONFREE  "Build ArrayFire nonfree algorithms"   OFF)
 option(AF_WITH_LOGGING  "Build ArrayFire with logging support" ON)
 option(AF_WITH_STACKTRACE  "Add stacktraces to the error messages." ON)
+option(AF_CACHE_KERNELS "Enable caching kernels on disk"       ON)
```
umar456, line 63: This option will not define anything when building ArrayFire. You will need to add the definition here:

https://github.com/arrayfire/arrayfire/blob/master/CMakeLists.txt#L221

Perhaps name it CACHE_KERNELS_TO_DISK? because we cache kernels in memory but its not persistent across multiple processes.

src/backend/common/util.cpp
```diff
@@ -91,3 +102,115 @@ std::string int_version_to_string(int version) {
...
+const string& getCacheDirectory() {
+    static std::once_flag flag;
+    static string cacheDirectory;
+
+    std::call_once(flag, []() {
+        const vector<string> pathList = {
+#if defined(OS_WIN)
+            getTemporaryDirectory() + "\\ArrayFire"
+#else
+            "/var/lib/arrayfire",
```
umar456, line 176: Lets remove this entry. I am not sure how to handle this scenario wrt permissions. I think its safe to assume that the home directory is sufficient.

cschreib-ibex: Ok. Shall I remove ``/tmp`` as well?

src/backend/common/util.cpp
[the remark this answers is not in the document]
umar456: Ahh. my mistake. never mind then.

CMakeLists.txt
cschreib-ibex: Done!

src/backend/common/util.cpp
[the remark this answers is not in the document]
cschreib-ibex: In my home directory on my Linux Mint machine, most directories have permission 0755, but ``.local`` is 0700. I'd go with 0755 if the directory is in the user's home folder, and 0777 otherwise.

src/backend/common/util.cpp
umar456: I think tmp is fine. There are some cases where a service will be running on a user without a home directory. I can't think of a case where tmp is not available.

cschreib-ibex: Ok.

src/backend/common/util.cpp
```diff
@@ -91,3 +102,114 @@ std::string int_version_to_string(int version) {
...
+}
+#else
+string getHomeDirectory() {
+    string home = getEnvVar("XDG_CACHE_HOME");
+    if (!home.empty()) return home;
+
+    home = getEnvVar("HOME");
+    if (!home.empty()) return home;
+
+    home = getpwuid(getuid())->pw_dir;
```
umar456, line 123: This value is not being returned. I was getting a warning when building on linux.

9prady9: Yeah, missing return for third case.

src/backend/common/util.hpp
```diff
@@ -21,3 +21,21 @@ void saveKernel(const std::string& funcName, const std::string& jit_ker,
...
+
+bool renameFile(const std::string& sourcePath, const std::string& destPath);
+
+bool isDirectoryWritable(const std::string& path);
+
+std::string makeTempFilename();
+
+std::size_t deterministicHash(const void* data, std::size_t byteSize);
+
+std::size_t deterministicHash(const std::string& data);
```
9prady9, line 41: All directory helpers are self-explanatory with their name. The following however would make understanding easier if they had one or two line documentation above them.
- `makeTempFilename` - something like what to expect in the output string as the name is also a little vague. Function says file name, but in actual it is returning just a string based on active thread ID and current callee counter value in a given thread. I understand the intent, but a couple of lines of docs above it and perhaps a name change could help for future developers.
- `deterministicHash` - You could say in docs that this fn currently return FNV1a hash
- `deterministicHash` - this version is just a wrapper around the other one.

src/backend/cuda/nvrtc/cache.cpp
```diff
@@ -334,21 +378,81 @@ Kernel buildKernel(const int device, const string &nameExpr,
...
+        vector<char> cubin(cubinSize);
+        in.read(cubin.data(), cubinSize);
+        in.close();
+
+        // check CUBIN binary data has not been corrupted
+        const size_t recomputedHash =
+            deterministicHash(cubin.data(), cubinSize);
+        if (recomputedHash != cubinHash) {
+            AF_ERROR("cached kernel data is corrupted", AF_ERR_LOAD_SYM);
+        }
```
9prady9, line 416: Nice, content hashing!

src/backend/common/util.cpp
cschreib-ibex: Fixed

src/backend/common/util.hpp
cschreib-ibex: I have added the documentation, please let me know if it is adequate!

9prady9: Looks good.

9prady9:
@cschreib-ibex Final one, formatting seems to be off in `/src/backend/common/util.hpp` Please correct it.

cschreib-ibex:
@9prady9 Man, clang-tidy is very particular :) It doesn't like the default behavior of Visual Studio. Anyways, I've fixed the formatting.

9prady9:
@cschreib-ibex I understand. Sorry, there seems to be some other thing off. `clang-format -i src/backend/common/util.hpp` should take care of the entire file as per config we have. Try that.

9prady9:
Finally! Thanks @cschreib-ibex

cschreib-ibex:
No worries :P Thanks for the tip on ``clang-format``, I actually had it installed, never ran it before...

jacobkahn:
@cschreib-ibex this is really cool. Out of curiosity, do you have any benchmarks with respect to a performance improvements?

cschreib-ibex:
On our systems, and with our particular application which makes heavy use of ArrayFire, the runtime when "cold" (i.e., the first run of our processing after the application has started) went down from 28s to 9s.

PR #2849, opened 2020-04-20
Improve the performance of CPU join and transpose

umar456:
This PR improves the performance of the transpose and join kernels in the CPU backend. The previous approach was naive and was not optimized for data locality. This new version is slightly improved and uses a tile based approach to speed up the operation.

The join kernel is improved by using a memcpy call instead of a for loop to perform the copy to the output matrix.

Fixed several warnings using the -Wall flag in GCC and enabled it by default in CMake

Fixed a matrix multiplication test where the output matrix was not being tested

Fixed a potential issue with mean where the optimizer could remove some operations that could reduce the accuracy of the result.

Use double to calculate the mean for the random engine uniform tests to avoid overflow issues with larger arrays

Fixed a bug in the DefaultMemoryManager introduced in f2112530c where I was dereferencing an iterator before checking if the find function returned an actual value.

### DIFF
--- a/src/backend/cpu/join.cpp
+++ b/src/backend/cpu/join.cpp
@@ -20,8 +20,8 @@ using common::half;
 
 namespace cpu {
 
-template<typename Tx, typename Ty>
-Array<Tx> join(const int dim, const Array<Tx> &first, const Array<Ty> &second) {
+template<typename T>
+Array<T> join(const int dim, const Array<T> &first, const Array<T> &second) {
     // All dimensions except join dimension must be equal
     // Compute output dims
     af::dim4 odims;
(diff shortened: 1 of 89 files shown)


src/backend/cpu/join.cpp
```diff
@@ -20,8 +20,8 @@ using common::half;
 
 namespace cpu {
 
-template<typename Tx, typename Ty>
-Array<Tx> join(const int dim, const Array<Tx> &first, const Array<Ty> &second) {
+template<typename Tx>
```
9prady9, line 23: nit: Can be just `T`

src/backend/opencl/surface.cpp
```diff
@@ -11,7 +11,6 @@
 #include <GraphicsResourceManager.hpp>
 #include <debug_opencl.hpp>
 #include <err_opencl.hpp>
-#include <join.hpp>
 #include <reduce.hpp>
 #include <reorder.hpp>
```
9prady9, line 15: reduce and reorder headers are also not needed here now.

PR #2851, opened 2020-04-20
Fix work group function name conflict in OpenCL 2.0

9prady9:


src/backend/opencl/kernel/reduce_blocks_by_key_dim.cl
```diff
@@ -7,7 +7,7 @@
  * http://arrayfire.com/licenses/BSD-3-Clause
  ********************************************************/
 
-Tk work_group_scan_inclusive_add(__local Tk *wg_tmp, __local Tk *arr) {
+Tk wgScanInclusiveAdd(__local Tk *wg_tmp, __local Tk *arr) {
```
umar456, line 10: If we can mimic the same functionality we should be able to just ifdef this out.

9prady9: Yeah, true - the thought crossed my mind. I think I saw different signatures, so postponed it for later wondering if we should push fix release soon :thinking:

9prady9: I believe we can't declare local memory inside a non-kernel function in OpenCL version we develop 1.2 or lower. I have to verify this, but I remember handling this few weeks back in another kernel.

If we can't do the above, I don't think we can mimic https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/work_group_reduce.html API, because it takes a single scalar much like sync_threads_or in CUDA.

umar456: Yeah, that wouldn't be possible in 1.2. nevermind.

umar456: Can you think of a way to implement the alternative 2.0 code path for this kernel?

9prady9: Let me look into it.

9prady9:
Windows github ci job fix is in PR #2852

PR #2852, opened 2020-04-21
Enable support for large kernels in morphological ops

### LINKED ISSUE #1941: Add support bigger CUDA kernel size for Morphological Operations (erode/dilate)
Up to 19x19 square kernels are only supported currently in CUDA, it's too small.

PS: Other libs (like OpenCV) works with 150x150px.

9prady9:
Resolves #1941 

Requires https://github.com/arrayfire/arrayfire-data/pull/18


9prady9:
I am looking into the regression failures on ubuntu(MKL) CPU jobs.

9prady9:
Was able to reproduce, seems to be happening with MKL based fft only. Looking into it.

PR #2854, opened 2020-04-23
Pass USE_(DOUBLE,HALF) definitions to OpenCL build in a uniform way

umar456:
* Creates the getTypeBuildDefinition which will pass the USE_DOUBLE
  and USE_HALF types to the OpenCL build step in a consistent way
* The reduce by key step was not passing the USE_HALF flag so
  things weren't compiling on the Intel OpenCL GPU implementation

### DIFF
--- a/src/backend/opencl/types.hpp
+++ b/src/backend/opencl/types.hpp
@@ -20,6 +20,8 @@
 #include <common/kernel_type.hpp>
 #include <common/traits.hpp>
 
+#include <algorithm>
+#include <array>
 #include <string>
 
 namespace common {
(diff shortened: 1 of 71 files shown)


src/backend/opencl/types.hpp
```diff
@@ -119,4 +121,32 @@ const char *getFullName() {
     return af::dtype_traits<T>::getName();
 }
 
+template<typename... ARGS>
+constexpr const char *getTypeBuildDefinition() {
+    using std::any_of;
+    using std::array;
+    using std::begin;
+    using std::end;
+    std::array<bool, sizeof...(ARGS)> is_half = {
```
9prady9, line 130: redundant namespace qualifier.

src/backend/opencl/types.hpp
```diff
@@ -119,4 +121,32 @@ const char *getFullName() {
...
+        std::is_same<ARGS, common::half>::value...};
```
9prady9, line 131: can loose `std::` if `using std::is_same` is added above as well.

umar456:
Addressed all feedback

PR #2855, opened 2020-04-23
Adjust JIT heuristics for the Intel GPU on the OpenCL backend

umar456:


src/backend/opencl/Array.cpp
```diff
@@ -317,11 +323,19 @@ kJITHeuristics passesJitHeuristics(Node *root_node) {
         constexpr size_t max_nvidia_param_size = (4096 - base_param_size);
         constexpr size_t max_amd_param_size    = (3520 - base_param_size);
 
+        // This value is really for the Intel HD Graphics platform. The CPU
+        // platform seems like it can handle unlimited parameters but the
+        // compile times become very large.
+        constexpr size_t max_intel_cpu_param_size =
```
9prady9, line 329: incorrect var name - can lead to confusion `max_intel_igpu_param_size` seems more appropriate

umar456: You are right it should be:
```suggestion
        constexpr size_t max_intel_igpu_param_size =
```

PR #2860, opened 2020-04-27
Create a test library to speed up test compilation

umar456:


test/var.cpp
```diff
@@ -114,8 +114,8 @@ TYPED_TEST(Var, DimCPPSmall) {
     vector<vector<TypeParam> > in;
     vector<vector<outType> > tests;
 
-    readTests<TypeParam, outType, double>(TEST_DIR "/var/var.data", numDims, in,
-                                          tests);
+    readTests<TypeParam, outType, float>(TEST_DIR "/var/var.data", numDims, in,
```
9prady9, line 117: Are these changes to reduce number of instantiations for `readTests` ?

test/CMakeLists.txt
```diff
@@ -70,6 +70,38 @@ if(AF_BUILD_UNIFIED)
...
+
+target_link_libraries(arrayfire_test
+  PRIVATE
+    gtest_main
+  PUBLIC
+    gtest
+  )
+
+if(AF_TEST_WITH_MTX_FILES)
+  target_link_libraries(arrayfire_test PRIVATE mmio)
```
9prady9, line 89: Shouldn't this be public ? I mean you can perhaps remove this from linking command in `make_test` macro and make it PUBLIC dependency here.

test/var.cpp
umar456: yes

PR #2865, opened 2020-04-29
adds bitwise not(~) operation to all backends

syurkevi:
This unary operation had been missing. Twos complement -x - 1 trick would fail for some types due to our upcasting behavior.

### DIFF
--- a/include/af/arith.h
+++ b/include/af/arith.h
@@ -741,6 +741,19 @@ extern "C" {
     */
     AFAPI af_err af_not   (af_array *out, const af_array in);
 
+#if AF_API_VERSION >= 38
+    /**
+       C Interface for performing bitwise not on input
+
+       \param[out] out will contain result of bitwise not of \p in.
+       \param[in] in is the input
+       \return \ref AF_SUCCESS if the execution completes properly
+
+       \ingroup arith_func_bitnot
+    */
+    AFAPI af_err af_bitnot   (af_array *out, const af_array in);
+#endif
+
     /**
        C Interface for performing bitwise and on two arrays
 
--- a/src/api/c/unary.cpp
+++ b/src/api/c/unary.cpp
@@ -560,6 +560,41 @@ af_err af_not(af_array *out, const af_array in) {
     return AF_SUCCESS;
 }
 
+template<typename T>
+static inline af_array bitOpNot(const af_array in) {
+    return unaryOp<T, af_bitnot_t>(in);
+}
+
+af_err af_bitnot(af_array *out, const af_array in) {
+    try {
+        const ArrayInfo &iinfo = getInfo(in);
+        const af_dtype type    = iinfo.getType();
+
+        dim4 odims = iinfo.dims();
+
+        if (odims.ndims() == 0) {
+            return af_create_handle(out, 0, nullptr, type);
+        }
+
+        af_array res;
+        switch (type) {
+            case s32: res = bitOpNot<int>(in); break;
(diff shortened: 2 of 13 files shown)


include/af/arith.h
```diff
@@ -741,6 +741,17 @@ extern "C" {
     */
     AFAPI af_err af_not   (af_array *out, const af_array in);
 
+    /**
```
umar456, line 744: #if AF_API_VERSION >= 38

9prady9: I don't understand what is different from the file and your comment ?

src/api/c/unary.cpp
```diff
@@ -560,6 +560,42 @@ af_err af_not(af_array *out, const af_array in) {
...
+        if (odims.ndims() == 0) {
+            return af_create_handle(out, 0, nullptr, type);
+        }
+
+        af_array res;
+        switch (type) {
+            case s32: res = bitOpNot<int>(in); break;
+            case u32: res = bitOpNot<uint>(in); break;
+            case u8: res = bitOpNot<uchar>(in); break;
+            case b8: res = bitOpNot<char>(in); break;
```
9prady9, line 585: bitwise not of b8 array is not needed since, b8 is boolean and `operator!` should do fine for that one right ?

src/api/c/unary.cpp
```diff
@@ -560,6 +560,42 @@ af_err af_not(af_array *out, const af_array in) {
     return AF_SUCCESS;
 }
 
+template<typename T>
+static inline af_array bitOpNot(const af_array in) {
+    af_array res = unaryOp<T, af_bitnot_t>(in);
+    return res;
```
9prady9, line 566: nit: can be clubbed into one line `return unaryOp<T, af_bitnot_t>(in)`;

src/api/cpp/array.cpp
```diff
@@ -988,6 +988,13 @@ array array::operator!() const {
     return array(out);
 }
 
+array array::operator~() const {
+    af_array lhs = this->get();
+    af_array out;
```
9prady9, line 993: out = nullptr;

test/binary.cpp
```diff
@@ -289,6 +289,28 @@ BITOP(bitxor, uintl, ^)
...
+        const int num = 10;                                           \
+        af::array a   = af::constant(vala, num, ty);                  \
+        af::array c   = op a;                                         \
+        T *h_a        = a.host<T>();                                  \
+        T *h_c        = c.host<T>();                                  \
+        for (int i = 0; i < num; i++)                                 \
+            ASSERT_EQ(h_c[i], valc)                                   \
+                << "for values: " << h_a[i] << endl;                  \
+        af_free_host(h_a);                                            \
+        af_free_host(h_c);                                            \
```
9prady9, line 306: I think we have a macro that takes two two arrays for comparison. ASSERT_ARRAYS_EQ I think

test/binary.cpp
```diff
@@ -289,6 +289,28 @@ BITOP(bitxor, uintl, ^)
...
+    }
+
+UBITOP(bitnot,   int, ~)
+UBITOP(bitnot,  uint, ~)
+UBITOP(bitnot,  intl, ~)
+UBITOP(bitnot, uintl, ~)
```
9prady9, line 312: add for short and ushort too. unsigned char(u8)

test/binary.cpp
9prady9: I took care of this in my local commit, will push soon after fixing the windows issue.

PR #2871, opened 2020-05-05
broadcasting in af_arith for binary operations

syurkevi:
Will prepend 1s until dimension sizes match, then 1s are treated as wildcards and are broadcasted to their matching sizes.
```
array A = range(dim4(32, 32, 3), 2);  // [32 32 3]
array B = -range(dim4(3)); //    [ 3 ] -> [1* 1* 3]

array C = A + B; // [32 32 3]
```

In cases where one of the arguments is a vector, prepending until dimensions match may not align well:
```
 array A = constant(-1, dim4(15, 3, 5)); // [15 3 5]
 array B = constant(1, dim4(3, 1));        // [ 1  1 3] - misaligned,
 // even though 3 intended to broadcast with second dimension of A
```
Upon a mismatch, the prepended vectors will be shifted left until there is a match. In the above example the broadcast will be a success after a single additional shift.

This manner of treating vectors is a workaround due to no difference between.
```
af::dim4 d1(3);
af::dim4 d2(3, 1);
af::dim4 d3(3, 1, 1);
af::dim4 d4(3, 1, 1, 1);
// d1 == d2 == d3 == d4
```
TODO: 
* n-ary operations?
* complex values


src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
...
+    }
+    return padded;
+}
+
+std::tuple<dim4, dim4, dim4, dim4, dim4> getBroadcastDims(
+    const ArrayInfo &linfo, const ArrayInfo &rinfo) {
+    const dim4 ldims = linfo.dims();
+    const dim4 rdims = rinfo.dims();
+
+    unsigned maxndims =
```
9prady9, line 66:
```c++
unsigned maxndims = std::max(ldims.ndims(), rdims.ndims());
```
Would this break logic ?

src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
     return res;
 }
 
+dim4 paddims(const dim4 &dims, unsigned offset) {
+    dim4 padded(1, 1, 1, 1);
+
+    const unsigned MAXDIMS = 4;
+    for (unsigned i = 0; i < MAXDIMS; ++i) {
```
9prady9, line 55: MAXDIMS -> AF_MAX_DIMS

We can then remove this new constant MAXDIMS

src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
     return res;
 }
 
+dim4 paddims(const dim4 &dims, unsigned offset) {
```
9prady9, line 51: paddims -> padDims to be consistent

src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
...
+
+    // pre-pad each dimension with 1s if necessary
+    dim4 ldims_padded = paddims(ldims, loffset);
+    dim4 rdims_padded = paddims(rdims, roffset);
+
+    dim4 bdims;
+    dim4 ltile(1, 1, 1, 1);
+    dim4 rtile(1, 1, 1, 1);
+    // check that dimensions are compatible
+    for (unsigned i = 0; i < 4; ++i) {
```
9prady9, line 81: `4` -> `AF_MAX_DIMS`

src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
...
+    const ArrayInfo &linfo = getInfo(lhs);
+    const ArrayInfo &rinfo = getInfo(rhs);
+
+    dim4 odims, ltile, rtile, lpad, rpad;
+    std::tie(odims, ltile, rtile, lpad, rpad) = getBroadcastDims(linfo, rinfo);
+
+    detail::Array<T> lhst =
+        detail::tile<T>(modDims(getArray<T>(lhs), lpad), ltile);
+    detail::Array<T> rhst =
+        detail::tile<T>(modDims(getArray<T>(rhs), rpad), rtile);
```
9prady9, line 135: Add using statements for the following to fit convention we are following in other files.
- `using detail::Array`
- `using detail::tile`
- `using std::tie`

src/api/c/binary.cpp
```diff
@@ -46,6 +48,96 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
...
+
+    af_array res = getHandle(arithOp<T, op>(lhst, rhst, odims));
+    return res;
```
9prady9, line 138: Above two lines can be clubbed.

test/array.cpp
```diff
@@ -582,3 +582,69 @@ TEST(Array, InitializerListAndDim4) {
 
     ASSERT_ARRAYS_EQ(A, B);
 }
+
+TEST(Broadcast, Simple1) {
```
umar456, line 586: Can you create parameterized tests for the input dims? You also want to compare this functionality to the tiled version of the same operation. Additionally you should make sure that the kernels generated and the operations performed by these functions are what you expect.

syurkevi:
As per the discussion during the maintainers meeting, support for matching vectors to corresponding matrix dimensions was dropped due to ambiguous cases with multiple matching dimensions. Instead, only the 1s wild-card style broadcasting will be supported:
```
array a = randu(5, 3) + randu(5); // Okay
array b = randu(5, 3) + randu(3); // Fails
array c = randu(5, 1, 3) + randu(5); // Okay
array d = randu(5, 1, 3) + randu(1, 4); // Okay
array e = randu(5, 1, 3) + randu(1, 1, 3); // Okay
array f = randu(5, 1, 3) + randu(5, 5, 3); // Okay
array g = randu(5, 1, 3) + randu(5, 3, 3); // Okay

array i = randu(3, 255, 255) + randu(3); // Okay
```

umar456 (changes requested):

src/api/c/binary.cpp
```diff
@@ -55,6 +60,36 @@ static inline af_array arithOp(const af_array lhs, const af_array rhs,
...
+        if (lshape[d] == rshape[d]) {
+            ltile[d] = rtile[d] = 1;
+        } else if (lshape[d] == 1 && rshape[d] > 1) {
+            ltile[d] = odims[d];
+        } else if (lshape[d] > 1 && rshape[d] == 1) {
+            rtile[d] = odims[d];
+        }
+    }
+
+    Array<T> lhst = tile<T>(modDims(getArray<T>(lhs), lshape), ltile);
```
umar456, line 87: This operation to tile will always call the kernel even if the tile operation could be performed by the JIT.

It would be better to move the tile operation to the common namespace. I have a commit that does this that I will be pushing soon.

test/array.cpp
```diff
@@ -641,23 +641,167 @@ TEST(Array, ReferenceCount2) {
     }
 }
 
-// This tests situations where the compiler incorrectly assumes the initializer
-// list constructor instead of the regular constructor when using the uniform
-// initilization syntax
+TEST(Broadcast, VectorMatrix2d) {
```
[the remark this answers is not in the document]
umar456: I think these tests should be moved to the binary test file. The array.cpp file should only test the af_array and af::array object member functions and operations. I have a commit that does this and should be pushing it soon.

umar456 (changes requested):

umar456 (approved):

Issue #2857, opened 2020-04-24
Create a common API to cache CUDA/OpenCL kernels


9prady9:
This would include caching to disk feature as well.

- [x] Create unified API for caching kernels - WIP 
- [x] Port CUDA kernels to use new API - #2560 #2790
- [x] Port OpenCL kernels to use new API
- [x] (Optional) Add disk store support for OpenCL binaries
- [x] ~(Optional) Unify disk store feature for compiled kernels~

9prady9:
Main features are done, will handle disk caching for OpenCL now.

PR #2873, opened 2020-05-05
Merge kernel caching logic for CUDA and OpenCL backends

9prady9:
Partially addresses tasks from #2857


src/backend/common/kernel_util.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <string>
+
+template<typename T>
+std::string toString(T value);
+
+struct TemplateArg {
+    std::string _tparam;
+
+    TemplateArg(std::string str) : _tparam(str) {}
```
umar456, line 20:
```suggestion
    TemplateArg(std::string str) : _tparam(std::move(str)) {}
```

src/backend/common/kernel_util.hpp
```diff
@@ -0,0 +1,51 @@
...
+ ********************************************************/
+
+#include <traits.hpp>
+
+#include <string>
+
+template<typename T>
+std::string toString(T value);
+
+struct TemplateArg {
```
umar456, line 17: maybe define this in its own header?

src/backend/common/kernel_util.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <string>
+
+using std::string;
+
+template<typename T>
+string toString(T value) {
+    return std::to_string(value);
+}
+
+template string toString<int>(int);
```
umar456, line 25: why not use the std::to_string

src/backend/common/kernel_interface.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <cstddef>
+#include <utility>
+
+namespace common {
+
+/// Kernel Interface that should be implemented by each backend
+template<typename Module, typename Kernel, typename Enqueuer, typename DevPtr>
+class KernelInterface {
+   public:
+    using ModuleType   = Module;
```
umar456, line 21: These using statements don't add anything. IMO

src/backend/common/kernel_interface.hpp
```diff
@@ -0,0 +1,84 @@
...
+    using KernelType   = Kernel;
+    using EnqueuerType = Enqueuer;
+    using DevPtrType   = DevPtr;
+
+    Module mProgram;
```
umar456, line 26: These should be private

src/backend/common/kernel_util.hpp
9prady9: Moving them in next commit.

PR #2875, opened 2020-05-05
Add an ArrayFire conanfile.py that pulls from the linux binary installer

jacobkahn:
[Conan](https://conan.io/) is a C++ package manager with a large ecosystem and several projects that have dependencies on CUDA/OpenCL (e.g. [OpenCV](https://github.com/conan-community/conan-opencv/blob/release/4.1.1/conanfile.py)).

This Conan configuration is minimal and only pulls from the Linux binary installer. I've tested it with other downstream projects (e.g. [flashlight](https://github.com/facebookresearch/flashlight)), and it works. The installer as it stands is backend agnostic since it installs all libraries in the binary installer (except for `debug` libraries). One could eventually add a version that builds from source (and is thus eligible for all of the configuration options that Conan provides), but the Linux installer is already quite generic/has many of these things already.

I'll also be adding the package provided by this installer to the [`conan-community` remote](https://docs.conan.io/en/latest/uploading_packages/remotes.html#conan-community) which has packages which are still experimental/being improved. We can easily push breaking changes to those at any point.

(I've added myself as author of the package config for now -- we should change that to be an appropriate author/a generic AF point of contact)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ compile_commands.json
 venv
 test/gtest
 src/backend/cuda/cub
+conanbuildinfo*
+conaninfo*
+conan.lock
+graph_info.json
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,0 +1,104 @@
+from conans import ConanFile, CMake, tools
+import os
+
+
+class ArrayFireConan(ConanFile):
+    name = "arrayfire"
+    version = "3.7.1"
+    license = "BSD"
+    author = "jacobkahn jacobkahn1@gmail.com"
+    url = "https://github.com/arrayfire/arrayfire"
+    requires = []
+    description = "ArrayFire: a general purpose GPU library"
+    topics = ("arrayfire", "gpu", "cuda", "opencl", "gpgpu",
+              "hpc", "performance", "scientific-computing")
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False], "backend": ["CPU", "CUDA", "OPENCL"]}
+    default_options = {"shared": True}
+    generators = "cmake"  # unused
+
+    def configure(self):
+        if self.settings.os == "Windows":
+            raise ConanInvalidConfiguration(
+                "Linux binary installer not compaible with Windows.")
+
+    def _download_arrayfire(self):
+        # Download ArrayFire 3.7.1
+        self.af_installer_local_path = 'ArrayFire-v3.7.1-1_Linux_x86_64.sh'
+        if not os.path.exists(self.af_installer_local_path):
+            self.output.info(
+                "Downloading the ArrayFire 3.7.1 binary installer...")
+            tools.download(
+                'https://arrayfire.s3.amazonaws.com/3.7.1/ArrayFire-v3.7.1-1_Linux_x86_64.sh', self.af_installer_local_path)
+            self.output.success(
+                f"ArrayFire 3.7.1 binary installer successfully downloaded to {self.af_installer_local_path}")
+        else:
+            self.output.info(
+                "ArrayFire 3.7.1 binary installer already exists - skipping download.")
+
+    def _unpack_arrayfire(self):
+        if not os.path.exists(self.af_unpack_path):
+            os.mkdir(self.af_unpack_path)
+        self.output.info("Unpacking ArrayFire 3.7.1 binary installer...")
+        cmd = f"bash {self.af_installer_local_path} --prefix={self.af_unpack_path} --skip-license"
+        self.run(cmd)
+        self.output.success("ArrayFire 3.7.1 successfully unpacked.")
+
+    def _process_arrayfire(self):
+        # Install ArrayFire to requisite path
+        self.af_unpack_path = os.path.join(self.source_folder, 'arrayfire')
+
+        # Only proceed if missing
+        if os.path.exists(os.path.join(self.af_unpack_path, 'include', 'arrayfire.h')):
+            self.output.info("ArrayFire 3.7.1 already unpacked - skipping.")
+        else:
+            self._download_arrayfire()
+            self._unpack_arrayfire()
+
+    def build(self):
+        self._process_arrayfire()
+
+    def package(self):
+        # libs
+        self.copy("*.so", dst="lib", keep_path=False, symlinks=True)
+        self.copy("*.so.*", dst="lib", keep_path=False, symlinks=True)
+
+        # headers
+        self.copy("*.h", dst="include", src="arrayfire/include")
+        self.copy("*.hpp", dst="include", src="arrayfire/include")
+
+    def package_info(self):
+        self.cpp_info.libs = [
+            "libaf.so.3.7.1",
+            "libforge.so.1.0.5",
+        ]
+        if self.options.backend == 'CUDA':
+            self.cpp_info.libs.extend([
+                "libnvrtc-builtins.so",
+                "libcudnn.so.10.0",
+                "libcusparse.so.10.0",
+                "libcublas.so.10.0",
+                "libafcuda.so.3.7.1",
+                "libcusolver.so.10.0",
+                "libnvrtc.so.10.0",
+                "libcufft.so.10.0",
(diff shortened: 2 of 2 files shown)


conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+    version = "3.7.1"
+    license = "BSD"
+    author = "jacobkahn jacobkahn1@gmail.com"
+    url = "https://github.com/arrayfire/arrayfire"
+    requires = []
+    description = "ArrayFire: a general purpose GPU library"
+    topics = ("arrayfire", "gpu", "cuda", "opencl", "gpgpu",
+              "hpc", "performance", "scientific-computing")
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False], "backend": ["CPU", "CUDA", "OPENCL"]}
```
9prady9, line 16: Shouldn't unified backend also be present ?

conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+                "libmkl_intel_lp64.so",
+                "libmkl_core.so",
+                "libmkl_avx.so",
+                "libmkl_def.so",
+                "libiomp5.so",
+                "libafopencl.so.3.7.1",
+                "libmkl_avx512.so",
+                "libmkl_intel_thread.so",
+                "libmkl_mc3.so",
+                "libafcpu.so.3.7.1",
```
9prady9, line 99: It would be nice if we can do string substitution for `3.7.1` from the member variable `version`. This way update to newer version can be handled by editing just a couple of lines.

conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+        self.cpp_info.libs = [
+            "libaf.so.3.7.1",
+            "libforge.so.1.0.5",
+        ]
+        if self.options.backend == 'CUDA':
+            self.cpp_info.libs.extend([
+                "libnvrtc-builtins.so",
+                "libcudnn.so.10.0",
+                "libcusparse.so.10.0",
+                "libcublas.so.10.0",
```
9prady9, line 80: similar to `version` member variable, we can have `cuda_toolkit_ver` set to `10.0` that can be string substituted for all cuda dependencies. This way update to newer version can be handled by editing just a couple of lines.

conanfile.py
```diff
@@ -0,0 +1,104 @@
+from conans import ConanFile, CMake, tools
+import os
+
```
9prady9, line 3: Is it a hard requirement for this file to be made available in the root directory of the repository ? 

If not, I am contemplating if it would be a good idea to move this to a folder `package_configs` at the root level of repository. This location may be potentially used any configuration files associated for other package managers too.

@umar456 @syurkevi What do you guys think ?

9prady9:
arrayfire build especially for CUDA backend for all computes takes significant time, I wonder if conan allows/handles such huge build times while building a package from source. Let's say we do provide such facility, I have question, let's say someone wants to use araryfire conan package, could  they potentially build cuda backend for a specific compute on their development device ? I am not sure how much of benefit that would provide :thinking: 

Idea to add to conan-community sounds good to me.

conanfile.py
jacobkahn: (It's not a hard requirement at all, no)

conanfile.py
jacobkahn: 👍 done

conanfile.py
jacobkahn: 👍 done

conanfile.py
jacobkahn: Done. Also changed so users can opt-in to multiple backends.

jacobkahn: added 👍

.gitignore
```diff
@@ -16,3 +16,7 @@ compile_commands.json
 venv
 test/gtest
 src/backend/cuda/cub
+conanbuildinfo*
+conaninfo*
+conan.lock
```
9prady9, line 21: :+1:

conanfile.py
```diff
@@ -0,0 +1,126 @@
...
+            tools.download(
+                f"https://arrayfire.s3.amazonaws.com/{ARRAYFIRE_VERSION}/{BINARY_INSTALLER_NAME}", self.af_installer_local_path)
+            self.output.success(
+                f"ArrayFire {ARRAYFIRE_VERSION} binary installer successfully downloaded to {self.af_installer_local_path}")
+        else:
+            self.output.info(
+                f"ArrayFire {ARRAYFIRE_VERSION} binary installer already exists - skipping download.")
+
+    def _unpack_arrayfire(self):
+        if not os.path.exists(self.af_unpack_path):
```
9prady9, line 53: I am curious as to where is this set `af_unpack_path`

conanfile.py
```diff
@@ -0,0 +1,126 @@
+from conans import ConanFile, CMake, tools
+import os
+
+
+ARRAYFIRE_VERSION = "3.7.1"
+BINARY_INSTALLER_NAME_SUFFIX = "-1"
+BINARY_INSTALLER_NAME = f"ArrayFire-v{ARRAYFIRE_VERSION}{BINARY_INSTALLER_NAME_SUFFIX}_Linux_x86_64.sh"
+CUDA_TOOLKIT_VERSION = "10.0"
```
9prady9, line 8: :+1:

conanfile.py
```diff
@@ -0,0 +1,126 @@
...
+        if self.options.cuda_backend:
+            self.cpp_info.libs.extend([
+                f"libafcuda.so.{ARRAYFIRE_VERSION}",
+                "libnvrtc-builtins.so",
+                f"libcudnn.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcusparse.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcublas.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcusolver.so.{CUDA_TOOLKIT_VERSION}",
+                f"libnvrtc.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcufft.so.{CUDA_TOOLKIT_VERSION}",
```
9prady9, line 105: Just a heads-up, with #2785 changes into master, next fix release onwards, the following libraries wouldn't be needed to be shipped. Thus, we can remove them for 3.7.2 or 3.8.0 onwards.
- cuFFT
- cuSolver
- cuSparse
- cuBlas

conanfile.py
```diff
@@ -0,0 +1,126 @@
...
+            self.cpp_info.libs.extend([
+                "libforge.so.1.0.5",
+            ])
+        if self.options.cuda_backend:
+            self.cpp_info.libs.extend([
+                f"libafcuda.so.{ARRAYFIRE_VERSION}",
+                "libnvrtc-builtins.so",
+                f"libcudnn.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcusparse.so.{CUDA_TOOLKIT_VERSION}",
+                f"libcublas.so.{CUDA_TOOLKIT_VERSION}",
```
9prady9, line 102: `libcublasLt` is missing. That should also be available in lib folder.

conanfile.py
jacobkahn: Line 64 in `_processArrayfire` before the download or unpack.

9prady9: Ah, missed that. Thanks.

conanfile.py
jacobkahn: @9prady9 - I don't see that as having been downloaded in the binary installer. There was an issue with it before that I debugged with @umar456 - maybe it was removed?

conanfile.py
jacobkahn: Sounds good. 👍

conanfile.py
9prady9: Yeah, I didn't realize earlier. cublasLt was shipped by toolkits starting from 10.1. With 3.7.1 binary on our website. We are good without it.

conanfile.py
umar456: Lets leave it here for now and we can move it to another spot once we add additional files.

PR #2884, opened 2020-05-11
Use read/write buffer in reduce by key instead of fill buffer

### LINKED ISSUE #2879: enqueueFillBuffer not existing in OpenCL 1.1
There are those calls in opencl/kernel/reduce_by_key.hpp
```
        // reset flags
        getQueue().enqueueFillBuffer<int>(*needs_another_reduction.get(), 0, 0,
                                          sizeof(int));
        getQueue().enqueueFillBuffer<int>(*needs_block_boundary_reduction.get(),
                                          0, 0, sizeof(int));
```

_Originally posted by @glavaux2 in https://github.com/arrayfire/arrayfire/issues/2803#issuecomment-602558969_

9prady9:
Resolves #2879

### DIFF
--- a/src/backend/opencl/kernel/reduce_by_key.hpp
+++ b/src/backend/opencl/kernel/reduce_by_key.hpp
@@ -484,11 +484,15 @@ int reduce_by_key_first(Array<Tk> &keys_out, Array<To> &vals_out,
                                      sizeof(int), &n_reduced_host);
 
         // reset flags
-        getQueue().enqueueFillBuffer<int>(*needs_another_reduction.get(), 0, 0,
-                                          sizeof(int));
-        getQueue().enqueueFillBuffer<int>(*needs_block_boundary_reduction.get(),
-                                          0, 0, sizeof(int));
+        needs_block_boundary_reduction_host = 0;
+        needs_another_reduction_host        = 0;
 
+        getQueue().enqueueWriteBuffer(*needs_another_reduction.get(), CL_FALSE,
+                                      0, sizeof(int),
+                                      &needs_another_reduction_host);
+        getQueue().enqueueWriteBuffer(*needs_block_boundary_reduction.get(),
+                                      CL_FALSE, 0, sizeof(int),
+                                      &needs_block_boundary_reduction_host);
         numBlocksD0 = divup(n_reduced_host, numThreads);
 
         launch_test_needs_reduction<Tk>(*needs_another_reduction.get(),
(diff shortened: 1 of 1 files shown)


src/backend/opencl/kernel/reduce_by_key.hpp
```diff
@@ -484,23 +484,27 @@ int reduce_by_key_first(Array<Tk> &keys_out, Array<To> &vals_out,
...
+                                      &needs_block_boundary_reduction_host);
         numBlocksD0 = divup(n_reduced_host, numThreads);
 
         launch_test_needs_reduction<Tk>(*needs_another_reduction.get(),
                                         *needs_block_boundary_reduction.get(),
                                         t_reduced_keys, n_reduced_host,
                                         numBlocksD0, numThreads);
 
-        getQueue().enqueueReadBuffer(*needs_another_reduction.get(), true, 0,
+        getQueue().enqueueReadBuffer(*needs_another_reduction.get(), CL_TRUE, 0,
```
umar456, line 503:
```suggestion
        getQueue().enqueueReadBuffer(*needs_another_reduction.get(), CL_FALSE, 0,
```

src/backend/opencl/kernel/reduce_by_key.hpp
```diff
@@ -612,11 +621,11 @@ int reduce_by_key_dim(Array<Tk> &keys_out, Array<To> &vals_out,
...
```
umar456, line 624:
```suggestion
        getQueue().enqueueReadBuffer(*needs_another_reduction.get(), CL_FALSE, 0,
```

src/backend/opencl/kernel/reduce_by_key.hpp
```diff
@@ -484,23 +484,27 @@ int reduce_by_key_first(Array<Tk> &keys_out, Array<To> &vals_out,
...
         // reset flags
-        getQueue().enqueueFillBuffer<int>(*needs_another_reduction.get(), 0, 0,
-                                          sizeof(int));
-        getQueue().enqueueFillBuffer<int>(*needs_block_boundary_reduction.get(),
-                                          0, 0, sizeof(int));
-
+        needs_block_boundary_reduction_host = 0;
+        needs_another_reduction_host        = 0;
+
+        getQueue().enqueueWriteBuffer(*needs_another_reduction.get(), CL_FALSE,
```
umar456, line 490: Would be nice to combine these two into an int2 or something. That way you would only need one Write instead of two. Its outside the scope of this PR.

src/backend/opencl/kernel/reduce_by_key.hpp
9prady9: This is second call after kernel launch and has to be blocking to read the value for later conditional check.

PR #2901, opened 2020-05-23
Refactor kernel cache to store CUmodule/cl_program

9prady9:
Refactor kernel cache to store CUmodule/cl_program
    
    - Rename common::compileKernel to common::compileModule
    - Rename common::loadKernelFromDisk to common::loadModuleFromDisk
    - Rename common::findKernel to common::getKernel
    - Rename common::lookupKernel to common::findModule
    
    With this change, kernels are cached using respective backend's context handle,
    CUmodule for CUDA; cl_program for OpenCL, a.k.a module in arrayfire jargon.
    This required renaming the relevant files to appropriate names like
    compile_module.[hpp|cpp]. Each backend has to implement common::compileModule
    that handles the backend specific compilation of jit source. They also
    have to implement common::loadModuleFromDisk to handle loading already
    cached(should be taken care of by compileModule) modules from disk.
    Loading modules from disk help with quick repopulation of cache without
    the need of remcompiling the kernels.
    
    Disk caching modules is not yet implemented in OpenCL.


9prady9:
@9prady9 Rename getModuleMutex -> getCacheMutex

src/backend/common/kernel_cache.cpp
```diff
@@ -82,15 +72,51 @@ Kernel findKernel(const string& kernelName, const vector<string>& sources,
...
+    const int device       = detail::getActiveDeviceId();
+    Module currModule      = findModule(device, moduleKey);
 
-    if (kernel.getModule() == nullptr || kernel.getKernel() == nullptr) {
-        kernel = compileKernel(kernelName, tInstance, sources, compileOpts,
-                               isKernelJIT);
-        cacheKernel(device, tInstance, kernel);
+    if (currModule.get() == nullptr) {
+#if defined(AF_CUDA) && defined(AF_CACHE_KERNELS_TO_DISK)
+        currModule = loadModuleFromDisk(device, moduleKey, sourceIsJIT);
```
umar456, line 94: couldn't this logic be moved into compileModule for the CUDA backend?

src/backend/common/kernel_cache.cpp
```diff
@@ -82,15 +72,51 @@ Kernel findKernel(const string& kernelName, const vector<string>& sources,
         tInstance += ">";
     }
 
-    int device    = detail::getActiveDeviceId();
-    Kernel kernel = lookupKernel(device, tInstance, sources);
+    const bool notJIT = !sourceIsJIT;
+
+    vector<string> everything;
```
umar456, line 77:
```suggestion
    vector<string> hashingValues;
```

src/backend/common/kernel_cache.cpp
```diff
@@ -82,15 +72,51 @@ Kernel findKernel(const string& kernelName, const vector<string>& sources,
...
+            // kernel. Keep the generated module.
+            getCache(device).emplace(moduleKey, currModule);
+        } else {
+            currModule.unload();  // dump the current threads extra compilation
+            currModule = iter->second;
+        }
     }
-    return kernel;
+#if defined(AF_CUDA)
+    return getKernel(currModule, tInstance, sourceIsJIT);
```
umar456, line 116: we can just do a substring to find the kernel name in tInstance. That way we can send the same variable to both. Not a big deal.

9prady9: That is not possible unfortunately, it would rather increase the logic since kInstance won't have < always. it is only present for regular kernels and that too for CUDA only

PR #2902, opened 2020-05-23
Reduce unnecessary template instantiations

9prady9:


src/backend/cpu/kernel/medfilt.hpp
```diff
@@ -97,9 +104,9 @@ void medfilt2(Param<T> out, CParam<T> in, dim_t w_len, dim_t w_wid) {
                     for (int wj = 0; wj < (int)w_wid; ++wj) {
                         bool isColOff = false;
 
-                        int im_col = col + wj - w_wid / 2;
-                        int im_coff;
-                        switch (Pad) {
+                        int im_col  = col + wj - w_wid / 2;
+                        int im_coff = 0;
+                        switch (pad) {
```
umar456, line 109: Should keep this as a template parameter since it in the inner loop of the kernel. This will avoid branching.

9prady9:
Some of the CPU kernels may need template paramters for better optimization. Will reintroduce them in the following places carefull.y
- fftconvolve (still need to verify if it can be avoided here)
- histogram
- match_template
- medfilt
- triangle

PR #2907, opened 2020-05-25
Fix mem leak caused in opencl::removeDeviceContext function

umar456:
* The OpenCL Wrapper objects were being leaked by OpenCL when calling the
  removeDeviceContext function. The handles were decremented correctly but
  the objects were not released so it caused a very small leak in the
  binary.

### DIFF
--- a/src/backend/opencl/compile_kernel.cpp
+++ b/src/backend/opencl/compile_kernel.cpp
@@ -131,9 +131,8 @@ Kernel compileKernel(const string &kernelName, const string &tInstance,
     auto compileBegin = high_resolution_clock::now();
     auto prog         = detail::buildProgram(sources, compileOpts);
     auto prg          = new cl::Program(prog);
-    auto krn =
-        new cl::Kernel(*static_cast<cl::Program *>(prg), kernelName.c_str());
-    auto compileEnd = high_resolution_clock::now();
+    auto krn          = new cl::Kernel(*prg, kernelName.c_str());
+    auto compileEnd   = high_resolution_clock::now();
 
     AF_TRACE("{{{:<30} : {{ compile:{:>5} ms, {{ {} }}, {} }}}}", kernelName,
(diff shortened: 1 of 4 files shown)


src/backend/opencl/compile_kernel.cpp
```diff
@@ -131,9 +131,8 @@ Kernel compileKernel(const string &kernelName, const string &tInstance,
     auto compileBegin = high_resolution_clock::now();
     auto prog         = detail::buildProgram(sources, compileOpts);
     auto prg          = new cl::Program(prog);
-    auto krn =
-        new cl::Kernel(*static_cast<cl::Program *>(prg), kernelName.c_str());
-    auto compileEnd = high_resolution_clock::now();
+    auto krn          = new cl::Kernel(*prg, kernelName.c_str());
```
9prady9, line 134: Curious, is this a nit pick ?

src/backend/opencl/platform.cpp
```diff
@@ -396,43 +395,41 @@ void devprop(char* d_name, char* d_platform, char* d_toolkit, char* d_compute) {
 
     DeviceManager& devMngr = DeviceManager::getInstance();
 
-    vector<cl::Context*> contexts;
     {
         common::lock_guard_t lock(devMngr.deviceMutex);
```
9prady9, line 399: This section is only reading data from device manager, exclusive access doesn't make sense. Since we have C++14 now, we should move to shared Mutex so that multiple threads can read this info if they want to without the wait.

Alternatively, we can use boost's shared_mutex instead of std::shared_timed_mutex but we have to link against boost's thread library for that. I wonder how much of a benefit that would be - but again I would expect it to avoid timing logic if we just use `lock` and `shared_lock` methods.

9prady9: This change has been deferred to another PR after discussion.

9prady9:
Don't merge yet. testing a windows specific change.

PR #2911, opened 2020-05-28
Add allocV2 and freeV2 which return cl_mem on OpenCL backends

umar456:
The af_alloc_device functions were returning C++ objects instead of C objects. This behavior is now decremented in favor of cl_mem objects in the OpenCL backends. The behavior of the CUDA and CPU backends have not changed.

Updated documentation to clarify proper usage and behavior

Deprecated af_alloc_device af_free_device

Added several memory tests.

### DIFF
--- a/src/backend/opencl/memory.cpp
+++ b/src/backend/opencl/memory.cpp
@@ -11,6 +11,7 @@
 #include <common/MemoryManagerBase.hpp>
 #include <common/half.hpp>
 #include <err_opencl.hpp>
+#include <errorcodes.hpp>
 #include <memory.hpp>
 #include <platform.hpp>
 #include <spdlog/spdlog.h>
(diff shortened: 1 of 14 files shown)


test/CMakeLists.txt
```diff
@@ -283,9 +283,43 @@ make_test(SRC nodevice.cpp CXX11)
...
+
+      set_target_properties(${target}
+        PROPERTIES
+        FOLDER "Tests"
+        OUTPUT_NAME "cuda_${backend}")
+
+      add_test(NAME ${target} COMMAND ${target})
+    endif()
+  endforeach()
+endif()
```
9prady9, line 320: What am I missing here ? It feels almost identical to `make_test` macro.

test/cuda.cu
```diff
@@ -0,0 +1,79 @@
...
+#include <testHelpers.hpp>
+#include <af/array.h>
+#include <af/device.h>
+
+using af::allocV2;
+using af::freeV2;
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+TEST(Memory, AfAllocDeviceCUDA) {
```
9prady9, line 20: Something like `AllocDevicePointerUsability` instead of `AfAllocDeviceCUDA` is more readable name.

test/cuda.cu
```diff
@@ -0,0 +1,79 @@
...
+    float host;
+    ASSERT_EQ(cudaSuccess,
+              cudaMemcpy(&host, ptr, sizeof(float), cudaMemcpyDeviceToHost));
+    ASSERT_SUCCESS(af_free_device(ptr));
+
+    ASSERT_EQ(5, host);
+}
+#pragma GCC diagnostic pop
+
+TEST(Memory, AfAllocDeviceV2CUDA) {
```
9prady9, line 43: Similarly `AllocDeviceV2PointerUsability` is more readable.

test/manual_memory_test.cpp
```diff
@@ -26,7 +26,7 @@ TEST(Memory, recover) {
             vec[i] = randu(1024, 1024, 256);  // Allocating 1GB
         }
 
-        ASSERT_EQ(true, false);  // Is there a simple assert statement?
+        FAIL();
```
9prady9, line 29: :+1:

test/memory.cpp
```diff
@@ -796,7 +802,166 @@ void remove_memory_management_fn(af_memory_manager manager, int id) {}
...
+    ASSERT_EQ(payload->table.size(), 1);
+    ASSERT_EQ(payload->table[a.device<float>()],
+              aSize * aSize * aSize * aSize * sizeof(double) * 2);
+    ASSERT_EQ(payload->lastElementSize, 16);
+
+    // Currently this is set to 1 because all allocations request linear memory
+    // This behavior will change in the future
+    ASSERT_EQ(payload->lastNdims, 1);
+    ASSERT_EQ(payload->lastDims, af::dim4(aSize * aSize * aSize * aSize));
+}
```
9prady9, line 932: Tests: `E2ETest*D` and  `E2ETest4DComplexDouble` can use a better name like `MostRecentArrayCreated_*D` or ``MostRecentArray_*D`

test/memory.cpp
```diff
@@ -796,7 +802,166 @@ void remove_memory_management_fn(af_memory_manager manager, int id) {}
...
+
+TEST_F(MemoryManagerApi, E2ETestMultipleAllocations) {
```
9prady9, line 934: nit: Don't feel like the prefix `E2ETest` is adding anything to the readability factor..

test/memory.cpp
```diff
@@ -893,3 +1058,55 @@ TEST(MemoryManagerApi, E2ETest) {
     ASSERT_EQ(payload->initializeCalledTimes, 1);
     ASSERT_EQ(payload->shutdownCalledTimes, af::getDeviceCount());
 }
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+TEST(Memory, AfAllocDeviceCPUC) {
```
9prady9, line 1064: `AfAllocDeviceCPUC` what does `C` in the end mean ? for this and the following tests

test/ocl_ext_context.cpp
```diff
@@ -136,3 +139,123 @@ TEST(OCLCheck, DevicePlatform) {
 #else
 TEST(OCLExtContext, NoopCPU) {}
 #endif
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
9prady9, line 144: We have to figure out a generic way (a custom drop in marco perhaps) to do this for gcc, clang and msvc if possible. Instead of adding compiler specific pragmas in multiple locations.

Not as a part of this PR, of course.

src/backend/opencl/memory.cpp
```diff
@@ -56,42 +57,69 @@ template<typename T>
...
+    if (buf) {
+        cl_mem mem = (*buf)();
+        delete buf;
+        memoryManager().unlock(static_cast<void *>(mem), false);
+    }
 }
 
-void memLock(const void *ptr) {
-    memoryManager().userLock(const_cast<void *>(ptr));
+void memLock(const cl::Buffer *ptr) {
```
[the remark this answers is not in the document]
umar456: I think cl::Buffers are fine internally as long as the native functions are using cl_mem. The mem* functions are used internally and they should be returning cl::Buffers because that is most convenient for our code base. The reason I moved to the nativeAlloc interface is because the memory manager should be compatible with the base OpenCL specification. The C++ interface is not stable enough for that.

test/CMakeLists.txt
umar456: `cuda_add_executable(${target} cuda.cu)`

test/cuda.cu
umar456: its testing af_alloc_device. I think the name is appropriate here.

test/memory.cpp
umar456: Most recent array created is not an appropriate name because it is only one of the many things it tests. in this case we are testing the behavior of the custom memory manager in the case we create a 4D array.

test/memory.cpp
umar456: Uses the C api instead of the C++ api

test/CMakeLists.txt
9prady9: Too much redundancy for single line difference :(

PR #2918, opened 2020-06-01
Update OpenCL interop page so they discuss deleting of memory

### LINKED ISSUE #2917: Memory leak in OpenCL backend
I think I've discovered a memory leak in the OpenCL backend which occurs during a call to the `array::device()` method.

Specifically, in `arrayfire/src/backend/opencl/api.cpp` at line 7

```cpp
namespace af {
template<>
AFAPI cl_mem *array::device() const {
    cl_mem *mem_ptr = new cl_mem;
    af_err err      = af_get_device_ptr((void **)mem_ptr, get());
    if (err != AF_SUCCESS)
        throw af::exception("Failed to get cl_mem from array object");
    return mem_ptr;
}
}  // namespace af
```

The allocation of a `new cl_mem` instance on the heap is not released, and the documentation does not indicate to the user that they need to call `delete` on the resulting pointer to release the resource. Subsequent calls to `array::unlock()` obviously don't release the resource either.

Apologies if I've misunderstood this (perhaps it is a reference counted resource?), but I would appreciate some clarification.

Many thanks.

umar456:
* Created snippets for examples in the document

Fixes #2917

### DIFF
--- a/test/interop_opencl_external_context_snippet.cpp
+++ b/test/interop_opencl_external_context_snippet.cpp
@@ -1,0 +1,104 @@
+/*******************************************************
+ * Copyright (c) 2020, ArrayFire
+ * All rights reserved.
+ *
+ * This file is distributed under 3-clause BSD license.
+ * The complete license agreement can be obtained at:
+ * http://arrayfire.com/licenses/BSD-3-Clause
+ ********************************************************/
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wignored-qualifiers"
+#pragma GCC diagnostic ignored "-Wignored-attributes"
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#if __GNUC__ >= 8
+#pragma GCC diagnostic ignored "-Wcatch-value="
+#endif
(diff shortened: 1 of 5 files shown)


test/interop_opencl_external_context_snippet.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    // 7. Obtain cl_mem references for important memory
+    cl_mem *af_mem = af_A.device<cl_mem>();
+    cl_A           = cl::Buffer(*af_mem, /*retain*/ true);
+
+    /// Delete the af_mem pointer. The buffer returned by the device pointer is
+    /// still valid
+    delete af_mem;
+
+    // 8. Continue your OpenCL application
```
9prady9, line 86: I don't quite understand retain in line 66 ?

test/interop_opencl_external_context_snippet.cpp
```diff
@@ -0,0 +1,104 @@
...
+    //    First, create a device from the current OpenCL device + context +
+    //    queue
+    afcl::addDevice(device(), context(), queue());
+    //    Next switch ArrayFire to the device using the device and context as
+    //    identifiers:
+    afcl::setDevice(device(), context());
+
+    // 4. Create ArrayFire arrays from OpenCL memory objects
+    af::array af_A = afcl::array(length, cl_A(), f32, true);
+    clRetainMemObject(cl_A());
```
9prady9, line 77: Feel like this unnecessary extra retain.

PR #2923, opened 2020-06-05
Fix errors and warnings with CUDA 9.0 builds

umar456:
* The rdc and dlink flags are not required because they are added
  by CMake for separable compilation and static linking respectively
* Add guards around libs that are not included in the CUDA 9.0
  Toolkit
* Only link with OpenMP when linking with cuSOLVER dynamically
* Fix error message when CUDNN is not found
* Address casts from double to __half which are missing in 9.0
* Thrust return_temporary_buffer function can accept void* pointers
  in older versions of Thrust. Use raw_pointer_cast to pass the
  pointer to memFree
* cublasGemmEx doesn't exist in CUDA 9.0. Add ifdefs to guard
  against older builds
* __float2half is not a host function so it needs to be removed
  from mean
* Add template instantiation for memFree to accept void* pointers
* CUSOLVER_CHECK error message printed "CUBLAS Error" instead of
CUSOLVER Error

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -7,7 +7,7 @@
 
 dependency_check(CUDA_FOUND "CUDA not found.")
 if(AF_WITH_CUDNN)
-  dependency_check(cuDNN_FOUND "CUDA not found.")
+  dependency_check(cuDNN_FOUND "CUDNN not found.")
 endif()
 
 include(AFcuda_helpers)
@@ -34,7 +34,7 @@ endif()
 
 # Find if CUDA Toolkit is at least 10.0 to use static
 # lapack library. Otherwise, we have to use regular shared library
-if(UNIX AND CUDA_VERSION_MAJOR VERSION_GREATER 10 OR CUDA_VERSION_MAJOR VERSION_EQUAL 10)
+if(UNIX AND (CUDA_VERSION_MAJOR VERSION_GREATER 10 OR CUDA_VERSION_MAJOR VERSION_EQUAL 10))
   set(use_static_cuda_lapack ON)
 else()
   set(use_static_cuda_lapack OFF)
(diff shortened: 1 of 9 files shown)


src/backend/cuda/CMakeLists.txt
```diff
@@ -301,13 +294,19 @@ if(UNIX)
...
       ${CUDA_cusparse_static_LIBRARY}
       ${cusolver_static_lib}
       -Wl,--end-group
   )
+
+  if(CUDA_VERSION VERSION_GREATER 9.5)
+    target_link_libraries(af_cuda_static_cuda_library
+      PRIVATE
+        ${CUDA_cublasLt_static_LIBRARY}
+        ${CUDA_lapack_static_LIBRARY})
```
9prady9, line 307: Makes sense, these are not there prior to toolkit 10 -  I should have thought about this

src/backend/cuda/CMakeLists.txt
```diff
@@ -301,13 +294,19 @@ if(UNIX)
...
       ${CUDA_cublas_static_LIBRARY}
-      ${CUDA_cublasLt_static_LIBRARY}
       ${CUDA_cufft_static_LIBRARY}
-      ${CUDA_lapack_static_LIBRARY}
       ${CUDA_cusparse_static_LIBRARY}
       ${cusolver_static_lib}
       -Wl,--end-group
   )
+
+  if(CUDA_VERSION VERSION_GREATER 9.5)
```
9prady9, line 303: nit: I think a better check to avoid any unexpected issues (with decimals and all against different cmake versions) with cmake is to do
```cmake
if(CUDA_VERSION_MAJOR VERSION_GREATER 9)
```

On another note, I will take care of this with some other change.

src/backend/cuda/jit/kernel_generators.hpp
```diff
@@ -71,8 +71,9 @@ void generateBufferRead(std::stringstream& kerStream, int id,
               << "];\n";
 }
 
-void generateShiftNodeOffsets(std::stringstream& kerStream, int id,
-                              bool is_linear, const std::string& type_str) {
+inline void generateShiftNodeOffsets(std::stringstream& kerStream, int id,
```
9prady9, line 74: nit: looks like this is just to make the compiler silent even though this won't be actually inlined.

PR #2931, opened 2020-06-12
Fix ccache launch scripts to use sh compatible syntax

9prady9:
Earlier to this change, I added bash based syntax which won't work
with /bin/sh or dash shells.

/usr/sh is available on most systems that use init.d scripts. So, it is
safe to assume it's availability on majority of linux distributions.


9prady9:
The change works fine on Arch(sh -> bash) and Ubuntu(sh -> dash) too.  This should be good for merge. @umar456

Issue #2944, opened 2020-06-25
[Build] OPENCL and CPU errors for Mac OS 


henry090:
Description
===========
This is related to https://github.com/arrayfire/arrayfire/issues/2938

Error Log OpenCL
---------

```
/Users/turgutabdullayev/Documents/arrayfire/src/backend/opencl/kernel/sparse_arith.hpp:36:26: error: 
      constexpr function's return type 'std::string' (aka 'basic_string<char,
      char_traits<char>, allocator<char> >') is not a literal type
AF_CONSTEXPR std::string getOpString() {
                         ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:4317:64: note: 
      'basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>
      >' is not literal because it is not an aggregate and has no constexpr
      constructors other than copy or move constructors
_LIBCPP_EXTERN_TEMPLATE(class _LIBCPP_EXTERN_TEMPLATE_TYPE_VIS basic_str...
                                                               ^
In file included from /Users/turgutabdullayev/Documents/arrayfire/src/backend/opencl/sparse_arith.cpp:10:
/Users/turgutabdullayev/Documents/arrayfire/src/backend/opencl/kernel/sparse_arith.hpp:62:28: error: 
      no matching function for call to 'getOpString'
        DefineKeyValue(OP, getOpString<op>()),
                           ^~~~~~~~~~~~~~~
/Users/turgutabdullayev/Documents/arrayfire/src/backend/common/TemplateArg.hpp:29:61: note: 
      expanded from macro 'DefineKeyValue'
#define DefineKeyValue(key, arg) " -D " #key "=" + toString(arg)
                                                            ^~~
/Users/turgutabdullayev/Documents/arrayfire/src/backend/opencl/kernel/sparse_arith.hpp:76:27: note: 
      in instantiation of function template specialization
      'opencl::kernel::fetchKernel<float, af_add_t>' requested here
    auto sparseArithCSR = fetchKernel<T, op>("sparseArithCSR", src);
                          ^
/Users/turgutabdullayev/Documents/arrayfire/src/backend/opencl/sparse_arith.cpp:68:21: note: 
      in instantiation of function template specialization
      'opencl::kernel::sparseArithOpCSR<float, af_add_t>' requested here
            kernel::sparseArithOpCSR<T, op>(out, lhs.getValues(),
[...]

PR #2945, opened 2020-06-25
Fixed several errors on OSX with Apple Clang

umar456:
Fixed several errors on OSX with Apple Clang

Description
-----------
This bugfix addresses errors compiling on Apple Clang. This PR

* Adds Boost::boost target to the unified backend for stacktrace
* Change getOpString to return a const char* to maintain constexpr specification
* Make Array<T> destructor default in the header instead of in the cpp file
* Make all Node classes movable and add static asserts

In the future we should consider adding move constructors to dim4.

Fixes: #2944 

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] ~Functions added to unified API~
- [ ] ~Functions documented~

### DIFF
--- a/src/api/unified/CMakeLists.txt
+++ b/src/api/unified/CMakeLists.txt
@@ -76,6 +76,7 @@ target_link_libraries(af
     cpp_api_interface
     spdlog
     Threads::Threads
+    Boost::boost
     ${CMAKE_DL_LIBS}
   )
 
--- a/src/backend/common/ArrayInfo.hpp
+++ b/src/backend/common/ArrayInfo.hpp
@@ -56,6 +56,10 @@ class ArrayInfo {
         , dim_strides(stride)
         , is_sparse(false) {
         setId(id);
+        static_assert(std::is_move_assignable<ArrayInfo>::value,
+                      "ArrayInfo is not move assignable");
+        static_assert(std::is_move_constructible<ArrayInfo>::value,
+                      "ArrayInfo is not move constructible");
         static_assert(
             offsetof(ArrayInfo, devId) == 0,
             "ArrayInfo::devId must be the first member variable of ArrayInfo. \
@@ -79,10 +83,24 @@ class ArrayInfo {
                    This is then used in the unified backend to check mismatched arrays.");
     }
 
-    // Copy constructors are deprecated if there is a
-    // user-defined destructor in c++11
     ArrayInfo()                       = default;
     ArrayInfo(const ArrayInfo& other) = default;
+    ArrayInfo(ArrayInfo&& other)      = default;
+
+    ArrayInfo& operator=(ArrayInfo other) noexcept {
+        swap(other);
+        return *this;
+    }
+
+    void swap(ArrayInfo& other) noexcept {
+        using std::swap;
+        swap(devId, other.devId);
+        swap(type, other.type);
+        swap(dim_size, other.dim_size);
+        swap(offset, other.offset);
+        swap(dim_strides, other.dim_strides);
+        swap(is_sparse, other.is_sparse);
+    }
 
     const af_dtype& getType() const { return type; }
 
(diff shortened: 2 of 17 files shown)


src/api/unified/CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ target_link_libraries(af
     cpp_api_interface
     spdlog
     Threads::Threads
+    Boost::boost
```
9prady9, line 79: We don't use boost headers anywhere in unified API. If this it absolutely necessary for OSX build for some reason. It's better to have a comment around here so that we don't again remove it later.

src/backend/common/ArrayInfo.hpp
```diff
@@ -79,10 +83,24 @@ class ArrayInfo {
                    This is then used in the unified backend to check mismatched arrays.");
     }
 
-    // Copy constructors are deprecated if there is a
-    // user-defined destructor in c++11
     ArrayInfo()                       = default;
     ArrayInfo(const ArrayInfo& other) = default;
+    ArrayInfo(ArrayInfo&& other)      = default;
+
+    ArrayInfo& operator=(ArrayInfo other) noexcept {
```
9prady9, line 90: Shouldn't this take `ArrayInfo &&other` if is meant to be move assignment operator ?

src/backend/common/jit/BufferNodeBase.hpp
```diff
@@ -28,7 +28,9 @@ class BufferNodeBase : public common::Node {
     bool m_linear_buffer;
 
    public:
-    BufferNodeBase(af::dtype type) : Node(type, 0, {}) {}
+    BufferNodeBase(af::dtype type) : Node(type, 0, {}) {
+        // This class is not movable because of std::once_flag
```
9prady9, line 32: Is this comment a placeholder for `static_asserts` we have in other places because `BufferNodeBase is not movable` ?

src/backend/common/jit/NaryNode.hpp
```diff
@@ -39,7 +39,30 @@ class NaryNode : public Node {
...
+    /// Default move assignment operator
+    NaryNode &operator=(NaryNode &&node) noexcept = default;
+
+    void swap(NaryNode &other) noexcept {
+        using std::swap;
+        Node::swap(other);
+        swap(m_num_children, other.m_num_children);
+        swap(m_op, other.m_op);
+        swap(m_op_str, other.m_op_str);
+    }
```
9prady9, line 65: What was the problem with compiler created defaults ? Do we have to explicitly add the ` = default ` lines on OSX ?

9prady9: This comment pertains to mostly all other changes in `common/jit` too.

src/api/unified/CMakeLists.txt
umar456: CHECK_ARRAYS uses AF_RETURN_ERROR which is defined in common/err_common.hpp which uses boost/stacktrace.hpp

src/backend/common/ArrayInfo.hpp
umar456: This is an implementation of the "Rule of 4 and a half". It handles both copy and move assignment using this approach.

src/backend/common/jit/BufferNodeBase.hpp
umar456: Yes

src/api/unified/CMakeLists.txt
9prady9: Clang caught it appropriately, hm.

src/backend/common/jit/NaryNode.hpp
umar456: They were some changes I wanted to add but didn't want to create a separate PR for because they are not significant. That's why they are in a separate commit.

9prady9: Hm, Apparently sometimes the compiler will add noexcept versions of the move constructor which is bad. So these nothrow static asserts are added but those fail if there are no defaulted constructors and move operators

src/backend/common/ArrayInfo.hpp
9prady9: I don't much about these rules by names. I will read up.

PR #2946, opened 2020-06-26
Add move constructor and operator= for dim4, Array<T> SparseArray<T>

umar456:
Add nothrow move constructors for dim4. Array<T>, SparseArray<T>

Description
-----------
Create move constructors and move assignment operators for Array<T>, SparseArray<T>,
dim4. This adds move constructors for dim4 so this cannot be back ported.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/include/af/dim4.hpp
+++ b/include/af/dim4.hpp
@@ -40,14 +40,29 @@ public:
     /// \param[in] other The dim4 that will be copied
     dim4(const dim4& other);
 
+#if AF_API_VERSION >= 38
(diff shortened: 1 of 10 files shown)


umar456:
That static check and constructor was added in the osx_fixes PR. I couldn't add it to these classes because of dim4's move assignment operator and constructor.

9prady9:
Got it. I didn't pull latest master after that PR merge, hence when I checked the file locally didn't find them in CPU backend.

PR #2952, opened 2020-06-29
FindMKL.cmake: allow double include

junghans:
Only create targets if they are not existing already, this happens
when find_package(MKL) is called twice for some reason.

Also supipress a warning from cmake-3.17 about mismatching package
names.

Description
-----------

Changes to Users
----------------

None

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [N/A] Functions added to unified API
- [ N/A] Functions documented

### DIFF
--- a/CMakeModules/FindMKL.cmake
+++ b/CMakeModules/FindMKL.cmake
@@ -204,6 +204,10 @@ function(find_mkl_library)
 
   cmake_parse_arguments(mkl_args "${options}" "${single_args}" "${multi_args}" ${ARGN})
 
+  if(TARGET MKL::${mkl_args_NAME})
+    return()
+  endif()
+
   add_library(MKL::${mkl_args_NAME}        SHARED IMPORTED)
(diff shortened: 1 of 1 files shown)


CMakeModules/FindMKL.cmake
```diff
@@ -351,6 +357,11 @@ set(MKL_RUNTIME_KERNEL_LIBRARIES "${MKL_RUNTIME_KERNEL_LIBRARIES_TMP}" CACHE STR
     "MKL kernel libraries targeting different CPU architectures")
 mark_as_advanced(MKL_RUNTIME_KERNEL_LIBRARIES)
 
+# Bypass developer warning that the first argument to find_package_handle_standard_args (MKL_...) does not match
+# the name of the calling package (MKL)
+# https://cmake.org/cmake/help/v3.17/module/FindPackageHandleStandardArgs.html
+set(FPHSA_NAME_MISMATCHED TRUE)
```
9prady9, line 363: :+1:

PR #2954, opened 2020-06-30
Add support for versioned loading of libs at runtime; Fix cudnn cmake install command

9prady9:
Description
-----------
- Added support for loading versioned libraries by DependencyModule class. If versioned library is not found, the search will fallback to non versioned name lookup.
- Changed cudnn module to lookup latest minor releases for each major version available.
- Earlier to this change, cuDNN install command failed on windows. On Linux, it was packaged with incorrect version suffix.
After this change, cuDNN library will be installed as `<lib prefix>cudnn<lib suffix><soname>` on unix platform.

Changes to Users
----------------
None

Checklist
---------
- [x] Code compiles
- ~[ ] Tests pass~
- ~[ ] Functions documented~

### DIFF
--- a/src/backend/common/DependencyModule.cpp
+++ b/src/backend/common/DependencyModule.cpp
@@ -10,6 +10,7 @@
 #include <common/DependencyModule.hpp>
 #include <common/Logger.hpp>
 #include <common/module_loading.hpp>
+
 #include <algorithm>
 #include <string>
 
@@ -19,29 +20,72 @@
 #include <dlfcn.h>
 #endif
 
+using common::Version;
+using std::make_tuple;
+using std::string;
+using std::to_string;
+using std::vector;
+
+constexpr Version NullVersion{-1, -1, -1};
+
 #ifdef OS_WIN
 #include <Windows.h>
+
 static const char* librarySuffix = ".dll";
-static const char* libraryPrefix = "";
+
+namespace {
+vector<string> libNames(const std::string& name, const string& suffix,
+                        const Version& ver = NullVersion) {
+    UNUSED(ver);  // Windows DLL files are not version suffixed
+    return {name + suffix + librarySuffix};
+}
+}  // namespace
+
 #elif defined(OS_MAC)
+
 static const char* librarySuffix = ".dylib";
 static const char* libraryPrefix = "lib";
+
+namespace {
+vector<string> libNames(const std::string& name, const string& suffix,
+                        const Version& ver = NullVersion) {
+    UNUSED(suffix);
+    const string noVerName = libraryPrefix + name + librarySuffix;
+    if (ver != NullVersion) {
+        const string infix = "." + to_string(std::get<0>(ver)) + ".";
+        return {libraryPrefix + name + infix + librarySuffix, noVerName};
+    } else {
+        return {noVerName};
+    }
+}
+}  // namespace
+
 #elif defined(OS_LNX)
+
 static const char* librarySuffix = ".so";
 static const char* libraryPrefix = "lib";
+
+namespace {
+vector<string> libNames(const std::string& name, const string& suffix,
+                        const Version& ver = NullVersion) {
+    UNUSED(suffix);
+    const string noVerName = libraryPrefix + name + librarySuffix;
+    if (ver != NullVersion) {
+        const string soname("." + to_string(std::get<0>(ver)));
+
+        const string vsfx = "." + to_string(std::get<0>(ver)) + "." +
+                            to_string(std::get<1>(ver)) + "." +
+                            to_string(std::get<2>(ver));
+        return {noVerName + vsfx, noVerName + soname, noVerName};
+    } else {
+        return {noVerName};
+    }
+}
+}  // namespace
+
 #else
 #error "Unsupported platform"
 #endif
-
-using std::string;
-using std::vector;
-
-namespace {
-
-std::string libName(const std::string& name) {
-    return libraryPrefix + name + librarySuffix;
-}
-}  // namespace
 
 namespace common {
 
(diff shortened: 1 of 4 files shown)


9prady9:
Leaving sample output of versioned loading for reference.
```Built target helloworld_cuda
[platform][1593503872][042598] [ ../src/backend/common/DependencyModule.cpp:62 ] Attempting to load: libforge.so
[platform][1593503872][042598] [ ../src/backend/common/DependencyModule.cpp:65 ] Found: libforge.so
[platform][1593503872][042598] [ ../src/backend/cuda/device_manager.cpp:427 ] CUDA Driver supports up to CUDA 10.2 ArrayFire CUDA Runtime 10.2
[platform][1593503872][042598] [ ../src/backend/cuda/device_manager.cpp:495 ] Found 1 CUDA devices
[platform][1593503872][042598] [ ../src/backend/cuda/device_manager.cpp:517 ] Found device: GeForce GTX 1060 3GB (2.94 GB | ~3754.03 GFLOPs | 9 SMs)
[platform][1593503872][042598] [ ../src/backend/cuda/device_manager.cpp:556 ] AF_CUDA_DEFAULT_DEVICE: 
[platform][1593503872][042598] [ ../src/backend/cuda/device_manager.cpp:574 ] Default device: 0(GeForce GTX 1060 3GB)
ArrayFire v3.8.0 (CUDA, 64-bit Linux, build 703a1fdc8)
Platform: CUDA Runtime 10.2, Driver: 440.100
[0] GeForce GTX 1060 3GB, 3012 MB, CUDA Compute 6.1
[mem][1593503873][042598] [ ../src/backend/cuda/memory.cpp:158 ] nativeAlloc:    1 KB 0x7f69c4600000
[platform][1593503873][042598] [ ../src/backend/common/DependencyModule.cpp:95 ] Attempting to load: libcudnn.so.7.6.5
[platform][1593503873][042598] [ ../src/backend/common/DependencyModule.cpp:98 ] Found: libcudnn.so.7.6.5
[platform][1593503873][042598] [ ../src/backend/cuda/cudnnModule.cpp:85 ] cuDNN Version: 7.6.5 cuDNN CUDA Runtime: 10.0
[platform][1593503873][042598] WARNING: ArrayFire CUDA Runtime(10.2) and cuDNN CUDA Runtime(10.0) do not match. For maximum compatibility, make sure the two versions match.(Ignoring check)
[platform][1593503873][042598] [ ../src/backend/cuda/platform.cpp:118 ] Initializing cuDNN
```

src/backend/common/DependencyModule.cpp
```diff
@@ -19,30 +20,74 @@
 #include <dlfcn.h>
 #endif
 
+using common::Version;
+using std::make_tuple;
+using std::string;
+using std::to_string;
+using std::vector;
+
+constexpr Version NullVersion{0, 0, 0};
```
umar456, line 29: -1 would be a better Null value

src/backend/common/DependencyModule.cpp
```diff
@@ -19,30 +20,74 @@
...
+    UNUSED(suffix);
+    if (ver != NullVersion) {
+        const string soname("." + to_string(std::get<0>(ver)));
 
-std::string libName(const std::string& name) {
-    return libraryPrefix + name + librarySuffix;
+        const string vsfx = "." + to_string(std::get<0>(ver)) + "." +
+                            to_string(std::get<1>(ver)) + "." +
+                            to_string(std::get<2>(ver));
+        return {libraryPrefix + name + librarySuffix + vsfx,
```
umar456, line 78: libraryPrefix + name+ librarySuffix  is used multiple times. You should store that in a temporary and add the values that are different for the values. This is will most likely require multiple allocations.

src/backend/cuda/cudnnModule.cpp
```diff
@@ -22,6 +22,22 @@ using std::string;
...
+    make_tuple(7, 4, 2),
+    make_tuple(7, 3, 1),
+    make_tuple(7, 2, 1),
+    make_tuple(7, 1, 4),
+    make_tuple(7, 0, 5),
+    make_tuple(6, 0, 21),
+    make_tuple(5, 1, 10),
+    make_tuple(4, 0, 7),
+    // clang-format off
+};
```
pavanky, line 39: would `constexpr std::array` work here ?

src/backend/common/DependencyModule.cpp
```diff
@@ -51,11 +96,11 @@ DependencyModule::DependencyModule(const char* plugin_file_name,
     // TODO(umar): Implement handling of non-standard paths
     UNUSED(paths);
     if (plugin_file_name) {
-        string filename = libName(plugin_file_name);
-        AF_TRACE("Attempting to load: {}", filename);
-        handle = loadLibrary(filename.c_str());
+        auto fileNames = libName(plugin_file_name, "");
```
pavanky, line 99: consider changing the name of the `libName` function because it is returning multiple file names.

src/backend/cuda/cudnnModule.cpp
9prady9: I don't think so because DependencyModule needs to work for any library that we want to load and the number of versions we can enlist for any such library is not fixed. So, the DependencyModule constructor should take a vector. Given this, if we store the version list of the libraries in a std::array then, we need to convert that into a std::vector which will essentially result in a copy. This much worse than passing a reference to `static const` vector.

There is however another way if we had something like with `array_view` we which don't until C++20's `std::span`. Having something like this, we can accept this `array_view` via DependencyModule and it doesn't matter what container is used for version enlisting.

9prady9: We can use boost if there is something like `array_view` in there.

src/backend/common/DependencyModule.cpp
9prady9: :+1: took care

src/backend/common/DependencyModule.cpp
9prady9: Sure

src/backend/common/DependencyModule.cpp
9prady9: :+1: took care

src/backend/cuda/cudnnModule.cpp
9prady9: It has boost::span

Issue #2955, opened 2020-06-30
[BUG] [CUDA] freeze with reduce by key


ebarjou:
Under circumstances that I do not fully understand, the CUDA backend can freeze when executing a reduce by key operation. I isolated the issue with the values contained in the key array, and it seems linked to the number of consecutive equal keys.
Under OpenCL, the function does not exhibit this behavior.

Description
===========

I use the release AF binaries v3.7.1, with the Cuda backend, tested on two computers with Windows and Linux.
The issue happened every time at exactly the same spot in my program. 
At first I saved the key and value arrays to files in order to test in a separate program, and it does act the same.
I then tried to replicate the issue by constructing array, and got the same issue when having two long sequences of equal keys.
It give no error log, no exception, the only information I have is that it freeze in the sumByKey operation (it behave the same with all ___ByKey operations).

Reproducible Code
------------------------------
Here's the test program that i came up with. Under OpenCL, it run flawlessly, but under Cuda it freeze each time at i=73.
```c
#include <arrayfire.h>
#include <iostream>

int main(int argc, char *argv[]) {
    int N = 1280*1280;
    int count = 200;
    try{
        af::sync();
        af::array val = af::randu(N);
        af::array key = af::range(af::dim4(N), 0, af::dtype::s32);
        af::array res1, res2;
        for(int i = 0; i < count; ++i) {
            std::cout << i << " consecutive key : ";
            key(i) = 0;
            key(count+i) = 1;
            af::sumByKey(res1, res2, key, val);
            res2.eval();
            res1.eval();
            af::sync();
            std::cout << "Ok !" << std::endl;
        }

        std::cout << "Finished" << std::endl;
    } catch  (af::exception& e) { 
        std::cout << e.what() << std::endl;
        return -1;
    }
    return 0;
}
```
System Information
------------------
ArrayFire version 3.7.1
Intel Core i7-9750H, 16Go RAM, GTX1650 4Go
Cuda info : https://pastebin.com/kLGvdUA0
Output of nvidia-smi : 
```console
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 446.14       Driver Version: 446.14       CUDA Version: 11.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name            TCC/WDDM | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
[...]

PR #2957, opened 2020-07-01
Support building ArrayFire with the device symbols flag

umar456:
ArrayFire was failing to build with NVCC's -G(device symbols) flag. 

Description
-----------
Support compiling of ArrayFire with the -G flag in NVCC. This flag embeds device symbols in the binary
so that you could debug CUDA kernels using cuda-gdb. Compiling with device debug symbols was failing
because we were calling host functions from functions that supported both device and host code.

This PR also addresses an issue where we were using the default stream when calling CUB functions
in reduce by key.

Tests for reduce by keys don't pass in the current master with debug symbols because of the newer
version of CUB used. The older version of CUB passes the tests but it is not compatible with CUDA 11.
This is still being investigated and I have **NOT** included the fix as part of this PR.

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [ ] Tests pass
- ~[ ] Functions documented~


umar456:
This PR also fixes #2955

9prady9:
fyi: opencl reduce failed.

PR #2965, opened 2020-07-06
Work around in floodfill for what seems to be bug in NVIDIA opencl

9prady9:
Description
-----------
* The kernel's while loop seems to go into an infinite loop for no apparent reason on NVIDIA
  platform which doesn't happen on AMD, BIEGNET and POCL. CCC Tests pass on AMD, BIEGNET but
  doesn't on POCL, fails with value difference errors.

* Also, cleaned up flood fill kernels in both CUDA and OpenCL backends.

Changes to Users
----------------
Confidence Connected Components is now enabled in OpenCL backend.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/src/api/c/confidence_connected.cpp
+++ b/src/api/c/confidence_connected.cpp
@@ -188,13 +188,6 @@ af_err af_confidence_cc(af_array* out, const af_array in, const af_array seedx,
                         const af_array seedy, const unsigned radius,
                         const unsigned multiplier, const int iter,
                         const double segmented_value) {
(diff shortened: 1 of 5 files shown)


9prady9:
I am yet to add note in the kernel how the loop threshold is decided. will add that with feedback changes.

9prady9:
Closing temporarily, want to avoid ci builds for WIP commits that I need to test on different setup.

PR #2970, opened 2020-07-09
Enable disk caching support for OpenCL kernel binaries

9prady9:
Description
-----------
Fixes: #2857 

Changes to Users
----------------
OpenCL kernel binaries are loaded from disk if available instead compiling from scratch on initialization of the library. In the tests with nvidia opencl implementation, observed a speedup of 2x loading vs compile time.

Checklist
---------
- [x] Code compiles
- [ ] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/src/backend/cuda/compile_module.cpp
+++ b/src/backend/cuda/compile_module.cpp
@@ -131,7 +131,7 @@ string getKernelCacheFilename(const int device, const string &key) {
         to_string(computeFlag.first) + to_string(computeFlag.second);
 
     return "KER" + key + "_CU_" + computeVersion + "_AF_" +
-           to_string(AF_API_VERSION_CURRENT) + ".cubin";
+           to_string(AF_API_VERSION_CURRENT) + ".bin";
 }
 
 namespace common {
--- a/src/backend/cuda/compile_module.cpp
+++ b/src/backend/cuda/compile_module.cpp
@@ -316,40 +316,47 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
                                  getKernelCacheFilename(device, moduleKey);
         const string tempFile =
             cacheDirectory + AF_PATH_SEPARATOR + makeTempFilename();
-
-        // compute CUBIN hash
-        const size_t cubinHash = deterministicHash(cubin, cubinSize);
-
-        // write module hash(everything: names, code & options) and CUBIN data
-        ofstream out(tempFile, std::ios::binary);
-        if (!sourceIsJIT) {
-            size_t mangledNamesListSize = retVal.map().size();
-            out.write(reinterpret_cast<const char *>(&mangledNamesListSize),
-                      sizeof(mangledNamesListSize));
-            for (auto &iter : retVal.map()) {
-                size_t kySize   = iter.first.size();
-                size_t vlSize   = iter.second.size();
-                const char *key = iter.first.c_str();
-                const char *val = iter.second.c_str();
-                out.write(reinterpret_cast<const char *>(&kySize),
-                          sizeof(kySize));
-                out.write(key, iter.first.size());
-                out.write(reinterpret_cast<const char *>(&vlSize),
-                          sizeof(vlSize));
-                out.write(val, iter.second.size());
+        try {
+            // write module hash(everything: names, code & options) and CUBIN
+            // data
+            ofstream out(tempFile, std::ios::binary);
+            if (!sourceIsJIT) {
+                size_t mangledNamesListSize = retVal.map().size();
+                out.write(reinterpret_cast<const char *>(&mangledNamesListSize),
+                          sizeof(mangledNamesListSize));
+                for (auto &iter : retVal.map()) {
+                    size_t kySize   = iter.first.size();
+                    size_t vlSize   = iter.second.size();
+                    const char *key = iter.first.c_str();
+                    const char *val = iter.second.c_str();
+                    out.write(reinterpret_cast<const char *>(&kySize),
+                              sizeof(kySize));
+                    out.write(key, iter.first.size());
+                    out.write(reinterpret_cast<const char *>(&vlSize),
+                              sizeof(vlSize));
+                    out.write(val, iter.second.size());
+                }
             }
-        }
-        out.write(reinterpret_cast<const char *>(&cubinHash),
-                  sizeof(cubinHash));
-        out.write(reinterpret_cast<const char *>(&cubinSize),
-                  sizeof(cubinSize));
-        out.write(static_cast<const char *>(cubin), cubinSize);
-        out.close();
 
-        // try to rename temporary file into final cache file, if this fails
-        // this means another thread has finished compiling this kernel before
-        // the current thread.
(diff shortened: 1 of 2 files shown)


src/backend/opencl/compile_module.cpp
```diff
@@ -141,10 +187,41 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
 
 Module loadModuleFromDisk(const int device, const string &moduleKey,
                           const bool isJIT) {
-    UNUSED(device);
-    UNUSED(moduleKey);
-    UNUSED(isJIT);
-    return {};
+    const string &cacheDirectory = getCacheDirectory();
```
umar456, line 190: This code is not handling the exceptions thrown by ifstream reads.

src/backend/opencl/compile_module.cpp
```diff
@@ -119,6 +122,16 @@ Program buildProgram(const vector<string> &kernelSources,
 
 }  // namespace opencl
 
+string getKernelCacheFilename(const int device, const string &key) {
+    auto &dev = opencl::getDevice(device);
+
+    unsigned vendorId = dev.getInfo<CL_DEVICE_VENDOR_ID>();
+    auto devName      = dev.getInfo<CL_DEVICE_NAME>();
```
umar456, line 129: Perhaps remove spaces or replace with underscores from the device name? Some file systems may interpret paths with spaces differently.

Another thing to consider is to use the same case.

src/backend/opencl/compile_module.cpp
```diff
@@ -141,10 +202,45 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
...
+        if (recomputedHash != clbinHash) {
+            AF_ERROR("Binary on disk seems to be corrupted", AF_ERR_LOAD_SYM);
+        }
+        Program progWithBin(opencl::getContext(), {dev}, {clbin});
+        progWithBin.build();
+
+        AF_TRACE("{{{:<30} : loaded from {} for {} }}", moduleKey, cacheFile,
+                 dev.getInfo<CL_DEVICE_NAME>());
+        return {progWithBin};
+    } catch (...) {
```
umar456, line 239: Can you be more specific which exceptions you catch. We should avoid doing this in the backend because it masks errors in the code.

src/backend/opencl/compile_module.cpp
```diff
@@ -131,6 +152,46 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
...
+            const char *clbin =
+                reinterpret_cast<const char *>(binaries[0].data());
+            const size_t clbinHash = deterministicHash(clbin, clbinSize);
+
+            out.write(reinterpret_cast<const char *>(&clbinHash),
+                      sizeof(clbinHash));
+            out.write(reinterpret_cast<const char *>(&clbinSize),
+                      sizeof(clbinSize));
+            out.write(static_cast<const char *>(clbin), clbinSize);
+            out.close();
```
umar456, line 181: nit: You don't need to do this because its going to be handled when out goes out of scope.

src/backend/opencl/compile_module.cpp
```diff
@@ -141,10 +202,45 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
...
+
+        // TODO Handle cases where program objects are created from contexts
+        // having multiple devices
+        size_t clbinHash = 0;
+        in.read(reinterpret_cast<char *>(&clbinHash), sizeof(clbinHash));
+        size_t clbinSize = 0;
+        in.read(reinterpret_cast<char *>(&clbinSize), sizeof(clbinSize));
+        vector<unsigned char> clbin(clbinSize);
+        in.read(reinterpret_cast<char *>(clbin.data()), clbinSize);
+        in.close();
```
umar456, line 226: nit: This will close when `in` goes out of scope

src/backend/cuda/compile_module.cpp
```diff
@@ -316,40 +316,46 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
...
-                out.write(reinterpret_cast<const char *>(&kySize),
-                          sizeof(kySize));
-                out.write(key, iter.first.size());
-                out.write(reinterpret_cast<const char *>(&vlSize),
-                          sizeof(vlSize));
-                out.write(val, iter.second.size());
+        try {
+            // write module hash(everything: names, code & options) and CUBIN
+            // data
+            ofstream out(tempFile, std::ios::binary);
```
umar456, line 322: I was looking at the files generated by our code and these aren't cubin files because we add a bunch of information in the header. We should either have a separate extension or put the information in another file. Making the extension cubin is misleading.

umar456: Not in this PR of course.

9prady9: Just an .bin extension should clear any confusion. It will be just a binary file saved to disk by arrayfire, the name has enough information o. What the binary is about already.

src/backend/opencl/compile_module.cpp
9prady9: The only two specific exceptions we can catch here are `ios_base::failure` and `cl::Error`. In both cases, we would end up returning `AF_ERR_INTERNAL` error code unless we expand our existing codes. Isn't it ? Like add `AF_ERR_DISK_IO`. Final catch would have to be `...` to avoid throwing exception from this location because the inability to cache binary to disk is not a blocker for ArrayFire to run.

src/backend/opencl/compile_module.cpp
```diff
@@ -119,6 +125,21 @@ Program buildProgram(const vector<string> &kernelSources,
...
+    unsigned vendorId = dev.getInfo<CL_DEVICE_VENDOR_ID>();
+    auto devName      = dev.getInfo<CL_DEVICE_NAME>();
+    string infix      = to_string(vendorId) + "_" + devName;
+
+    transform(infix.begin(), infix.end(), infix.begin(),
+              [](unsigned char c) { return std::toupper(c); });
+    std::replace(infix.begin(), infix.end(), ' ', '_');
+
+    return "KER" + key + "_CL_" + infix + "_AF_" +
+           to_string(AF_API_VERSION_CURRENT) + ".clbin";
```
pavanky, line 140: May be use git hash or something else  instead of version? It may get tricky debugging issues when using a development branch.

src/backend/opencl/compile_module.cpp
9prady9: True but the file is being renamed after this line, so we want the file write to be complete before the rename.

src/backend/opencl/compile_module.cpp
9prady9: true again, but there are quite a few operations going on after reading the file contents, recomputation of hash, cl::program construction and there is no reason to keep the stream open while doing those.

src/backend/opencl/compile_module.cpp
9prady9: That shouldn't be a problem because we hash contents of source and associated compile options for that kernel - this hash is included via `key`.

Issue #2972, opened 2020-07-14
af::randn creates duplicates with CUDA backend [BUG]


willat343:
When creating large af::arrays with af::randn, the resulting sequence is not a truly random sequence, but in fact contains at least 2 of every number for large arrays.

Description
===========

* I originally became aware of the issue because I was listening to sequence, expected white noise however it had a distinct "harsh" audio quality compared to af::randu and using my workaround, this quality disappeared. Testing with an array of length 32768, I found that not a single number in the sequence was present an odd number of times, but some multiple of 2. I was able to verify this with a very small bin histogram plot of the data. I'm using the CUDA backend with a 2080.
* My theory: The CUDA backend generates this sequence by creating blocks of normally distributed random numbers in many threads and then recombines the data together, BUT each of these threads appears to be initiated with the same seed, and therefore the random sequences in the threads are the same. Possible then a fix is to set different seeds in each thread.
* I originally produced this error on 3.6.2 but then confirmed the issue is indeed still present in 3.7.2.
* I used the official installers to install arrayfire.
* I have reproduced this error using all three Random Engines; PHILOX, MERSENNE and THREEFLY.
* As a workaround, I am creating many blocks of size 128 and using af::join to combine them, with each block having a different random seed. This is obviously slow (by ~100 times on my machine).
* I have comfirmed that af::randu does not seem to suffer the same issue.
* This is very reproducible - it always occurs.
* I've put a screenshot below. Top left is the raw signal with a moving average in yellow. Top right is a fft done in matlab which seems somewhat suspicous. Bottom left is the clincher - histogram with tiny tiny bin width showing that each number occurs in a multiple of 2 (possibly pairs)

![arrayfire-randn-32768-points](https://user-images.githubusercontent.com/9509635/87372108-1daa7f00-c5ca-11ea-92b4-3c1caf24d969.png)

Reproducible Code and/or Steps
------------------------------

* `af::array my_array = af::randn(32768, f32);`
* Plot a histogram (e.g. with small bin size) to see even number in every bin.

System Information
------------------

* 3.7.2 and 3.6.2 tested
* Output of `nvidia-smi --query-gpu="name,memory.total,driver_version" --format=csv -i 0`: `name, memory.total [MiB], driver_version GeForce RTX 2080 SUPER, 7981 MiB, 440.100`

Checklist
---------

- [x ] Using the latest available ArrayFire release
- [x ] GPU drivers are up to date

umar456:
[...]

Issue #2007, opened 2017-12-04
RNG quality with OpenCL


rstub:
As part of https://github.com/RInstitute/rcpparrayfire I am trying to advertise ArrayFire's capability of producing random numbers real fast. But before doing so I wanted to check the RNGs and ran into problems when doing chi^2 tests with lots of random numbers. The original analysis used R in some parts, but I have recoded it in C++ here:
```C++
#include <iostream>
#include <iomanip>
#include <cassert>

#include <arrayfire.h>

// normalized chi^2 statistic for uniform distribution
// multiply with number of samples to get chi^2 statistic
float uchi2(af::array input) {
    assert(input.isvector());
    double expected = 1.0 / input.elements();
    af::array diff = input - expected;
    return af::sum<float>((diff * diff) / expected);
}

int main(int argc, char ** argv) {
    int backend = argc > 1 ? atoi(argv[1]) : 0;
    af::setBackend(static_cast<af::Backend>(backend));
    int device = argc > 2 ? atoi(argv[2]) : 0;
    af::setDevice(device);
    af::info();

    double samples = 1e8;
    int steps = 300;
    int bins = 100;

    af::array total_hist = af::constant(0, bins, f32);
    af::array step_hist = af::constant(0, bins, f32);

    af::setDefaultRandomEngineType(AF_RANDOM_ENGINE_PHILOX);
    //af::setDefaultRandomEngineType(AF_RANDOM_ENGINE_THREEFRY);
    //af::setDefaultRandomEngineType(AF_RANDOM_ENGINE_MERSENNE);
    std::cout << std::setw(4) << "step"
              << std::setw(7) << "chi2_i"
              << std::setw(7) << "chi2_t"
              << std::setprecision(2) << std::fixed << std::endl;
    for(int i = 0; i < steps; ++i) {
        step_hist = af::histogram(af::randu(samples), bins, 0.0, 1.0) / samples;
        total_hist = (total_hist * i + step_hist) / (i + 1);
        std::cout << std::setw(4) << i
                  << std::setw(7) << uchi2(step_hist) * samples
                  << std::setw(7) << uchi2(total_hist) * samples * (i +1)
                  << std::endl;
    }
}
```
[...]

PR #2980, opened 2020-07-30
Fixes random number generator with normal distribution

umar456:
The randn function was generating low quality floating point random numbers. This PR fixes and issue with the Box Muller parameters for floating point numbers.

Description
-----------

This bug fix addresses duplicate values when generating random numbers with floats. This issue was caused because the boxMullerTransform function was passed incorrect values for the second set of 2 floating point numbers. 

This PR also enables the Chi squared tests for random number generators. These are heavier tests but we should start testing the quality to avoid issues like this. We are unfortunately not testing the randn function but its a step in the right direction. @rstub How difficult would it be to implement a similar test for the normal distribution.

Fixes: #2972 

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~`[ ] Functions added to unified API~
- ~[ ] Functions documented~


rstub:
Concerning the chi-squared test: That is also possible for the normal distribution:

> Another test variant is the χ^2 test which was also used by Leong et al. (2005). The basic idea is as follow:
> 1. The real line is divided into B bins, equally spaced between (symmetric) values distant enough from zero so that no N(0, 1) draw should exceed them.
> 2. Here, we follow Leong et al. (2005) and use a range from -7 to 7 with a total of 200 bins.
> 3. A large number of N(0, 1) random variates is drawn, and for each of these numbers a counter in the bin corresponding to
the draw is increased.
> 4. After the N draws, the empirical distribution is compared to the theoretical (provided by the corresponding value of the Normal density function) using a standard chi2 test.

(source: https://cran.r-project.org/web/packages/RcppZiggurat/vignettes/RcppZiggurat.pdf)

I have tried to code this as follows:

```cpp
#include <iostream>
#include <iomanip>
#include <cassert>

#include <arrayfire.h>
#include <af/traits.hpp>

using af::array;
using af::randomEngine;
using af::randomEngineType;

template <typename T>
T chi2_statistic(array input, array expected) {
    expected *= af::sum<T>(input) / af::sum<T>(expected);
    array diff = input - expected;
    return af::sum<T>((diff * diff) / expected);
}

// should be used only for x <= 5 (roughly)
array cnd(array x)
{
    return 0.5 * erfc(- x * sqrt(0.5));
}

template <typename T>
bool testRandomEngineNormalChi2(randomEngineType type)
{
    af::dtype ty = (af::dtype)af::dtype_traits<T>::af_type;

    int elem = 1024*1024;
    int steps = 256*32;
    int bins = 100;
    T lower_edge = -7.0;
    T upper_edge = 7.0;

    array total_hist = af::constant(0.0, 2 * bins, ty);
    array edges = af::seq(bins + 1)/bins * lower_edge;
    array expected = -af::diff1(cnd(edges));
    expected = af::join(0, expected(af::seq(bins-1, 0, -1)), expected);

    af::randomEngine r(type, 0);

    // R> qchisq(c(5e-6, 1 - 5e-6), 197)
    // [1] 121.3197 297.2989
    T lower = 121.3197;
    T upper = 297.2989;

    bool prev_step = true;
    bool prev_total = true;
    af::setSeed(0x76fa214467690e3c);
    std::cout << std::setw(4) << "step"
              << std::setw(7) << "chi2_i"
              << std::setw(7) << "chi2_t"
              << std::setprecision(2) << std::fixed << std::endl;
    for (int i = 0; i < steps; ++i) {
        array step_hist = af::histogram(af::randn(elem, ty, r), 2 * bins, lower_edge, upper_edge);
        T step_chi2 = chi2_statistic<T>(step_hist, expected);
        std::cout << std::setw(4) << i
                  << std::setw(7) << step_chi2;
        bool step = step_chi2 > lower && step_chi2 < upper;
        if (!(step || prev_step)) return false;
        prev_step = step;

        total_hist += step_hist;
        T total_chi2 = chi2_statistic<T>(total_hist, expected);
        std::cout << std::setw(7) << total_chi2
                  << std::endl;
        bool total = total_chi2 > lower && total_chi2 < upper;
        if (!(total || prev_total)) return false;
        prev_total = total;
    }
    return true;
}



int main(int argc, char ** argv) {
    int backend = argc > 1 ? atoi(argv[1]) : 0;
    af::setBackend(static_cast<af::Backend>(backend));
    int device = argc > 2 ? atoi(argv[2]) : 0;
    af::setDevice(device);
    af::info();

    std::cout << "PHILOX" << std::endl;
    if (!testRandomEngineNormalChi2<float>(AF_RANDOM_ENGINE_PHILOX))
        std::cout << "FAILURE" << std::endl;
    std::cout << "THREEFRY" << std::endl;
    if (!testRandomEngineNormalChi2<float>(AF_RANDOM_ENGINE_THREEFRY))
        std::cout << "FAILURE" << std::endl;
    std::cout << "MERSENNE" << std::endl;
    if (!testRandomEngineNormalChi2<float>(AF_RANDOM_ENGINE_MERSENNE))
        std::cout << "FAILURE" << std::endl;

    if (af::isDoubleAvailable(device)) {
        std::cout << "PHILOX" << std::endl;
        if (!testRandomEngineNormalChi2<double>(AF_RANDOM_ENGINE_PHILOX))
            std::cout << "FAILURE" << std::endl;
        std::cout << "THREEFRY" << std::endl;
        if (!testRandomEngineNormalChi2<double>(AF_RANDOM_ENGINE_THREEFRY))
            std::cout << "FAILURE" << std::endl;
        std::cout << "MERSENNE" << std::endl;
        if (!testRandomEngineNormalChi2<double>(AF_RANDOM_ENGINE_MERSENNE))
            std::cout << "FAILURE" << std::endl;
    }
}
```

Getting the `expected` correct is a bit tricky. One needs to go far from the mean value, where the density of the normal distribution is almost zero. I am computing the density within these ranges via the difference of the cummulated density function (c.f. #2078), but that only works well for values close to zero, not close to one. I am sure there are more elegant ways to do this ...

Anyway, for me this fails *very* quickly

```
 ArrayFire v3.8.0 (OpenCL, 64-bit Linux, build 64855cbd)
[0] INTEL: Intel(R) Gen8 HD Graphics NEO, 9399 MB
-1- BEIGNET: Intel(R) HD Graphics 5500 BroadWell U-Processor GT2, 4096 MB
-2- POCL: pthread-Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz, 9701 MB
PHILOX
step chi2_i chi2_t
   0 287.13 287.13
   1 454.45 371.92
   2 371.33FAILURE
THREEFRY
step chi2_i chi2_t
   0 355.77 355.77
   1 464.06FAILURE
MERSENNE
step chi2_i chi2_t
   0 367.78 367.78
   1 350.76FAILURE
```

I am not sure if this indicates an error in my code or is an example of the problem you have identified here. I have not tried to build a patched version.

umar456:
@rstub This is great. I will work with this and see if I can improve the results. I am able to go further(fails after 13 steps) with the latest changes but not by much.  Thanks for your input!

umar456:
@rstub I pushed my latest changes to this branch and the normal distributions look better. The issue was with the box muller transform function we had been using to generate the normal distribution. For floats we were incorrectly passing in the wrong arguments and there was a subtle bug in one of the conversion methods. I made a couple changes to the test you posed above and I would appreciate if you could run and review the tests.

@willat343 Could you also give this branch a try and give me any feedback on the results?

test/rng_quality.cpp
```diff
@@ -0,0 +1,240 @@
...
+    testRandomEngineUniformChi2<TypeParam>(AF_RANDOM_ENGINE_THREEFRY_2X32_16);
+}
+
+TYPED_TEST(RandomEngine, mersenneRandomEngineUniformChi2) {
+    testRandomEngineUniformChi2<TypeParam>(AF_RANDOM_ENGINE_MERSENNE_GP11213);
+}
+
+// should be used only for x <= 5 (roughly)
+
+array cnd(array x) { return 0.5 * erfc(-x * sqrt(0.5)); }
```
rstub, line 141: Nitpick: I would remove the empty line between the comment and the function.

test/rng_quality.cpp
```diff
@@ -0,0 +1,240 @@
...
+        af::join(0, expected(af::seq(bins - 1, 0, -1)), expected).as(f32);
+    // af_print(expected);
+
+    af::randomEngine r(type, 0);
+
+    // R> qchisq(c(5e-6, 1 - 5e-6), 197)
+
+    // [1] 121.3197 297.2989
+    float lower(121.3197);
+    float upper(297.2989);
```
rstub, line 170: Nitpick: I would remove the empty line between the comments.

Explanation: In the chi^2 test one computes the test statistic and compares the value with the chi^2 distribution with appropriate number of degrees of freedom. For th euniform distribution one has "number of bins minus 1" degrees of freedom. For the normal distribution it is "number of bins minus 3", since there are two parameters mu and sigma. Here I used the `qchisq()` function from R to compute "suitable" values from the chi^2 distribution.

test/rng_quality.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    // R> qchisq(c(5e-6, 1 - 5e-6), 199)
+    // [1] 121.3197 297.2989
+    // float lower = 118.1094;
+    // float upper = 308.6010;
```
rstub, line 175: I think these lines can be removed. BTW, the actual result of this computation would be
```r
R> qchisq(c(5e-6, 1 - 5e-6), 199)
[1] 122.8777 299.7415
```

rstub:
I have checked out the branch. Works very well! I have add some nitpicky comments above.

BTW, good that you switched back to `int elem  = 256 * 1024 * 1024;`. The lower number of elements had been an artifact of my testing. And with lower elements, it gets more likely to get "odd" results in one of the tries.

umar456:
Thanks @rstub. I was curious about your choice for the dof as well. Your explanation makes sense. Can you give me your reasoning behind the probability values you chose for that function. I am trying to figure out if those values apply to other types like f16 or if they need to be modified for the limited precision of that type. The actual distribution for f16 looks reasonable when you print out the histogram of the random number array but quickly fails the chi test. I suspect it is because of the reduced resolution of the type.

src/backend/cpu/kernel/random_engine.hpp
```diff
@@ -31,89 +32,137 @@ static const double PI_VAL =
...
+#define SIGNED_HALF_FLT_FACTOR ((0.5f) * SIGNED_FLT_FACTOR)
+
+#define DBL_FACTOR \
+    ((1.0) / (std::numeric_limits<unsigned long long>::max() + (1.0)))
 #define HALF_DBL_FACTOR ((0.5) * DBL_FACTOR)
 
+// Conversion to floats adapted from Random123
+#define SIGNED_DBL_FACTOR \
+    ((1.0) / (std::numeric_limits<long long>::max() + (1.0)))
+#define SIGNED_HALF_DBL_FACTOR ((0.5) * SIGNED_DBL_FACTOR)
```
9prady9, line 60: Can't these also use numeric limits like CUDA backend ?

rstub:
@umar: The range of chi^2 is just a heuristic. Large enough to not trigger spurious failures in the cases I looked at but small enough to catch the errors identified at that time (c.f. #2007).

Have you tried running the test for `f16` for a longer period without the failure conditions? What does the local and global test statstics look like when plotted as a function of iteration? See https://cran.r-project.org/web/packages/RcppZiggurat/vignettes/RcppZiggurat.pdf for an example of such graphs.

src/backend/opencl/kernel/random_engine_write.cl
```diff
@@ -398,15 +398,22 @@ void partialBoxMullerWriteOut128Bytes_double(global double *out, uint index,
 #define SIGNED_HALF_FACTOR ((1.h) / (SHRT_MAX + (1.h)))
 #define SIGNED_HALF_HALF_FACTOR ((0.5h) * SIGNED_HALF_FACTOR)
 
+/// This is the largest integer representable by fp16. We need to
+/// make sure that the value converted from ushort is smaller than this
+/// value to avoid generating infinity
+#define MAX_INT_BEFORE_INFINITY (ushort)65504u
```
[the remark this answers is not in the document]
umar456: We don't have constexpr in OpenCL C

src/backend/cuda/compile_module.cpp
```diff
@@ -85,7 +85,7 @@ using std::chrono::milliseconds;
     do {                                                                \
         CUresult res = (fn);                                            \
         if (res == CUDA_SUCCESS) break;                                 \
-        array<char, 2048> cu_err_msg;                                   \
+        array<char, 2048 + 512> cu_err_msg;                             \
```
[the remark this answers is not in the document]
umar456: This fixes a warning in the new NVCC compiler. The derived message from sprintf could have been longer than the length specified here. It's just a buffer size and it needs to hold enough data for the log message. We don't even need the entire log message for this to be useful. I don't think this requires a comment.

src/backend/cuda/compile_module.cpp
```diff
@@ -270,7 +270,7 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
     ptx.resize(ptx_size);
     NVRTC_CHECK(nvrtcGetPTX(prog, ptx.data()));
 
-    const size_t linkLogSize    = 4096;
+    const size_t linkLogSize    = 2048;
```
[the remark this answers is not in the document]
umar456: 4096 is too much imo. this would also require that much space every time we check for an error message. 2048 seems more reasonable.

PR #2986, opened 2020-08-08
Add new standard deviation APIs with bias parameter control

9prady9:
Description
---------------
- Add new std dev APIs with bias parameter
- Add new covariance APIs with bias enum instead of bool
- Add new variance APIs with bias enum argument instead of bool

Fixes: #2881 

Changes to Users
----------------
New APIs to control bias for variance, covariance and standard deviation functions.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- [x] Functions documented
- [x] Test deprecated API of new API

### DIFF
--- a/include/af/statistics.h
+++ b/include/af/statistics.h
@@ -121,18 +121,36 @@ AFAPI void meanvar(array& mean, array& var, const array& in, const array& weight
 */
 AFAPI array stdev(const array& in, const dim_t dim=-1);
 
-
 /**
    C++ Interface for covariance
 
    \param[in] X is the first input array
    \param[in] Y is the second input array
-   \param[in] isbiased is boolean specifying if biased estimate should be taken (default: false)
+   \param[in] isbiased is boolean specifying if biased estimate should be
+              taken (default: false)
    \return    the covariance of the input arrays
 
    \ingroup stat_func_cov
+
+   \deprecated Use af::cov(const array&, const array& const af_var_bias)
 */
+AF_DEPRECATED("Use af::cov(const af::array&, const array&, conv af_var_bias)")
 AFAPI array cov(const array& X, const array& Y, const bool isbiased=false);
+
+#if AF_API_VERSION >= 38
+/**
+   C++ Interface for covariance
+
+   \param[in] X is the first input array
+   \param[in] Y is the second input array
+   \param[in] bias The type of bias used for variance calculation. Takes of
+              value of type \ref af_var_bias.
+   \return the covariance of the input arrays
+
+   \ingroup stat_func_cov
+*/
+AFAPI array cov(const array &X, const array &Y, const af_var_bias bias);
+#endif
 
 /**
    C++ Interface for median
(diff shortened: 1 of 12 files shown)


include/af/statistics.h
```diff
@@ -232,6 +237,80 @@ AFAPI void topk(array &values, array &indices, const array& in, const int k,
...
+                  const af_var_bias bia = AF_VARIANCE_POPULATION,
+                  const dim_t dim = -1);
+
+/**
+   C++ Interface for variance of all elements
+
+   \param[in] in is the input array
+   \param[in] bias The type of bias used for variance calculation. Takes of
+              value of type \ref af_var_bias. Defaults to population variance.
+   \return variance of the entire input array
```
umar456, line 264:
```suggestion
   \return variance of the \p in array
```

test/stdev.cpp
```diff
@@ -90,7 +90,7 @@ void stdevDimTest(string pFileName, dim_t dim = -1) {
 
     array a(dims, &(input.front()));
 
-    array b = stdev(a, dim);
+    array b = stdevV2(a, dim, AF_VARIANCE_POPULATION);
```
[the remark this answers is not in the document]
9prady9: Yes, that is still pending - hence the item in issue checklist.

include/af/statistics.h
```diff
@@ -43,6 +43,8 @@ AFAPI array mean(const array& in, const dim_t dim=-1);
 AFAPI array mean(const array& in, const array& weights, const dim_t dim=-1);
 
 /**
+   \deprecated Use \ref af::varV2 instead
```
umar456, line 46: This is supposed to be at the bottom of the comment. It is like that in all other places in the headers.

include/af/statistics.h
```diff
@@ -232,6 +247,80 @@ AFAPI void topk(array &values, array &indices, const array& in, const int k,
                 const int dim = -1, const topkFunction order = AF_TOPK_MAX);
 #endif
 
+#if AF_API_VERSION >= 38
```
umar456, line 250: It would be nice if we moved these definitions closer to the deprecated versions.

include/af/statistics.h
9prady9: I understand consistency concern but ideally I would expect to see if a function is deprecated right at the front of it's documentation rather than towards end after spending time reading it.

include/af/statistics.h
9prady9: Sure.

include/af/statistics.h
```diff
@@ -97,22 +118,61 @@ AFAPI void meanvar(array& mean, array& var, const array& in, const array& weight
    \ingroup stat_func_stdev
 
    \note \p dim is -1 by default. -1 denotes the first non-singleton dimension.
+
+   \deprecated Use \ref af::stdev that takes \ref af_var_bias instead
 */
+AF_DEPRECATED("Use af::stdev(const array&, const af_var_bias, const dim_t")
```
umar456, line 124:
```suggestion
AF_DEPRECATED("Use af::stdev(const array&, const af_var_bias, const dim_t)")
```

PR #2987, opened 2020-08-11
Move initializer_list constructor implementation to the header

umar456:
Moves the initialization list constructor to the header so that
we avoid version conflicts between compilers for the implemenation
of the initializer list. This approach will generate the initializer
list constructor for the user's compiler and avoid such conflicts.

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -50,7 +50,8 @@ namespace af
         ///
         /// \brief Intermediate data class. Used for assignment and indexing.
         ///
-        /// \note This class is for internal book keeping while indexing. This class is not intended for use in user code.
+        /// \note This class is for internal book keeping while indexing. This
+        ///       class is not intended for use in user code.
         ///
         class AFAPI array_proxy
         {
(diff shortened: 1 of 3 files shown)


include/af/array.h
```diff
@@ -494,10 +521,37 @@ namespace af
...
+        }
+
+        /// \brief Initializer list constructor
         template <typename T>
-        array(const af::dim4 &dims, std::initializer_list<T> list);
+        array(const af::dim4 &dims, std::initializer_list<T> list)
+            : arr(nullptr) {
+          const dim_t *size = dims.get();
+          if (af_err __aferr = af_create_array(
+              &arr, list.begin(), AF_MAX_DIMS, size,
```
9prady9, line 545: nit: Would be better if `dims.ndims()` - doesn't depend on `Dim4` having space allocated for `AF_MAX_DIMS`.

PR #2991, opened 2020-08-11
Fix failure with git apply after make clean

umar456:
OpenCL builds fail after you try to rebuild after you do a make clean because
of the patch command in the build_clblast.cmake script. This PR fixes this
issue.

Description
-----------

After the clean target is run, the patch on CLBlast is applied again.
this fails because the patch has already been applied. In order to
fix this we need to reset to the original tag and apply again.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9:
This is odd, I do remember doing a clean and rebuilding just few days back and it worked as expected.

PR #2996, opened 2020-08-15
Compute capability fix

umar456:
Fixes checkAndSet

Description
-----------
Fixes an issue where the device compute capability is larger than
the supported maximum of the CUDA runtime used to build ArrayFire.
This happens for example when you run the Turing card with a CUDA
runtime of 9.0. The compute capability of Turing is 7.5 and the
maximum supported by the runtime is 7.0/7.2. Before this change
we were only checking the major compute capability and not checking
the minor version to set the max compute capability of the device.
This caused errors like:

```
In file src/backend/cuda/compile_module.cpp:266
NVRTC Error(5): NVRTC_ERROR_INVALID_OPTION
Log:
nvrtc: error: invalid value for --gpu-architecture (-arch)
```
This PR also updates the error messages for failure cases.

The utility header in cuda_fp16.hpp is not included automatically
in CUDA 9. Additionally we need to pass the
--device-as-default-execution-space flag to nvrtc for JIT and
non-JIT kernels

* The moduleKey is an size_t object so the maximum number of digits
  it can have is 20 so the format length for that value is updated
* The runtime check messages are always logged (but not displayed)
  Errors are still only thrown in debug modes
* Display the compute capability of the CUDA device along with
  its name and other stats

  example:
```
  Found device: Quadro T2000 (sm_75) (3.82 GB | ~3164.06 GFLOPs | 16 SMs)
```

Changes to Users
----------------
Better error messages and better support for newer devices with older CUDA toolkits

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/src/backend/common/half.hpp
+++ b/src/backend/common/half.hpp
@@ -832,9 +832,7 @@ class alignas(2) half {
 #endif
 
    public:
-#if CUDA_VERSION >= 10000
     AF_CONSTEXPR
-#endif
     half() = default;
 
     /// Constructor.
--- a/src/backend/cuda/compile_module.cpp
+++ b/src/backend/cuda/compile_module.cpp
@@ -143,20 +143,29 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
                      const vector<string> &kInstances, const bool sourceIsJIT) {
     nvrtcProgram prog;
     if (sourceIsJIT) {
-        array<const char *, 2> headers = {
+        constexpr const char *header_names[] = {
+            "utility",
+            "cuda_fp16.hpp",
+            "cuda_fp16.h",
+        };
+        constexpr size_t numHeaders = extent<decltype(header_names)>::value;
+        array<const char *, numHeaders> headers = {
+            "",
             cuda_fp16_hpp,
             cuda_fp16_h,
         };
-        array<const char *, 2> header_names = {"cuda_fp16.hpp", "cuda_fp16.h"};
+        static_assert(headers.size() == numHeaders,
+                      "headers array contains fewer sources than header_names");
         NVRTC_CHECK(nvrtcCreateProgram(&prog, sources[0].c_str(),
-                                       moduleKey.c_str(), 2, headers.data(),
-                                       header_names.data()));
+                                       moduleKey.c_str(), numHeaders,
+                                       headers.data(), header_names));
     } else {
         constexpr static const char *includeNames[] = {
             "math.h",          // DUMMY ENTRY TO SATISFY cuComplex_h inclusion
             "stdbool.h",       // DUMMY ENTRY TO SATISFY af/defines.h inclusion
             "stdlib.h",        // DUMMY ENTRY TO SATISFY af/defines.h inclusion
             "vector_types.h",  // DUMMY ENTRY TO SATISFY cuComplex_h inclusion
+            "utility",         // DUMMY ENTRY TO SATISFY cuda_fp16.hpp inclusion
             "backend.hpp",
             "cuComplex.h",
             "jit.cuh",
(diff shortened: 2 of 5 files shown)


src/backend/cuda/compile_module.cpp
```diff
@@ -143,20 +143,29 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
...
+            "cuda_fp16.h",
+        };
+        constexpr size_t numHeaders = extent<decltype(header_names)>::value;
+        array<const char *, numHeaders> headers = {
+            "",
             cuda_fp16_hpp,
             cuda_fp16_h,
         };
-        array<const char *, 2> header_names = {"cuda_fp16.hpp", "cuda_fp16.h"};
+        static_assert(headers.size() == numHeaders,
```
9prady9, line 157: :+1:

src/backend/cuda/compile_module.cpp
```diff
@@ -230,11 +240,11 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
             sourceStrings[22].c_str(), sourceStrings[23].c_str(),
             sourceStrings[24].c_str(), sourceStrings[25].c_str(),
             sourceStrings[26].c_str(), sourceStrings[27].c_str(),
-        };
-        static_assert(extent<decltype(headers)>::value == NumHeaders,
+            sourceStrings[28].c_str()};
+        static_assert(extent<decltype(headers)>::value == numHeaders,
```
9prady9, line 244: :+1:

src/backend/cuda/compile_module.cpp
```diff
@@ -246,6 +256,7 @@ Module compileModule(const string &moduleKey, const vector<string> &sources,
     vector<const char *> compiler_options = {
         arch.data(),
         "--std=c++14",
+        "--device-as-default-execution-space",
```
9prady9, line 259: Hm, I thought I added the unconditionally to nvrtc flags

src/backend/common/half.hpp
```diff
@@ -832,9 +832,7 @@ class alignas(2) half {
 #endif
 
    public:
-#if CUDA_VERSION >= 10000
     AF_CONSTEXPR
```
9prady9, line 835: I believe this was added to circumvent another issue.

src/backend/cuda/device_manager.cpp
```diff
@@ -105,50 +126,66 @@ bool checkDeviceWithRuntime(int runtime, pair<int, int> compute) {
...
-            "cuda runtime compute version (%d.%d). Using %d.%d.";
-        fprintf(stderr, errMsg, originalCompute.first, originalCompute.second,
-                prop.first, prop.second, prop.first, prop.second);
-#endif
-    } else if (prop.first > Toolkit2MaxCompute[0].major) {
+        computeCapability.first >= tkitMaxCompute->major) {
+        int minorVersion = embeddedDevice ? tkitMaxCompute->embedded_minor
+                                          : tkitMaxCompute->minor;
+
+        if (computeCapability.second > minorVersion) {
```
9prady9, line 146: Does this condition actually help? I mean, if major version is greater, minor version doesn't matter I think.

Similar check in the next elseif give path

umar456: for 9.0 the max compute capability is 7.0 but my card is 7.5. This configuration causes compilation errors when building with nvrtc like the one I posted in my description.

src/backend/common/half.hpp
umar456: I think the problem was the --device-as-default-execution-space flag for jit kernels. It was treating this function as a host only function in that case causing errors.

src/backend/cuda/device_manager.cpp
9prady9: Right, I didn't notice the equal sign in >=. My bad

tvandera:
Hi,

I'm having a similar issue on this configuration:

```
ArrayFire v3.7.3 (CUDA, 64-bit Linux, build 59ac7b980)
Platform: CUDA Runtime 10.1, Driver: 460.32.03
[0] A100-SXM4-40GB, 40537 MB, CUDA Compute 8.0
```

This is the error:
```
In function cuda::Module common::compileModule(const string&, const std::vector<std::__cxx11::basic_string<char> >&, const std::vector<std::__cxx11::basic_string<char
> >&, const std::vector<std::__cxx11::basic_string<char> >&, bool)
In file src/backend/cuda/compile_module.cpp:277
NVRTC Error(5): NVRTC_ERROR_INVALID_OPTION
Log:
nvrtc: error: invalid value for --gpu-architecture (-arch)
```

I do not understand what I should do to fix this.

9prady9:
@tvandera That is expected outcome given that v3.7.3 installers aren't built with CUDA 11 support. Please check the latest 3.8 release which has CUDA 11 support.

https://arrayfire.com/arrayfire-v3-8-release/

PR #3000, opened 2020-08-19
Add functions to set and get the cache directory for runtime kernels

umar456:
Add functions to set and get the cache directory for the kernels compiled at runtime.

Description
-----------
* Adds an enviornment variable AF_JIT_KERNEL_CACHE_DIRECTORY which
  can change the path where ArrayFire will store the runtime generated
  kernels. If the path is not writeable ArrayFire will fallback to defaults
* Add functions af_get_kernel_cache_directory and af_set_kernel_cache_directory
  which can also set the directory in the code.
* The set function is capablie of overriding the environment variable values
  based on the override_env variable
* Add CUDA 11 to Toolkit2MaxCompute. Add Ampere sm to compute2cores
* Fix variance warning in naive_bayes example 

Changes to Users
----------------
* Add the ability to set and get the directory where ArrayFire will store the kernels generated at runtime
* Add an environment variable that can set where ArrayFire will store the kernels generated at runtime

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/include/af/device.h
+++ b/include/af/device.h
@@ -575,6 +575,49 @@ extern "C" {
     */
     AFAPI af_err af_get_device_ptr(void **ptr, const af_array arr);
 
+#if AF_API_VERSION >= 38
+    /**
+       Sets the path where the kernels generated at runtime will be cached
+
+       Sets the path where the kernels generated at runtime will be stored to
+       cache for later use. The files in this directory can be safely deleted.
+       The default location for these kernels is in $HOME/.arrayfire on Unix
+       systems and in the ArrayFire temp directory on Windows.
+
+       \param[in] path The location where the kernels will be stored
+       \param[in] override_env if true this path will take precedence over the
+                               AF_JIT_KERNEL_CACHE_DIRECTORY environment variable.
+                               If false, the environment variable takes precedence
+                               over this path.
+
+       \returns AF_SUCCESS if the variable is set. AF_ERR_ARG if path is NULL.
+       \ingroup device_func_mem
+    */
+    AFAPI af_err af_set_kernel_cache_directory(const char* path,
+                                               int override_env);
+
+    /**
+       Gets the path where the kernels generated at runtime will be cached
+
+       Gets the path where the kernels generated at runtime will be stored to
+       cache for later use. The files in this directory can be safely deleted.
+       The default location for these kernels is in $HOME/.arrayfire on Unix
+       systems and in the ArrayFire temp directory on Windows.
+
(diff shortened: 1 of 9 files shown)


include/af/device.h
```diff
@@ -575,6 +575,46 @@ extern "C" {
...
+
+       \param[out] length The length of the path array. If \p path is NULL, the
+                          length of the current path is returned
+       \param[out] path The path of the runtime generated kernel cache
+                         variable
+       \returns AF_SUCCESS if the variable is set. AF_ERR_ARG if path and length
+                           are null at the same time
+       \ingroup device_func_mem
+    */
+    AFAPI af_err af_get_kernel_cache_directory(size_t *length, char *path);
```
9prady9, line 615: We have to mention the behaviour  when size of the char buffer is not sufficient to store the path as  note.

src/backend/common/util.hpp
```diff
@@ -14,6 +14,11 @@
 #include <string>
 #include <vector>
 
+/// The environment variable that determines where the runtime kernels
+/// will be stored on the file system
+constexpr const char* JIT_KERNEL_CACHE_DIRECTORY =
+    "AF_JIT_KERNEL_CACHE_DIRECTORY";
```
9prady9, line 20: nit: Would be nice if it is something like `JIT_KERNEL_CACHE_DIR_ENV_NAME`

include/af/device.h
```diff
@@ -575,6 +575,46 @@ extern "C" {
...
     AFAPI af_err af_get_device_ptr(void **ptr, const af_array arr);
 
+#if AF_API_VERSION >= 38
+    /**
+       Sets the path where the kernels generated at runtime will be cached
+
+       Sets the path where the kernels generated at runtime will be stored to
+       cache for later use. The files in this directory can be safely deleted.
+       The default location for these kernels is in $HOME/.arrayfire on Unix
+       systems and in the ArrayFire temp directory on Windows.
```
9prady9, line 585: I think we can better illustrate the lookup logic as an ordered list of entries such as below. We can then hyperlink this order of lookup from environment variables page documentation also.
1. Environment variable path if writable
2. List of predetermined paths as per the operation system.
    - Windows
        1. ArrayFire application Temp folder 
    - Linux
       1. Under `.arrayfire` folder of user home directory.
       2. `/tmp/arrayfire`

9prady9 (changes requested):
Is missing C++ API intentional ?

9prady9 (approved):

Issue #2937, opened 2020-06-23
[Perf] OpenCL performance issue with sparse matmul


ebarjou:
Sparse-dense matrix multiplication under OpenCL perform terribly compared to CPU and Cuda backend

Description
===========

Trying to debug an issue where consecutive matmulNT under CUDA take more an more time (up to 10 times), I faced a far more severe issue trying to compare to the OpenCL backend.
Compared to other backend, the matmul operation under OpenCL is extremly slow, and the matmulTN even more : 

OpenCL : 
        matmulNN : 1510.16ms
        matmulTN : 15950.30ms

CUDA : 
        matmulNN : 2.25ms
        matmulTN : 4.07ms

CPU : 
        matmulNN : 95.73ms
        matmulTN : 177.07ms

Reproducible Code
-----------------
```c
#include <arrayfire.h>
#include <iostream>

int main(int argc, char *argv[]) {
    int run_count = 10;
    int M = 1280*1280;    //
    int N = 1280*512*20; // Size of the matrices
    double res[] = {0, 0};
    try{
        af::array A = af::floor(af::randu(M)*M).as(af::dtype::s32);
        af::array B = af::range(af::dim4(M+1), 0, af::dtype::s32);
        af::array W = af::randu(M);

        af::array P = af::randu(N);
        af::array L = af::randu(M);

        af::array S = af::sparse(M, N, W, B, A, af::storage::AF_STORAGE_CSR);

        af::array res1 = af::eval(af::matmul(S, P));
        af::array res2 = af::eval(af::matmulTN(S, L));

        for(int i = 0; i < run_count; ++i) {
            af::timer::start();
            res1 = af::eval(af::matmul(S, P));
            af::sync();
            res[0] += af::timer::stop();

            af::timer::start();
            res2 = af::eval(af::matmulTN(S, L));
            af::sync();
            res[1] += af::timer::stop();
        }
    } catch  (af::exception& e) { 
        std::cout << e.what() << std::endl;
        return -1;
    }

    std::cout << "Results : " << std::endl;
    std::cout << "\tmatmulNN : " << 1000*res[0]/run_count << "ms" << std::endl;
    std::cout << "\tmatmulTN : " << 1000*res[1]/run_count << "ms" << std::endl;
    return 0;
}
```

System Information
------------------
ArrayFire Version: 3.7.1
Device: i7-9750H, 16Go RAM, GTX 1650 4Go
Operating System: Windows 10
Driver version: 442.19

[...]

PR #3010, opened 2020-09-08
Improve csrmv sparse-dense matmul performance slightly

9prady9:
Description
-----------
Fixes: #2937 

Changes to Users
----------------
Sparse-Dense matrix multiplication should have improved performance when dense matrix is a vector.
 
Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9:
- Note that this change is not kernel change, it reorganized and/or removed some unnecessary code which brought about 6x speed improvement in OpenCL backend's for sparse-matrix--dense-vector matmulNN operation.
- There has been no change in speed for matmulTN operation though.

I am going to look into kernel improvements of NN and TN ops now.

9prady9:
>     * Note that this change is not kernel change, it reorganized and/or removed some unnecessary code
>        which brought  about 6x speed improvement in OpenCL backend's for sparse-matrix--dense-vector
>        matmulNN operation.
> 
>     * There has been no change in speed for matmulTN operation though.
> 
> 
> I am going to look into kernel improvements of NN and TN ops now - differed for later.

@umar456 

I was unable to reproduce the 6x performance difference with this code reorg on lastest master with latest versions of dependencies. It seems like some sort of driver or upstream fix addressed the problem of optimizing out the unused global counter being passed for `use_greedy` code-path.

Nevertheless, the cleanup of the wrapper code is still valid. We don't need to pass the global counter when not used and we most definitely don't need to compile an additional kernel when not using it at all. I am going to reword the PR title to appropriately reflect the change.

umar456 (approved):

PR #3023, opened 2020-10-05
Fix indexing error and add support for CUDA 11.1

umar456:
Fix an issue with indexing where uncommon usage of the indexing object caused a segfault.

Description
-----------
When copying indexing objects, the reference count was not incremented. this caused a segfault. This becomes an issue when you are directly using the indexing object.

Added support for CUDA 11 and compute capability 8.6(GTX 3090)

Changes to Users
----------------
N/A

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9 (approved):

jacobkahn:
Thanks again for this @umar456.

PR #3031, opened 2020-10-23
JIT optimization for sequential casts that are idempotent

jacobkahn:
Adds a JIT optimization which does a NOOP in the case of sequential cases that don't result in a differently-typed result.

Description
-----------

The following code is technically a noop:
```c++
af::array a = af::randu(10, 1, 1, 1, af::dtype::f32);
af::array b = a.as(af::dtype::f64);
af::array c = b.as(af::dtype::f32);
```
No casting kernels should be generated for any of the above operations, especially for `c`, but they are. The solution here is to, when creating the `CastOp`/`CastWrapper` for `c`, to check to see if the previous operation was a cast. If it was, and the previous operation's previous operation's output type is the same output type as the current cast, create a `__noop` node between the prev-prev operation and the current one.

This also precludes tricky cases like:
```c++
af::array a = af::randu(10, 1, 1, 1, af::dtype::f32);
af::array b = a.as(af::dtype::f64);
af::array d = b + 2;
af::array c = b.as(af::dtype::f32);
c.eval();
d.eval();
```
where the result of `b` could be used, in which case the intermediate casting operation can't be discounted completely.

With the change, running `AF_JIT_KERNEL_TRACE=stderr ./test/cast_cuda --gtest_filter="*Test_JIT_DuplicateCastNoop"` produces the following kernels:
- CUDA: https://gist.github.com/jacobkahn/c59884ee1d64a996bf9e81002ed46c9f.
- OpenCL: https://gist.github.com/jacobkahn/628659c9d7adbfc7c713fb44b1afed4c

Before the change, the generated kernel used wasteful casts:
- CUDA: https://gist.github.com/jacobkahn/ec66b1645a5e7c69fbc457dd7b16461b,

This PR also adds op, type, and children accessors to `Node`/`NaryNode` to facilitate inspecting the JIT tree for optimization.

Further optimization could be had by recursively checking previous operations until an operation has no previous operations are casts - this would fix arbitrarily long chains of casts that were noops on a particular subtree of JIT operations.

Changes to Users
----------------
No changes to user behavior.

Checklist
---------
- [X] Code compiles
- [X] Tests pass
- [X] Functions added to unified API
- [X] Functions documented

### DIFF
--- a/src/backend/cuda/cast.hpp
+++ b/src/backend/cuda/cast.hpp
@@ -16,7 +16,6 @@
 #include <optypes.hpp>
 #include <types.hpp>
 #include <af/dim4.hpp>
-#include <complex>
 
 namespace cuda {
 
--- a/include/af/arith.h
+++ b/include/af/arith.h
@@ -822,6 +822,35 @@ extern "C" {
     /**
        C Interface for casting an array from one type to another
 
+       This function casts an af_array object from one type to another. If the
+       type of the original array is the same as \p type then the same array is
+       returned.
+
+       \note Consecitive casting operations may be may be optimized out if the
+       original type of the af_array is the same as the final type. For example
+       if the original type is f64 which is then cast to f32 and then back to
+       f64, then the cast to f32 will be skipped and that operation will *NOT*
+       be performed by ArrayFire. The following table shows which casts will
+       be optimized out. outer -> inner -> outer
+       | inner-> | f32 | f64 | c32 | c64 | s32 | u32 | u8 | b8 | s64 | u64 | s16 | u16 | f16 |
+       |---------|-----|-----|-----|-----|-----|-----|----|----|-----|-----|-----|-----|-----|
+       | f32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+       | f64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+       | c32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+       | c64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+       | s32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
+       | u32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
+       | u8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
+       | b8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
+       | s64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
+       | u64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
+       | s16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
+       | u16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
+       | f16     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+       If you want to avoid this behavior use af_eval after the first cast
+       operation. This will ensure that the cast operation is performed on the
+       af_array
+
        \param[out] out will contain the values in the specified type
        \param[in] in is the input
        \param[in] type is the target data type \ref af_dtype
--- a/include/af/array.h
+++ b/include/af/array.h
@@ -933,9 +933,34 @@ namespace af
         const array::array_proxy slices(int first, int last) const; ///< \copydoc slices
         /// @}
 
-        /// \brief Converts the array into another type
+        /// \brief Casts the array into another data type
         ///
-        ///  \param[in] type is the desired type(f32, s64, etc.)
+        /// \note Consecitive casting operations may be may be optimized out if
+        /// the original type of the af::array is the same as the final type.
+        /// For example if the original type is f64 which is then cast to f32
+        /// and then back to f64, then the cast to f32 will be skipped and that
+        /// operation will *NOT* be performed by ArrayFire. The following table
+        /// shows which casts will be optimized out. outer -> inner -> outer
+        /// | inner-> | f32 | f64 | c32 | c64 | s32 | u32 | u8 | b8 | s64 | u64 | s16 | u16 | f16 |
+        /// |---------|-----|-----|-----|-----|-----|-----|----|----|-----|-----|-----|-----|-----|
+        /// | f32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+        /// | f64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+        /// | c32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+        /// | c64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
+        /// | s32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
+        /// | u32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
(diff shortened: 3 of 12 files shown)


src/backend/cuda/cast.hpp
```diff
@@ -89,9 +89,40 @@ struct CastWrapper {
...
+        // TODO: handle arbitrarily long chains of casts
+        auto in_node_nary =
+            std::dynamic_pointer_cast<common::NaryNode>(in_node);
+        if (in_node_nary && in_node_nary->getOp() == af_cast_t) {
+            // The only way to get the input type of the child node if it's a
+            // cast is to get the output type of the child's child.
+            auto in_node_children = in_node_nary->getChildren();
+            // Check if any children are casts with the same type - if so,
+            // insert a shortcut noop
+            for (size_t i = 0; i < in_node_children.size(); ++i) {
```
umar456, line 105: use `NaryNode::m_num_childern instead of size here. The size function will always return `Node::kMaxChildren`. You may need to create an accessor function.

src/backend/opencl/cast.hpp
```diff
@@ -73,11 +73,41 @@ struct CastOp<cdouble, cdouble> {
 template<typename To, typename Ti>
 struct CastWrapper {
     Array<To> operator()(const Array<Ti> &in) {
-        CastOp<To, Ti> cop;
```
umar456, line 76: I think removing this line is causing build failures on the OpenCL backend.

src/backend/cuda/cast.hpp
```diff
@@ -89,9 +89,40 @@ struct CastWrapper {
...
+                common::Node_ptr in_in_node = in_node_nary->getChildren()[i];
+                if (in_in_node->getType() == to_dtype) {
+                    // If the output of the child's child is the same as the
+                    // output of this node, ignore the input node and simply
+                    // connect a noop node from the child's child to produce
+                    // this op's output
+
+                    // TODO: including unary.hpp to use
+                    // detail::unaryName<af_noop_t> breaks some other stuff
+                    return createNodeArray<To>(
```
umar456, line 116: Instead of creating a noop node. Is it possible to just pass the in_in_node ptr to the createNodeArray? That way you don't need to call a new kernel to do the noop step.

umar456 (changes requested):
Thanks for sending this in. I made a couple of suggestions which are required to get the OpenCL  backend working again.

src/backend/cuda/cast.hpp
umar456: Since cast is a unary operation, you cast to UnaryNode instead of NaryNode. That way you can remove the loop.

src/backend/cuda/cast.hpp
jacobkahn: If you're referring to `return createNodeArray<To>(in.dims(), in_node);`, that doesn't seem to work — the output isn't equal to the input. I'll have to read into these functions to understand why that's the case - do you have any idea?

umar456:
There is a invalid read access error with the sparse test in the CUDA backend. I am able to reproduce it using the following code:

```
TEST(Cast, abs) {
    using namespace af;
    array a = randu(100, 100, f64);
    array b = a.as(f32);

    array c = max<double>(abs(a - b));
}
```

There is something odd going on with the implicit casts in the subtraction operation. It looks like the buffer object's shape is not set during the conversion. I am not sure where this is happening and I am investigating it.

jacobkahn:
@umar456 any update on this? Can I help in any way?

umar456:
@jacobkahn
I fixed the issues I referenced in my previous comment. I am thinking of a couple of scenarios where this optimization could be an issue. For example, what if we cast a floating point type to an integer type and back to float. This should floor all the values in the array but that may not happen with this change. Do you think it would be a good idea to perform this operation in non-destructive casting operations?

We can limit this optimization to casts between integer types or floating point types. This way it behaves like C++ types.

Alternatively, we could keep the current behavior and allow destructive casts and expect the user to use functions like floor or ceil to get the same behavior.

jacobkahn:
@umar456 — revisiting this after some time — I think for now, destructive casts that emulate floor/ceiling operations probably aren't good to implicitly-implement. The casts that I think are more interesting to optimize away are casts between similar types with different precisions — f16 <> f32 <> f64 or u32 <> u64, etc. While some of these casts are destructive, there isn't really an operation to emulate them. A user who casts f32 --> f16 --> f32 almost certainly isn't doing so to intentionally lose precision.

Thoughts?

umar456 (approved):

umar456:
I modified the PR to remove the cast in limited scenarios.  Currently I am removing the intermediate casts for casts between floating point values. Floating point to integer types are not removed and integer casts from larger to smaller types are not removed. Here are all the combinations of casts that will be removed. The x cell indicates that the cast will be removed if the left hand column type is the outer type and the top row is the inner type. for example outer -> inner -> outer

```
       | inner-> | f32 | f64 | c32 | c64 | s32 | u32 | u8 | b8 | s64 | u64 | s16 | u16 | f16 |
       |---------|-----|-----|-----|-----|-----|-----|----|----|-----|-----|-----|-----|-----|
       | f32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | f64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | c32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | c64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | s32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
       | u32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
       | u8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
       | b8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
       | s64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
       | u64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
       | s16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
       | u16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
       | f16     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
```

umar456:
I modified the PR to remove the cast in limited scenarios.  Currently I am removing the intermediate casts for casts between floating point values. Floating point to integer types are not removed and integer casts from larger to smaller types are not removed. Here are all the combinations of casts that will be removed. The x cell indicates that the cast will be removed if the left hand column type is the outer type and the top row is the inner type. for example outer -> inner -> outer

```
       | inner-> | f32 | f64 | c32 | c64 | s32 | u32 | u8 | b8 | s64 | u64 | s16 | u16 | f16 |
       |---------|-----|-----|-----|-----|-----|-----|----|----|-----|-----|-----|-----|-----|
       | f32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | f64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | c32     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | c64     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
       | s32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
       | u32     | x   | x   | x   | x   | x   | x   |    |    | x   | x   |     |     | x   |
       | u8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
       | b8      | x   | x   | x   | x   | x   | x   | x  | x  | x   | x   | x   | x   | x   |
       | s64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
       | u64     | x   | x   | x   | x   |     |     |    |    | x   | x   |     |     | x   |
       | s16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
       | u16     | x   | x   | x   | x   | x   | x   |    |    | x   | x   | x   | x   | x   |
       | f16     | x   | x   | x   | x   |     |     |    |    |     |     |     |     | x   |
```

PR #3040, opened 2020-11-02
JIT optimization: Faster generation of an unique funcName

willyborn:
The performance of JIT commands are improved up to 2x, dependent on the length of the JIT tree.
The difference is more pronounced for cached JIT kernels.

Description
-----------
The unique funcName is used to identify a JIT tree combination.  This name is generated, even when the corresponding kernel is cached.
The usage of std::string and eliminating of unnecessary formatting is a few times faster than the std::streamstring (which is very slow at construction).

Since some formatting is changed, the resulting KER number will be different resulting in a once recompilation of the kernel and a new cache file.

* Is this a new feature or a bug fix? NO, Performance improvement
* More detail if necessary to describe all commits in pull request: Each Node type has its virtual function to generate this part of the name, so they are all updated.
* Why these changes are necessary: With the improved GPU speed, the bottleneck is shifting to the CPU.  The GPU utilization increases.
* Potential impact on specific hardware, software or backends: The faster the connected GPU, the bigger the speed improvement.
* New functions and their functionality: None
* Future changes not implemented in this PR: A new PR will be launched to improve the caching/hashing mechanism for all kernels.

Changes to Users
----------------
No changes, except on the clean-up of the caching directory.
Perhaps a note necessary in the documentation??

Checklist
---------
- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/common/jit/BufferNodeBase.hpp
+++ b/src/backend/common/jit/BufferNodeBase.hpp
@@ -53,11 +53,11 @@ class BufferNodeBase : public common::Node {
         return m_linear_buffer && same_dims;
     }
 
-    void genKerName(std::stringstream &kerStream,
+    void genKerName(std::string &kerString,
                     const common::Node_ids &ids) const final {
-        kerStream << "_" << getNameStr();
-        kerStream << std::setw(3) << std::setfill('0') << std::dec << ids.id
-                  << std::dec;
+        kerString += '_';
+        kerString += getNameStr();
+        kerString += std::to_string(ids.id);
     }
 
     void genParams(std::stringstream &kerStream, int id,
--- a/src/backend/common/jit/BufferNodeBase.hpp
+++ b/src/backend/common/jit/BufferNodeBase.hpp
@@ -57,6 +57,7 @@ class BufferNodeBase : public common::Node {
                     const common::Node_ids &ids) const final {
         kerString += '_';
         kerString += getNameStr();
+        kerString += ',';
         kerString += std::to_string(ids.id);
     }
 
--- a/src/backend/common/jit/NaryNode.hpp
+++ b/src/backend/common/jit/NaryNode.hpp
@@ -68,7 +68,7 @@ class NaryNode : public Node {
                     const common::Node_ids &ids) const final {
         // Make the dec representation of enum part of the Kernel name
         kerString += '_';
-        kerString += m_op_str;
+        kerString += std::to_string(m_op);
         kerString += ',';
         for (int i = 0; i < m_num_children; i++) {
             kerString += std::to_string(ids.child_ids[i]);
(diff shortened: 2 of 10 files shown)


src/backend/common/jit/BufferNodeBase.hpp
```diff
@@ -53,11 +53,11 @@ class BufferNodeBase : public common::Node {
         return m_linear_buffer && same_dims;
     }
 
-    void genKerName(std::stringstream &kerStream,
+    void genKerName(std::string &kerString,
                     const common::Node_ids &ids) const final {
-        kerStream << "_" << getNameStr();
-        kerStream << std::setw(3) << std::setfill('0') << std::dec << ids.id
-                  << std::dec;
+        kerString += '_';
```
umar456, line 58: The filled in zeros are required to avoid naming conflicts. It is used to distinguish between 1, 3 and 13.

umar456 (changes requested):
I can see why this would be faster. I suspect its the reserve function that improves the performance of this approach over the string stream approach. You will need to add the formatting of the IDs back because it will cause a naming conflict with some kernels.

willyborn:
umar,

The primary function of this string, is to generate an unique name for this
JIT combination.  Padding with zeros or adding commas serve the same
purpose, although the comma's are much faster because it is a static
operation.

All IDs (Numbers) are separated by a comma (inside a node) or underscore
(start of node).
- NaryNode has a comma separating all the IDs.  (This line is not in your
snippet)
- BufferNodeBase is only 1 ID.
All the Nodes are starting with an underline, also separating possible
numbers there.

Perhaps getNameStr could start or end with a number, I will add an extra
separator here as well when concatenated with an ID, just to be sure
though I did not encounter such a case in the current code.
------

I will check an alternative, if making the stringstream static will give
the same improvement, since we do not have any construction then.  I have
read that the construction of stringstream (due to the locale) is the
reason.

BR,
Willy

willyborn:
Umar,

Please find an overview of my findings attached in the spreadsheet.
The string# version corresponds with the JIToverhead PR (3040)

The generated unique function names are the following:
- master: L_s__float000_00100000001_000000000002_002001002003
- string#: L_s_float,0_1,0,0,1_0,0,0,2_2,1,2,3
The same information is in both strings, only the IDs from the same node
are separated by commas which avoids the risk that numbers concatenate.
The advantage of the latter, is that the formatting is static and no
calculation is wasted.

I also included 2 versions based on stringstream.
1. have the stringstream as thread_local so that the construction happens
only once.  The stringstream is initialized before each function.
2. same as 1 + only static formatting.
Both stringstream versions result in similar performance as the master.

As a conclusion, I propose the string# version (with ID number) because
this is the fastest and has the same reliability as the current solution.


PS: The PR is updated to correspond with the string# version.

BR,
Willy Born

umar456 (approved):
That sounds reasonable. I had missed the comma in the first run through your code. Do you still have your performance benchmarks for this change?

willyborn:
I do not understand your question.
The performance numbers (for all combinations) are in the attached
spreadsheet.

umar456:
This is great! Thanks for your contribution.

9prady9:
I am little late to the party, simple but efficient improvement, thanks @willyborn

PR #3068, opened 2020-12-17
CL_DEVICE_HALF_FP_CONFIG returns CL_INVALID_VALUE

willyborn:
CL_DEVICE_HALF_FP_CONFIG only exists when the extension cl_khr_fp16 is available.

Description
-----------
Instead of testing on the CONFIGURATION, it is better to directly test if the extensions (fp16 & fp64) are available so that no exceptions are thrown.

While updating the file, I added explicit conversions to eliminate compiler warnings, so that later real warnings jump in the eye.

* Is this a new feature or a bug fix?  bug
* Why these changes are necessary:  Without this correction, OCLGRIND HALTS reporting the error due to incompatibility with standard.
* Potential impact on specific hardware, software or backends: Most older devices, that do not support fp16.
* New functions and their functionality: None
* Can this PR be backported to older versions? Yes
* Future changes not implemented in this PR: None

Changes to Users
----------------
No changes to end-users.

Checklist
---------
- [X] Code compiles
- [X] Tests pass
~~- [ ] Functions added to unified API~~
~~- [ ] Functions documented~~

### DIFF
--- a/src/backend/opencl/platform.cpp
+++ b/src/backend/opencl/platform.cpp
@@ -184,6 +184,7 @@ string getDeviceInfo() noexcept {
             nDevices++;
         }
     } catch (const AfError& err) {
+        UNUSED(err);
         info << "No platforms found.\n";
         // Don't throw an exception here. Info should pass even if the system
         // doesn't have the correct drivers installed.
@@ -215,8 +216,9 @@ int getDeviceCount() noexcept try {
     DeviceManager& devMngr = DeviceManager::getInstance();
 
     common::lock_guard_t lock(devMngr.deviceMutex);
-    return devMngr.mQueues.size();
+    return static_cast<int>(devMngr.mQueues.size());
 } catch (const AfError& err) {
+    UNUSED(err);
     // If device manager threw an error then return 0 because no platforms
     // were found
     return 0;
@@ -233,7 +235,7 @@ int getDeviceIdFromNativeId(cl_device_id id) {
 
     common::lock_guard_t lock(devMngr.deviceMutex);
 
-    int nDevices = devMngr.mDevices.size();
+    int nDevices = static_cast<int>(devMngr.mDevices.size());
     int devId    = 0;
     for (devId = 0; devId < nDevices; ++devId) {
         if (id == devMngr.mDevices[devId]->operator()()) { break; }
@@ -359,8 +361,9 @@ bool isDoubleSupported(unsigned device) {
         common::lock_guard_t lock(devMngr.deviceMutex);
         dev = *devMngr.mDevices[device];
     }
-
-    return (dev.getInfo<CL_DEVICE_DOUBLE_FP_CONFIG>() > 0);
+    // 64bit fp is an optional extension
+    return (dev.getInfo<CL_DEVICE_EXTENSIONS>().find("cl_khr_fp64") !=
+            string::npos);
 }
 
 bool isHalfSupported(unsigned device) {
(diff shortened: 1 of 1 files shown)


src/backend/opencl/platform.cpp
```diff
@@ -507,7 +499,7 @@ void setDeviceContext(cl_device_id dev, cl_context ctx) {
 
     common::lock_guard_t lock(devMngr.deviceMutex);
 
-    const int dCount = devMngr.mDevices.size();
+    const size_t dCount = devMngr.mDevices.size();
     for (int i = 0; i < dCount; ++i) {
```
9prady9, line 503:
```suggestion
    for (size_t i = 0; i < dCount; ++i) {
```

9prady9: that would avoid the warning of signed and unsigned type comparison

src/backend/opencl/platform.cpp
```diff
@@ -529,7 +521,7 @@ void removeDeviceContext(cl_device_id dev, cl_context ctx) {
     {
         common::lock_guard_t lock(devMngr.deviceMutex);
 
-        const int dCount = devMngr.mDevices.size();
+        const size_t dCount = devMngr.mDevices.size();
         for (int i = 0; i < dCount; ++i) {
```
9prady9, line 525:
```suggestion
        for (size_t i = 0; i < dCount; ++i) {
```

src/backend/opencl/platform.cpp
```diff
@@ -233,13 +235,13 @@ int getDeviceIdFromNativeId(cl_device_id id) {
...
-    int nDevices = devMngr.mDevices.size();
-    int devId    = 0;
+    size_t nDevices = devMngr.mDevices.size();
+    size_t devId    = 0;
     for (devId = 0; devId < nDevices; ++devId) {
         if (id == devMngr.mDevices[devId]->operator()()) { break; }
     }
 
-    return devId;
+    return static_cast<int>(devId);
```
9prady9, line 244: nit pick: Perhaps just do `static_cast` on return of `mDevices.size()` itself, that way no need alter other lines.

src/backend/opencl/platform.cpp
```diff
@@ -359,8 +361,9 @@ bool isDoubleSupported(unsigned device) {
         common::lock_guard_t lock(devMngr.deviceMutex);
         dev = *devMngr.mDevices[device];
     }
-
-    return (dev.getInfo<CL_DEVICE_DOUBLE_FP_CONFIG>() > 0);
+    // 64bit fp is an optional extension
+    return (dev.getInfo<CL_DEVICE_EXTENSIONS>().find("cl_khr_fp64") !=
+            string::npos);
```
9prady9, line 366: 👍🏾 

Although I wonder if the vendor implementation caches the device type support in internal cache. If cached in such manner, then fetching that compared to doing a string search is more efficient.

src/backend/opencl/platform.cpp
willyborn: When trying, I get a new warning at setActiveContext(i).
As result, I decided to stay at the original int and to add a static_cast<int> at the first assignment instead.  It also solves the warning you have, and which MSVC does not report.

src/backend/opencl/platform.cpp
willyborn: Here I get now a warning at the deleteIdx assignment, which has to remain int.

As above, I will go back to the original with an extra static_cast<int>.

src/backend/opencl/platform.cpp
willyborn: Just incredible.  This MSVC reports, and nothing for the previous bigger ones.

I changed everything to int with a static_cast as suggested.

src/backend/opencl/platform.cpp
9prady9: You mean, you directly did this `const int dCount = static_cast<int>(devMngr.mDevices.size());` ?

src/backend/opencl/platform.cpp
9prady9: Perhaps you missed that line in the huge lot of warnings, to be fair - we do have quite a few warnings that needs to be addressed with MSVC compiler. We largely develop using GCC/clang, so we often miss the MSVC warnings unless we test/debug on windows for specific issues.

src/backend/opencl/platform.cpp
willyborn: I will get through the debugger step by step and report what I find.

src/backend/opencl/platform.cpp
willyborn: That's right.  Gives this a problem somewhere else?

9prady9: Nope, sure do it.

src/backend/opencl/platform.cpp
willyborn: I have the impression that it is straight copy into a vector<char>.
For NVIDIA
- Total length of vector: 453 chars
- cl_khr_fp16 NA
- cl_khr_fp64 at pos 138
For AMD
- Total length of vector: 661 chars
- cl_khr_fp16 NA
- cl_khr_fp64 at pos 0

The complete string is cached in the driver.  Copy happens in cl2.hpp at line 1427.
Although I can not see inside the function, I come to this conclusion because the exact length is provided upfront during the construction of the vector.

umar456 (approved):

Issue #3065, opened 2020-12-17
[BUG] Setting AF_CUDA_MAX_JIT_LEN to values < 2 produces SIGSEGV


cschreib-ibex:
Setting the environment variable ``AF_CUDA_MAX_JIT_LEN`` to any value less than 2 makes any ArrayFire code produce SIGSEGV (Segmentation Fault).

Description
===========
* The problem also occurs with the OpenCL backend and ``AF_OPENCL_MAX_JIT_LEN``.
* The problem does *not* occur with the CPU backend and ``AF_CPU_MAX_JIT_LEN``.
* Using a debugger, I can see the error appears to be some form of infinite recursion:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffcfa6ae179 in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
(gdb) bt
#0  0x00007ffcfa6ae179 in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
#1  0x00007ffcfa6a9b2b in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
#2  0x00007ffcfa6a9c1c in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
#3  0x00007ffcf9516e1d in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
#4  0x00007ffcf9517009 in afcuda!??5af@@YA?AVarray@0@AEBV10@AEBE@Z () from C:\Program Files\ArrayFire\v3\lib\afcuda.dll
...
```
* Did you build ArrayFire yourself or did you use the official installers: Official installers 
* Which backend is experiencing this issue: CUDA and OpenCL
* Do you have a workaround: No (besides the obvious "don't do that")
* Can the bug be reproduced reliably on your system: Yes
* A clear and concise description of what you expected to happen: I expect it to either disable the JIT entirely, or if this is not possible, at least throw an exception to signal an invalid environment variable
[...]

PR #3072, opened 2020-12-30
Fix a infinite recursion bug in NaryNode JIT Node

9prady9:
Description
-----------
When the maximum JIT tree height is two, createNaryNode goes into
infinite recursion due to missing base case check inside the function.
Fixes: #3065 

Changes to Users
----------------
Fixes the infinite recursion bug when maximum jit length is set to 2.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,9 @@ mark_as_advanced(
   AF_BUILD_FRAMEWORK
   AF_INSTALL_STANDALONE
   AF_WITH_CPUID
+  Boost_INCLUDE_DIR
   CUDA_HOST_COMPILER
+  CUDA_SDK_ROOT_DIR
   CUDA_USE_STATIC_CUDA_RUNTIME
   CUDA_rt_LIBRARY
   SPDLOG_BUILD_EXAMPLES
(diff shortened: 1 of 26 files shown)


test/jit.cpp
```diff
@@ -828,3 +828,12 @@ TEST(JIT, setKernelCacheDirectory) {
...
 }
+
+TEST(JIT, UnitTreeSize) {
+    af::array a = af::constant(1, 10);
+    af::array b = af::constant(2, 10);
+    af::array c = a * b;
+    af::array d = b * c;
+    std::cout << af::sum<double>(c) << std::endl;
+    std::cout << af::sum<double>(d) << std::endl;
+}
```
9prady9, line 839: I am going to write a better test if possible or remove this one. Accidentally left this print outs here.

9prady9: To test the bug without any environment change, we need a programmatic way of setting the maximum JIT length (per backend).

9prady9:
I think the clang-format error should be fixed once related fix from #3071 is merged. Locally I have made sure the formatting is good.

umar456 (approved):

Issue #3073, opened 2021-01-04
[BUG] Different indexing behavior for const vs non-const array


cschreib-ibex:
Depending on whether an ``af::array`` is ``const`` or not, the same indexing operation produces an array with different dimensions.

Description
===========

* I encountered this bug while refactoring some of my code. The code was a single big function, which I started breaking down into several smaller functions. I carefully marked ``const`` the function parameters that should be constant, and was surprised when this simple refactoring (which shouldn't have changed the output at all) all of a sudden made my program crash.
* Did you build ArrayFire yourself or did you use the official installers: Official installers
* Which backend is experiencing this issue? All
* Do you have a workaround? Yes (don't use ``const``, which isn't great)
* Can the bug be reproduced reliably on your system? Yes
* A clear and concise description of what you expected to happen: I expect that operations on a non-const ``af::array`` form a strict super-set of operations on a const ``af::array``. In other words, if I can do operation ``B`` a const array, I should be able to do the same operation ``B``, with the exact same output, on a non-const array.
* Run your executable with AF_TRACE=all and AF_PRINT_ERRORS=1 environment
  variables set, screenshot or terminal output of the results:
```
[platform][1609772142][007696] [ ..\src\backend\common\DependencyModule.cpp(99) ] Attempting to load: forge.dll
[platform][1609772142][007696] [ ..\src\backend\common\DependencyModule.cpp(102) ] Found: forge.dll
[platform][1609772142][007696] [ ..\src\backend\cuda\device_manager.cpp(428) ] CUDA Driver supports up to CUDA 10.1 ArrayFire CUDA Runtime 10.1
[platform][1609772142][007696] [ ..\src\backend\cuda\device_manager.cpp(495) ] Found 1 CUDA devices
[platform][1609772142][007696] [ ..\src\backend\cuda\device_manager.cpp(521) ] Found device: GeForce GTX 1070 (8 GB | ~6256.71 GFLOPs | 15 SMs)
[platform][1609772142][007696] [ ..\src\backend\cuda\device_manager.cpp(556) ] AF_CUDA_DEFAULT_DEVICE:
[platform][1609772142][007696] [ ..\src\backend\cuda\device_manager.cpp(575) ] Default device: 0(GeForce GTX 1070)
[mem][1609772143][007696] [ ..\src\backend\cuda\memory.cpp(158) ] nativeAlloc:    1 KB 0xb04e00000
```

Reproducible Code and/or Steps
------------------------------
```c++
[...]

PR #3078, opened 2021-01-13
Fix const array indexing inside gfor

9prady9:
Description
-----------
Indexing a const array inside GFOR produced results which aren't identical to results generated from indexing a non-const array. 
Fixes: #3073 

Changes to Users
----------------
Identical behavior when indexing const vs non-const array inside gfor.

Checklist
---------
- [x] Code compiles
- [ ] Tests pass - `test_inverse_dense_opencl` seems to fail after this change, looking into it why, although the specific failing test seems to run fine if run via filter argument to the test program.
- ~[ ] Functions documented~


9prady9:
Doesn't seem like `inverse_dense` failure on my local machine is related to this change. I think somehow `cpu_randu` helper is causing it because if I don't use it and use GPU function to generate data for inverse_dense, the tests pass in every run just fine.

umar456 (approved):

PR #3088, opened 2021-02-11
Fix cmake arguments for external projects for msvc

9prady9:
Without the additional toolset argument being forwarded to msvc
toolchain, cmake/msvc is free to choose a toolset as per their
respective logic. This causes build issues.

Also fixed some conditions that are based on CMakeBuildType variable -
not recommended to use checks based on that variable and often resulted
in issues when used with untested multi-config generators.

Changes to Users
----------------
None unless it is an user building ArrayFire on Windows platform.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,15 +9,6 @@ cmake_minimum_required(VERSION 3.5)
 
 project(ArrayFire VERSION 3.8.0 LANGUAGES C CXX)
 
-if(NOT CMAKE_VERSION VERSION_LESS 3.9)
-  get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -9,6 +9,15 @@ cmake_minimum_required(VERSION 3.5)
 
 project(ArrayFire VERSION 3.8.0 LANGUAGES C CXX)
 
+if(NOT CMAKE_VERSION VERSION_LESS 3.9)
```
umar456, line 12: Can you move this to InternalUtils? I don't I want software requirements to be more prominent here and move this type of logic in to the other file.

umar456 (approved):

PR #3106, opened 2021-03-09
Fix glad add_subdirectory to fix out of tree builds

umar456:
This was a problem on the arrayfire-benchmark repo where the repository
is built as a subproject

Description
-----------
Add the binary directory to the add_subdirectory command that adds the glad repository.
This was causing errors with the arrayfire-benchmark repository when building master.

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ if(NOT LAPACK_FOUND)
     endif()
 endif()
 
-add_subdirectory(${${glad_prefix}_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


syurkevi (approved):

9prady9 (approved):
Won't change anything. I must have excluded the explicit binary directory since the glad target is an interface and doesn't get built directly.

PR #3119, opened 2021-04-08
vcpkg manifest file for ease of development; support to use as many vcpkg ports as possible

9prady9:
**Needs https://github.com/microsoft/vcpkg/pull/17099 to be merged so that ArrayFire's manifest baseline is correct**

Description
-----------
    Developers can now invoke cmake as shown below to install dependencies automatically when using vcpkg and cmake.
    
    ```cmake
    cmake .. -DVCPKG_ROOT:PATH=<path to vcpkg repository tool>
    ```
    or
    ```cmake
    export VCPKG_ROOT=<path to vcpkg repository tool>
    cmake ..
    ```
    
    One may add `-DAF_BUILD_CUDA:BOOL=ON` command line argument to enable CUDA dependency check.
    Even if not provided, ArrayFire will silently check for CUDA and enable the backend if available.
    
    There are couple of caveats though for the following dependencies
    - cuda
    - cudnn
    - intel-mkl
    
    As these libraries have complex installation mechanisms, their respective vcpkg dependency is merely a check
    for user. They have to be installed using respective vendor provided installers.
    
    A few important notes regarding using vcpg manifest file:
    
    1. For linux developers, currently full support for only Intel MKL compute backend is availalbe.
    2. As x64-linux triplet creates static builds only as of now, forge cannot be part of vcpkg dependency list
       on non windows platforms. Nevertheless, the user doesn't need to do anything as fetchcontent workflow is
       the fallback.
    3. vcpkg manifest is for development puporses only and isn't intended to be production ready dependency
       management for arrayfire as there are dependencies that don't get built with vcpkg at all.

Changes to Users
----------------
ArrayFire development using vcpkg becomes a better experience. I am going to send in a another PR with cmake presets which should further improve development experience of ArrayFire.

Checklist
---------
- [x] Code compiles
- [x ] Tests pass
- ~[ ] Functions documented~


9prady9:
@jacobkahn Please let me know your thoughts on the manifest file content.

jacobkahn:
@9prady9 we spoke offline a bit about this -- the manifest looks good to me though!

9prady9:
- [x] Local windows and linux build completed successfully
- [x] Tests on Arch Linux except the expected ones passed
- [ ] Windows tests are in progress
- [ ] Test using cmake 3.5.1

umar456 (approved):

PR #3139, opened 2021-05-25
Use system/vcpkg spdlog when available otherwise fallback to fetchcontent

9prady9:
Description
-----------
This change enables using system/vcpkg spdlog when available otherwise fallback to fetchcontent workflow. I have also udpated the spdlog version to 1.8.5 from the archaic 1.0.0 version. Fortunately, there were no issues like we encountered with fix releases during the version update of spdlog.

Apart from this, I have corrected the extern arrayfire deps download location; This is just a folder organization change, didn't fix any build issue.

Changes to Users
----------------
None

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9:
Turns out the issue is with the dependency of spdlog, fmt whose latest version on vcpkg has compilation issues. I am working on a workaround. Will update the PR once it is ready

9prady9:
the work around seemed to work on local windows setup, I wonder why it failed on CI after the change - I will look into the windows problem.

9prady9:
> 
> 
> the work around seemed to work on local windows setup, I wonder why it failed on CI after the change - I will look into the windows problem.

Never mind, I think this is due to a100 system being blocked by me for running another benchmark. I will shut that down now that I don't need it further.

CMakeLists.txt
```diff
@@ -247,7 +254,11 @@ else()
                              ${ArrayFire_SOURCE_DIR}/include
                              ${ArrayFire_BINARY_DIR}/include
                              ${ArrayFire_SOURCE_DIR}/src/backend)
-  target_link_libraries(bin2cpp PRIVATE spdlog)
+  if(TARGET spdlog::spdlog_header_only)
```
umar456, line 257: af_spdlog?

Also why is bin2cpp using spdlog anyway. It doesn't look like it uses it.

9prady9: bin2cpp uses `common/util.hpp` and that indirectly causes this I think.

umar456 (approved):

PR #3141, opened 2021-06-01
The compare function should return false, for equal elements.

willyborn:
When compiling in debug mode, the MSVC compiler returns an non-compliance error.
The provided method should simulate the '<' function, not '<='.

Description
-----------
bug fix, could result in increase sort time or even incorrect results.

Changes to Users
----------------
None

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/cpu/kernel/sift.hpp
+++ b/src/backend/cpu/kernel/sift.hpp
@@ -91,7 +91,7 @@ bool feat_cmp(feat_t i, feat_t j) {
         if (i.f[k] != j.f[k]) return (i.f[k] < j.f[k]);
     if (i.l != j.l) return (i.l < j.l);
 
-    return true;
+    return false;
 }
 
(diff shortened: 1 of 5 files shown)


9prady9 (changes requested):
I think this should also be changed https://github.com/arrayfire/arrayfire/blob/master/test/sift.cpp#L49

Can you also please change the following as well
https://github.com/arrayfire/arrayfire/blob/master/test/orb.cpp#L48
https://github.com/arrayfire/arrayfire/blob/master/test/orb.cpp#L48

9prady9:
@willyborn there is a merge commit in the PR commit history, please remove it.

willyborn:
I found an extra one in topk

9prady9 (approved):

PR #3151, opened 2021-06-22
Free unlocked buffers before tests run in rng quality tests

9prady9:
Description
-----------

This is needed when running rng quality tests on lesser memory cards
where higher memory usage is causing out of memory issues.

Changes to Users
----------------
None

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/test/rng_quality.cpp
+++ b/test/rng_quality.cpp
@@ -18,7 +18,7 @@ template<typename T>
 class RandomEngine : public ::testing::Test {
    public:
     virtual void SetUp() {
-        //Ensure all unlocked buffers are freed
+        // Ensure all unlocked buffers are freed
         deviceGC();
     }
 };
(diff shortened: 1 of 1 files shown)


test/rng_quality.cpp
```diff
@@ -18,7 +18,7 @@ template<typename T>
 class RandomEngine : public ::testing::Test {
    public:
     virtual void SetUp() {
-        //Ensure all unlocked buffers are freed
+        // Ensure all unlocked buffers are freed
         deviceGC();
```
willyborn, line 22: If you want to unlock all buffers, you need to add af::sync() before the deviceGC().
Without the sync, the buffers which are still in use by executing kernels will not be freed.

9prady9: That is not the case with `deviceGC` in this location. When test finishes, all arrays involved are already evaluated because they need to be brought to host for comparison. So, they will be marked as unused by the time this line hits before the next test starts.

umar456: Additionally all allocation and free function are device synchronization events.

PR #3159, opened 2021-07-20
Turn OFF cmake fetchcontent update step

9prady9:
Description
-----------

This was ON earlier which necessiated the need of an internet connection
while re-running cmake configure command after an initial successful
run.

This change eliminates the cloud connection requirement after the first
successful cmake configure run. Note, that changing certain options that
use fetchcontent workflow while there is no cloud connection may result
in failed cmake configure step due to change in options.

Changes to Users
----------------
None. Helps devs in a narrow use case where development is done on no network setup after running cmake configure.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9:
Closing this PR in favor of a better change that I am going raise towards end of this week. It improves offline build experience in multiple scenarios.

On another note: I wonder why mtx dataset download on github actions is only failing. I have verified that the data source is online and that is not the issue. CI jobs and local jobs are running successfully as expected. I will look into this once the other PR is up and if similar behavior repeats on that PR as well.

PR #3162, opened 2021-07-27
Improve offline build experience for developers

9prady9:
Description
-----------
The following common scenarios(majority we think) are covered with this change.
- Developer has cloud connection always.
- Developer has cloud connection for initial cmake run, but not later.
- Developer has lost cloud connection for a while after the initial successful cmake run but regained the connection later.
- Developer is doing an completely disconnected build using the source tarball we generate and attach to our release assets.

When the developer wants to do an offline build for a specific commit other than release tags, they would have to generate the relevant source tarball themselves. The commands required to do the same can be found from the following ci workflow file in our repository.
```
.github/workflows/release_src_artifact.yml
```

The source tarball generation CI job has also been changed to reflect the change in external dependencies location.

Changes to Users
----------------
None to users, but Offline development made more easy for developers.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/.github/workflows/release_src_artifact.yml
+++ b/.github/workflows/release_src_artifact.yml
@@ -70,7 +70,14 @@ jobs:
                   done
                   shopt -u extglob
                   rm -rf matrixmarket
-                  cd ../../..
+                  cp -r ./* ../../extern/
+                  cd ..
+                  wget https://github.com/arrayfire/forge/releases/download/v1.0.8/forge-full-1.0.8.tar.bz2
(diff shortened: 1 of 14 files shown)


9prady9:
Lets wait for https://github.com/microsoft/vcpkg/pull/19166 to be merged so that arrayfire master doesn't break for developers who use vcpkg manifest file.

umar456 (approved):

PR #3167, opened 2021-08-06
Fix reference count if array used in JIT operations.

umar456:
Better manage the reference count of an Array<T> object when it is used in a JIT operation

Description
-----------
Previously when an af::array was used in a jit operation and it was backed by a
buffer, a buffer node was created and the internal shared_ptr was stored in the
Array for future use and returned when getNode was called. This increased the
reference count of the internal buffer. This reference count never decreased
because of the internal reference to the shared_ptr.

This commit changes this behavior by storing a weak_ptr instead of a
shared_ptr inernally to the Array<T> object. This change reduces the
reference count once the Array<T> is evaluated in an expression. If
the weak_ptr can be converted into a

Changes to Users
----------------
Enjoy lower reference counts. This may increase performance under certain scenarios

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/src/api/c/median.cpp
+++ b/src/api/c/median.cpp
@@ -8,7 +8,7 @@
  ********************************************************/
 
 #include <backend.hpp>
-#include <cast.hpp>
+#include <common/cast.hpp>
 #include <common/err_common.hpp>
 #include <handle.hpp>
 #include <math.hpp>
@@ -36,7 +36,7 @@ static double median(const af_array& in) {
 
     af_array temp = 0;
     AF_CHECK(af_moddims(&temp, in, 1, dims.get()));
-    const Array<T> input = getArray<T>(temp);
+    const Array<T>& input = getArray<T>(temp);
 
     // Shortcut cases for 1 or 2 elements
     if (nElems == 1) {
--- a/src/backend/common/jit/BufferNodeBase.hpp
+++ b/src/backend/common/jit/BufferNodeBase.hpp
@@ -12,8 +12,6 @@
 #include <common/jit/Node.hpp>
 #include <jit/kernel_generators.hpp>
 
-#include <iomanip>
-#include <mutex>
 #include <sstream>
 
 namespace common {
@@ -24,25 +22,20 @@ class BufferNodeBase : public common::Node {
     DataType m_data;
     ParamType m_param;
     unsigned m_bytes;
-    std::once_flag m_set_data_flag;
     bool m_linear_buffer;
 
    public:
-    BufferNodeBase(af::dtype type) : Node(type, 0, {}) {
-        // This class is not movable because of std::once_flag
-    }
+    BufferNodeBase(af::dtype type)
+        : Node(type, 0, {}), m_bytes(0), m_linear_buffer(true) {}
 
     bool isBuffer() const final { return true; }
 
     void setData(ParamType param, DataType data, const unsigned bytes,
                  bool is_linear) {
-        std::call_once(m_set_data_flag,
-                       [this, param, data, bytes, is_linear]() {
-                           m_param         = param;
-                           m_data          = data;
-                           m_bytes         = bytes;
-                           m_linear_buffer = is_linear;
-                       });
+        m_param         = param;
+        m_data          = data;
+        m_bytes         = bytes;
+        m_linear_buffer = is_linear;
     }
 
     bool isLinear(dim_t dims[4]) const final {
--- a/src/backend/common/jit/BufferNodeBase.hpp
+++ b/src/backend/common/jit/BufferNodeBase.hpp
@@ -92,6 +92,29 @@ class BufferNodeBase : public common::Node {
     }
 
     size_t getBytes() const final { return m_bytes; }
+
+    size_t getHash() const noexcept {
+        size_t out = 0;
+        auto ptr   = m_data.get();
+        memcpy(&out, &ptr, std::max(sizeof(Node *), sizeof(size_t)));
+        return out;
+    }
+
+    /// Compares two BufferNodeBase objects for equality
+    bool operator==(
+        const BufferNodeBase<DataType, ParamType> &other) const noexcept;
+
+    /// Overloads the equality operator to call comparisons between Buffer
+    /// objects. Calls the BufferNodeBase equality operator if the other
+    /// object is also a Buffer Node
+    bool operator==(const common::Node &other) const noexcept final {
+        if (other.isBuffer()) {
+            return *this ==
+                   static_cast<const BufferNodeBase<DataType, ParamType> &>(
+                       other);
+        }
+        return false;
+    }
 };
 
 }  // namespace common
--- a/src/backend/common/jit/Node.hpp
+++ b/src/backend/common/jit/Node.hpp
@@ -15,12 +15,13 @@
 #include <types.hpp>
 #include <af/defines.h>
 
+#include <algorithm>
 #include <array>
 #include <functional>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <unordered_map>
-#include <utility>
 #include <vector>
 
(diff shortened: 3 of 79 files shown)


src/api/c/median.cpp
```diff
@@ -36,7 +36,7 @@ static double median(const af_array& in) {
 
     af_array temp = 0;
     AF_CHECK(af_moddims(&temp, in, 1, dims.get()));
-    const Array<T> input = getArray<T>(temp);
+    const Array<T>& input = getArray<T>(temp);
```
9prady9, line 39: I think we can avoid such mistakes if we use auto in locations such as these where return type is obvious and evident from the function name.

src/backend/opencl/jit/BufferNode.hpp
```diff
@@ -9,14 +9,27 @@
...
+
+template<typename DataType, typename ParamType>
+bool BufferNodeBase<DataType, ParamType>::operator==(
+    const BufferNodeBase<DataType, ParamType> &other) const noexcept {
+    if (m_data.get() == other.m_data.get()) {
+        if (m_bytes == other.m_bytes) {
+            if (m_param.offset == other.m_param.offset) { return true; }
+        }
+    }
+    return false;
```
9prady9, line 32:
```suggestion
    return m_data.get() == other.m_data.get() &&
           m_bytes == other.m_bytes &&
           m_param.offset == other.m_param.offset;
```
Perhaps both generate same binary, but isn't above version much easier to read ?

src/backend/cuda/jit/BufferNode.hpp
```diff
@@ -16,4 +16,20 @@ namespace jit {
...
+
+template<typename DataType, typename ParamType>
+bool BufferNodeBase<DataType, ParamType>::operator==(
+    const BufferNodeBase<DataType, ParamType> &other) const noexcept {
+    if (m_data.get() == other.m_data.get()) {
+        if (m_bytes == other.m_bytes) {
+            if (m_param.ptr == other.m_param.ptr) { return true; }
+        }
+    }
+    return false;
```
9prady9, line 32:
```suggestion
    return m_data.get() == other.m_data.get() &&
           m_bytes == other.m_bytes &&
           m_param.offset == other.m_param.offset;
```
I think this is easier to read and less lines

src/backend/cpu/jit/BufferNode.hpp
```diff
@@ -126,7 +126,34 @@ class BufferNode : public TNode<T> {
...
+        return false;
+    }
 };
 
+template<typename T>
+bool BufferNode<T>::operator==(const BufferNode<T> &other) const noexcept {
+    if (m_ptr == other.m_ptr) {
+        if (m_bytes == other.m_bytes) { return true; }
+    }
+    return false;
```
9prady9, line 155:
```suggestion
    return m_ptr == other.m_ptr && m_bytes && other.m_bytes;
```

test/jit.cpp
```diff
@@ -53,9 +53,11 @@ TEST(JIT, CPP_JIT_HASH) {
...
     // Creating a kernel
     {
-        array d    = a + b;
-        array e    = a + c;
-        array f1   = d * e - e;
+        array d  = a + b;
+        array e  = a + c;
+        array f1 = d * e - e;
+
+        f1.eval();
```
9prady9, line 60: why is eval needed here ?

src/backend/common/jit/Node.hpp
```diff
@@ -228,6 +248,20 @@ class Node {
...
+    /// Returns the hash of the node. For all Nodes other than the Buffer node,
+    /// this is the pointer of the object
+    virtual size_t getHash() const noexcept {
+        size_t out = 0;
+        auto ptr   = this;
+        memcpy(&out, &ptr, std::max(sizeof(Node *), sizeof(size_t)));
+        return out;
+    }
+
+    /// A very bad equality operator used only for the has function.
```
9prady9, line 261:
```suggestion
    /// A very bad equality operator used only for the hash function.
```
bad indeed :)

9prady9 (changes requested):
The following commits can be merged
- https://github.com/arrayfire/arrayfire/pull/3167/commits/3081bb8957c5b633bb3710ef82885b70b7541899
- https://github.com/arrayfire/arrayfire/pull/3167/commits/1fc0aecaac6707c3f314b78fcece708a2a056395

I have left comments in other respective files/commits.

There are some typos in Node Hash commit I think.

src/api/c/median.cpp
umar456: auto will also cause this because by default auto is by value which would result in the same behavior.

9prady9: yeah, my bad. forgot that even that needs `&` at the end.

9prady9:
Also I forgot to mention this earlier. median_cpu and canny_cpu are failing consistently across all CPU jobs on Ubuntu 20.04 now.

test/jit.cpp
umar456: It was left over from debugging. Removed.

src/backend/common/jit/Node.hpp
```diff
@@ -228,6 +248,20 @@ class Node {
...
     /// Default destructor
     virtual ~Node() noexcept = default;
+
+    /// Returns the hash of the node. For all Nodes other than the Buffer node,
+    /// this is the pointer of the object
+    virtual size_t getHash() const noexcept {
+        size_t out = 0;
+        auto ptr   = this;
+        memcpy(&out, &ptr, std::max(sizeof(Node *), sizeof(size_t)));
+        return out;
```
9prady9, line 258:
```suggestion
        constexpr size_t PtrSizeInBytes = sizeof(Node *) > sizeof(size_t) ? sizeof(Node *) : sizeof(size_t);
        size_t out = 0;
        std::memcpy(&out, &ptr, PtrSizeInBytes);
        return out;
```
This should help evaluate the copy size at compile time.

src/backend/cpu/jit/BufferNode.hpp
```diff
@@ -126,7 +126,38 @@ class BufferNode : public TNode<T> {
     }
 
     bool isBuffer() const final { return true; }
+
+    size_t getHash() const noexcept final {
+        size_t out = 0;
+        memcpy(&out, &m_ptr, std::max(sizeof(T *), sizeof(size_t)));
+        return out;
```
9prady9, line 133:
```suggestion
        constexpr size_t PtrSizeInBytes = sizeof(T *) > sizeof(size_t) ? sizeof(T *) : sizeof(size_t);
        size_t out = 0;
        memcpy(&out, &m_ptr, PtrSizeInBytes);
        return out;
```
Here also

src/backend/common/jit/BufferNodeBase.hpp
```diff
@@ -92,6 +87,29 @@ class BufferNodeBase : public common::Node {
     }
 
     size_t getBytes() const final { return m_bytes; }
+
+    size_t getHash() const noexcept {
+        size_t out = 0;
+        auto ptr   = m_data.get();
+        memcpy(&out, &ptr, std::max(sizeof(Node *), sizeof(size_t)));
+        return out;
```
9prady9, line 95:
```suggestion
        constexpr size_t PtrSizeInBytes = sizeof(Node *) > sizeof(size_t) ? sizeof(Node *) : sizeof(size_t);
        auto ptr   = m_data.get();
        size_t out = 0;
        std::memcpy(&out, &ptr, PtrSizeInBytes);
        return out;
```

9prady9 (changes requested):
Couple of more minor improvements.

9prady9 (approved):
Looks good but I have some queries about the node hashing, but we can discuss those over chat

PR #3170, opened 2021-08-17
Create ASSERT_IMAGE_NEAR which compares two images for equality

umar456:
Add an image comparison assertion to the tests that compares two images
and if there is an error, uploads the result and the gold image to CDash
for comparison. Useful for when image tests fail

Description
-----------
ASSERT_IMAGE_NEAR will compare two images for similarity and upload
the image to ci.arrayfire.org if the test fails. This allows us to quickly view
the errors by going to the site and viewing the differences to determine the
error.

This assertion works on both af::array and af_array objects. It currently
supports f32, f64, u8, and b8 images but can support additional types
fairly easily.

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -128,6 +128,7 @@ endif()
 
 target_compile_definitions(arrayfire_test
   PRIVATE
+    TEST_RESULT_IMAGE_DIR="${CMAKE_BINARY_DIR}/test/"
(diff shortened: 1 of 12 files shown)


9prady9 (approved):
looks good

PR #3182, opened 2021-10-13
kernel crashing, due to mismatch with disk cache for updated JIT code.

willyborn:
Disk caching continued loading expired JIT code, resulting in exceptions.  Clearing the disk cached, hided the problem.
For interface changes, an exception was generated.
For all other changes, the old code (initial cached) was loaded and executed.

Description
-----------
JIT code uses the hashing of the function name (based on parameters for JIT) for fast retrieval of the corresponding kernel.  The same hashing was also used to save the code on disk.  When JIT code was updated, the same function name still applies, resulting in an mismatch with a previous saved/cached version.
Static code uses the hashing based on the full code including options.

Now for JIT code, the hashing for disk caching is based on the full code including options (same as static code).  Once the kernel is saved into memory everything remains the same:
  . JIT code remains using the hashing based on function name (no code generation)
  . static code remains using the hashing based on full code + options (hashing is also static)
Result, JIT is now reacting exactly the same as the static code and the correct kernel is always loaded.

* This is a bug fix.
* No new functions, 
* This PR can be back-ported to previous versions. 

Changes to Users
----------------
* No changes to the end-user.  Clearing of the disk cache is no longer needed to eliminate the exception.

Checklist
---------
- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/common/kernel_cache.cpp
+++ b/src/backend/common/kernel_cache.cpp
@@ -73,39 +73,60 @@ Kernel getKernel(const string& kernelName,
     UNUSED(targs);
 #endif
 
-    size_t moduleKey = 0;
+    // The JIT kernel uses the hashing of the kernelName (tInstance) only to
+    // speed up to search for its cached kernel.  All the other kernels have the
+    // full source code linked in, and will hash the full code + options
+    // instead.
+    size_t moduleKeyCache = 0;
     if (sourceIsJIT) {
-        moduleKey = deterministicHash(tInstance);
+        moduleKeyCache = deterministicHash(tInstance);
     } else {
-        moduleKey = (sources.size() == 1 && sources[0].hash)
-                        ? sources[0].hash
-                        : deterministicHash(sources);
-        moduleKey = deterministicHash(options, moduleKey);
+        moduleKeyCache = (sources.size() == 1 && sources[0].hash)
+                             ? sources[0].hash
+                             : deterministicHash(sources);
+        moduleKeyCache = deterministicHash(options, moduleKeyCache);
 #if defined(AF_CUDA)
-        moduleKey = deterministicHash(tInstance, moduleKey);
+        moduleKeyCache = deterministicHash(tInstance, moduleKeyCache);
 #endif
     }
     const int device  = detail::getActiveDeviceId();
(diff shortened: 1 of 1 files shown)


9prady9:
👍🏾 It is awesome that you finally found out the reason for the JIT kernel cache issue. I am just concerned that it increases the startup cost slightly in the case of loading from disk. Did you happen to time by how much the loading time increased with this change ?

@umar456 Is there a way we can actually fix the kernel name generation of JIT ? I was under the impression that JIT kernel names encoded the some form of kernel operations info.

willyborn:
Regarding the JIT kernel names.
It does contain a description of the dynamic calling parameters, incl types etc.
It does not include, the real code or a description of the always present parameters.

As result, fixes in the code, without dynamic interface changes, do result in the same JIT kernel name.
For my opinion, I would keep this, because each time generating the code of those kernels is really time consuming and are static inside one run.  (not between runs/disk caches, since code could have changed).

umar456:
@willyborn I couldn't see the times in the log you posted. I assume its pretty small based on your comments. I can't imagine this hashing would increase the times significantly. I am curious if we can't achieve the same thing by including the AF_REVISION when storing the names. This will allow us to refresh the cache for each commit and avoid issues like this. Can you think of a reason why this would fail?

willyborn:
@umar
I first included the AF_REVISION, as proposed.
l still had to clear the cache for each test run, because this version
number is constant during development recompiles.
I assume the version nr changes after a commit in GitHub, which is not the
case during development.  On my system they never changed.

With the static kernels, I did not have this problem so I searched further
to understand why, with this PR as result.

Regarding the timings: the 1st run gives a theoretical time almost equal to
the static kernel (any run), where the combining of the hashes is always
performed.
I did not give times, because as soon as you put it inside a loop, the
extra steps are no longer executed.  The measured times, before or after
the change, are as result equal.  (I believe we are talking here over
microseconds, if not picoseconds).

PS: Only the combining is performed because the hashing on the code text is
already performed before the call, also before the change.  These hashes
are stored in common::Source struct and are present in the parameters of
the getKernel call.  (also for JIT)

PR #3199, opened 2021-12-08
reduce all -> array

syurkevi:
This PR adds af::array as a return type to the reduction functions. ( reduce<af::array>(arr); ) 

Description
Previously, the reduce all functions would return a scalar value. The return of an af::array type was missing. Furthermore, the return of a scalar causes a mandatory synchronization at the function call.

New feature
Partial commits contain separate API changes, kernels for each backend, and tests as described in each commit message. Will be squashed before merging.
To improve performance by avoiding unnecessary synchronization at each reduce<>() call.

Adds af_sum_array, af_sum_nan_array, af_product_array, af_product_nan_array, af_min_array, af_max_array, af_count_array, af_all_true_array, af_any_true_array,  and the corresponding C++ versions such as af::max<af::array>(arr);
* Can this PR be backported to older versions?
 No, breaks API

Changes to Users
----------------
Existing code should be compatible. Users can now specify <af::array> as a template argument for reduction functions.

Checklist
---------

- [x] Code compiles
- [ ] Tests pass
- [x] Functions documented

### DIFF
--- a/src/api/c/reduce.cpp
+++ b/src/api/c/reduce.cpp
@@ -10,6 +10,7 @@
 #include <backend.hpp>
 #include <common/err_common.hpp>
 #include <common/half.hpp>
+#include <copy.hpp>
 #include <handle.hpp>
 #include <ireduce.hpp>
 #include <math.hpp>
@@ -25,6 +26,7 @@ using detail::Array;
 using detail::cdouble;
 using detail::cfloat;
 using detail::createEmptyArray;
+using detail::getScalar;
 using detail::imag;
 using detail::intl;
 using detail::real;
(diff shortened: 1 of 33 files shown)


src/api/c/reduce.cpp
```diff
@@ -686,34 +762,133 @@ static af_err reduce_all_promote(double *real_val, double *imag_val,
...
 
+af_err af_sum_array(af_array *out, const af_array in) {
+    return reduce_all_promote_array<af_add_t>(out, in);
+}
+
 af_err af_product_all(double *real, double *imag, const af_array in) {
     return reduce_all_promote<af_mul_t>(real, imag, in);
 }
 
+af_err af_product_array(af_array *out, const af_array in) {
```
syurkevi, line 864: af_product_all_array()

syurkevi: TODO: all c api

umar456 (dismissed):

umar456 (approved):

jacobkahn:
Very happy to see this committed - thanks for the work @syurkevi!

PR #3204, opened 2022-01-13
fix multiprocess filename collisions in imageio tests

syurkevi:
This PR fixes test failures in imageio when using ctest.
Ctest creates multiple processes to run tests in parallel. In some cases, a collision between the filenames can occur due to parallel access and cause random failures.

Description
-----------
  Bug fix
* More detail if necessary to describe all commits in pull request
  Fix added uuids to the filenames which collided 
* Why these changes are necessary
  These changes should allow the multiple processes of ctest to work in parallel without the chance of collision.

Changes to Users
----------------
Should reduce number of failures in CI

Checklist
---------
- [x] Code compiles
- [ ] Tests pass

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -101,10 +101,6 @@ add_library(arrayfire_test OBJECT
   testHelpers.hpp
   arrayfire_test.cpp)
 
-set_target_properties(arrayfire_test
-PROPERTIES
-    CXX_STANDARD 11)
-
 target_include_directories(arrayfire_test
   PRIVATE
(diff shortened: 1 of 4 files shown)


9prady9 (commented):
I have one concern, let's say a developer works on this, there is a possibility that the build folder keeps accumulating multiple files across multiple runs and can potentially (unlikely perhaps) generate same hex 4 digit code in later runs. At which point, we will still have name collision.

The CI pipeline won't have any issue because it will be cleaned up regularly.

9prady9 (commented):
looks better I think

umar456 (approved):

PR #3205, opened 2022-01-13
Build libafcuda with dynamically loaded CUDA numeric binaries

umar456:
Dynamically link CUDA numeric libraries instead of always staticly linking.

Description
-----------
ArrayFire's CUDA backend linked against the CUDA numeric libraries
staticly before this change. This caused the libafcuda library to be
in the 1.1GB range for CUDA 11.5 even if you were targeting one compute
capability. This is partially due to the fact that the linker does not
remove the compute capabilities of older architectures when linking.

One way around this would be to use nvprune to remove the architectures
that are not being used by the compute cability when building. This
approach is not yet implemented.

This commit will revert back to dynamically linking the CUDA numeric
libraries by default. You can still select the old behavior by setting
the AF_WITH_STATIC_CUDA_NUMERIC_LIBS option in CMake

Changes to Users
----------------
Then binary sizes are significantly smaller but requires you to have
the cuda libraries in the library paths. This is only an issue when building
installers.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ option(AF_WITH_LOGGING  "Build ArrayFire with logging support" ON)
 option(AF_WITH_STACKTRACE  "Add stacktraces to the error messages." ON)
 option(AF_CACHE_KERNELS_TO_DISK "Enable caching kernels to disk" ON)
 option(AF_WITH_STATIC_MKL "Link against static Intel MKL libraries" OFF)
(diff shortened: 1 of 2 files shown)


WilliamTambellini:
@umar456  any way for afcuda to dynload such nvidia libs only if needed (especially cufft, ...)?

umar456:
It could be done but would require a good bit of work. We have done this for other libraries but not for things like cublas and cufft.

9prady9:
This is unfortunate, is this still the same with CUDA 11.6 ?

syurkevi (approved):

PR #3210, opened 2022-02-02
Deterministic topK

syurkevi:
topK would not use a stable sort, so in the case of ties results would be non-deterministic.
This PR adds new topk function types AF_TOPK_STABLE_MIN, AF_TOPK_STABLE_MAX that will perform the topk search in a deterministic manner. 

 New feature
 The CPU and openCL backends needed tweaks to their comparator functions. Now the comparators will perform an additional index comparison if the values are equal to deterministically sort ties.
The cuda backend used radix sort which is stable, however the order of loading the data would assign each granule(n items per thread) with spread out data, resulting in an output depending on grid-size. Shared memory is now used to load data to specific threads.

Cuda backend now uses more shared memory. Doesn't affect performance much since bottleneck is SMEM bandwidth in radix sort. CPU and OpenCL backends now need an extra comparison in some cases, shouldn't affect performance much.
new options are  AF_TOPK_STABLE_MIN, AF_TOPK_STABLE_MAX, which are the previous AF_TOPK_MAX | AF_TOPK_STABLE and AF_TOPK_MIN | AF_TOPK_STABLE

Changes to Users
----------------
Updated API options

Checklist
---------

- [x] Code compiles
- [ ] Tests pass
- [x] Functions documented

### DIFF
--- a/include/af/defines.h
+++ b/include/af/defines.h
@@ -508,8 +508,11 @@ typedef enum {
 } af_diffusion_eq;
 
 typedef enum {
-    AF_TOPK_MIN     = 1,  ///< Top k min values
-    AF_TOPK_MAX     = 2,  ///< Top k max values
+    AF_TOPK_MIN         = 1,  ///< Top k min values
+    AF_TOPK_MAX         = 2,  ///< Top k max values
+    AF_TOPK_STABLE      = 4,  ///< Preserve order of indices for equal values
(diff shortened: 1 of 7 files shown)


9prady9:
Rebase from latest master should fix the windows GA job failure

umar456 (approved):

umar456 (approved):

PR #3212, opened 2022-02-07
Use boost's epsilon difference when comparing floating point values

umar456:
Use epsilon difference to compare floating point values in tests

Description
-----------
This commit changes the way we compare floating point values in the
tests to use the boost math's epsilon difference to compare two floating
point values for equality. This is a more accurate form of equality and
handles differences in half float values when the values reach a
certain threshold.

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9 (approved):
looks good to me

Issue #3208, opened 2022-01-21
 release-src-artifact defaults to mkl causing failure in CI


chenrui333:
👋  it looks like `arrayfire-full-3.8.1.tar.bz2` artifact did not get uploaded for `3.8.1` release? Thanks!

syurkevi:
The release-src-artifact job is failing, still needs to be fixed, keeping issue up until that is fixed. The artifact has been uploaded though.

PR #3214, opened 2022-02-11
Forge dep install fix

syurkevi:
If forge was found in vcpkg, AF_BUILD_FORGE would be ignored. This PR fixes that behavior.
This PR also fixes:
* missing glfw3.dll during AF_BUILD_FORGE standalone installation
* ci workflows for non-intel build during src_artifact_release

Is this a new feature or a bug fix?
bug fix  

Future changes not implemented in this PR.
glfw should also be installed if forge is prebuilt, can currently be done with X_VCPKG_APPLOCAL_DEPS_INTALL, perhaps there is an elegant way to do this in the CMakeModules.
AF_COMPUTE_BACKEND should have a default fallback if intel fails

Fixes: #3208

Checklist
---------

- [x] Code compiles
- [ ] Tests pass

### DIFF
--- a/.github/workflows/release_src_artifact.yml
+++ b/.github/workflows/release_src_artifact.yml
@@ -46,7 +46,7 @@ jobs:
               run: |
                   cd ${GITHUB_WORKSPACE}/arrayfire-full-${AF_VER}
                   mkdir build && cd build
-                  cmake .. -DAF_BUILD_FORGE:BOOL=ON
+                  cmake .. -DAF_BUILD_FORGE:BOOL=ON -DAF_COMPUTE_LIBRARY="FFTW/LAPACK/BLAS"
 
             - name: Create source tarball
               id: create-src-tarball
--- a/CMakeModules/AFconfigure_forge_dep.cmake
+++ b/CMakeModules/AFconfigure_forge_dep.cmake
@@ -8,34 +8,16 @@
 set(FG_VERSION_MAJOR 1)
 set(FG_VERSION_MINOR 0)
 set(FG_VERSION_PATCH 8)
+set(FG_VERSION "${FG_VERSION_MAJOR}.${FG_VERSION_MINOR}.${FG_VERSION_PATCH}")
+set(FG_API_VERSION_CURRENT ${FG_VERSION_MAJOR}${FG_VERSION_MINOR})
 
-find_package(Forge
-  ${FG_VERSION_MAJOR}.${FG_VERSION_MINOR}.${FG_VERSION_PATCH}
-  QUIET
-)
 
-if(TARGET Forge::forge)
-  get_target_property(fg_lib_type Forge::forge TYPE)
-  if(NOT ${fg_lib_type} STREQUAL "STATIC_LIBRARY")
-      install(FILES
-          $<TARGET_FILE:Forge::forge>
-          $<$<PLATFORM_ID:Linux>:$<TARGET_SONAME_FILE:Forge::forge>>
-          $<$<PLATFORM_ID:Darwin>:$<TARGET_SONAME_FILE:Forge::forge>>
-          $<$<PLATFORM_ID:Linux>:$<TARGET_LINKER_FILE:Forge::forge>>
-          $<$<PLATFORM_ID:Darwin>:$<TARGET_LINKER_FILE:Forge::forge>>
-          DESTINATION "${AF_INSTALL_LIB_DIR}"
-          COMPONENT common_backend_dependencies)
-  endif()
-else()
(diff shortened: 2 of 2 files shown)


9prady9:
If forge was found on the system (irrespective of the mechanism of it's installation), it will be picked up and hence it is supposed to avoid using anything else. What issue is this behaviour causing ?

> glfw should also be installed if forge is prebuilt, can currently be done with X_VCPKG_APPLOCAL_DEPS_INTALL, perhaps there is an elegant way to do this in the CMakeModules.

Do you mean when forge is already installed using vcpkg ? then glfw will be already installed as it's own dependency right ?

umar456:
This was done to fix the installer mechanism which is not packaging Forge's dependencies.

9prady9:
> This was done to fix the installer mechanism which is not packaging Forge's dependencies.

Which OS ?

umar456:
Windows. @syurkevi can explain this better.

9prady9:
> Windows. @syurkevi can explain this better.

Okay, interesting because I never had to manually do anything earlier.

syurkevi:
For AF_BUILD_STANDALONE on windows, the vcpkg forge would be picked up as you're saying, however that packag d one isn't quite right and the resulting installers would be incapable of running graphics examples. If I wanted to manually rebuild forge with correct dependencies I'd be unable to since the AF_BUILD_FORGE is ignored if the vcpkg one is found.

.github/workflows/release_src_artifact.yml
```diff
@@ -46,7 +46,7 @@ jobs:
               run: |
                   cd ${GITHUB_WORKSPACE}/arrayfire-full-${AF_VER}
                   mkdir build && cd build
-                  cmake .. -DAF_BUILD_FORGE:BOOL=ON
+                  cmake .. -DAF_BUILD_FORGE:BOOL=ON -DAF_COMPUTE_LIBRARY="FFTW/LAPACK/BLAS"
```
9prady9, line 49: 👍🏾

CMakeModules/AFconfigure_forge_dep.cmake
```diff
@@ -67,20 +49,43 @@ else()
     set(FETCHCONTENT_QUIET ${af_FETCHCONTENT_QUIET})
     set(FETCHCONTENT_FULLY_DISCONNECTED ${af_FETCHCONTENT_FULLY_DISCONNECTED})
     set(FETCHCONTENT_UPDATES_DISCONNECTED ${af_FETCHCONTENT_UPDATES_DISCONNECTED})
-
     install(FILES
         $<TARGET_FILE:forge>
+        $<TARGET_RUNTIME_DLLS:forge>
```
9prady9, line 54: Unfortunately, `TARGET_RUNTIME_DLLS`  is only available from cmake 3.21

9prady9 (dismissed):

CMakeModules/AFconfigure_forge_dep.cmake
```diff
@@ -51,13 +51,21 @@ if(AF_BUILD_FORGE)
...
-        $<TARGET_RUNTIME_DLLS:forge>
         $<$<PLATFORM_ID:Linux>:$<TARGET_SONAME_FILE:forge>>
         $<$<PLATFORM_ID:Darwin>:$<TARGET_SONAME_FILE:forge>>
         $<$<PLATFORM_ID:Linux>:$<TARGET_LINKER_FILE:forge>>
         $<$<PLATFORM_ID:Darwin>:$<TARGET_LINKER_FILE:forge>>
         DESTINATION "${AF_INSTALL_LIB_DIR}"
         COMPONENT common_backend_dependencies)
+
+    if(AF_INSTALL_STANDALONE)
+        cmake_minimum_required(VERSION 3.21)
```
9prady9, line 62: Anyone who wants to do an stand alone installation would need newer cmake ? Is that the intended behaviour going forward ?

umar456: Yeah, I think that is an acceptable requirement. I think we are the only people interested in that functionality.

WilliamTambellini: We are too @umar456  
tks.

umar456: @WilliamTambellini Is this a difficult requirement for you? Also I don't think you are interested in the forge dependency. This is only required if you are interested in forge and want to include forge as part of the install.

WilliamTambellini: no, we are cool on it: we can just 'make install' and copy manually the libs/includes we are into.

umar456 (approved):

PR #3230, opened 2022-03-10
Allow multiple root nodes in passJitHeuristics function

umar456:
Allow multiple nodes to be passed into passJitHeuristics. This allows the ability to catch errors associated with large kernels when constructing the JIT kernels.

Description
-----------
* Fixes errors associated with kernels with multiple output
* Add span-lite header which allows you to use C++20 std::span in earlier versions of C++

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


9prady9:
@umar456 We can avoid adding such a large file to the code base using `FetchContent`. It seems  https://github.com/martinmoene/span-lite is header only and is on github too. It should be easy to using FetchContent

umar456:
@9prady9 Yeah, that is a good idea. I have updated the repo with your suggestion.

9prady9 (approved):
LGTM did I miss something or am I correct if I say the actual multiple root scenario isn't actually used anywhere yet ? I see NaryNode uses it but it didn't look like the `N` was anything other than `1` anywhere.

PR #3240, opened 2022-03-28
Make cuSparse a runtime dependency

umar456:
Load cuSparse at runtime instead of at link time.

Description
-----------
This PR makes cuSparse a runtime dependency. You do not need the library
to the installed on the system to run ArrayFire code. You only need to have it
in the load path when you use any of the cusparse functions.

Changes to Users
----------------
* This could reduce the install size of the CUDA backend

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,17 +85,21 @@ if(MKL_FOUND)
   set(default_compute_library "Intel-MKL")
 endif()
 
-set(AF_COMPUTE_LIBRARY ${default_compute_library}
-    CACHE STRING "Compute library for signal processing and linear algebra routines")
-set_property(CACHE AF_COMPUTE_LIBRARY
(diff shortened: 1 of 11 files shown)


WilliamTambellini:
@umar456  we like libafcuda.so to be standalone so could you keep a cmake option to link statically with cusparse ?

umar456:
Are you using sparse arrays? I didn't think you were. I will have to think about how this would work with this approach.

WilliamTambellini:
yes, we are using af sparse, meaning cusparse when using afcuda.

umar456:
@WilliamTambellini I have updated the PR to allow static linking if you set the AF_cusparse_LINK_LOADING to "Static" when you run CMake. By default we will load cuSparse at runtime.

syurkevi (approved):

Issue #3256, opened 2022-05-17
fp16 norm is not supported


ktdq:
Half precision float is not supported in `af_norm` : ArrayFire Error (204) : Function does not support this data type

PR #3258, opened 2022-06-10
Add f16 support for modulus and norm

umar456:
Add support for f16 for modulus and norm. This PR also adds tests for the norm function

Description
-----------
* Adds support for modulus operation for half type. This operation is missing in the CUDA backend because it is not supported by the CUDA math library. We are now implementing this function in software
* Add support for norm for half type.
* Add tests for norm
* We now use f<op>f functions that accept float types as inputs in CUDA kernels.
* Add tests for additional binary operations

Fixes: #3256 

Changes to Users
----------------
* Users can now call modulus and norm from user code

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


test/norm.cpp
```diff
@@ -0,0 +1,183 @@
+/*******************************************************
+ * Copyright (c) 2014, ArrayFire
```
syurkevi, line 2: 2022

test/norm.cpp
```diff
@@ -0,0 +1,183 @@
...
+    norm_params param = GetParam();
+    if (get<1>(param) == f16) SUPPORTED_TYPE_CHECK(half_float::half);
+    if (get<1>(param) == f64) SUPPORTED_TYPE_CHECK(double);
+    array in        = af::identity(get<0>(param), get<1>(param));
+    double result   = norm(in, AF_NORM_MATRIX_INF);
+    double norm_inf = cpu_norm_inf(in);
+
+    ASSERT_DOUBLE_EQ(norm_inf, result);
+}
+
```
syurkevi, line 172: vector norm?

syurkevi (dismissed):

src/backend/cpu/binary.hpp
```diff
@@ -88,8 +88,11 @@ LOGIC_CPLX_FN(double, af_or_t, ||)
 
 template<typename T>
 static T __mod(T lhs, T rhs) {
-    T res = lhs % rhs;
```
edwinsolisf, line 91: This was introduced a while ago: check commit [81bd8840fcf385f380d304fecbafcb45baf388a7](https://github.com/arrayfire/arrayfire/commit/81bd8840fcf385f380d304fecbafcb45baf388a7). It seems that here that the distinction between `rem` (remainder) and `mod` (modulus) was introduced and is still mentioned in the [ArrayFire wiki](https://arrayfire.org/docs/group__arith__func__rem.htm#gsc.tab=0). However, CUDA and OpenCL define `mod` and `rem`. See [`src/backend/cuda/kernel/jit.cuh:59`](src/backend/cuda/kernel/jit.cuh:59) and [`src/backend/opencl/jit.cl:107`](src/backend/opencl/jit.cl:107).

From the current state of things, the procedure should be:
1) Make this change to be consistent with the other backends.
2) Update the documentation
3) Possibly deprecate `rem`?

edwinsolisf:
For bookkeeping and future reference regarding the change of `mod` in the cpu backend and the history of the changes.

The original intent of `af::mod` was to produce only positive values while `af::rem` to produce the same sign as the dividend(left handside) input. However, that distinction was only made on the `cpu` backend while not implemented on the others. This can be seen in the state of `rem` and `mod` in the `cuda` and `opencl` backends. See [`src/backend/cuda/kernel/jit.cuh:59`](https://github.com/arrayfire/arrayfire/tree/master/src/backend/cuda/kernel/jit.cuh#L59) and [`src/backend/opencl/kernel/jit.cl:107`](https://github.com/arrayfire/arrayfire/tree/master/src/backend/opencl/kernel/jit.cl#L107).

There was intent to implement it in other backends but the changes never went through. An example of this is can be seen in the code in [src/backend/cuda/jit.cuh:75](https://github.com/arrayfire/arrayfire/blob/553c38d4c2cb6e6943dfa053168085b57f106346/src/backend/cuda/kernel/jit.cuh#L75) introduced in this [commit]
(https://github.com/arrayfire/arrayfire/blame/553c38d4c2cb6e6943dfa053168085b57f106346/src/backend/cuda/kernel/jit.cuh#L75) were fremf is implemented as using `std::remainder` (which technically does not  always produce an output with the same sign, that function is `fmodf`); however, it is not used as the implementation of `af::rem`.

The current decision reflected by the code is that to keep compatibility and consistency between all backends `af::mod` and `af::rem` will now produce the same output such that it isconsistent with what `%` and `fmod` produce in `c++/cuda`. This means both will give a result that is the remainder and has the same sign as the dividend (left handside input).

christophe-murphy (approved):

PR #3284, opened 2022-09-01
Better support for system installed dependencies

umar456:
This PR allows ArrayFire to use system installed packages for some dependencies.

Description
-----------
Allow the use of system installed packages for some dependencies that ArrayFire uses. This PR is useful for package managers like debian package managers, spack and vcpkg. This approach allows for smaller binaries and faster build times. The following packages can be found by the current build:

* fmt
* spdlog
* CLBlast
* Google Test
* Boost
* span-lite

The following packages will still need to be downloaded:

* CLFFT: We have a couple of changes that have not been incorporated in the main repo
* af_threads: There are no packages for this library. It is used and maintained by ArrayFire
* 

This PR introduces the following CMake variables

* AF_WITH_EXTERNAL_PACKAGES_ONLY: When true CMake will fail if the packages are not found. If false, if packages are not found, CMake will attempt to download the package using FetchContent
* AF_WITH_SPDLOG_HEADER_ONLY: Use the header only version of spdlog. This reduces the number of link time dependencies but increases build times
* AF_WITH_FMT_HEADER_ONLY: Use the header onlly version of fmt

Additionally this PR also fixes a couple of CMake warning by updating the FindOpenMP.cmake and FindLAPACKE.cmake files.

Changes to Users
----------------
Users must make sure that the installed packages are compatible. If packages are not found then the current behavior will fall back to downloading repositories using fetch content. Users can use AF_WITH_EXTERNAL_PACKAGES_ONLY CMake variable to prevent the downloading of some of the packages. Some internal and special forks will still be downloaded even when this option is set to true.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


syurkevi (approved):

src/backend/common/util.cpp
```diff
@@ -162,25 +173,26 @@ bool removeFile(const string& path) {
...
 
     const string testPath = path + AF_PATH_SEPARATOR + "test";
-    if (!std::ofstream(testPath).is_open()) { return false; }
+    if (!ofstream(testPath).is_open()) { return false; }
     removeFile(testPath);
 
     return true;
 }
 
+#ifndef NOSPDLOG
```
syurkevi, line 189: double negative? let's update this if we touch other backends

Issue #3293, opened 2022-09-19
[BUG] Problem with non us Locales and OpenCL wrapper.


GuillaumeSchmid:
There is a problem when the local is non C when rendering the arguments to OpenCL kernels.

Description
===========
The problem occurred when I integrated arrayfire through the rust wrapper to my app using Tauri for user interface.
Some OpenCL refused to compile when Tauri was initialized in the application. And worked fine when Tauri init was commented out.

The problem is that Tauri initializes somewhere the locale for the whole software, and it impacts how arrayfire gegerates parameter strings provided to the OpenCL kernels.

The French locale uses "," instead of "." to render floats decimal part, but there might also be some other local that pause problems (such as using the "," to separate thousands).

The output should use the "C" locale.

I am working on a PR, I have a working version.

System Information
------------------
I am on Linux Ubuntu 22.04 using MKL + OpenCL ( with nvidia drivers ).

~ > uname -a
Linux merelam 5.15.0-47-generic #51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
using 
* latest oneapi MKL 2022
* nvidia NVIDIA-SMI 515.65.01    Driver Version: 515.65.01    CUDA Version: 11.7
* Arrayfire 3.8.2
* rust-arrayfire 3.8.0
*

PR #3294, opened 2022-09-19
This patch fixes bug #3293 - OpenCL kernel argument problem with some locales

GuillaumeSchmid:
**Short Description:**

There is a problem when the locale is non C when rendering the arguments to OpenCL kernels.

**Short description of the change:**

The patch is pretty simple, it changes the toString() function to use the stringstream imbued with C locale in src/backend/common/TemplateArg.cpp and changed the to_string calls to this toString function in types.cpp.

**Description**

The arguments provided to OpenCL uses the C++ standard library function std::to_string(). This function uses the locale to render it's argument to a string.

It is a problem when arrayfire is used in a
software initialised with non "C" locale.

For instance, on a French computer, to_string(1.0) will output the string "1,0000000".
This string is provided to OpenCL kernels, generating a syntax error.

The most portable way to fix this problem is to use a local ostringstream imbued withe "C" locale.

An Other way would be to use C++17 to_chars function, as it only renders it argument with "C" locale, without impact from the application or system locale.
I did not use this latest solution as it would force the use of C++17.

- [ yes] Rebased on latest master
- [yes ] Code compiles
- [yes] Tests pass
- [no ] Functions added to unified API
- [no ] Functions documented


umar456 (approved):

umar456:
@GuillaumeSchmid I made some changes to your PR to address a couple of other issues related to the structure of the codebase. Could you give these changes another try to see if addresses the issue you are seeing. I was able to incorporate the to_chars feature suggested in your comment.

GuillaumeSchmid:
@umar456 I tested your implementation with to_char on my code, it works.

umar456 (approved):

PR #3296, opened 2022-09-22
std::cout << "hello OneAPI" << std::endl;

syurkevi:
Adds a OneAPI backend to ArrayFire. This PR is a starting point for further development and collaboration.

Description
-----------

New feature. New backend!
Lots of commits. Commits should probably be squashed. Enough scaffolding is in place to start porting kernels.
This should allow performant support for upcoming Intel hardware.
This new backend may eventually need to utilize existing JIT implementations. Not sure if changes will be necessary if so.
Sooooo many kernels need to be ported.

Changes to Users
----------------
Current source should remain unaffected when building. If new backend is targeted the icx compiler will be necessary with a command similar to:
```
cmake -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icx  ..
```

Checklist
---------

- [x] Code compiles
- [ ] Tests pass
- [ ] Functions documented


CMakeModules/InternalUtils.cmake
```diff
@@ -122,7 +122,7 @@ macro(arrayfire_set_cmake_default_variables)
   set(CMAKE_PREFIX_PATH "${ArrayFire_BINARY_DIR};${CMAKE_PREFIX_PATH}")
   set(BUILD_SHARED_LIBS ON)
 
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
```
9prady9, line 125: nice

include/af/oneapi.h
```diff
@@ -0,0 +1,440 @@
...
+#include <af/array.h>
+#include <af/dim4.hpp>
+#include <af/exception.h>
+#include <af/device.h>
+#include <stdio.h>
+
+namespace afoneapi
+{
+
+#if 0
```
9prady9, line 165: Ah, I missed this one in the first look :)

src/backend/common/EventBase.hpp
```diff
@@ -36,7 +36,8 @@ class EventBase {
 
     /// \brief Event destructor. Calls the destroy event call on the native API
     ~EventBase() noexcept {
-        if (e_) NativeEventPolicy::destroyEvent(&e_);
+      //if (e_)
```
9prady9, line 39: why leave and comment it out ?

src/backend/common/Logger.hpp
```diff
@@ -13,8 +13,40 @@
...
+/* GNU GCC/G++ */
+#pragma GCC diagnostic pop
+#elif defined(_MSC_VER)
+/* Microsoft Visual Studio */
+#pragma warning(pop)
+#else
+/* Other */
+#endif
+
+
```
9prady9, line 49: This entire pragma based suppression for spdlog is better clubbed in separate header which can be resued from multiple locations in the code base which wants to use spdlog.

src/backend/common/forge_loader.hpp
```diff
@@ -10,10 +10,39 @@
...
+#elif defined(__GNUC__) || defined(__GNUG__)
+/* GNU GCC/G++ */
+#pragma GCC diagnostic pop
+#elif defined(_MSC_VER)
+/* Microsoft Visual Studio */
+#pragma warning(pop)
+#else
+/* Other */
+#endif
+
```
9prady9, line 45: Perhaps we can do the same to glad header too (like spdlog)

9prady9 (commented):
Interesting, is this a copy of opencl backend with few fixes and commented out kernel launches ?

src/backend/common/forge_loader.hpp
syurkevi: There's only two places it happens, perhaps a macro could work 🤔

umar456 (dismissed):

umar456 (approved):

PR #3312, opened 2022-10-17
Fix the backend id encoding for the oneAPI backend

umar456:
The way we were formatting the backend ID was incorrect and failed when we had
more than 3 backends. With the new oneAPI backend, this mechanism was failing
and causing errors.

Description
-----------
* Fix the way we encode the beckend ID in ArrayInfo
* Remove the use of vectors in the getKernel API and use span instead. This allows the use of std::array which are allocated on the stack instead of the heap.
* Remove an extraneous print in the oneAPI memcpy kernel

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/src/backend/common/ArrayInfo.cpp
+++ b/src/backend/common/ArrayInfo.cpp
@@ -38,26 +38,18 @@ unsigned ArrayInfo::getDevId() const {
 }
 
 void ArrayInfo::setId(int id) const {
-    // 1 << (backendId + 8) sets the 9th, 10th or 11th bit of devId to 1
-    // for CPU, CUDA and OpenCL respectively
-    // See ArrayInfo.hpp for more
-    unsigned backendId =
-        detail::getBackend() >> 1U;  // Convert enums 1, 2, 4 to ints 0, 1, 2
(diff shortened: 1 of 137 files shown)


src/backend/common/ArrayInfo.cpp
```diff
@@ -41,18 +41,15 @@ void ArrayInfo::setId(int id) const {
...
     // for CPU, CUDA and OpenCL respectively
     // See ArrayInfo.hpp for more
-    unsigned backendId =
-        detail::getBackend() >> 1U;  // Convert enums 1, 2, 4 to ints 0, 1, 2
-    const_cast<ArrayInfo *>(this)->setId(id | 1 << (backendId + 8U));
+    const_cast<ArrayInfo *>(this)->setId(id);
 }
 
 void ArrayInfo::setId(int id) {
     // 1 << (backendId + 8) sets the 9th, 10th or 11th bit of devId to 1
```
syurkevi, line 48: set 12th bit for oneapi

syurkevi (approved):

PR #3332, opened 2022-11-17
reorder oneapi port

pv-pterab-s:
reorder ported to oneapi. all tests pass. horribly, this only works with an auxilliary debug buffer for a memset(..,0). otherwise, reorder fails %50 of the time with a segfault. this works, so it is a valid port, but work will continue to determine the source of the failure

Checklist
---------

- [X] Code compiles
- [X] Tests pass
- [ ] Functions documented


src/backend/oneapi/kernel/reorder.hpp
```diff
@@ -0,0 +1,142 @@
...
+    const int d1_;
+    const int d2_;
+    const int d3_;
+    const int blocksPerMatX_;
+    const int blocksPerMatY_;
+    sycl::accessor<float, 2, sycl::access::mode::write> debugAcc_;
+};
+
+#include "/home/gpryor/new-dev/io.hpp"
+#include "/home/gpryor/new-dev/msg.hpp"
```
umar456, line 111:
```suggestion
```
I think you want to remove these

syurkevi:
I've noticed in previous kernels that having an early return statement such as:
```
if((xx < op_.dims[0] && yy < op_.dims[1] && oz < op_.dims[2] &&
            ow < op_.dims[3]) { 
    return; //weird behavior including crashing
}
//vs
  bool valid = (xx < op_.dims[0] && yy < op_.dims[1] && oz < op_.dims[2] &&
            ow < op_.dims[3]);
  // if(valid) dowrite[i] = someval;
```
will cause the kernels to crash. I searched the sycl spec for this behavior and couldn't find if this is explicitly disallowed. Using a conditional boolean instead fixes the mystery crashes that the "debug accessor" hack was avoiding... not to say this isn't a hack... will need to investigate if the compiler is wrong or the spec.

syurkevi (approved):

PR #3334, opened 2022-11-22
Compiler flag adjustments for Intel Compiler

umar456:
The Intel compiler uses fast math by default in order to perform some optimization. These
optimizations causes some of our tests to fail. This PR handles several of the issues 
brought up with the new compiler

Description
-----------
* Add AF_WITH_FAST_MATH to options which enable the use of fast math flags in all compilers. This is off by default
* Fix several warnings with clang and icx
* Refactor GitHub actions so that the builds only execute if clang-format and docs pass
* Add additional compilers to GitHub actions

Changes to Users
----------------
* Users can now enable AF_WITH_FAST_MATH for additional performance at the cost of precision

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~


src/backend/common/TemplateArg.hpp
```diff
@@ -16,7 +16,7 @@
 #include <utility>
 
 template<typename T>
-class TemplateTypename;
+struct TemplateTypename;
```
9prady9, line 19: good catch :)

test/testHelpers.hpp
```diff
@@ -223,9 +223,17 @@ bool noDoubleTests(af::dtype ty);
...
+#define SUPPORTED_TYPE_CHECK(type)                                \
+    if (noDoubleTests((af_dtype)af::dtype_traits<type>::af_type)) \
+        GTEST_SKIP() << "Device doesn't support Doubles";         \
+    if (noHalfTests((af_dtype)af::dtype_traits<type>::af_type))   \
+        GTEST_SKIP() << "Device doesn't support Half";
+
+#ifdef AF_WITH_FAST_MATH
+#define NOFASTMATH() GTEST_SKIP() << "ArrayFire compiled with AF_WITH_FAST_MATH"
+#else
+#define NOFASTMATH()
```
9prady9, line 235: I would say `SKIP_TEST_IF_FAST_MATH_ENABLED` is more clear on what the macro accomplishes. It is a bit long but clear I think.

9prady9 (commented):
did a quick pass, lgtm

umar456:
Addressed all feedback.

syurkevi (approved):

PR #3337, opened 2022-11-25
More fast math

umar456:
Fix tests and improve the AF_WITH_FAST_MATH option in OpenCL and CUDA

Description
-----------
* Pass fast math flag to nvcc
* Pass fast math flag to nvrtc
* Pass fast math flag to OpenCL
* Fix min and max function when AF_WITH_FAST_MATH is set because infinity doesn't exist with fast math
* Set cuBLAS math mode to CUBLAS_TF32_TENSOR_OP_MATH when AF_WITH_FAST_MATH is set

Changes to Users
----------------
N/A

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~[ ] Functions documented~

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -636,6 +636,7 @@ endif()
 
 target_compile_options(afcuda
   PRIVATE
+    $<$<BOOL:${AF_WITH_FAST_MATH}>:$<$<COMPILE_LANGUAGE:CUDA>:-use_fast_math>>
     $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr>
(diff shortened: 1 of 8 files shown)


9prady9 (approved):
do we have any fast mode specific tests ?

umar456:
We don't have fast math specific tests. I am not sure how we could write one. We can lower precision but I don't think that will be helpful. A lot of tests still fail with these changes. I am not sure if we can fix all of them.

WilliamTambellini:
Perhaps you could add some fastmath specific refs values using the macro :
https://stackoverflow.com/questions/43140581/how-to-use-the-gcc-preprocessor-macro-fast-math
But what I ve seen is that results with fastmath depend on many vars (os, cpu arch, compiler, ...)

PR #3344, opened 2022-12-08
tile oneapi port

pv-pterab-s:
tile oneapi port. passes up to jit. double not supported

Checklist
---------
- [X] Code compiles
- [ ] Tests pass
- [ ] Functions documented


src/backend/oneapi/kernel/tile.hpp
```diff
@@ -0,0 +1,110 @@
...
+    auto local = sycl::range(TX, TY);
+
+    int blocksPerMatX = divup(out.info.dims[0], TILEX);
+    int blocksPerMatY = divup(out.info.dims[1], TILEY);
+    auto global       = sycl::range(local[0] * blocksPerMatX * out.info.dims[2],
+                                    local[1] * blocksPerMatY * out.info.dims[3]);
+
+    try {
+        getQueue().submit([&](auto &h) {
+            sycl::accessor d_out{*out.data, h};
```
syurkevi, line 99: These should be with sycl::access::mode semantics

src/backend/oneapi/tile.cpp
```diff
@@ -25,8 +25,9 @@ Array<T> tile(const Array<T> &in, const af::dim4 &tileDims) {
 
     Array<T> out = createEmptyArray<T>(oDims);
 
-    ONEAPI_NOT_SUPPORTED("tile Not supported");
-    // kernel::tile<T>(out, in);
+    if constexpr(!(std::is_same_v<T, double> || std::is_same_v<T, cdouble>)) {
```
syurkevi, line 28: Let's wait for the resolution of #3342 to determine the best approach here

src/backend/oneapi/kernel/tile.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <common/kernel_cache.hpp>
+#include <debug_oneapi.hpp>
+
+#include <string>
+#include <vector>
+
+namespace oneapi {
+namespace kernel {
+
+constexpr sycl::specialization_id<bool> compilingAsDouble;
```
syurkevi, line 23: leftover from debug? don't see this used

src/backend/oneapi/tile.cpp
pv-pterab-s: umar confirmed this approach. without it, arrayfire outright crashes

src/backend/oneapi/kernel/tile.hpp
pv-pterab-s: good catch

src/backend/oneapi/kernel/tile.hpp
pv-pterab-s: done

pv-pterab-s:
dropping

PR #3352, opened 2023-01-01
Add support for CUDA 12

### LINKED ISSUE #3353: [BUG] CU_CHECK does not check for validity of CUDA messages extracted from error code
The `CU_CHECK` macro, used to report errors from CUDA, uses CUDA functions to convert the error code into human-readable messages. The macro currently does not check if the conversion was successful, and uses the returned `const char*` blindly. This causes crashes when the CUDA environment is badly setup (bad drivers, incorrect initialization, etc.) and the basic error message functions do not succeed.

Description
===========
I arrived there by a convoluted path; I still don't know if this bug report is related to my original problem. But here is the full story.

I have been trying to run a Windows application built with AF, with AF built from source against CUDA 11.8. The desktop on which I run the application has drivers that are below the recommended driver version for CUDA 11.8 (which is 522.6, and I am running 512.15). I expected the application to either run, or error gracefully. But instead it hard-crashed with an access violation inside ArrayFire. 

I have not been able to reproduce this with a simple application, only with our full product. I also wasn't able to debug the ArrayFire build, somehow the `.pdb` don't load. But I noticed that the hard crash disappeared if I cleared my ArrayFire kernel cache on disk. This suggested to me that the crash happened somewhere in the code that loads the kernels from disk. There isn't much code there, so I focused on the main CUDA call:
https://github.com/arrayfire/arrayfire/blob/a230ef46c27588cffe204bb5c465dba30bc08cdd/src/backend/cuda/compile_module.cpp#L458

Looking at the implementation of `CU_CHECK`:
https://github.com/arrayfire/arrayfire/blob/a230ef46c27588cffe204bb5c465dba30bc08cdd/src/backend/cuda/cu_check_macro.hpp#L18-L30

I noticed that there is no error checking on the calls to `cuGetErrorName()` and `cuGetErrorString()`. The documentation of these functions says that they can error only if the provided error code is invalid. So in principle, unless CUDA itself is returning invalid error codes, this should be alright. However the CUDA documentation appears to be inaccurate here.

Indeed, I experimented with this, and managed to trigger an error in a simple CUDA application outside of ArrayFire, by trying to run `cuModuleLoadData()` without having previously called `cuInit()`. In this case, when trying to convert the returned error code from `cuModuleLoadData()`, both `cuGetErrorName()` and `cuGetErrorString()` returned `CUDA_ERROR_NOT_INITIALIZED`, and actually didn't set the output messages to `nullptr`. If used in the `CU_CHECK` macro, this means the strings are left uninitialised, and passed to `snprintf`, which could cause a hard-crash.

I suspect this could be what is happening in my larger application. It may not be, but at any rate, the existing code should be improved to include all the necessary error checking. 

Suggestion:
```c++
    CUresult res1 = cuGetErrorName(res, &cu_err_name);
    CUresult res2 = cuGetErrorString(res, &cu_err_string);
    if (res1 == CUDA_SUCCESS && res2 == CUDA_SUCCESS)
    {
        snprintf(cu_err_msg, sizeof(cu_err_msg), "CU Error %s(%d): %s\n", cu_err_name, (int)(res), cu_err_string);
    }
    else
    { 
        snprintf(cu_err_msg, sizeof(cu_err_msg), "CU Error UNKNOWN(%d): unknown error\n", (int)(res)); 
    }
```

Reproducible Code and/or Steps
------------------------------
 - Build AF from source against CUDA 11.8
 - Run application with older drivers, 512.15.

System Information
------------------
Please provide the following information:
1. ArrayFire version: 3.8.2
2. Devices installed on the system: Nvidia GeForce GTX 1070 8GB
3. (optional) Output from the af::info() function if applicable. 
```
ArrayFire v3.8.2 (CUDA, 64-bit Windows, build 5752f2dc)
Platform: CUDA Runtime 11.8, Driver: 11060
[0] NVIDIA GeForce GTX 1070, 8192 MB, CUDA Compute 6.1
```
5. Output from the following scripts:
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 512.15       Driver Version: 512.15       CUDA Version: 11.6     |
|-------------------------------+----------------------+----------------------+
| GPU  Name            TCC/WDDM | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ... WDDM  | 00000000:65:00.0 Off |                  N/A |
|  0%   43C    P8    13W / 151W |   1921MiB /  8192MiB |      3%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
```

Checklist
---------

- [x] Using the latest available ArrayFire release
- [ ] GPU drivers are up to date

umar456:
Changes to support CUDA 12

Description
-----------
* Remove deprecated cuSPARSE API
* Move all internal namespaces in the `arrayfire` namespace
* Add Version class to support multiple versions of loaded libraries
* Add support for sparse arrays in `ASSERT_ARRAYS_*` test macros

Fixes #3353 #3354 

Changes to Users
----------------
* Compile ArrayFire with CUDA 12
* Currently CUDA 10 support is removed because of C++17 changes.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented


test/arrayfire_test.cpp
```diff
@@ -1390,6 +1391,116 @@ INSTANTIATE(long long);
...
+
+template<typename T>
+std::ostream &operator<<(std::ostream &os, const sparseCooValue<T> &val) {
+    os << "(" << val.row << ", " << val.col << "): " << val.value;
+    return os;
+}
+
+template<typename T>
+bool isZero(const sparseCooValue<T> &val) {
+    return val.value == 0.;
```
syurkevi, line 1428: For future reference when updating tests for sparse solve, improve comparison w/float to within threshold

test/arrayfire_test.cpp
```diff
@@ -1597,6 +1796,67 @@ ::testing::AssertionResult elemWiseEq(std::string aName, std::string bName,
...
+
+    if (aItr == a.end()) {
+        return ::testing::AssertionSuccess();
+    } else {
+        dim_t idx       = std::distance(b.begin(), bItr);
+        af::dim4 coords = unravelIdx(idx, bDims, calcStrides(bDims));
+
+        af::dim4 aStrides = calcStrides(aDims);
+
+        //::testing::AssertionResult result = ::testing::AssertionFailure() <<
```
syurkevi, line 1837: //TODO:remove

syurkevi (dismissed):

syurkevi (approved):

PR #3361, opened 2023-02-02
memcopy

pv-pterab-s:
memcopy implemented to support reshape and finally fft. very dirty implementation including all potential input types, but will simplify with jit replacement when we get to it.

Checklist
---------

- [X] Code compiles
- [ ] Tests pass
- [ ] Functions documented


pv-pterab-s:
This pull request did not work out. Doubles seem to creep in causing errors on a770. When run on straight host, etc, OpenCL barfs with unknown symbols in the "link" at runtime.

Issue #3369, opened 2023-02-25
field.cpp example on Arch linux breaks on all backends


bitstormFA:
building field.cpp in the examples and executing field_cuda, field_cpu, field_opencl results in errors
Seems to be caused by the join operation in line 41
```array hbowl = join(1, constant(1, x.elements()), flat(bvals));```

Description
===========
After installing ArrayFire 3.8.3 from the community package in Arch linux I did check the examples from the corresponding github release tar. All executables for the field.cpp example produce errors

* Additional details regarding the bug
field_cpu breaks with 
```
ArrayFire v3.8.3 (CPU, 64-bit Linux, build default)
malloc(): invalid size (unsorted)
[1]    113747 IOT instruction (core dumped)  ./field_cpu
```
field_cuda breaks with
```
ArrayFire v3.8.3 (CUDA, 64-bit Linux, build default)
Platform: CUDA Runtime 11.8, Driver: 525.89.02
[0] NVIDIA GeForce RTX 4090, 24151 MB, CUDA Compute 8.9
ArrayFire Exception (Internal error:998):
In function void arrayfire::cuda::join(Array<T>&, int, const std::vector<Array<T> >&) [with T = float]
In file src/backend/cuda/join.cpp:177
CUDA Error (1): invalid argument

 0# 0x00007F7D85C3161F in /usr/lib/libafcuda.so.3
 1# 0x00007F7D86475888 in /usr/lib/libafcuda.so.3
 2# 0x00007F7D8646C8AB in /usr/lib/libafcuda.so.3
 3# af::Window::vectorField(af::array const&, af::array const&, char const*) in /usr/lib/libafcuda.so.3
 4# main in ./field_cuda
 5# 0x00007F7D8363C790 in /usr/lib/libc.so.6
 6# __libc_start_main in /usr/lib/libc.so.6
 7# _start in ./field_cuda

In function void af::Window::vectorField(const af::array&, const af::array&, const char*)
In file src/api/cpp/graphics.cpp:135
terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Internal error:998):
In function void arrayfire::cuda::join(Array<T>&, int, const std::vector<Array<T> >&) [with T = float]
In file src/backend/cuda/join.cpp:177
CUDA Error (1): invalid argument

 0# 0x00007F7D85C3161F in /usr/lib/libafcuda.so.3
 1# 0x00007F7D86475888 in /usr/lib/libafcuda.so.3
[...]

PR #3375, opened 2023-03-14
Corrected field example (#3369)

willyborn:
Uninitialized buffer used to collect partial join (2D -> 3D)
Join now accepts any buffer as out array, as long as it is large enough.

Description
-----------
* This is a BUG fix reported in issue 3369
* The array provided to join was insufficient dimensioned, resulting in buffer overflow and wrong results.
* This can be back-ported to 3.8.3 (where the issue is first reported)
* No new functions added, only corrected.

Fixes: #3369

Changes to Users
----------------
Non

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [-] Functions added to unified API
- [-] Functions documented

### DIFF
--- a/examples/graphics/field.cpp
+++ b/examples/graphics/field.cpp
@@ -22,7 +22,7 @@ int main(int, char**) {
         af::info();
         af::Window myWindow(1024, 1024, "2D Vector Field example: ArrayFire");
 
-        myWindow.grid(1, 2);
+        myWindow.grid(2, 2);
 
         array dataRange = seq(MINIMUM, MAXIMUM, STEP);
 
@@ -38,12 +38,21 @@ int main(int, char**) {
             array saddle = join(1, flat(x), -1.0f * flat(y));
 
             array bvals = sin(scale * (x * x + y * y));
-            array hbowl = join(1, constant(1, x.elements()), flat(bvals));
+            array hbowl = join(1, constant(1., x.elements()), flat(bvals));
             hbowl.eval();
 
+            // 2D points
             myWindow(0, 0).vectorField(points, saddle, "Saddle point");
             myWindow(0, 1).vectorField(
                 points, hbowl, "hilly bowl (in a loop with varying amplitude)");
+
+            // 2D coordinates
+            myWindow(1, 0).vectorField(2.0 * flat(x), flat(y), flat(x),
+                                       -flat(y), "Saddle point");
+            myWindow(1, 1).vectorField(
+                2.0 * flat(x), flat(y), constant(1., x.elements()), flat(bvals),
+                "hilly bowl (in a loop with varying amplitude)");
+
             myWindow.show();
 
             scale -= 0.0010f;
--- a/src/api/c/vector_field.cpp
+++ b/src/api/c/vector_field.cpp
@@ -50,20 +50,21 @@ fg_chart setup_vector_field(fg_window window, const vector<af_array>& points,
     vector<Array<T>> pnts;
     vector<Array<T>> dirs;
 
-    for (unsigned i = 0; i < points.size(); ++i) {
-        pnts.push_back(getArray<T>(points[i]));
-        dirs.push_back(getArray<T>(directions[i]));
+    Array<T> pIn = getArray<T>(points[0]);
+    Array<T> dIn = getArray<T>(directions[0]);
+    if (points.size() > 1) {
+        for (unsigned i = 0; i < points.size(); ++i) {
+            pnts.push_back(getArray<T>(points[i]));
+            dirs.push_back(getArray<T>(directions[i]));
+        }
+
+        // Join for set up vector
+        const dim4 odims(pIn.dims()[0], points.size());
+        pIn = createEmptyArray<T>(odims);
+        dIn = createEmptyArray<T>(odims);
+        detail::join<T>(pIn, 1, pnts);
+        detail::join<T>(dIn, 1, dirs);
     }
-
-    // Join for set up vector
-    dim4 odims(3, points.size());
-    Array<T> out_pnts = createEmptyArray<T>(odims);
-    Array<T> out_dirs = createEmptyArray<T>(odims);
-    detail::join(out_pnts, 1, pnts);
-    detail::join(out_dirs, 1, dirs);
-    Array<T> pIn = out_pnts;
-    Array<T> dIn = out_dirs;
-
     // do transpose if required
     if (transpose_) {
         pIn = transpose<T>(pIn, false);


src/api/c/vector_field.cpp
```diff
@@ -50,25 +51,23 @@ fg_chart setup_vector_field(fg_window window, const vector<af_array>& points,
...
-    // do transpose if required
-    if (transpose_) {
-        pIn = transpose<T>(pIn, false);
-        dIn = transpose<T>(dIn, false);
-    }
+    dim4 odims(transpose_ ? ndots : 3, transpose_ ? 3 : ndots);
+    // initialize the full out_ptns & out_dirs arrays, because join of 2D points
+    // will only partially fill those arrays
+    Array<T> out_pnts = createValueArray<T>(odims, 0.);
+    Array<T> out_dirs = createValueArray<T>(odims, 0.);
```
9prady9, line 66: Since dimensions are similar, I don't see the need for `createValueArray`, it adds an unnecessary mem set call.

src/api/c/vector_field.cpp
```diff
@@ -50,25 +51,23 @@ fg_chart setup_vector_field(fg_window window, const vector<af_array>& points,
...
         dirs.push_back(getArray<T>(directions[i]));
+        ndots += getInfo(points[i]).dims().dims[0];
     }
 
     // Join for set up vector
-    dim4 odims(3, points.size());
-    Array<T> out_pnts = createEmptyArray<T>(odims);
-    Array<T> out_dirs = createEmptyArray<T>(odims);
-    detail::join(out_pnts, 1, pnts);
-    detail::join(out_dirs, 1, dirs);
```
9prady9, line 63: > Uninitialized buffer used to collect partial join (2D -> 3D)

It has been a while I looked at this section. Can you please elaborate since you have been looking at this recently. The dimensions of the arrays are identical here and join writes to output arrays, overwriting uninitialized values.

src/api/c/vector_field.cpp
willyborn: I had the same thought when looking into the problem, although will debugging I discovered the unexpected usage.

We have 4 entry scenario's (and I use the quantities of field example)
1. transpose=true
   a) pnts.dims()=[1156,2,1,1], dirs.dims()=[1156,2,1,1]  --> odims.T()=[3,1156,1,1]  --> field example
   b) pnts.dims()=[1156,3,1,1], dirs.dims()=[1156,3,1,1]  --> odims.T()=[3,1156,1,1]
2. transpose=false
   a) pnts.dims()=[2,1156,1,1], dirs.dims()=[2,1156,1,1]  --> odims=[3,1156,1,1]
   b) pnts.dims()=[3,1156,1,1], dirs.dims()=[3,1156,1,1]  --> odims=[3,1156,1,1]
PS: For multiple array points, the output array will by [3,sum of all points of all input arrays, 1,1]

For 2D inputs + transponse, the join will copy X & Y and leaving the Z uninitialized.
   input array storage (X11,X12,...,X1n, Y11,Y12,...,Y1n) & (X21,X22,...,X2, Y21,Y22,...,Y2n) & ...
       --> odims      (X11,X12,...X1n,X21,X22,..X2n,... , Y11,Y12,..Y1n,Y21,Y22,..Y2n,..., ?,?,?,?...)  !!Z-axis uninitialized!!
       --> odims.T() (X11,Y11,?, X12,Y12,?, ...)
For 3D inputs the full result array will be written.


As alternative to pre-initializing the full output array, we can also add the constant array of 0.  I did not do so, because I wanted to remain as close to possible to the original code.


PS: The original calculation of odims took the number of arrays iso the total number of points.  This could generate a memory overflow on the GPU buffer.

willyborn: Your previous statement, did make me doubt so I returned to a very old version of arrayfire 3.7.2, where I saw that the conversion to 3D wasn't done neither although It displayed a better result.

Finally, only the dimensions were wrong, resulting in a possible buffer overrun on the GPU.

So I willy push a new version, with lesser changes although with better results.

src/api/c/vector_field.cpp
willyborn: Sorry for the late response, I missed this question.

Depending on the request for transpose or not, the dimensions are different.
With transpose == true --> [N,1,1] + [N,1,1] + [N,1,1] --> [N,3,1]
With transpose == false --> [1,N,1] + [1,N,1] + [1,N,1] --> [3,N,1]
Remark: the order of the elements in the data buffer is the same for both options.  The dimensions are however different.

In the old implementation of the join methods, the dimensions were not verified with the risk of "out of buffer" writes, with unpredictable results as a consequence.  In many cases the reserve of an oversized buffer is used (unnoticed) or the next array in memory is overwritten (unpredictable effect).

christophe-murphy:
@9prady9 , I'll pick up this review if you have no objections

christophe-murphy:
@willyborn I'm happy with the changes you have made to the setup_vector_field routine. The field example is working with that fix. I see that you have made other changes to the join method to check the dimensions and handle the case where the stride is not 1 in the first dimension. I think there is more work to be done on the other versions of the join routine to fully support this. There should probably also be additional tests added to the join test suite for this. I suggest that we just merge the setup_vector_field changes here and open a separate issue for the join method changes.

willyborn:
OK.
I have to resync to the latest env. so this could take some time looking at
the recent build problems, some experience.

Willy

examples/graphics/field.cpp
```diff
@@ -38,12 +38,21 @@ int main(int, char**) {
...
+            array hbowl = join(1, constant(1., x.elements()), flat(bvals));
             hbowl.eval();
 
+            // 2D points
             myWindow(0, 0).vectorField(points, saddle, "Saddle point");
             myWindow(0, 1).vectorField(
                 points, hbowl, "hilly bowl (in a loop with varying amplitude)");
+
+            // 2D coordinates
+            myWindow(1, 0).vectorField(2.0 * flat(x), flat(y), flat(x),
```
christophe-murphy, line 50: What are these extra plots for?

willyborn:
vectorField is a graphical function, so it is not included in the test
series.  The only way to test it, is by using it in examples.

The are 3 versions of vectorField, dependent on the provision of the
coordinates
1. full coordinates in 1 array
2. X & Y axis as separate arrays
3. X, Y & Z axis as separate arrays

The corrected error is the part that combines all those different
provisions (join) into 1 common set that is forwarded to forge to display.

So in the example I added the 2nd form so that it can be verified.

PS: The last form 3D is not included since forge has problems displaying
those.  I noticed that multiple 3D examples are broken.

Willy

christophe-murphy (approved):

PR #3376, opened 2023-03-14
multi threaded OPENBLAS

willyborn:
Adds the threads options to OPENBLAS & FFT VCPKG compilation.

Description
-----------
* Improvement, by specializing options during VCPKG compilation

Changes to Users
----------------
Improved speed when using OPENBLAS.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [-] Functions added to unified API
- [-] Functions documented

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -46,8 +46,14 @@
         "openblasfftw": {
             "description": "Build with OpenBLAS/FFTW",
             "dependencies": [
-                "fftw3",
-                "openblas",
+                {
+                    "name": "fftw3",
+                    "features": [ "threads" ]
+                },
+                {
+                    "name": "openblas",
+                    "features": [ "threads" ]
+                },
                 "lapack"
             ]
         },


vcpkg.json
```diff
@@ -61,6 +67,7 @@
             "description": "Build OpenCL backend",
             "dependencies": [
                 "boost-compute",
+                "boost-program-options",
```
umar456, line 70: I don't think this is a good dependency for the project. It is only required for some of the tests in the clFFT project and those aren't built. I think the vcpkg should minimize the number of dependencies installed whenever possible.

willyborn: OK.
I get it out, and resend the PR.

willyborn: Or should I move it to the openblasfftw section ?

umar456 (approved):

vcpkg.json
willyborn: OK, done

PR #3380, opened 2023-03-22
include colab link in readme

pv-pterab-s:
include colab link in readme


umar456:
The problem we have is that the link is to a Python example and it was placed in the wrong location with an inconsistent style.  If you moved it next to the "Languages" section and matched their style I think it would work better perhaps.

pv-pterab-s:
Note how flashlight does this ... https://github.com/flashlight/flashlight/tree/main/flashlight/app/asr/tutorial   its c++ and it runs flawlessly up there.

PR #3389, opened 2023-03-31
Oneapi lapack

syurkevi:
This PR adds several missing lapack functions - svd, cholesky, lu to the oneapi backend, as well as the identity kernel required for the svd tests.

Description
-----------
* A few tests are still reliant on the indexing kernel and it is unknown if they are passing. 
* Adds lu, with kernel::lu_split
* Adds svd (inplace, out-of-place)
* Adds cholesky
Tests also rely on blas functions in outstanding blas PR.

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/oneapi/CMakeLists.txt
+++ b/src/backend/oneapi/CMakeLists.txt
@@ -216,10 +216,12 @@ target_sources(afoneapi
     kernel/bilateral.hpp
     kernel/diagonal.hpp
     kernel/diff.hpp
+    kernel/histogram.hpp
+    kernel/identity.hpp
     kernel/interp.hpp
     kernel/iota.hpp
     kernel/ireduce.hpp
-    kernel/histogram.hpp
+    kernel/lu_split.hpp
     kernel/memcopy.hpp
     kernel/mean.hpp
     kernel/random_engine.hpp
@@ -293,6 +295,7 @@ target_compile_options(afoneapi
 target_compile_definitions(afoneapi
   PRIVATE
     AF_ONEAPI
+    WITH_LINEAR_ALGEBRA
     CL_TARGET_OPENCL_VERSION=300
     CL_HPP_TARGET_OPENCL_VERSION=300
     CL_HPP_MINIMUM_OPENCL_VERSION=110
--- a/src/backend/oneapi/CMakeLists.txt
+++ b/src/backend/oneapi/CMakeLists.txt
@@ -266,9 +266,13 @@ add_library(ArrayFire::afoneapi ALIAS afoneapi)
 
 arrayfire_set_default_cxx_flags(afoneapi)
 
+find_package(MKL CONFIG REQUIRED)
+message(STATUS "${MKL_IMPORTED_TARGETS}")
+
 target_include_directories(afoneapi
   SYSTEM PRIVATE
-    ${SYCL_INCLUDE_DIR})
+    ${SYCL_INCLUDE_DIR}
+)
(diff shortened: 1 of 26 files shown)


vcpkg.json
```diff
@@ -23,10 +23,7 @@
     ],
     "features": {
         "tests": {
-            "description": "Build with tests",
-            "dependencies": [
-                "gtest"
-            ]
+            "description": "Build with tests"
```
umar456, line 26: I don't remember why we did this. Can you add this back?

test/iota.cpp
```diff
@@ -102,6 +102,8 @@ using af::array;
 using af::iota;
 
 TEST(Iota, CPP) {
+    af::setDevice(0);
```
umar456, line 105: I think this was an unintended change. please remove.

test/blas.cpp
```diff
@@ -119,6 +127,11 @@ TYPED_TEST(MatrixMultiply, NonSquare) {
 }
 
 TYPED_TEST(MatrixMultiply, SquareVector) {
+    af::setDevice(0);
+    af::info();
+    af::array a = af::randu(3, 3);
+    a.eval();
+    af_print(a);
```
umar456, line 134: Debugging changes should be removed from tests

src/backend/oneapi/CMakeLists.txt
```diff
@@ -264,15 +266,20 @@ add_library(ArrayFire::afoneapi ALIAS afoneapi)
 
 arrayfire_set_default_cxx_flags(afoneapi)
 
+find_package(MKL CONFIG REQUIRED)
+message(STATUS "${MKL_IMPORTED_TARGETS}")
```
umar456, line 270: Master now has the changes to correctly include the MKLConfig file from the oneapi install. Please remove

umar456 (changes requested):
Looks like you are still working on this PR. I left a few comments but I think you know about most of them.

vcpkg.json
syurkevi: gtest binaries are incompatible w/IntelLLVM and cause SEH exceptions. we will need a separate toolchain file. to be addressed in separate PR

src/backend/oneapi/platform.cpp
```diff
@@ -172,7 +172,8 @@ string getDeviceInfo() noexcept {
                 device->get_info<sycl::info::device::driver_version>();
             info << devVersion;
             info << " -- Device driver " << driVersion;
-
+            info << " -- FP64 Support: "
```
umar456, line 175: Can you remove this change?

vcpkg.json
umar456: Can you not add this to the main vcpkg and only do it locally while we figure out a solution. This change will break the CI and all the other backends.

src/backend/oneapi/kernel/iota.hpp
```diff
@@ -95,7 +95,7 @@ void iota(Param<T> out, const af::dim4& sdims) {
 
     try {
         getQueue()
-            .submit([=](sycl::handler& h) {
+            .submit([&](sycl::handler& h) {
```
umar456, line 98: we want this by value here right? otherwise the objects will be deleted when they go out of scope.

umar456 (approved):

umar456 (dismissed):

umar456 (approved):

PR #3390, opened 2023-03-31
oneapi sort

syurkevi:
This PR adds sort, sort_by_key, and topk to the oneapi backend.

Description
-----------
Adds sort, and sort_by_key functions based on oneDPL.
A few gotchas w/oneDPL
* oneDPL defines global static variable which cause runtime issues, so #define ONEDPL_USE_PREDEFINED_POLICIES 0 is necessary to avoid the globals
* oneDPL is  also unable to use the default std::less, std::greater functions without a endless recursive name-mangling error. The workaround is to manually define a comparison lambda function which requires overloads of the comparison operators for arrayfire::common::half types. These are decorated w/SYCL_EXTERNAL to allow calling them from kernel code.

The sorting functions also allow the topk functions to be enabled

Checklist
---------
- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/common/half.hpp
+++ b/src/backend/common/half.hpp
@@ -920,18 +920,10 @@ AF_CONSTEXPR __DH__ static inline bool operator==(
 AF_CONSTEXPR __DH__ static inline bool operator!=(
     arrayfire::common::half lhs, arrayfire::common::half rhs) noexcept;
 
-#ifdef AF_ONEAPI
-SYCL_EXTERNAL inline bool operator<(common::half lhs,
-                                    common::half rhs) noexcept;
-SYCL_EXTERNAL inline bool operator<(common::half lhs, float rhs) noexcept;
-SYCL_EXTERNAL inline bool operator>(common::half lhs,
-                                    common::half rhs) noexcept;
-#else
 __DH__ static inline bool operator<(arrayfire::common::half lhs,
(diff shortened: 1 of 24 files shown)


src/backend/common/half.hpp
```diff
@@ -919,10 +919,20 @@ AF_CONSTEXPR __DH__ static inline bool operator==(
     arrayfire::common::half lhs, arrayfire::common::half rhs) noexcept;
 AF_CONSTEXPR __DH__ static inline bool operator!=(
     arrayfire::common::half lhs, arrayfire::common::half rhs) noexcept;
+
+#ifdef AF_ONEAPI
+SYCL_EXTERNAL inline bool operator<(common::half lhs,
```
umar456, line 924: I don't see the difference between this and the other branch. can you explain? if its because of SYCL_EXTERNAL it would be better to define it for the other backends rather duplicating the code here.

syurkevi: external linkage required to call function from within a sycl kernel, also:
'sycl_device' attribute cannot be applied to a static function or function in an anonymous namespace

umar456:
You will need to split up the sort translation units into multiple objects like we do in the other backends because its taking too long to compilt(not link) the files.

umar456 (approved):

PR #3409, opened 2023-04-22
convolve2 and convolve_separable oneapi ports

pv-pterab-s:
oneapi port of convolve2 and convolve_separable. should close out necessary ports on the convolutions. this is missing support for the half data type. see comments on `src/backend/oneapi/convolve.cpp`.

Checklist
---------
- [X] Code compiles
- [ ] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/oneapi/convolve.cpp
+++ b/src/backend/oneapi/convolve.cpp
@@ -110,8 +110,38 @@ template<typename T>
 Array<T> convolve2_unwrap(const Array<T> &signal, const Array<T> &filter,
                           const dim4 &stride, const dim4 &padding,
                           const dim4 &dilation) {
-    Array<T> out =
-        convolve2_unwrap<T>(signal, filter, stride, padding, dilation);
+    dim4 sDims = signal.dims();
+    dim4 fDims = filter.dims();
+
+    dim_t outputWidth =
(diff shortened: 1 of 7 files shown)


src/backend/oneapi/convolve.cpp
```diff
@@ -110,18 +110,54 @@ template<typename T>
...
     return out;
 }
 
 template<typename T>
 Array<T> convolve2(Array<T> const &signal, Array<T> const &filter,
                    const dim4 stride, const dim4 padding, const dim4 dilation) {
-    ONEAPI_NOT_SUPPORTED("");
-    Array<T> out = createEmptyArray<T>(dim4(1));
-    return out;
+    if constexpr (!std::is_same<T, half>::value) {
```
pv-pterab-s, line 152: if this compiles with half enabled (dropping the if constexpr), the compiler complains with:
```
/home/gpryor/arrayfire/src/backend/oneapi/blas.hpp:33:24: error: no viable conversion from 'double' to 'const arrayfire::common::half'
    static constexpr T alpha = 1.0;
```
casting or `1.0f` has no effect on the error message. i vote we move on from half from here on out since this seems like a deeper issue

syurkevi (approved):

PR #3411, opened 2023-04-26
corrects missing oneapi accessor semantics

syurkevi:
Corrects missing oneapi accessor semantics

Description
---------
This PR goes back to the oneapi kernels and audits them for correct read/write/local semantics. This will allow the sycl runtime to create an out of order task-graph for better parallelism between unrelated(through memory accesses) kernels. This PR also removes a few uneccessary .wait() statements, further reducing host-side blocking. A few kernels had read/write buffers that were left as default sycl::accessor<T>. The mean kernel is still outstanding as there are other issues that need to be addressed in that kernel.

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/oneapi/CMakeLists.txt
+++ b/src/backend/oneapi/CMakeLists.txt
@@ -210,6 +210,7 @@ add_library(afoneapi
 target_sources(afoneapi
   PRIVATE
     kernel/KParam.hpp
+    kernel/accessors.hpp
     kernel/approx1.hpp
     kernel/approx2.hpp
     kernel/assign.hpp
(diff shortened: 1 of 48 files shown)


src/backend/oneapi/kernel/accessors_defs.hpp
```diff
@@ -0,0 +1,17 @@
+/*******************************************************
```
umar456, line 1: 2023 and can you rename this to accessor.hpp?

src/backend/oneapi/kernel/accessors_defs.hpp
```diff
@@ -0,0 +1,17 @@
...
+ ********************************************************/
+
+#pragma once
+#include <sycl/sycl.hpp>
+
+template<typename T>
+using read_accessor = sycl::accessor<T, 1, sycl::access::mode::read>;
+
+template<typename T>
+using write_accessor = sycl::accessor<T, 1, sycl::access::mode::write>;
```
umar456, line 17: What about local_accessor?

syurkevi: part of the spec, no need

umar456 (approved):

PR #3440, opened 2023-05-31
reduce local memory usage in iir

syurkevi:
Reduces local memory requirements for the oneapi iir kernel

The oneapi implementation of iir would greedliy allocate local memory following the opencl implementation. This would result in too much local memory being required. This PR reduces local memory requirements by dynamically allocating local memory, better following common sycl practices. 

Checklist
---------

- [X] Code compiles
- [X] Tests pass

### DIFF
--- a/src/backend/oneapi/kernel/iir.hpp
+++ b/src/backend/oneapi/kernel/iir.hpp
@@ -21,8 +21,6 @@ namespace arrayfire {
 namespace oneapi {
 namespace kernel {
 
-constexpr int MAX_A_SIZE = 1024;
-
 template<typename T, bool batch_a>
 class iirKernel {
    public:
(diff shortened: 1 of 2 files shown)


src/backend/oneapi/kernel/iir.hpp
```diff
@@ -124,10 +126,26 @@ void iir(Param<T> y, Param<T> c, Param<T> a) {
...
-        auto s_z = sycl::local_accessor<T>(MAX_A_SIZE, h);
-        auto s_a = sycl::local_accessor<T>(MAX_A_SIZE, h);
+        unsigned num_a = a.info.dims[0];
+
+        auto s_z = sycl::local_accessor<T>(num_a, h);
+        auto s_a = sycl::local_accessor<T>(num_a, h);
         auto s_y = sycl::local_accessor<T>(1, h);
 
+        size_t local_bytes_req = (num_a * 2 + 1) * sizeof(T);
+        if (local_bytes_req >
```
umar456, line 136: can you move this to the src/backend/oneapi/level. This is not the right place to handle these types of errors. How do we handle this scenarios in the other backends?

PR #3442, opened 2023-06-02
Oneapi lapack scratchsize

syurkevi:
Uses exactly sized buffers for scratch space

This PR removes the use of memAlloc, or createArray for scratch space since these methods return memory managed buffers that may be larger than what MKL expect. Reducing the size of these through the use of sub-buffers may work in most cases, but occasionally will fail if the memory returned by the memory manager is already a sub-buffer (which cannot be used to create a sub-sub-buffer in sycl)

The solve functions were also missing from the lapack functionality. This PR implements solve and pinverse for the oneapi backend.

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/oneapi/solve.cpp
+++ b/src/backend/oneapi/solve.cpp
@@ -30,6 +30,7 @@
 using arrayfire::common::cast;
 using std::min;
 using std::vector;
+using sycl::buffer;
 
 namespace arrayfire {
 namespace oneapi {
@@ -59,32 +60,13 @@ Array<T> solveLU(const Array<T> &A, const Array<int> &pivot, const Array<T> &b,
         ::oneapi::mkl::lapack::getrs_scratchpad_size<compute_t<T>>(
             getQueue(), opts, N, NRHS, LDA, LDB);
 
-    // TODO: which one?
-    Array<intl> ipiv              = cast<intl, int>(pivot);
-    sycl::buffer<int64_t> ipivBuf = ipiv.get()->reinterpret<int64_t, 1>();
-    sycl::buffer<compute_t<T>> scratchpad(scratchpad_size);
+    Array<intl> ipiv        = cast<intl, int>(pivot);
+    buffer<int64_t> ipivBuf = ipiv.get()->reinterpret<int64_t, 1>();
+    buffer<compute_t<T>> scratchpad(scratchpad_size);
 
-    /*
-    sycl::buffer<int64_t> ipivBuf(pivot.elements());
-    getQueue()
-        .submit([&](sycl::handler &h) {
-            auto ipivIn =
-                pivot.get()->template get_access<sycl::access_mode::read>(h);
-            auto ipivOut =
-                ipivBuf.get_access<sycl::access_mode::write>(h);
-            h.parallel_for(pivot.elements(),
-                    [=](sycl::id<1> i) {
-                        ipivOut[i] = static_cast<int64_t>(ipivIn[i]);
-                    });
-
-        });
-    */
-
-    Array<compute_t<T>> B = copyArray<compute_t<T>>(b);
-    sycl::buffer<compute_t<T>> aBuf =
-        A.template getBufferWithOffset<compute_t<T>>();
-    sycl::buffer<compute_t<T>> bBuf =
-        B.template getBufferWithOffset<compute_t<T>>();
+    Array<compute_t<T>> B     = copyArray<compute_t<T>>(b);
+    buffer<compute_t<T>> aBuf = A.template getBufferWithOffset<compute_t<T>>();
+    buffer<compute_t<T>> bBuf = B.template getBufferWithOffset<compute_t<T>>();
 
     ::oneapi::mkl::lapack::getrs(getQueue(), opts, N, NRHS, aBuf, LDA, ipivBuf,
                                  bBuf, LDB, scratchpad, scratchpad_size);
@@ -102,10 +84,10 @@ Array<T> generalSolve(const Array<T> &a, const Array<T> &b) {
     int K      = bDims[1];
     int MN     = std::min(M, N);
 
-    int lda     = a.strides()[1];
-    int astride = a.strides()[2];
-
-    sycl::buffer<int64_t> ipiv(MN * batches);
+    int lda      = a.strides()[1];
+    int astride  = a.strides()[2];
+    auto ipivMem = memAlloc<int64_t>(MN * batches);
+    buffer<int64_t> ipiv(*ipivMem, 0, MN * batches);
     int ipivstride = MN;
 
     int ldb     = b.strides()[1];
@@ -113,19 +95,17 @@ Array<T> generalSolve(const Array<T> &a, const Array<T> &b) {
 
     vector<int> info(batches, 0);
 
-    Array<T> A = copyArray<T>(a);
-    Array<T> B = copyArray<T>(b);
+    Array<T> A = copyArray<T>(a);  // A will be overwritten by L,U
+    Array<T> B = copyArray<T>(b);  // will be overwritten with solution
 
     std::int64_t scratchpad_size =
         ::oneapi::mkl::lapack::getrf_batch_scratchpad_size<compute_t<T>>(
             getQueue(), M, N, lda, astride, ipivstride, batches);
 
-    sycl::buffer<compute_t<T>> scratchpad(scratchpad_size);
+    buffer<compute_t<T>> scratchpad(scratchpad_size);
 
-    sycl::buffer<compute_t<T>> aBuf =
-        A.template getBufferWithOffset<compute_t<T>>();
-    sycl::buffer<compute_t<T>> bBuf =
-        B.template getBufferWithOffset<compute_t<T>>();
+    buffer<compute_t<T>> aBuf = A.template getBufferWithOffset<compute_t<T>>();
+    buffer<compute_t<T>> bBuf = B.template getBufferWithOffset<compute_t<T>>();
     ::oneapi::mkl::lapack::getrf_batch(getQueue(), M, N, aBuf, lda, astride,
                                        ipiv, ipivstride, batches, scratchpad,
                                        scratchpad_size);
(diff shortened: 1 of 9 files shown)


syurkevi:
@umar456  There's a few //TODOs that need your input.

src/backend/oneapi/solve.cpp
```diff
@@ -11,111 +11,152 @@
...
+        ::oneapi::mkl::lapack::getrs_scratchpad_size<compute_t<T>>(
+            getQueue(), opts, N, NRHS, LDA, LDB);
+
+    // TODO: which one?
+    Array<intl> ipiv              = cast<intl, int>(pivot);
+    sycl::buffer<int64_t> ipivBuf = ipiv.get()->reinterpret<int64_t, 1>();
+    sycl::buffer<compute_t<T>> scratchpad(scratchpad_size);
+
+    /*
+    sycl::buffer<int64_t> ipivBuf(pivot.elements());
```
umar456, line 68: Remove comment

src/backend/oneapi/solve.cpp
```diff
@@ -11,111 +11,152 @@
...
+    int M      = aDims[0];
+    int N      = aDims[1];
+    int K      = bDims[1];
+    int MN     = std::min(M, N);
+
+    int lda     = a.strides()[1];
+    int astride = a.strides()[2];
 
-    if (OpenCLCPUOffload()) { return cpu::solveLU(A, pivot, b, options); }
+    sycl::buffer<int64_t> ipiv(MN * batches);
```
umar456, line 108: This can be memAlloced right?

src/backend/oneapi/solve.cpp
```diff
@@ -11,111 +11,152 @@
...
-    int N    = A.dims()[0];
-    int NRHS = b.dims()[1];
+    int ldb     = b.strides()[1];
+    int bstride = b.strides()[2];
 
-    vector<int> ipiv(N);
-    copyData(&ipiv[0], pivot);
+    vector<int> info(batches, 0);
 
+    Array<T> A = copyArray<T>(a);
```
umar456, line 116: Why are we copying here? Is it modified by getrf? If so can you make a comment.

src/backend/oneapi/solve.cpp
```diff
@@ -11,111 +11,152 @@
...
-    // int batchz = aDims[2];
-    // int batchw = aDims[3];
+    scratchpad_size =
+        ::oneapi::mkl::lapack::getrs_batch_scratchpad_size<compute_t<T>>(
+            getQueue(), ::oneapi::mkl::transpose::nontrans, N, K, lda, astride,
+            ipivstride, ldb, bstride, batches);
 
-    // Array<T> A = copyArray<T>(a);
-    Array<T> B = copyArray<T>(b);
+    // TODO: reuse? or single large scratchpad?
```
umar456, line 138: You can't reuse right? the size is going to be different every time. One thing we can do is create a memory manager that allocates the buffers of the exact size, but I don't think you want to do that at this time.

src/backend/oneapi/solve.cpp
```diff
@@ -127,71 +168,81 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
...
+        B.resetDims(dim4(M, K));
 
-        OPENCL_BLAS_CHECK(
-            gpu_blas_trsm(OPENCL_BLAS_SIDE_LEFT, OPENCL_BLAS_TRIANGLE_UPPER,
-                          OPENCL_BLAS_CONJ_TRANS, OPENCL_BLAS_NON_UNIT_DIAGONAL,
-                          B.dims()[0], B.dims()[1], scalar<T>(1), (*dA)(),
-                          A.getOffset(), A.strides()[1], (*dB)(), B.getOffset(),
-                          B.strides()[1], 1, &queue, 0, nullptr, &event));
+        sycl::buffer<compute_t<T>> bBuf =
+            B.template getBufferWithOffset<compute_t<T>>();
```
umar456, line 200: Maybe we should make getBufferWithOffset as a free function? I don't really like the usage here but its your call.

src/backend/oneapi/solve.cpp
```diff
@@ -127,71 +168,81 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
...
+        // TODO: move to helper? trsm<T>(A, B, AF_MAT_CTRANS, true, true,
```
umar456, line 201: That would be nice.

src/backend/oneapi/solve.cpp
```diff
@@ -204,56 +255,65 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
...
+        sycl::buffer<compute_t<T>> tBuf =
+            t.template getBufferWithOffset<compute_t<T>>();
+        // In place Perform in place QR
+        ::oneapi::mkl::lapack::geqrf(getQueue(), M, N, aBuf, A.strides()[1],
+                                     tBuf, scratchpad, scratchpad_size);
+
+        // matmul(Q1, B)
+        sycl::buffer<compute_t<T>> bBuf =
+            B.template getBufferWithOffset<compute_t<T>>();
+        if constexpr (std::is_same_v<compute_t<T>, float> ||
```
umar456, line 277: we should create a variadic version of is_same_v that accepts any number of types. Reading this is hard on the eyes. something lke this:

is_any_of<T, double, float>

src/backend/oneapi/svd.cpp
```diff
@@ -38,15 +38,31 @@ void svdInPlace(Array<Tr> &s, Array<T> &u, Array<T> &vt, Array<T> &in) {
...
+    // MKL is finicky about exact scratch space size so we'll need to
+    // create sycl::buffer of exact size. if we use memAlloc, this might
+    // require a sub-buffer of a sub-buffer returned by memAlloc which is
+    // currently illegal in sycl
+    int64_t scratch_size =
+        ::oneapi::mkl::lapack::gesvd_scratchpad_size<compute_t<T>>(
+            getQueue(), ::oneapi::mkl::jobsvd::vectors,
+            ::oneapi::mkl::jobsvd::vectors, M, N, LDA, LDU, LDVt);
+
+    sycl::buffer<compute_t<T>> scratchpad(scratch_size);
```
umar456, line 50: can you make using statements in cpp files? i don't like namespaces sprinkled everywhere.

umar456 (changes requested):

src/backend/oneapi/solve.cpp
syurkevi: avoiding for now since no reuse anywhere else

src/backend/oneapi/solve.cpp
syurkevi: had to bring out the candles and soft jazz for this one

src/backend/oneapi/svd.cpp
syurkevi: keeping ::oneapi::mkl since those don't make sense on their own

src/backend/oneapi/solve.cpp
```diff
@@ -211,31 +190,28 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
...
             ::oneapi::mkl::lapack::ormqr(
                 getQueue(), ::oneapi::mkl::side::left,
                 ::oneapi::mkl::transpose::nontrans, B.dims()[0], B.dims()[1],
                 A.dims()[0], aBuf, A.strides()[1], tBuf, bBuf, B.strides()[1],
                 scratchpad_ormqr, scratchpad_size);
-        } else if constexpr (std::is_same_v<compute_t<T>,
-                                            std::complex<float>> ||
-                             std::is_same_v<compute_t<T>,
-                                            std::complex<double>>) {
+        } else if constexpr (is_any_of<compute_t<T>, std::complex<float>,
```
umar456, line 206: `arrayfire::common::is_complex<compute_t<T>>`

src/backend/oneapi/solve.cpp
```diff
@@ -260,44 +236,41 @@ Array<T> leastSquares(const Array<T> &a, const Array<T> &b) {
...
         ::oneapi::mkl::lapack::geqrf(getQueue(), M, N, aBuf, A.strides()[1],
                                      tBuf, scratchpad, scratchpad_size);
 
         // matmul(Q1, B)
-        sycl::buffer<compute_t<T>> bBuf =
+        buffer<compute_t<T>> bBuf =
             B.template getBufferWithOffset<compute_t<T>>();
-        if constexpr (std::is_same_v<compute_t<T>, float> ||
-                      std::is_same_v<compute_t<T>, double>) {
+        if constexpr (is_any_of<compute_t<T>, float, double>()) {
```
umar456, line 253: `std::is_floating_point<compute_t<T>>`

src/backend/oneapi/solve.hpp
```diff
@@ -11,6 +11,16 @@
 
 namespace arrayfire {
 namespace oneapi {
+
+template<typename T, typename U, typename... Args>
```
umar456, line 15: You should move it to common/traits.hpp

src/backend/oneapi/solve.cpp
```diff
@@ -59,32 +60,13 @@ Array<T> solveLU(const Array<T> &A, const Array<int> &pivot, const Array<T> &b,
...
-        });
-    */
-
-    Array<compute_t<T>> B = copyArray<compute_t<T>>(b);
-    sycl::buffer<compute_t<T>> aBuf =
-        A.template getBufferWithOffset<compute_t<T>>();
-    sycl::buffer<compute_t<T>> bBuf =
-        B.template getBufferWithOffset<compute_t<T>>();
+    Array<intl> ipiv        = cast<intl, int>(pivot);
+    buffer<int64_t> ipivBuf = ipiv.get()->reinterpret<int64_t, 1>();
```
umar456, line 64: Is it not accepting intl? It sucks that we need pivot to be converted. Can you update the API to accept s64 types as the pivot?

src/backend/oneapi/solve.cpp
```diff
@@ -102,30 +84,28 @@ Array<T> generalSolve(const Array<T> &a, const Array<T> &b) {
...
     int MN     = std::min(M, N);
 
-    int lda     = a.strides()[1];
-    int astride = a.strides()[2];
-
-    sycl::buffer<int64_t> ipiv(MN * batches);
+    int lda      = a.strides()[1];
+    int astride  = a.strides()[2];
+    auto ipivMem = memAlloc<int64_t>(MN * batches);
+    buffer<int64_t> ipiv(*ipivMem, 0, MN * batches);
```
umar456, line 90: Can you dereference the ipivMem instead of creating a buffer from it?

umar456 (changes requested):
A couple of more things. I think we should work on using standard functions whenever possible. is_floating_point. Additionallly can you move the is_any_of implementation to a common location?

umar456 (approved):

PR #3467, opened 2023-07-07
fix fftconvolve one2many tests

syurkevi:
This PR fixes fftconvolve one2many tests

This should not work. The descriptor does not seem to take a stride 0 offset parameter in standalone tests w/oneMKL. Using getOffset() for sub-arrays will result in a invalid descriptor for gpu, while it will work on cpu, but output incorrect results. Setting the descriptor to 0 sized offset makes tests pass, however this doesn't make sense since the backward convolution should only happen to the RHS batches which are offset wrt/the original packed array...

 bug fix???
The offset parameter doesn't seem to work in oneMKL, once correct behavior is determined corrections in corresponding commented code will need to be addressed.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/src/backend/oneapi/fft.cpp
+++ b/src/backend/oneapi/fft.cpp
@@ -95,7 +95,7 @@ std::vector<std::int64_t> computeStrides(const int rank, const dim4 istrides,
     if (rank == 3) return {offset, istrides[2], istrides[1], istrides[0]};
     if (rank == 4)
         return {offset, istrides[3], istrides[2], istrides[1], istrides[0]};
-    return {offset};
+    return {offset, istrides[0]};
 }
 
 template<::oneapi::mkl::dft::precision precision,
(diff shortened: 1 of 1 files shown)


src/backend/oneapi/fft.cpp
```diff
@@ -180,9 +180,14 @@ void fft_inplace(Array<T> &in, const int rank, const bool direction) {
         ::oneapi::mkl::dft::descriptor<precision,
                                        ::oneapi::mkl::dft::domain::COMPLEX>;
 
+    // TODO[STF]: WTF
+    // getOffset() for s0 throwing Invalid Descriptor when targeting gpu
+    // on CPU, results are wrong but does not throw
+    // strides not working? TODO: test standalone oneMKL
+    // perhaps in.getDataDims() needed instead of in.dims()?
```
pv-pterab-s, line 187: I believe it. The descriptor is fussy about many things.

pv-pterab-s (approved):

Issue #3511, opened 2023-10-13
[BUG] found on AF 3.8.3 and 3.9.0 for OPENCL, possibly with af::join


sebastienleclaire:
Hi,

First, congragulation for the release of AF 3.9.0.

I found a bug on OpenCL for AF version 3.8.3 and 3.9.0. Everything is fine with 3.8.2 on CPU and OPENCL. I did not test CUDA or ONEAPI.

Here the snippet:
```
float column_host[] = { 1, 2, 3 };
af::array column(3, 1, column_host); 

af::array result1 = af::join(1, -column, af::constant(0, 3, 1, f32), column);
af_print(result1)

af::array result2 = af::join(1, -column, af::constant(0, 3, 1, f32));
result2 = af::join(1, result2, column);
af_print(result2)
```
I'm not sure where the problem is comming from but result1 and result2 should be identical between CPU and OPENCL, however I do get:
**FOR OPENCL:**
result1
[3 3 1 1]
    1.0000    -1.0000     0.0000
    2.0000    -2.0000     0.0000
    3.0000    -3.0000     0.0000
**FOR CPU:**
result1 and result2
[3 3 1 1]
   -1.0000     0.0000     1.0000
   -2.0000     0.0000     2.0000
   -3.0000     0.0000     3.0000

The correct result should be "result2" as in AF 3.8.2 and before.

If you can't reproduce the problem, please let me know and I'll post more information.

Sébastien Leclaire

umar456:
Hey thanks for submitting this issue. I was able to reproduce this. I am investigating the issue and will follow up with a  solution.

Thanks

willyborn:
@umar

Since I updated the join part in 3.8.3, I looked into it and found the
error in the jit module (is also present in CUDA).
If you like, I can upload the latest version of master and create a PR.

Willy Born

umar456:
Thanks @willyborn for submitting the PR. I am going to try to get it in later today so we can fix this issue.

twesterhout:
Hello @umar456! Is there any progress on this? I've just got bitten by this bug, and a fix/workaround would be greatly appreciated.

AustinMooreT:
I think I've encountered this as well. I worked around it by flattening then joining.
Look at my following example for what I did.

`
const auto dim0     = pc1.dims(0) + pc2.dims(0);
`
`
[...]

PR #3513, opened 2023-10-15
Join does not always respect the order of provided parameters (#3511)

willyborn:
Arrays were be joined in the order the JIT engine generated the arrays iso the order of the parameters.

Description
-----------
1. A wrong assumption was made that the JIT engine would generate the arrays in the same order as provided by the parameters.
This assumption is violated when one of the generated arrays is an intermediate result of a previous specified JIT generated array.
Example of error condition: 
```
af::array A{af::constant(1.,10)};
af::array R{af::join(0,-A,A)};  --> generated af::join(0,A,-A) because A is an intermediate result of -A.
```
In this fix, the order of the parameters is imposed, independent from the order of JIT generation.
PS: This bug appears in OPENCL and CUDA.

2. The same 'unique' identifier "funcName" was generated for JIT kernels, with outputs as only difference.
As result, a different kernel could be executed than intended, dependent on the order of the JIT kernel generation.

* bug fix
* can be back ported
* extra test is added covering a large set of possible combinations of similar outputs (567 combinations)

Fixes: #3511

Changes to Users
----------------
None

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/common/jit/Node.cpp
+++ b/src/backend/common/jit/Node.cpp
@@ -42,6 +42,7 @@ int Node::getNodesMap(Node_map_t &node_map, vector<Node *> &full_nodes,
 }
 
 std::string getFuncName(const vector<Node *> &output_nodes,
+                        const vector<int> &output_ids,
                         const vector<Node *> &full_nodes,
                         const vector<Node_ids> &full_ids, const bool is_linear,
                         const bool loop0, const bool loop1, const bool loop2,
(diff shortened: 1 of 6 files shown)


willyborn:
Do not merge yet.
I discovered a new combination giving unexpected results.
The testing case will be expended to include more logical combinations (JIT nodes, Block nodes, repeating, dependencies, ...)
You will get an updated PR in the coming days.

willyborn:
Code is updated and join tests are expanded with the generation of multiple join combinations.

For some joins, the kernel of a previous join could be used since both generated the same 'unique' funcName (while having a different kernel) resulting in the reuse of the previously cached kernel iso generating the new one.
The chance of happening this is very rare, explaining why I added a systematic combination generator to the tests.

christophe-murphy (approved):

PR #3535, opened 2024-02-21
Fix race condition in OpenCL kernel

frasercrmck:
Description
-----------

Without the barrier at the end of barrierOR, it is possible for work-item 0 to start the next loop iteration and update predicates[0] while other work-items are still inside barrierOR reading `predicates`, meaning they read the next loop iteration's exit condition. This results in a divergent loop, where not all work-items reach the same barriers.

A previous fix identified this as a problem only on NVIDIA platforms, but strictly speaking a barrier is required in all cases to avoid a spec violation and undefined behaviour.

Changes to Users
----------------

The kernel should produce correct results on more OpenCL implementations.

Locally I tested both `Intel(R) FPGA Emulation Device` and various [oneAPI Construction Kit](https://github.com/codeplaysoftware/oneapi-construction-kit) devices, which all previously failed the `confidence_connected_opencl --gtest_filter="SingleSeed/ConfidenceConnectedDataTest.SegmentARegion/_prefix_background_radius_0_multiplier_1_iterations_5_replace_255"` unit test.

I'm unable to test other OpenCL implementations, sorry.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- ~~[ ] Functions added to unified API~~
- ~~[ ] Functions documented~~

### DIFF
--- a/src/backend/opencl/kernel/flood_fill.cl
+++ b/src/backend/opencl/kernel/flood_fill.cl
@@ -42,13 +42,7 @@ int barrierOR(local int *predicates) {
         barrier(CLK_LOCAL_MEM_FENCE);
     }
     int retVal = predicates[0];
-#if AF_IS_PLATFORM_NVIDIA
-    // Without the extra barrier sync after reading the reduction result,
-    // the caller's loop is going into infinite loop occasionally which is
-    // in turn randoms hangs. This doesn't seem to be an issue on non-nvidia
-    // hardware. Hence, the check.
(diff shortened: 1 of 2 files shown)


umar456:
Took me a bit to figure out the problem but I see the issue now. The we can ignore the errors in the CI because they are not related. I will test it on a couple of other systems before merge this PR. Thank you for your contribution!

christophe-murphy (approved):

Issue #3534, opened 2024-02-16
[BUG] Elementwise value assignment silently fails after in-place array downsize using `rows()`


warnellg:
In some scenarios, it seems that, after downsizing an array using an in-place call to `rows()` (ie, overwrite an array with only subset of its current rows), elementwise value assignment silently fails (ie, the element's value does not change and there are no compile or runtime errors).

Description
===========
* Did you build ArrayFire yourself or did you use the official installers: **Built myself.**
* Which backend is experiencing this issue? **CUDA.**
* Do you have a workaround? **No.**
* Can the bug be reproduced reliably on your system? **Yes.**
* A clear and concise description of what you expected to happen. **Expected elementwise value assignment to succeed (eg, in example below, `Dcols(0,0)` to have value `1234`). Given that that fails, expected a compile or runtime error.**
* Run your executable with AF_TRACE=all and AF_PRINT_ERRORS=1 environment variables set:
```
# AF_TRACE=all AF_PRINT_ERRORS=1 aftest
[platform][1708117940][9998] [ /tmp/arrayfire/src/backend/common/DependencyModule.cpp:102 ] Attempting to load: libforge.so
[platform][1708117940][9998] [ /tmp/arrayfire/src/backend/common/DependencyModule.cpp:107 ] Unable to open forge
[platform][1708117942][9998] [ /tmp/arrayfire/src/backend/cuda/device_manager.cpp:497 ] CUDA Driver supports up to CUDA 12.3.0 ArrayFire CUDA Runtime 11.3.0
[platform][1708117942][9998] [ /tmp/arrayfire/src/backend/cuda/device_manager.cpp:478 ] CUDA driver version(12.3.0) not part of the CudaToDriverVersion array. Please create an issue or a pull request on the ArrayFire repository to update the CudaToDriverVersion variable with this version of the CUDA runtime.

[platform][1708117942][9998] [ /tmp/arrayfire/src/backend/cuda/device_manager.cpp:566 ] Found 1 CUDA devices
[platform][1708117942][9998] [ /tmp/arrayfire/src/backend/cuda/device_manager.cpp:588 ] Found device: NVIDIA RTX A3000 Laptop GPU (sm_86) (5.80 GB | ~12187.5 GFLOPs | 32 SMs)
[platform][1708117942][9998] [ /tmp/arrayfire/src/backend/cuda/device_manager.cpp:652 ] AF_CUDA_DEFAULT_DEVICE: 
[...]

PR #3538, opened 2024-02-28
SubArray of subArray now returns subArray of original parent

willyborn:
left-hand subArrays always return the parent Array with a new dimensions/strides (copy parameter == false).
When the parent Array however already was an subArray, an internal copy was executed and a subArray on the internal copy was returned.  All updates performed on this left-hand subArray were lost.

Reason:
the subArray calculation assumes that the parent array is linear.  If not, a copy is created before starting the calculation.
For left-hand subArrays, a copy is not allowed because we would no longer write into the parent data buffer.  It is also not necessary for parent subArrays since the original parent array still is linear (although the parent subArray is not).

Solution:
The linear check is changed so that we now verify the original parent array on linearity.  In case of non-linear a copy is still performed when the copy parameter is true, otherwise an AF_ERROR is thrown.

Description
-----------

  * bug fix
  * Without this change, left-hand subArray of an existing subArray would reference an internal copy of the parent.  All data written remains invisible since it is an internal copy.
  * None
  * Existing function
* Can this PR be backported to older versions?
  * Yes
-->
Fixes: #3534

Changes to Users
----------------
No changes

Checklist
---------
- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/cpu/Array.cpp
+++ b/src/backend/cpu/Array.cpp
@@ -276,10 +276,23 @@ Array<T> createSubArray(const Array<T> &parent, const vector<af_seq> &index,
                         bool copy) {
     parent.eval();
 
-    dim4 dDims          = parent.getDataDims();
-    dim4 parent_strides = parent.strides();
+    const dim4 &dDims          = parent.getDataDims();
+    const dim4 &parent_strides = parent.strides();
 
-    if (parent.isLinear() == false) {
+    // (sub)Arrays remain linear when the strides corresponds to dataStrides
+    bool parent_isLinear = (parent_strides[0] == 1);
+    for (dim_t i = parent.ndims() - 1; i > 0; i--) {
+        parent_isLinear &=
+            parent_strides[i] == dDims[i - 1] * parent_strides[i - 1];
+    }
+
+    if (!parent_isLinear) {
+        if (!copy) {
+            // Linearizing parent through copy, is in conflict with the request
+            // of remaining inLine.
+            AF_ERROR("createSubArray inLine is impossible on non-Linear arrays",
+                     AF_ERR_INVALID_ARRAY);
+        }
         const Array<T> parentCopy = copyArray(parent);
         return createSubArray(parentCopy, index, copy);
     }
--- a/src/backend/cpu/Array.hpp
+++ b/src/backend/cpu/Array.hpp
@@ -239,7 +239,7 @@ class Array {
     dim_t getOffset() const { return info.getOffset(); }
     shared_ptr<T> getData() const { return data; }
 
-    dim4 getDataDims() const { return data_dims; }
+    const dim4 &getDataDims() const { return data_dims; }
 
     void setDataDims(const dim4 &new_dims);
 
--- a/test/array.cpp
+++ b/test/array.cpp
@@ -420,6 +420,23 @@ TEST(Array, ISSUE_951) {
     array b       = a.cols(0, 20).rows(10, 20);
 }
 
+TEST(Array, ISSUE_3534) {
+    // This works
+    // array a = range(dim4(5,5));
+    // a = a.rows(0,3).copy();
+    // Following assignment failed silently without above copy()
+    // a(0,0) = 1234;
+
+    array a = range(dim4(5, 5));
+    a       = a.rows(1, 4);
+    a(1, 1) = 1234;
+
+    array b = range(dim4(4, 5)) + 1.0;
+    b(1, 1) = 1234;
+
+    ASSERT_ARRAYS_EQ(a, b);
+}
+
 TEST(Array, CreateHandleInvalidNullDimsPointer) {
     af_array out = 0;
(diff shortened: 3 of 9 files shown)


src/backend/cpu/Array.hpp
edwinsolisf, line 1: Could you revert the const dim4& changes for all backends? I think changing this doesn't bring much benefit and breaks consistency with the way member variables are currently being returned.

src/backend/cpu/Array.cpp
```diff
@@ -276,10 +276,23 @@ Array<T> createSubArray(const Array<T> &parent, const vector<af_seq> &index,
...
 
-    dim4 dDims          = parent.getDataDims();
-    dim4 parent_strides = parent.strides();
+    const dim4 &dDims          = parent.getDataDims();
+    const dim4 &parent_strides = parent.strides();
+
+    // (sub)Arrays remain linear when the strides corresponds to dataStrides
+    bool parent_isLinear = (parent_strides[0] == 1);
+    for (dim_t i = parent.ndims() - 1; i > 0; i--) {
+        parent_isLinear &=
```
edwinsolisf, line 285: This check for linear solves the issue. Is the currently `isLinear` member function not correct?

edwinsolisf:
Works, tested on Windows and Ubuntu 22.04 on all backends

src/backend/cpu/Array.hpp
willyborn: I just wanted to make the interface consistent between dims(), strides() & getDataDims().
I will reverse the changes, in next upload.

src/backend/cpu/Array.cpp
willyborn: The loop is not a linear check on the parent array.  It is a linear check on the data written in the buffer.
For a subarray of an array, the parent (dims & strides) or equal to those in the data array.
For a subarray of a subarray, the parent is already a subarray and therefor likely non-linear.  However as long as the data array is linear, the operation is still possible.
Difference between code isLinear() and this:
 - isLinear(): Dims and Strides are from array
 - this code: Dims are from data array and Strides are from array

I will update the comments to clarify this further, and I will change the code so that it resembles more on the isLinear() code.  The difference will be noticed easier.

willyborn:
The PR is rebased on the last commit in master and force pushed.

The 2 previous requested changes are applied.

edwinsolisf (dismissed):

christophe-murphy:
It looks like this issue only occurs in the special case where a subarray is assigned to the original variable for the parent array. In other cases the reference count will be greater than 1 and a copy is made in af_assign_gen/af_assign_seq. While this fix works for this particular test I think it would be safer to add a linear check earlier in src/api/c/assign.cpp and make a copy there before the createSubArray call. What do you think @umar456 ?

willyborn:
The cause of the original bug, was the fact that a copy was taken, because the first subArray was considered non-linear.  This resulted in assign the rhs to the copy, which is discarded in the end.  The original parent remained unchanged.

Making a copy a level higher and building a subArray on the copy will have the same result as the original code.

christophe-murphy:
> The cause of the original bug, was the fact that a copy was taken, because the first subArray was considered non-linear. This resulted in assign the rhs to the copy, which is discarded in the end. The original parent remained unchanged.
> 
> Making a copy a level higher and building a subArray on the copy will have the same result as the original code.

It is discarded because the parent argument to createSubArray is a const reference and is left unchanged when it returns. That is why I suggested a linear check before the function call, see: https://github.com/christophe-murphy/arrayfire/tree/alternative_fix_issue_3534

christophe-murphy:
I think it is safer to use the fix I suggested instead ^ however if we are going to use this fix we will need more tests to make sure it handles all possible sub-arrays properly. All combinations of indexing by sequence, span, and by an arbitrary array of non-contiguous indices  along each dimension should be tried to make sure that the sub-array is created correctly.

christophe-murphy:
I have extended the testing. Right now there are failures on all the back ends except the CPU back end. I'm looking into why but it seems to be because we assume that dims is equivalent to data_dims in the assign kernels. Not sure if this is an oversight or intentional.

edwinsolisf (approved):
Tested on all backends

test/array.cpp
```diff
@@ -420,11 +426,290 @@ TEST(Array, ISSUE_951) {
...
+        const float *before_dev_pointer = a.device<float>();
+
+        a                              = a.cols(1, 4);
+        af_array after_handle          = a.get();
+        const float *after_dev_pointer = a.device<float>();
+
+        // We expect a copy to be created, so the handle and
+        // array pointers should be different
+        ASSERT_NE(before_handle, after_handle);
+        ASSERT_NE(before_dev_pointer, after_dev_pointer);
```
syurkevi, line 472: There is a potential for these pointers to be the same because the device pointer is linear even though its calling a sub array. This is not the case for the rows based tests above because the subarray would not be linear.

Issue #3545, opened 2024-03-20
[BUG] FP16 Types Not Allowed for Atan2 Method


AzeezIsh:
Manual type assertions not allowing FP16 types not allowing for atan2. Has not been updated to allow FP16. 

Description
===========
https://github.com/arrayfire/arrayfire/blob/4061db86e66306995175a14cf906c55c35373918/src/api/c/binary.cpp#L479

Reproducible Code and/or Steps
------------------------------
shape = (4, 7, 4, 9), dtype_name = arrayfire.float16(typecode<e>)

    @pytest.mark.parametrize(
        "shape",
        [
            (),
            (random.randint(1, 10),),
            (random.randint(1, 10), random.randint(1, 10)),
            (random.randint(1, 10), random.randint(1, 10), random.randint(1, 10)),
            (random.randint(1, 10), random.randint(1, 10), random.randint(1, 10), random.randint(1, 10)),
        ],
    )
    @pytest.mark.parametrize("dtype_name", util.get_float_types())
    def test_atan2_shape_dtypes(shape: tuple, dtype_name: dtype.Dtype) -> None:
        """Test inverse tan operation across all supported data types."""
        util.check_type_supported(dtype_name)
        lhs = wrapper.randu(shape, dtype_name)
        rhs = wrapper.randu(shape, dtype_name)
>       result = wrapper.atan2(lhs, rhs)

tests/test_trig.py:121: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
arrayfire_wrapper/lib/mathematical_functions/trigonometric_functions.py:30: in atan2
    return binary_op(atan2.__name__, lhs, rhs)
arrayfire_wrapper/lib/_utility.py:15: in binary_op
    call_from_clib(c_func_name, ctypes.pointer(out), lhs, rhs, bcast_var.get())
arrayfire_wrapper/lib/_utility.py:32: in call_from_clib
    _process_error(backend)
[...]

PR #3559, opened 2024-06-05
3545 bug fp16 types not allowed for atan2 method

christophe-murphy:
Add ability to use atan2 and hypot functions with fp16 arguments.

Description
-----------
Cases have been added to the atan2 and hypot functions to handle fp16 arguments. Previously an exception was thrown when trying to use fp16 arguments with these functions.

Tests have been added to verify that these functions are working correctly with fp16 arguments. Note that the OpenCL back end does not support half precision.

Several other functions have been identified as not supporting fp16, a new issue will be created for these.

Fixes: #3545 

Changes to Users
----------------
atan2 and hypot functions can now be used on fp16 arrays

Checklist
---------
- [ x] Rebased on latest master
- [ x] Code compiles
- [ x] Tests pass
- [ x] Functions added to unified API
- [ x] Functions documented

### DIFF
--- a/src/api/c/binary.cpp
+++ b/src/api/c/binary.cpp
@@ -476,7 +476,7 @@ af_err af_atan2(af_array *out, const af_array lhs, const af_array rhs,
     try {
         const af_dtype type = implicit(lhs, rhs);
 
-        if (type != f32 && type != f64) {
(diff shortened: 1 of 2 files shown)


edwinsolisf (approved):
Works, tested on Windows

Issue #3560, opened 2024-06-06
[BUG] Incorrect results when using the pow function with float16 arguments with CUDA backend


christophe-murphy:
Incorrect results are produced when using the pow function with float16 arguments with the CUDA backend

Description
===========

The following definition for the pow function is used in the JIT compiled kernel when the arguments are float16:

```
#ifdef AF_WITH_FAST_MATH
#define __pow(lhs, rhs)  \\
static_cast<double>( \\
pow(static_cast<double>(lhs), static_cast<double>(rhs)));
#else
#define __pow(lhs, rhs) \\
__float2int_rn(powf(__int2float_rn((int)lhs), __int2float_rn((int)rhs)))
#endif
```

As a result, if AF_WITH_FAST_MATH is false, the arguments are cast as integers and the result is rounded to the nearest integer.

Reproducible Code and/or Steps
------------------------------

```
#include <arrayfire.h>

using namespace af;

int main(int, char **) {
        info();

        array A = randu(64, 64, f16);
        array B = randu(64, 64, f16);
        array A_32(64, 64, f32);
        A_32(span,span) = A(span,span);
        array B_32(64, 64, f32);
        B_32(span,span) = B(span,span);
        array E = pow(A, B);
        array E_32 = pow(A_32, B_32);

        array E_d = abs((E - E_32) / E_32);

        float tol = 1.0e-3;
        float max_d;
        unsigned loc_d;
        max(&max_d, &loc_d, E_d);
        if(max_d > tol) {
            throw;
        }

    return 0;
}
```

Expected result: max_d <= 0.001 when comparing half precision with float
Actual result: max_d > 0.001

System Information
------------------

ArrayFire v3.9.0 (CUDA, 64-bit Linux, build 5e1e3362a)
Platform: CUDA Runtime 12.4, Driver: 550.54.15

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
name, memory.total [MiB], driver_version
NVIDIA GeForce RTX 4060 Laptop GPU, 8188 MiB, 550.54.15

Checklist
---------

- [ x] Using the latest available ArrayFire release
- [ x] GPU drivers are up to date

PR #3561, opened 2024-06-06
3560 bug incorrect results when using the pow function with float16 arguments with cuda backend

christophe-murphy:
When AF_WITH_FAST_MATH is disabled, the pow function definition incorrectly casts float16 arguments to integers.

Description
-----------
This error occurs when AF_WITH_FAST_MATH is disabled however when it is enabled the arguments are cast to double precision. This has been changed to always cast to double precision.

Fixes: #3560 

Checklist
---------

- [ x] Rebased on latest master but Depends on fix #3559 
- [ x] Code compiles
- [x ] Tests pass
- [x ] Functions added to unified API
- [ x] Functions documented


edwinsolisf (dismissed):
Works, tested on windows with an RTX3070 Ti Mobile

edwinsolisf (approved):

Issue #3539, opened 2024-02-28
[Build] OneAPI version 2024 incompatible for MKL


willyborn:
Description
===========
Intel only provides a download of the latest version of oneAPI (now 2024.0.1)
The arrayfire installer reports following error: compatible MKL package not found.  2024 is apparently not compatible.
During linking of afoneapi library DPCPP could not be found (as notified during the installation)

Apparently Intel changed this library to DPCPP-CT, which is an intel implementation.
oneAPI (2023+) requires MSVC 2019 (https://www.intel.com/content/www/us/en/developer/articles/reference-implementation/intel-compilers-compatibility-with-microsoft-visual-studio-and-xcode.html?wapkw=oneapi%20msvc%20table)

versions: 
  * system: windows 10
  * arrayfire: master (3.9.0)
  * oneAPI: 2024.0.0
![oneAPI installed modules](https://github.com/arrayfire/arrayfire/assets/70607676/5540fe88-3b43-4b2c-a55c-1d85f51af0fa)

  

Error Log
---------

```
D:\github\willyborn\arrayfire>CD build

D:\github\willyborn\arrayfire\build>cmake -G="Visual Studio 16 2019" -A=x64 ^
-DAF_BUILD_OPENCL:BOOL=ON ^
-DAF_BUILD_CPU:BOOL=ON ^
-DAF_BUILD_CUDA:BOOL=ON ^
-DAF_BUILD_UNIFIED:BOOL=ON ^
-DAF_BUILD_FORGE:BOOL=ON ^
-DAF_BUILD_ONEAPI:BOOL=ON ^
-DBUILD_TESTING:BOOL=ON ^
-DAF_WITH_FAST_MATH:BOOL=OFF ^
-DCUDA_NVCC_FLAGS:STRING="-Wno-deprecated-gpu-targets" ^
-DAF_COMPUTE_LIBRARY:STRING="FFTW/LAPACK/BLAS" ^
-DCMAKE_TOOLCHAIN_FILE:STRING="D:\github\vcpkg/scripts/buildsystems/vcpkg.cmake" ..
-- Running vcpkg install
Detecting compiler hash for triplet x64-windows...
All requested packages are currently installed.

...

-- Running vcpkg install - done
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19045.
-- The C compiler identification is MSVC 19.29.30154.0
-- The CXX compiler identification is MSVC 19.29.30154.0
[...]

PR #3573, opened 2024-07-16
3539 build oneapi version 2024 incompatible for mkl

christophe-murphy:
Modifications and bug fixes to get the OneAPI backend working with Intel OneAPI version 2024.1 This has only been tested on Linux and not Windows.

Description
-----------
Several changes needed to be made to the CMake files to support OneAPI version 2024. Calls to the Sycl queue also needed to be modified to prevent nested submissions which are now unsupported.

A number of tests were failing and bugs were identified and fixed including the following:
- In the wrap function the 4th dimension was not being wrapped
- The reduce function was being called with an incorrect number of threads in certain circumstances
- Half precision support check was erroneously indicating support for some devices

A number of functions are not supported on the OneAPI backend, a build option was added to skip tests for unsupported functions rather than failing.

Fixes: #3539 

Changes to Users
----------------
Additional CMake build option: AF_SKIP_UNSUPPORTED_TESTS
This will cause tests unsupported by the backend to be skipped rather than fail

Checklist
---------

- [ x] Rebased on latest master
- [ x] Code compiles
- [ x] Tests pass
- [ x] Functions added to unified API
- [ x] Functions documented


test/testHelpers.hpp
```diff
@@ -389,9 +394,30 @@ ::testing::AssertionResult assertRefEq(std::string hA_name,
...
+#else
 #define ASSERT_SUCCESS(CALL) ASSERT_EQ(AF_SUCCESS, CALL)
+#endif
+
+/// Catches function unsupported exception thrown by C++ API
+/// arrayfire function and skips test if SKIP_UNSUPPORTED_TESTS is true
+///
+/// \param[in]  CALL This is the arrayfire C++ function
+#ifdef SKIP_UNSUPPORTED_TESTS
+#define FUNCTION_UNSUPPORTED \
```
umar456, line 415: Not a fan of this approach to unsupported exceptions: Can you try using http://google.github.io/googletest/advanced.html#asserting-on-subroutines-with-an-exception

This should allow you to add a listener to catch these exceptions.

test/testHelpers.hpp
```diff
@@ -242,6 +242,11 @@ bool noHalfTests(af::dtype ty);
     if (noHalfTests((af_dtype)af::dtype_traits<type>::af_type))   \
     GTEST_SKIP() << "Device doesn't support Half"
 
+static af_err CHECK_SUPPORTED_ERR_RETURN;
+#define CHECK_SUPPORTED(af_stat) \
+    if (af_stat == AF_ERR_NOT_SUPPORTED) \
+        GTEST_SKIP() << "Function not supported on this backend"
```
umar456, line 248: would be good to print the backend name.

test/medfilt.cpp
```diff
@@ -406,7 +408,8 @@ TEST(MedianFilter, Docs) {
     //    2.0000        6.0000       10.0000       14.0000
     //    3.0000        7.0000       11.0000       15.0000
     //    4.0000        8.0000       12.0000       16.0000
-    array b = medfilt(a, 3, 3, AF_PAD_ZERO);
+    array b;
+    try { b = medfilt(a, 3, 3, AF_PAD_ZERO); } catch FUNCTION_UNSUPPORTED
```
umar456, line 412: These are examples in documentations. You should never put test related stuff in these types of code snippets.

src/backend/oneapi/kernel/index.hpp
```diff
@@ -133,11 +133,14 @@ void index(Param<T> out, Param<T> in, IndexKernelParam& p,
     blocks[0] *= threads[0];
 
     sycl::nd_range<3> marange(blocks, threads);
+    sycl::buffer<uint> *idxArrs_get[4];
```
umar456, line 136: Wouldn't this go out of scope when you call submit?

src/backend/oneapi/join.cpp
```diff
@@ -125,16 +127,18 @@ Array<T> join(const int jdim, const Array<T> &first, const Array<T> &second) {
     if (second.isReady()) {
         if (1LL + jdim >= second.ndims() && second.isLinear()) {
             // second & out are linear
+            auto second_get = second.get();
```
umar456, line 130: You need better names here.

CMakeModules/select_compute_arch.cmake
```diff
@@ -92,6 +92,23 @@ if(CUDA_VERSION VERSION_GREATER_EQUAL "11.1")
   set(CUDA_LIMIT_GPU_ARCHITECTURE "9.0")
 endif()
 
+if(CUDA_VERSION VERSION_GREATER_EQUAL "11.8")
```
umar456, line 95: Was any architecture removed during these releases?

umar456 (changes requested):

src/backend/oneapi/kernel/index.hpp
christophe-murphy: This is captured by the lambda function which is passed as an argument to submit.

CMakeModules/select_compute_arch.cmake
christophe-murphy: Support for CC < 5.0 was removed in CUDA 12. I've updated the CUDA_ALL_GPU_ARCHITECTURES list.

src/backend/oneapi/join.cpp
christophe-murphy: changed to "second_array", etc..

test/testHelpers.hpp
christophe-murphy: I've looked into this and an error message with the back end name is generated in backend/common/err_common.cpp in the set_global_error_string function. However it is only printed if the AF_PRINT_ERRORS environment variable is set. Not sure the reason for this.

christophe-murphy: I also notice that in the ***_NOT_SUPPORTED macros a string is being passed called 'message' however this string is then assigned to the 'backend' member of the SupportError class. So when getBackendName() is called the 'message' is being returned instead of the back end name.

christophe-murphy: It is difficult to obtain the back end name at this point in the code. If a C++ exception is triggered it will print out the back end as part of the verbose error message when the AF_PRINT_ERRORS environment variable is set. For the C API you just get a AF_ERR_NOT_SUPPORTED returned from the function.

test/testHelpers.hpp
christophe-murphy: The documentation you linked to seems to be for turning a failed assert into an exception, not for listening for an exception but I will investigate more to see if we can do what you suggest.

umar456 (approved):

Issue #3575, opened 2024-08-01
[BUG] __shfl_down_sync bug causes undefined behavior


christophe-murphy:
Bug in call to __shfl_down_sync in reduce_blocks_by_key results in Warp Illegal Instruction Error in CUDA > 12.2

Description
===========

- Built code from main
- Affects CUDA backend when using reduce_blocks_by_key
- Workaround is to use CUDA <=12.2

Reproducible Code and/or Steps
------------------------------
Run the following test case: reduce_cuda --gtest_filter=Reduce.SNIPPET_sum_by_key

System Information
------------------

1. ArrayFire version: main
2. Devices installed on the system: NVIDIA GeForce RTX 4060 Laptop GPU, 8188 MiB,
3. Driver version: 560.28.03

Checklist
---------

- [ x] Using the latest available ArrayFire release
- [ x] GPU drivers are up to date

PR #3576, opened 2024-08-05
3575 bug   shfl down sync bug causes undefined behavior

christophe-murphy:
Fix for bug where warp primitives were called by threads outside mask, triggering a warp illegal instruction exception in CUDA versions > 12.2

Description
-----------
A bug has been uncovered by CUDA versions newer than 12.2 where calls to the __warp_down_sync primitive are being made by threads outside of the thread mask in the reduce_by_key header. Although this results in undefined behavior, in prior CUDA versions the primitive would execute successfully. This has changed since in newer versions since 12.2 where a warp illegal instruction exception is thrown.

The primitive wrapper functions header has been modified so that the full mask is always used to ensure the same behavior with old and new versions of CUDA. If a different mask is required, then the specific new version of the primitive can be called.

Fixes: #3575 

Checklist
---------

- [ x] Code compiles
- [x ] Tests pass

### DIFF
--- a/src/backend/cuda/kernel/reduce_by_key.hpp
+++ b/src/backend/cuda/kernel/reduce_by_key.hpp
@@ -25,8 +25,6 @@
 
 using std::unique_ptr;
 
-const static unsigned int FULL_MASK = 0xFFFFFFFF;
-
 namespace arrayfire {
 namespace cuda {
 namespace kernel {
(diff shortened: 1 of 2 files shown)


src/backend/cuda/device_manager.cpp
edwinsolisf, line 1: Might be good to remove this version changes from here and move it to its own PR

edwinsolisf (approved):
Tested on Windows and Ubuntu-22.04, passes all tests

edwinsolisf (approved):

Issue #3525, opened 2023-12-19
[BUG] indexing the second dim with af::seq ignores step


FloopCZ:
Indexing the first dimension with `af::array` and the second dimension with `af::seq` ignores the `step` parameter of the `af::seq`. Casting the `af::seq` to `af::array` is a workaround for this issue.

Minimal example:
```c++
#include <arrayfire.h>

int main() {
    af::info();

    af::array A = af::seq(10);
    A = A.T();
    af::print("A", A);

    af::array selector_x = af::constant(0, 1);
    af::seq selector_y(0, 9, 2);  // select every second element
    // af::array selector_y = af::seq(0, 9, 2);  // this is a workaround

    af::print("A select", A(selector_x, selector_y));
}
```

Output:
```
ArrayFire v3.8.3 (CUDA, 64-bit Linux, build default)
Platform: CUDA Runtime 12.0, Driver: 545.29.02
[0] NVIDIA RTX A500 Laptop GPU, 3905 MB, CUDA Compute 8.6
A
[1 10 1 1]
   Offset: 0
   Strides: [1 1 10 10]
    0.0000     1.0000     2.0000     3.0000     4.0000     5.0000     6.0000     7.0000     8.0000     9.0000
A select
[1 5 1 1]
   Offset: 0
   Strides: [1 1 5 5]
    0.0000     1.0000     2.0000     3.0000     4.0000
```

Expected behavior (also the workaround behavior) would be that `A select` is equal to:
```
    0.0000     2.0000     4.0000     6.0000     8.0000
```

Arrayfire version:
```
ArrayFire v3.8.3 (CUDA, 64-bit Linux, build default)
```

PR #3585, opened 2024-08-15
Fix issue 3525: correct handling of array indexing with sequence

edwinsolisf:
Fixes issue where array indexing with a sequence whose step is not 1 is handled as a sequence with step 1.

Description
-----------
* Is this a new feature or a bug fix?: Bug fix
* More detail if necessary to describe all commits in pull request.: Adds a member variable to all index structs to pass the step into the different kernels and compute the correct index.
* Why these changes are necessary: Incorrect behavior of `af::seq` in array indexing
* Potential impact on specific hardware, software or backends: affects all backends
* New functions and their functionality.: None
* Can this PR be backported to older versions?: All backends except oneAPI can be backported
* Future changes not implemented in this PR.: None

Fixes: #3525

Changes to Users
----------------
* Additional options added to the build.: None
* No action required by the user

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented

### DIFF
--- a/src/backend/cuda/kernel/index.cuh
+++ b/src/backend/cuda/kernel/index.cuh
@@ -43,13 +43,17 @@ __global__ void index(Param<T> out, CParam<T> in, const IndexKernelParam p,
         gw < out.dims[3]) {
         // calculate pointer offsets for input
         int i =
-            p.strds[0] * trimIndex(s0 ? gx + p.offs[0] : ptr0[gx], in.dims[0]);
+            p.strds[0] *
+            trimIndex(s0 ? gx * p.steps[0] + p.offs[0] : ptr0[gx], in.dims[0]);
         int j =
-            p.strds[1] * trimIndex(s1 ? gy + p.offs[1] : ptr1[gy], in.dims[1]);
+            p.strds[1] *
+            trimIndex(s1 ? gy * p.steps[1] + p.offs[1] : ptr1[gy], in.dims[1]);
         int k =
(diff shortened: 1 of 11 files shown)


src/backend/cuda/kernel/index.cuh
```diff
@@ -43,13 +43,17 @@ __global__ void index(Param<T> out, CParam<T> in, const IndexKernelParam p,
         gw < out.dims[3]) {
         // calculate pointer offsets for input
         int i =
-            p.strds[0] * trimIndex(s0 ? gx + p.offs[0] : ptr0[gx], in.dims[0]);
+            p.strds[0] *
+            trimIndex(s0 ? gx * p.steps[0] + p.offs[0] : ptr0[gx], in.dims[0]);
```
umar456, line 47: Do you need to update the condition on 42? looks like you would read out of bounds here.

edwinsolisf: The condition is still correct because `gx gy gz gw` refer to indices in the output array, and the given check makes sure they are always inside the output array dimensions. The changes affect the indexing of the input array being sampled and the function `trimIndex` makes sure it's inside the input array dimensions.

christophe-murphy: Ran tests with compute sanitizer. No issues found.

christophe-murphy (approved):
Looks good, tests are passing.

PR #3587, opened 2024-08-20
Added testing for sequence indexing with non-unitary steps

edwinsolisf:
When sequences are mixed in array indexing, `af::backend::index` is called. This pull request adds a test for this function call for the case where the sequences have a step that is not 1.

Description
-----------

- [x] Code compiles
- [x] Tests pass
- [x] Functions documented


christophe-murphy:
New test data in https://github.com/arrayfire/arrayfire-data/pull/21 will be merged with this test.

christophe-murphy (approved):

PR #3590, opened 2024-08-21
Fix for incorrect x axis values for histogram.

christophe-murphy:
Rounding was being applied to x axis min and max values but this should not be done for a histogram where the values are in fact bin labels.

xmin and xmax are now always equal to the first and last bin respectively.

Checklist
---------

- [x ] Code compiles


edwinsolisf (approved):
Tested on windows, works

Issue #3597, opened 2024-08-31
[Build] Build failure when setting AF_BUILD_CUDA=OFF


christophe-murphy:
If CUDA_FOUND is true but AF_BUILD_CUDA is set to OFF and AF_BUILD_UNIFIED is set to ON then cmake will attempt to build the cuda_unified test and fail when it tries to link to the cuda arrayfire library.

Description
===========

* What do you suspect is causing the issue:
* What operating system and/or distro are you using: Ubuntu 24.04
* Versions of the packages related to this bug: master

PR #3598, opened 2024-08-31
Fix build failure awhen 

christophe-murphy:
Add condition so that cuda_* tests (including cuda_unified) are only built if AF_BUILD_CUDA is true

Fixes: #3597 

Checklist
---------

- [x ] Code compiles
- [ x] Tests pass


edwinsolisf (approved):
Tested on Windows and Ubuntu 22.04, works

Issue #1693, opened 2016-12-28
BUG: accessing deleted memory in array_proxy index


vakopian:
The following code should work:
```
    array a = constant(1.0f, 3, 5, f32);
    array cond = constant(1, 5, b8);
    auto X = a(span, cond);
    array Y = X;
```
but it produces the error below:
```
terminate called after throwing an instance of 'af::exception'
  what():  ArrayFire Exception (Unknown error:208):
In function const ArrayInfo& getInfo(af_array, bool, bool)
In file src/api/c/array.cpp:30
Input Array not created on current device

In function af::array::array_proxy::operator af::array() const
In file src/api/cpp/array.cpp:686
```

I think this is a bug in the way memory is managed for the `af_index_t` structure within the `array_proxy`.
Below is what I think happens:
1. the `array::operator()` is called to create the `X` proxy
2. the `cond` array is first converted to a **temporary** `index` object (since `operator()` can only take `index`)
3. the `index` constructor calls `where` which creates a new array that holds the indices ([index.cpp:69](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/index.cpp#L69))
4. then the array created by `where` is copied over into a new array ([index.cpp:71](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/index.cpp#L71)), that's used as the `arr` member of the index ([index.cpp:72](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/index.cpp#L72))
5. the `array::operator()` calls `gen_indexing` [array.cpp:350](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/array.cpp#L350), which then just copies the indices into a local container [array.cpp:312-316](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/array.cpp#L312-L316).

So at this point the pointer to the array allocated in step 4 within a **temporary** `index` is copied over into the index of the `array_proxy`. Then, once we exit the `array::operator()`, the temporary index is deleted, together with its `arr`: [index.cpp:84](https://github.com/arrayfire/arrayfire/blob/devel/src/api/cpp/index.cpp#L84). And we end-up with an `array_proxy` that's holding a dangling pointer in one if its indices. Because the `ArrayInfo` memory is now freed, it points to bad memory and we get the error above.

This analysis is confirmed by just making the automatic temporary index explicit, i.e. the following code works fine:
```
    array a = constant(1.0f, 3, 5, f32);
[...]

PR #3600, opened 2024-09-05
Fix for 1693 bug access to deleted memory in array proxy index

christophe-murphy:
Arrays can now be used as an index for creating an array proxy without failure.

Description
-----------

When using an array as an index when creating an array proxy a temporary index object is created and passed to the constructor for the array proxy. However the array underlying the index was not being kept after the temporary index was destroyed, resulting in undefined behavior. This fix ensures that the array is retained until the array proxy is destroyed. Users can now use an array as an index when creating an array proxy without needing to explicitly define an index object.

Fixes: #1693 

Checklist
---------
- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/api/cpp/array.cpp
+++ b/src/api/cpp/array.cpp
@@ -389,6 +389,14 @@ static array::array_proxy gen_indexing(const array &ref, const index &s0,
     inds[1] = s1.get();
     inds[2] = s2.get();
     inds[3] = s3.get();
+    
+    for(int i; i < AF_MAX_DIMS; ++i) {
+        if(!inds[i].isSeq) {
+            af_array arr = 0;
+            af_retain_array(&arr, inds[i].idx.arr);
+	    inds[i].idx.arr = arr;
+        }
+    }
 
     return array::array_proxy(const_cast<array &>(ref), inds, linear);
 }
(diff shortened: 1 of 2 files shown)


src/api/cpp/array.cpp
```diff
@@ -164,6 +164,9 @@ struct array::array_proxy::array_proxy_impl {
     void delete_on_destruction(bool val) { delete_on_destruction_ = val; }
 
     ~array_proxy_impl() {
+        for(int i = 0; i < AF_MAX_DIMS; ++i) {
```
umar456, line 167: Can you run this through the address-sanitizer for all backends?

christophe-murphy: I compiled and ran with the address sanitizer and no memory leak issues were reported. Couldn't get it to work with the CUDA backend so I ran with the compute sanitizer and no issues either. I did get an ODR violation for the unified back end but this seems to be unrelated to this.

christophe-murphy:
OK, I reran address sanitizer on the CUDA backend with the following options and it ran without any issues reported: `protect_shadow_gap=0:replace_intrin=0:detect_stack_use_after_return=0`

christophe-murphy:
I  have done some more testing, see latest commit. I ran these through the leak sanitizer for all back ends and, aside from an unrelated memory leak in the random number generator engine, there are no issues.

edwinsolisf (approved):

Issue #1918, opened 2017-09-08
Sparse matrix not updated in for loop in OpenCL


villekf:
When you create a sparse matrix in for loop with OpenCL backend, the matrix is always the first one, i.e. it doesn't get updated. On CUDA and CPU backend everything works normally. The following minimal example demonstrates this:

	array output;
	float value[] = { 1, 1, 2, 2 };
	int index[] = { -1, 1, 2 };
	int row[] = { 0, 2, 2, 0, 0, 2 };
	int col[] = { 0, 1, 0, 1 };
	array values(4, 1, value, afHost);
	array rows(6, 1, row, afHost);
	array cols(4, 1, col, afHost);
	array S;

	for (int i = 0; i < 2; i++) {

		S = sparse(2, 2, values(seq(index[i] + 1, index[i + 1])), rows(seq(3 * i, 3 * (i + 1) - 1)), cols(seq(index[i] + 1, index[i + 1])));
		output = dense(S);

		af_print(output);

	}

af:: info OpenCL:
ArrayFire v3.5.0 (OpenCL, 64-bit Windows, build 05999f3)
[0] NVIDIA: GeForce GTX TITAN, 6144 MB

CUDA:
ArrayFire v3.5.0 (CUDA, 64-bit Windows, build 05999f3)
Platform: CUDA Toolkit 8, Driver: CUDA Driver Version: 9000
[0] GeForce GTX TITAN, 6144 MB, CUDA Compute 3.5

christophe-murphy:
Reopen as issue has also been found for the COO storage format and for the oneAPI backend.

PR #3602, opened 2024-09-11
1918 sparse matrix not updated in for loop in opencl

christophe-murphy:
Fix for issue in the opencl backend where array offsets for the values and/or row/cols arrays are not accounted for when converting a sparse array to a dense one.

Description
-----------

A sparse csr matrix may be constructed using arrays for the values and the row and column indices. When these arrays are indexed using the seq method, the offsets into them are stored. These offsets were not being accounted for in the sparse to dense conversion function which can result in incorrect values in the resulting dense matrix. This PR fixes that.
Fixes: #1918 

Checklist
---------

- [x] Code compiles
- [x] Tests pass


edwinsolisf (approved):
Tested changes on windows with OpenCL

PR #3604, opened 2024-09-19
GMRES solver example

amirakb89:
Description
-----------
A GMRES solver example is added. In this example a 2D Laplace equation is solved with the GMRES solver. 
The GMRES method is an iterative algorithm for solving large, sparse, and non-symmetric systems of linear equations. Unlike other iterative methods like the Conjugate Gradient (CG) method, which is suitable only for symmetric positive-definite matrices, GMRES is designed for general, non-symmetric systems. The method approximates the solution by the vector in a Krylov subspace with minimal residual.

Fixes: #<issue number> ...

Changes to Users
----------------

Checklist
---------

- [ ] Code compiles
- [ ] Tests pass
- [ ] Functions documented

### DIFF
--- a/.github/workflows/release_src_artifact.yml
+++ b/.github/workflows/release_src_artifact.yml
@@ -9,7 +9,7 @@ name: ci
 jobs:
     upload_src_tarball:
         name: Upload release source tarball
-        runs-on: ubuntu-18.04
+        runs-on: ubuntu-latest
         steps:
             - name: Fetch Repo Info
               run: |
@@ -40,7 +40,7 @@ jobs:
                                           libopenblas-dev \
                                           ocl-icd-opencl-dev \
                                           nvidia-cuda-toolkit \
-                                          libboost1.68-dev
+                                          libboost-dev
 
             - name: CMake Configure
               run: |


examples/lin_algebra/gmres.cpp
```diff
@@ -0,0 +1,274 @@
...
+          
+          A_host.push_back(1);
+          row_indices.push_back(row_id);
+          col_indices.push_back(row_id);
+        }
+        else
+        {   
+          row_indices.push_back(row_id);
+          col_indices.push_back(row_id-n);
+          A_host.push_back(-1.0);
```
christophe-murphy, line 191: -1.0 should be -0.25 and 4.0 should be 1.0 here to be consistent with the example matrix given above.

examples/lin_algebra/gmres.cpp
```diff
@@ -0,0 +1,274 @@
...
+    s[0] = beta;
+    for ( int i=0; i< R; i++ )
+    {
+      iter++;
+      // XXLiu: w = solve(A * v[i]);
+      w = matmul(A, v.col(i));
+
+      int h_idx;
+      for (int k=0; k<=i; k++)
+      {
```
christophe-murphy, line 100: Code could use some more comments here to explain what is going. Looks like we are doing Arnoldi iteration here. Looks like Givens rotations are being performed on the Hessenberg matrix.

examples/lin_algebra/gmres.cpp
```diff
@@ -0,0 +1,274 @@
...
+  std::vector<float> iter;
+  std::vector<float> time;
+  std::vector<float> nnz;
+};
+template <class T>
+void Update(array& x, const int k, const std::vector<T>&H,
+            const int R, const std::vector<T>&s,
+            const array v)
+{
+  T y[k+1];
```
christophe-murphy, line 30: Use of variable length arrays is not technically valid C++ though GCC at least supports it anyway. I recommend using a vector instead.

Issue #3606, opened 2024-09-28
[BUG] Results for bagging different for opencl


willyborn:
The bagging example has an accuracy of 91.45% for CPU and CUDA.  For OpenCL, the accuracy is only 27.06%

Description
===========
The result should be independent from used platform (except rounding errors).

Reproducible Code and/or Steps
------------------------------

System Information
------------------
All run in Debug mode, explaining the long prediction times.
```
ArrayFire v3.9.0 (CPU, 64-bit Windows, build f4157374e)
[0] AMD: AMD A10-7870K Radeon R7, 12 Compute Cores 4C+8G
Accuracy on testing  data: 91.45
Prediction time: 1829.4632

D:\github\willyborn\arrayfire\build\examples\machine_learning\Debug\bagging_cpu.exe (process 13640) exited with code 0.
------------------
ArrayFire v3.9.0 (CUDA, 64-bit Windows, build f4157374e)
Platform: CUDA Runtime 11.8, Driver: 12060
[0] NVIDIA GeForce GTX 750 Ti, 2048 MB, CUDA Compute 5.0
Accuracy on testing  data: 91.45
Prediction time: 668.5332

D:\github\willyborn\arrayfire\build\examples\machine_learning\Debug\bagging_cuda.exe (process 11996) exited with code 0.
------------------
ArrayFire v3.9.0 (OpenCL, 64-bit Windows, build f4157374e)
[0] AMD: Spectre, 7211 MB -- OpenCL 2.0 AMD-APP (3224.5) -- Device driver 3224.5 -- FP64 Support: True
-1- NVIDIA: NVIDIA GeForce GTX 750 Ti, 2047 MB -- OpenCL 3.0 CUDA -- Device driver 560.81 -- FP64 Support: True
Accuracy on testing  data: 27.06
Prediction time: 15.0551

D:\github\willyborn\arrayfire\build\examples\machine_learning\Debug\bagging_opencl.exe (process 12884) exited with code 0.
```

Checklist
---------

- [x] Using the latest available ArrayFire release
- [x] GPU drivers are up to date

PR #3611, opened 2024-10-03
lookup disgarded idx offset  in OpenCL

willyborn:
The OpenCL code did not take the idx offset into account resulting in the usage of the wrong data set, without reporting any error.
This was the reason for the differences of the bagging example, on the different platforms.

Description
-----------
Following commits are included
- e4de910 Fixed missing offset handling in lookup

> This contains the fix in the OpenCL code

- 80dc4ed Extends ARG checking & buffer leak for af_lookup

> For all platforms
> Extra checks included on arguments: nullptr as address of output
> ASAP release of output array if previously allocated
> Returning NULL_ARRAY when input array or idx array are a NULL_ARRAY

- 98631b8 Adds argument checking for internal debugging

> For all platforms
> Internal code calls directly the back-end code, without passing a check on their arguments.  This check is performed with asserts so that performance of Release code is unaffected.

- 3e28fff Extends test cases for lookup

> For all platforms
> All tests on lookup are now including a check on buffer leaks
> Arrays of type: NULL_ARRAY, FULL_ARRAY, SUB_ARRAY and REORDERED_ARRAY are used as input type
> Check for nullptr for address of output array
> Check for assignment of an argument array to the output array

* Is this a new feature or a bug fix?   Bug
* Why these changes are necessary.    To produce a correct result in all situations
* Potential impact on specific hardware, software or backends.   Only OpenCL is impacted
* New functions and their functionality.   None.
* Can this PR be backported to older versions?   Yes
-->
Fixes: #3606

Changes to Users
----------------
No changes for programs using the FULL_ARRAY (offset = 0).  The correct result will be given for all other case as well now.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/cuda/device_manager.cpp
+++ b/src/backend/cuda/device_manager.cpp
@@ -101,6 +101,8 @@ static const int jetsonComputeCapabilities[] = {
 
 // clang-format off
 static const cuNVRTCcompute Toolkit2MaxCompute[] = {
+    {12080, 9, 0, 0},
+    {12070, 9, 0, 0},
     {12060, 9, 0, 0},
     {12050, 9, 0, 0},
     {12040, 9, 0, 0},
@@ -144,6 +146,8 @@ struct ComputeCapabilityToStreamingProcessors {
 // clang-format off
 static const ToolkitDriverVersions
     CudaToDriverVersion[] = {
+        {12080, 525.60f, 528.33f},
+        {12070, 525.60f, 528.33f},
         {12060, 525.60f, 528.33f},
         {12050, 525.60f, 528.33f},
         {12040, 525.60f, 528.33f},


willyborn:
I noticed that in many tests, some basics are missing, like
- nullptr checks for addresses
- support for all types of arrays
- buffer leak checks

If desired, I can go over all tests and make sure those base tests are included.

willyborn:
Should empty arrays "af_create_handle(&array,0,NULL,f32)" also be supported?
The created array has dims=[0,1,1,1] and strides =(1,0,0,0) --> elements() = 0

The above is currently not supported by af_lookup.

src/backend/cpu/lookup.cpp
```diff
@@ -21,12 +24,13 @@ namespace cpu {
 template<typename in_t, typename idx_t>
 Array<in_t> lookup(const Array<in_t> &input, const Array<idx_t> &indices,
                    const unsigned dim) {
-    const dim4 &iDims = input.dims();
+    assert(dim <= 3);
```
edwinsolisf, line 27: Could you replace the `assert`s with the `ARG_ASSERT` macro to make it consistent with our current argument checking? For example, for this one: `ARG_ASSERT(1, (dim <= 3))`

willyborn: Changes done.
Rebased to master and force pushed.

edwinsolisf (approved):
Works, tested on Windows

src/backend/cpu/lookup.cpp
```diff
@@ -21,12 +22,13 @@ namespace cpu {
 template<typename in_t, typename idx_t>
 Array<in_t> lookup(const Array<in_t> &input, const Array<idx_t> &indices,
                    const unsigned dim) {
-    const dim4 &iDims = input.dims();
+    ARG_ASSERT(1, (af_dtype)dtype_traits<idx_t>::af_type != c32 &&
```
christophe-murphy, line 25: These checks are already performed in src/api/c/index.cpp

src/api/c/index.cpp
```diff
@@ -174,7 +179,13 @@ af_err af_lookup(af_array* out, const af_array in, const af_array indices,
         ARG_ASSERT(2, (idxType != c64));
         ARG_ASSERT(2, (idxType != b8));
 
-        af_array output = 0;
+        // Try to release ASAP in the hope that the data buffer, if any, will be
```
christophe-murphy, line 182: Have you seen a performance improvement by doing this?

willyborn: `af_array output=0;` is replaced by  `af_array output=nullptr;`
In the snapshot above, only the upper section of the update is shown.

Code wise, I assume both will have the similar instructions and certainly the same timing.  
Type wise, the first is an assignment of an integer to a pointer.  In my eyes a nullptr needs special attention/treatment, because it is frequently the cause of many crashes (seg faults).  Therefor they should be clearly visible iso hidden behind a standard integer initialization.

src/backend/cpu/lookup.cpp
christophe-murphy: OK, I will discuss with @umar456 how we should go about this.

src/api/c/index.cpp
christophe-murphy: Sorry, should have been more clear. I see you added extra logic to try to reuse a buffer where possible. I was wondering if you saw any benefit from doing so.

willyborn: While testing, I fell myself into the trap by reusing the out array in the next loop.  As result, I ran out of GPU memory.
I called however af_release_array at the end of the function, although not inside the loop, so I was sure that all memory was correctly freed.  I see regularly people complaining about the high memory usage of arrayfire, perhaps this is a possible cause.  By the way, nowhere in the documentation is mentioned that you should provided a nullptr iso and not the address of a real array.

Searching further for the cause of my out of memory situation, I remarked that the provided out array could be released at the start of the af_function, resulting in possible reuse if size permits.

There is 1 downside.  If the provided out array is not initialized (only af_array arr;) you will get an error because a random address will be used in the internal releasing.  


PS: I discovered some cases of non-initialization in other sections of arrayfire.  In case you guys, agree with my reasoning I will go over all functions including checks and extra tests to increase the robustness and certainty of result.  I already discovered 2 other cases where random results are returned (missing offset in OCL & sync error with JIT array in CPU).  More after this commit, since it is a further development on this.

umar456:
The initial fix looks good.

With regard to the af_lookup changes. I don't think we can take that as the way you have it. Even though you are not changing the API, the contract we have with the user changes. With most other ArrayFire C functions we accept uninitialized af_array objects(i.e. `af_array blah = 0;`)  and this implies that ArrayFire will allocate a new af_array. 

With the new function you are accepting an already allocated array. This is fine but you are changing its meaning and it would break user code. I suggest we make a new function if you want this type of functionality. Either af_lookup_v2 or af_lookup_inplace. This function willl write to an already allocated array and will not allocate internally. This would require you to make sure that the types and size/shape are acceptable. I am okay if you want to make this change but I would prefer it in a different PR.

willyborn:
OK.

I drop the release_array inside the function and return back to the original version.

edwinsolisf:
To move forward with this PR we have decided to keep the solution to the bug and the tests for it, so could just leave in the fix commit [e4de910](https://github.com/arrayfire/arrayfire/commit/e4de9103297281294176df8b831d53f694f46703 and the tests specific to this problem? I see [3e28fff](https://github.com/arrayfire/arrayfire/commit/3e28fff9f40ab0fe969cb953ea422ce1bf811166) modifies some of our previous tests/setup,  could you point which specific test checks this problem? 

Regarding the other changes, you may submit them in another PR so we can discuss such changes separately.

willyborn:
The existing tests are not modified.  It generates different types of arrays to be overwritten by the function.

I extended the concept for input arrays, since multiple functions fail when a none common array is provided as input array.  The most common errors are with sub arrays, empty arrays, JIT arrays, etc.  I added some methods to TestOutputArrayInfo for reporting purposes, like current array type generated.
To remain consistent with the current way of testing I thought that it would be best to extend the same idea by creating a new class for TestInputArrayInfo although with same usage as the existing TestOutputArrayInfo class.

On my local version, I started including the tests for other functions and already found following errors as a result:
- reduce is having problems with JIT arrays on CPU and sub arrays on OpenCL.

I did not continue, because I wanted first feedback on the lookup case (sub array problems on OpenCL).

I will submit a separate PR, after the new release is finished.
You can include only e4de910 and leave all the changes on tests out for 3.10 .
Later for 3.11 you can decide for a more general approach (also for lookup).  As mentioned for reduce, I am convinced that more errors will be discovered.  The same go for the argument checking at the internal level, which also generated multiple errors in the past (most recent join wrongly called by field example: PR3605).

edwinsolisf:
@willyborn Yeah, we prefer that being its own PR. Sounds good, I'll take that commit. I will also add some tests for lookup with different offsets.

edwinsolisf (approved):
Tests pass on all backends, checked machine learning examples using lookup as well

edwinsolisf:
Hey @willyborn, I mistakenly closed this PR, could you reopen it again? Thanks!

willyborn:
@edwinsolisf .  I like to reopen, although I do not see any option allowing me.  I hope this comment will reopen as a consequence.
Perhaps lack of permissions?

edwinsolisf:
I have recreated this PR in #3650 that includes and references your fix . I apologize for the incovenience this caused.

willyborn:
Thanks.

The important thing is that it is fixed.

Issue #3276, opened 2022-07-19
[BUG] Program crash with no error message when using Unified Backend af.lib


FrMar:
I am having troubles using the Unified Backend library `af.lib`, but my program is working when I am using other libraries.

Description
===========
My program is perfectly working if I am using `afcpu.lib`: the code I use is presented below with the output obtained. But it won't work if I am using `af.lib`. I also put the code and the output right after.

Reproducible Code and/or Steps (working)
------------------------------
1. Add `$(AF_PATH)/include;` to _Project Properties_ -> _C/C++_ -> _General_ -> _Additional Include Directories_.
2. Add `$(AF_PATH)/lib;` to _Project Properties_ -> _Linker_ -> _General_ -> _Additional Library Directories_.
3. Add `afcpu.lib;` to _Project Properties_ -> _Linker_ -> _Input_ -> _Additional Dependencies_.

Here is the code used:

    #include <iostream>
    #include <arrayfire.h>

    int main()
    {
        std::cout << "Main running..." << std::endl;
        
        std::cout << "Available backends: " << af::getAvailableBackends() << std::endl;

       for (int i = 0; i <  af::getDeviceCount();  i++) {
            af::setDevice(i);
            af::info();
        }

        return 0;
    }

Here is the output:

    Main running...
    Available backends: 1
    ArrayFire v3.8.2 (CPU, 64-bit Windows, build 5752f2dc)
    [0] Intel: Intel(R) Core(TM) i5-3337U CPU @ 1.80 GHz

Reproducible Code and/or Steps (not working)
------------------------------

1. Add `$(AF_PATH)/include;` to _Project Properties_ -> _C/C++_ -> _General_ -> _Additional Include Directories_.
2. Add `$(AF_PATH)/lib;` to _Project Properties_ -> _Linker_ -> _General_ -> _Additional Library Directories_.
3. Add `af.lib;` to _Project Properties_ -> _Linker_ -> _Input_ -> _Additional Dependencies_.

Here is the code used:

    #include <iostream>
    #include <arrayfire.h>

    int main()
    {
        std::cout << "Main running..." << std::endl;

        try
        {
            std::cout << "Available backends: " << af::getAvailableBackends() << std::endl;
        }
        catch (af::exception& e)
        {
            std::cout << "Caught exception when trying reading backends" << std::endl;
            std::cerr << e.what() << std::endl;
        }

        af::setBackend(AF_BACKEND_CPU);
[...]

Issue #2346, opened 2018-11-13
Create a better error message for unified backend failures


umar456:
The unified backend loads several libraries and its difficult to figure out what is the cause of the failure. Th

```
ERROR app - ArrayFire Exception (Failed to load dynamic library. :501):
Error in af_err unified::AFSymbolManager::setBackend(af::Backend)
In file src/api/unified/symbol_manager.cpp:239
Failed to load dynamic library. See http://www.arrayfire.com/docs/unifiedbackend.htm for instructions to set up environment for Unified backend.
```

We should return an error message that describes the source of the failure. af_get_last_error should return this message and the exception should have this set as the message.

9prady9:
@umar456 With stack trace, trace output showing which backend library failed to load. Can we save this is complete or are we still missing something ?

PR #3617, opened 2024-10-18
unified reports filled af::exception errors.

willyborn:
The unified version of arrayfire, an std::exception is throw while collecting the arrayfire global error message.
Now the arrayfire error message is collected from the correct platform, resulting in the throwing of an af::exception filled with info.

When checking the unified tests: 18 tests failed because an unexpected exception was thrown

Description
-----------
When collecting the error message from the underlying platform, the address of the corresponding function was incorrect, resulting in an access violation.
Result: 
- Yes, an exception was thrown
- No, af::exception was not thrown, so no info on the real error

Fixes: #3276 #2346

Changes to Users
----------------
None

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/api/unified/error.cpp
+++ b/src/api/unified/error.cpp
@@ -42,7 +42,7 @@ void af_get_last_error(char **str, dim_t *len) {
         typedef void (*af_func)(char **, dim_t *);
         void *vfn    = LOAD_SYMBOL();
         af_func func = nullptr;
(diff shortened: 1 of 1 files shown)


christophe-murphy (approved):
Good catch, thanks!

PR #3627, opened 2024-12-08
Correct the conversion from float/double to half on CUDA

willyborn:
Corrects the conversion from a float/double to half.
This affects following functions: af::min, af::max, af::dot, af::mean, af::mean_var, af::topk, af::var for those GPU, where CUDA supports slow-rate FP16 (type __half is supported, although compute is in float)

Description
-----------
In the device function float2half_impl the float/double is converted to an unsigned integer on which bit operations are performed.
Finally we get a uint16 containing the bit presentation of the same value in native_half_t format.
During the return, an extra implicit casting was performed, this time from uint16 to the native_half_t format.

GPU supporting fully __half (type & compute) will not convert a float to __half, and will not show this error.
My GPU is a GTX1080 with CC6.1, with low-rate FP16 support.

Reproducible code, CUDA engine.
```
{
    float harr[] = {1, 2};
    af::array arr(2, harr);
    af::array arr_f16 = arr.as(f16);
    af::array out = min(arr_f16);
	
	af::print("array_f16", arr_f16);
	af::print("min of array_f16", out);
}
```

Output:
```
array_f16
[2 1 1 1]
   Offset: 0
   Strides: [1 2 2 2]
    1.0000
    2.0000

min of array_f16
[1 1 1 1]
   Offset: 0
   Strides: [1 1 1 1]
15360.0000
```

* Is this a new feature or a bug fix?   Bug fix
* Why these changes are necessary.  This corrects following tests: test_var_cuda, test_topk_dua, test_reduce_cuda, test_blas_cuda, test_dot_cuda, test_mean_cuda & test_meanvar_cuda.
* Potential impact on specific hardware, software or backends: All GPU's having slow_rate FP16 support
* Can this PR be backported to older versions?  Yes
* Future changes not implemented in this PR.  None

Changes to Users
----------------
No changes to API.

Checklist
---------

- [x] Code compiles
- [x] Tests pass, not tested on ONEAPI
- [ ] Functions documented

### DIFF
--- a/src/backend/common/half.hpp
+++ b/src/backend/common/half.hpp
@@ -87,6 +87,7 @@ using uint16_t = unsigned short;
 #define AF_CONSTEXPR constexpr
 #else
 #include <af/compilers.h>
+#include <algorithm>
 #include <cmath>
 #include <cstdint>
 #include <cstring>
@@ -245,9 +246,9 @@ AF_CONSTEXPR __DH__ native_half_t int2half_impl(T value) noexcept {
 /// \return binary representation of half-precision value
 template<std::float_round_style R = std::round_to_nearest>
 __DH__ native_half_t float2half_impl(float value) noexcept {
-    uint32_t bits = 0;  // = *reinterpret_cast<uint32*>(&value);
-                        // //violating strict aliasing!
-    std::memcpy(&bits, &value, sizeof(float));
+    alignas(std::max(alignof(uint32_t), alignof(float))) float _value = value;
+    uint32_t bits = *reinterpret_cast<uint32_t*>(&_value);
+
     constexpr uint16_t base_table[512] = {
         0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
         0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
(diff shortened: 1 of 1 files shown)


src/backend/common/half.hpp
```diff
@@ -379,9 +387,12 @@ __DH__ native_half_t float2half_impl(float value) noexcept {
 /// \return binary representation of half-precision value
 template<std::float_round_style R>
 __DH__ native_half_t float2half_impl(double value) {
-    uint64_t bits{0};  // = *reinterpret_cast<uint64*>(&value);		//violating
-                       // strict aliasing!
-    std::memcpy(&bits, &value, sizeof(double));
+    union {
```
edwinsolisf, line 390: According to [cppreference](https://stackoverflow.com/questions/67904738/is-it-undefined-behaviour-to-read-a-different-member-than-was-written-in-a-union), this type of conversion is Undefined Behavior. I think the more standard compliant way to do it would be using `reinterpret_cast`.

willyborn: All explicit conversions are now updated to reinterpret_cast.

edwinsolisf (approved):
Works, tested on Windows with an RTX 3070 Ti Mobile and on Ubuntu 22.04 LTS with a Tesla P4

christophe-murphy:
When compiling ArrayFire with CUDA Compute Capability 6.1 I am still getting the wrong answer for min.

Issue #3652, opened 2025-04-04
[Build] CUDA nvrtc-builtins not found in windows


edwinsolisf:
Description
===========

The `find_cuda_helper_libs(nvrtc-builtins)` from [`src/backend/cuda/CMakeLists.txt:52`](https://github.com/arrayfire/arrayfire/commit/a13dcb64287355ee100351827f069d4bbf5fe471#diff-87cdf3f7c76eec3a800bdd92a6737080038b14b4c55ac61f1a4657fb39f8b4c6R53) does not find the nvrtc-builtins library in windows as the library is prepended with cuda toolkit version. That is, cmake is searching for `nvrtc-builtins.lib` but it does not exist in windows. The alternative file to search is `nvrtc-builtins_static.lib`. See [nvidia-docs](https://docs.nvidia.com/cuda/nvrtc/#installation)

Without [this modification](https://github.com/arrayfire/arrayfire/commit/a13dcb64287355ee100351827f069d4bbf5fe471#diff-87cdf3f7c76eec3a800bdd92a6737080038b14b4c55ac61f1a4657fb39f8b4c6R53) the cuda backend builds without any issue on Windows.

Proposed solution is to not add this dependency when building on windows

Error Log
---------

```
CUDA_nvrtc-builtins_LIBRARY is set to Not Found
```

Build Environment
-----------------
Compiler version: 
Operating system: 
Build environment: 
CMake variables:

PR #3653, opened 2025-04-15
if not Win check added

ved1beta:
Skip nvrtc-builtins dependency on Windows

This PR modifies the CMake configuration to handle the nvrtc-builtins library
differently on Windows platforms where the library naming convention differs.

Description
-----------
This is a bug fix that addresses the issue with CUDA nvrtc-builtins library
not being found on Windows systems. The root cause is that Windows CUDA toolkit
prepends version information to the library name, making the standard library
search fail.

The changes:
- Add platform-specific condition to skip nvrtc-builtins on Windows
- Maintain existing behavior on other platforms
- Fix build issues on Windows without affecting other platforms

Impact:
- Windows builds will now complete successfully
- No impact on other platforms or backends
- No changes to existing functionality or API

This PR can be backported to older versions as it's a build system fix that
doesn't affect the core functionality.

Fixes: #3652

Changes to Users
----------------
No changes required for users. This is a build system fix that doesn't affect
the API or functionality. Existing code will continue to work as before.

Checklist
---------
- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/cuda/CMakeLists.txt
+++ b/src/backend/cuda/CMakeLists.txt
@@ -49,8 +49,12 @@ set(CUDA_architecture_build_targets "Auto" CACHE
   STRING "The compute architectures targeted by this build. (Options: Auto;3.0;Maxwell;All;Common)")
 
 find_cuda_helper_libs(nvrtc)
-find_cuda_helper_libs(nvrtc-builtins)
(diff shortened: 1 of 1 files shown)


christophe-murphy:
Thanks for this. We are currently overhauling the Windows installer build process so will be making some more changes and fixes soon including for this issue.

PR #3658, opened 2025-06-03
Fixes to the indexed reduce function for the cpu, opencl and cuda back ends

christophe-murphy:
These back ends were incorrectly assuming a linear array when finding the max/min value and its index. This causes an issue with a sub-array.

Description
-----------
In the case of the cuda and opencl back end this was just for the cpu-fallback methods which are used when the total number of elements in the array is less than or equal to 4096. There was no issue with the oneapi back end because it does not have this fall-back method. The fix for the cuda and opencl back ends was to only use the fall-back method for linear arrays. This is to avoid doing a complex sub-array copy operation from device to host memory.

Changes to Users
----------------
af_imax_all / af_imin_all functions now works for sub-arrays

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/cpu/ireduce.cpp
+++ b/src/backend/cpu/ireduce.cpp
@@ -58,6 +58,7 @@ void rreduce(Array<T> &out, Array<uint> &loc, const Array<T> &in, const int dim,
 
 template<af_op_t op, typename T>
 T ireduce_all(unsigned *loc, const Array<T> &in) {
+    in.eval();
     getQueue().sync();
 
     af::dim4 dims    = in.dims();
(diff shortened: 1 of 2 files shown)


syurkevi (dismissed):

src/backend/cpu/ireduce.cpp
```diff
@@ -63,6 +63,7 @@ T ireduce_all(unsigned *loc, const Array<T> &in) {
     af::dim4 dims    = in.dims();
     af::dim4 strides = in.strides();
     const T *inPtr   = in.get();
+    dim_t idx = 0;
```
willyborn, line 67: The  getQueue().sync(); of line 62 has to move to line 66.

For incomplete JIT arrays, the evaluation is initiated on line 65 in.get().  The sync has to be afterwards to have this completed before the following loop.

Test case:
array jit(dim4(2),{1.0f, 2.0f};
jit += constant(1.0f,dim4(2));
float val; unsigned idx;
min(&val, &idx, jit);

The min function will be calculated on the original content, without having added the 1.0;

PR #3671, opened 2025-06-23
Fix for evaluation containing array and its inverse.

christophe-murphy:
A bug was found in the jit calculation tree where under certain circumstances the incorrect result would be found when both an array and a transpose of the same array are used in the same evaluation. When the transpose method is used on an array the treatment depends on the structure of the jit tree of that array. In the particular case that the array is linear (e.g. not a sub-array) and is not simply a buffer (i.e. a combination of buffers and/or scalars with some operators) a moddim node will be added to the root of the tree with the new dimensions. When the tree is evaluated the dimensions of the child buffer(s) of a moddim node are changed and the moddim node is deleted. If an evaluation happens to include both an array and the transpose of that same array then when the moddim node is applied it changes the dimensions of the children for the remaining lifetime of the evaluation including any subsequent use of these nodes. As a result if the transpose of an array is used in an expression followed by the array itself the second instance will also be transposed.

In this fix instead of creating a moddims node a deep copy of the calculation tree is made at that point with the transpose applied to the child buffer nodes. These buffer node copies will have the transposed dimensions and strides but will still point to the same memory location for the data so no copy of the underlying data needs to be made.

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/common/jit/BufferNodeBase.hpp
+++ b/src/backend/common/jit/BufferNodeBase.hpp
@@ -119,6 +119,18 @@ class BufferNodeBase : public common::Node {
         }
         return false;
     }
+
+    virtual void modDims(const af::dim4 &newDim) override {
+        af::dim4 strides(1, 1, 1, 1);
+        for(dim_t i = 1; i < 4; ++i) {
+            strides[i] = strides[i - 1] * newDim[i - 1];
+        }
+
+        for(dim_t i = 0; i < 4; ++i) {
+            m_param.dims[i] = newDim[i];
+            m_param.strides[i] = strides[i];
+        }
+    }
 };
 
 }  // namespace common
--- a/src/backend/common/jit/Node.hpp
+++ b/src/backend/common/jit/Node.hpp
@@ -13,6 +13,7 @@
 #include <optypes.hpp>
 #include <types.hpp>
 #include <af/defines.h>
+#include <af/dim4.hpp>
 
 #include <nonstd/span.hpp>
 #include <algorithm>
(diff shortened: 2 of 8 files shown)


umar456 (approved):

edwinsolisf:
I tested the effects of this PR on the OpenCL backend on the Intel Arc B580 by comparing the output of the jit kernels from the moddims and the jit test. (Note: I removed the cache kernels to make sure they are regenerated from scratch)

These are the results for moddims:
[test_output_after_changes.txt](https://github.com/user-attachments/files/21202360/test_output_after_changes.txt)
[test_output_before_changes.txt](https://github.com/user-attachments/files/21202361/test_output_before_changes.txt)

These are the results for jit:
[before_jit.txt](https://github.com/user-attachments/files/21202814/before_jit.txt)
[after_jit.txt](https://github.com/user-attachments/files/21202815/after_jit.txt)


Things to note:

- The tests pass in both cases
- Some of the jit kernels are different after the changes. The main difference observed in the moddims test is that more __noop operations are added which from the rough test runtime seem to get optimize or have minimal impact in the Release build. (More test to confirm optimization is needed). However, on the jit test, specifically the added test evaluateBothArrayAndItsTranspose, with the changes two more concise kernels are created instead of three.
  - The size of the compiled kernels are very similar but very slightly larger (which is expected?). While I could not compare individual kernels due to the name hashing, I compared the total space occupied by all the cached kernels generated. In my case I observed that it increased from 31.8 MB (33,386,496 bytes) to 31.9 MB (33,488,896 bytes), an increase of 0.3%.


I also tested these changes on the oneAPI backend and they also pass the tests (no jit kernels generated)

PR #3672, opened 2025-06-27
Fix half precision pow function for oneAPI back end

christophe-murphy:
An incorrect power function was being used for half precision variables when building the JIT kernel for the oneAPI back end which lead to a fallback to an integer power function causing the result to be rounded. This fixes that.

Checklist
---------

- [x] Code compiles
- [x] Tests pass

### DIFF
--- a/src/backend/oneapi/binary.hpp
+++ b/src/backend/oneapi/binary.hpp
@@ -9,6 +9,9 @@
 
 #pragma once
 #include <optypes.hpp>
+#include <common/half.hpp>
+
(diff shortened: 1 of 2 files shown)


edwinsolisf (dismissed):
The same must be done on the OpenCL Backend

christophe-murphy:
@edwinsolisf Could you re-approve. Didn't see that you had already put in a PR for the OpenCL version. Thanks

edwinsolisf (approved):

PR #3687, opened 2025-08-04
Fixes to mean and better precision on sum (cpu)

willyborn:
When strided weights are used, the mean function typically returned random results.
The produced result of mean is no longer dependent on platform or processing style (linear, strided, on CPU or GPU)

Description
-----------
https://github.com/arrayfire/arrayfire/commit/b74f28f6b9137fadaa42538aa24476fc3ae6a491 [sum on CPU now has the same precision as other platforms](https://github.com/arrayfire/arrayfire/commit/b74f28f6b9137fadaa42538aa24476fc3ae6a491)
* Introduction of Kahan summation algorithm for series.  This algorithm reduced the error when a series of summations are performed with values of varying sizes.
With this introduction, the difference in result of a linear array and strided array is minimal.  The precision in the tests are as a result improved.
As result, the precision of many tests can be improved now since all platforms generate similar results, independent from parallel or serial processing.

https://github.com/arrayfire/arrayfire/commit/9307d6f08b740f694c9f8df19b1e23727e2ed9b0 [Fixes to mean, on all platforms](https://github.com/arrayfire/arrayfire/commit/9307d6f08b740f694c9f8df19b1e23727e2ed9b0)
All: 
* Introduction of Kahan summation algorithm for series.  This algorithm reduced the error when a series of summations are performed with values of varying sizes.
With this introduction, the difference in result of a linear array and strided array is minimal.  The precision in the tests are as a result improved.
* Negative weights can no longer generate a division by zero.
* In most cases sub-arrays would generated random results with strided weights.

CPU:
   - the idxs of the weights array is now calculated according to its own strides iso the strides of the input array.

CUDA:
   - Param<T> were used without full initialisation.  Because the construction of Param<T>’s are deprecated, they are replaced by the createEmptyArray() constructions.
   - Idxs of weights are now calculated according to its own strides.  Idem for the offsets.
   - Small arrays are processed on the CPU.  Strides arrays are now processed on the GPU, independent from size since the CPU version only handles linear arrays.

ONEAPI:
   - Idxs of weights are now calculated according to its own strides.  Idem for offsets.
   - Param<T> were used without full initialisation.  Because the construction of Param<T>’s are deprecated, they are replaced by the createEmptyArray() constructions.
   - Small arrays are processed on the CPU.  Strides arrays are now processed on the GPU, independent from size since the CPU version only handles linear arrays.
   - Offsets are handle for the linear processing.

OPENCL:
   - Idxs of weights are now calculated according to its own strides.  Idem for offsets.
   - When converting a linear array to a vector, the weights offsets were overwritten.
   - The size of the tmpOut and tmpWeight arrays are now correctly sized.  Intermediate results were overwritten resulting in a mean of a part of the array.
   - Small arrays are processed on the CPU.  Strides arrays are now processed on the GPU, independent from size since the CPU version only handles linear arrays.
   - Param<T> were used without full initialisation.  Because the construction of Param<T>’s are deprecated, they are replaced by the createEmptyArray() constructions.

TEST:
    - extra tests added on all temporary formats.
    - allowed fault tolerance for tests is lowered, since now the correct mean is calculated for all backends.
    - MeanAllTest now uses random data iso constants for testing.  This blocked the detection of partial processing of the input.

* Is this a new feature or a bug fix?   BUG
* Why these changes are necessary.  CORRECT RESULTS WERE NOT GUARANTEED
* Potential impact on specific hardware, software or backends.  NONE
* New functions and their functionality.  NONE
* Can this PR be backported to older versions?  NO
* Future changes not implemented in this PR.  NO
Fixes: 

Changes to Users
----------------
Better precision in all circumstances, and independent form platform or array format.

Checklist
---------

- [x] Code compiles
- [x] Tests pass
- [ ] Functions documented

### DIFF
--- a/src/backend/cpu/kernel/mean.hpp
+++ b/src/backend/cpu/kernel/mean.hpp
@@ -38,6 +38,45 @@ struct MeanOp {
     }
 };
 
+template<typename Ti, typename To, typename Tw>
+struct MeanOpWithCorrection {
+    common::Transform<Ti, To, af_add_t> transform;
+    To runningMean;
+    Tw runningCount;
+    To correction;
+    MeanOpWithCorrection(Ti mean, Tw count)
+        : transform()
+        , runningMean(transform(mean))
+        , runningCount(count)
+        , correction(scalar<To>(0)) {}
+
+    void operator()(Ti newMean, Tw newCount) {
+        runningCount += newCount;
+        if ((newCount != 0) || (runningCount != 0)) {  //
+            // Since only 1 pass is used, the rounding errors will become
+            // important because the longer the serie the larger the
+            // difference between the 2 numbers to sum See:
+            // https://en.wikipedia.org/wiki/Kahan_summation_algorithm to
+            // reduce this error
+            // correction is zero for the first time around
+            To y =
+                (transform(newMean) - runningMean) * (newCount / runningCount) -
+                correction;
+            // Alas, runningMean is big, y small, so low-order digits of y
+            // are lost
+            To t = runningMean + y;
+            // (t - runningMean) cancels the high order part of y
+            // subtracting y recovers negative (low part of y)
+            correction = (t - runningMean) - y;
+            // Algebraically, correction should always be zero.  Beware
+            // overly-agressive optimizing compilers!
+            runningMean = t;
+            // Next time around, the lost low part will be added to y in a
+            // fresh attempt
+        }
+    }
+};
+
 template<typename T, typename Tw, int D>
 struct mean_weighted_dim {
     void operator()(Param<T> output, const dim_t outOffset,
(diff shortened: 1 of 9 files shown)


christophe-murphy:
Thanks for this willyborn. FYI, our Buildbot CI system failed to build the PR with the following error:

`/var/lib/buildbot/workers/ubuntu2204-cuda-toolkit-12-9/linux-all/build/src/backend/cuda/kernel/mean.hpp(65): error: more than one operator "*" matches these operands:
            built-in operator "arithmetic * arithmetic"
            function "operator*(const __half &, const __half &)" (declared at line 223 of /usr/local/cuda/include/cuda_fp16.hpp)
            operand types are: __half * float
          To y = (rhs - (*lhs)) * r_wt / (*l_wt) - (*c);`

willyborn:
I'm on ubuntu 22.04LTS with GCC, and I can not reproduce any warning/error you mentioned.  
As result, I encapsulated every element inside the stable_mean so that misunderstandings by the compiler now are eliminated.  I hope this solves the error.

PS: Just to be certain I performed the same change for all platforms.

christophe-murphy:
This is happening when NVCC is compiling the stable_mean CUDA device kernel. Which version of CUDA are you using? Could you try with 12.9

willyborn:
I am using CUDA 12.9.86 on ubuntu 24.04.3 LTS kernel 6.14.
I have a GTX1080 which is on architecture 6.1 (stores halves, calculates in float).

As soon as I included all major architectures to the nvcc options, I got the error also, which is now also fixed and committed.

PR #3705, opened 2026-05-13
build: add missing direct includes

LwhJesse:
This PR adds missing direct includes needed by the files that use those declarations.

Changes:
- include <cstdint> in deterministicHash.cpp for uint8_t
- include <thrust/distance.h> in set.cu for thrust::distance

This is an include-hygiene build fix and does not change runtime behavior.


jailop:
I hit the same issue (gcc15). Added the cstdint header in order to compile

PR #3708, opened 2026-06-08
[ROCm] Add HIP backend (full gtest suite GPU-green, incl. sparse)

jeffdaily:
## Summary

ArrayFire has no native HIP/ROCm backend; AMD GPUs are reachable only through the OpenCL path. This adds a HIP backend as a new sibling of the CUDA backend (`src/backend/hip`, cloned from `src/backend/cuda`), keeping the NVIDIA/CUDA path byte-for-byte unchanged. The HIP backend reports `AF_BACKEND_CUDA` and builds as the library `afcuda`, so the unified dispatcher and the gtest suite treat it as the CUDA-model backend on AMD. There is no public ABI or enum change; `src/backend/{cuda,cpu,opencl,oneapi}`, `src/api`, and `include/` are untouched, and 380 of the 386 changed files are the additive `src/backend/hip` tree.

Enable it with `-DAF_BUILD_HIP=ON` (mutually exclusive with `AF_BUILD_CUDA`), and select the target GPU with `-DCMAKE_HIP_ARCHITECTURES=<arch>` (e.g. `gfx90a`, `gfx1100`); it defaults to a common arch when unset.

## What is implemented

- The runtime-JIT engine ported from NVRTC + the CUDA driver link step to hipRTC's direct-code-object flow (`hiprtcGetCode` -> `hipModuleLoadData`), with the arch taken from the device `gcnArchName` (`--offload-arch`).
- Library swaps: hipBLAS, hipSOLVER (via its cuSOLVER-compatible `hipsolverDn*` API), hipFFT, hipSPARSE (generic API + legacy sort/conversion + `csrgeam2`), and rocThrust / hipCUB.
- The full sparse subsystem on hipSPARSE, the int8 (schar) gemm path (int8 x int8 -> int32 accumulate, cast to f32), and FreeImage image IO.
- Wave64 (CDNA) and wave32 (RDNA) handling: the warp width comes from per-arch `__GFX*__` device guards and the runtime `warpSize`, never a hardcoded constant.
- Complex element types kept as plain PODs on HIP (host and JIT paths) so the project complex operators are unambiguous.

The commit message describes the recommended review order and the per-file rationale (hipRTC vs NVRTC device-code rules, the void* handle/descriptor aliasing, the complex-kernel fixes).

## Validation

Full `CUDA.*` gtest suite (`ctest -R '_cuda$' -j1`) on AMD GPUs:

| GPU | Arch | Result |
| --- | --- | --- |
| MI250X | gfx90a (CDNA2, wave64), Linux, ROCm 7.2.1 | 132 / 132 |
| Radeon Pro W7800 | gfx1100 (RDNA3, wave32), Linux, ROCm 7.2.1 | 132 / 132 |
| Radeon RX 9070 XT | gfx1201 (RDNA4, wave32), Windows, ROCm 7.14 | 128 / 131 |

The three gfx1201 exceptions are environmental rather than backend logic: image IO was not built into that configuration (FreeImage), a Windows ROCm sparse-library issue in `csrgeam2`, and a test-harness timeout.

The NVIDIA/CUDA path is unchanged and not affected by this backend.

## Test Plan

```
# Build (per arch), CPU backend kept on as the in-process gtest reference
cmake -S . -B build-hip -DAF_BUILD_HIP=ON -DAF_BUILD_CUDA=OFF \
  -DAF_BUILD_CPU=ON -DAF_WITH_IMAGEIO=ON \
  -DCMAKE_HIP_ARCHITECTURES=gfx90a \
  -DCMAKE_HIP_COMPILER=/opt/rocm/llvm/bin/clang++
cmake --build build-hip -j16

# Run the CUDA-tagged suite on one isolated GPU
HIP_VISIBLE_DEVICES=0 ctest --test-dir build-hip -R '_cuda$' -j1 --output-on-failure
# => 100% tests passed, 0 tests failed out of 132   (gfx90a, gfx1100)
```

Authored with the assistance of Claude (Anthropic).


villekf:
@jeffdaily did you use the included CMakeLists also on Windows? I either get errors about missing subsystem during linking or undefined symbols, such as `hipblasSgetrfBatched` (and other batched ones) if I specifically define Windows subsystem. I'm using gfx1100 (7900 XT). 

I specifically defined Clang as the compiler in CMake as the default is VS.

EDIT: I'm using HIP SDK 7.1.1. Also, what's the minimum supported ROCm?

jeffdaily:
Thanks for trying it. The short version: on Windows we build and validate against a full ROCm runtime from TheRock, not the packaged AMD HIP SDK, and that is the difference you are hitting.

**`hipblasSgetrfBatched` and the other batched symbols.** Those are the batched LU functions (`getrfBatched` / `getrsBatched`) used by `solve.cu`. hipBLAS only exports them when it is built with rocSOLVER support. The hipBLAS in TheRock exports all of them (I confirmed `hipblasS/D/C/Zgetrf/getrsBatched` are present in its `hipblas.dll`); the hipBLAS in HIP SDK 7.1.1 does not, which is why the link fails. You can confirm with `dumpbin /exports <HIP SDK>\bin\hipblas.dll | findstr getrfBatched` (it will come back empty).

**Recommended setup on Windows: use TheRock.** It is the ROCm build we test against and the one I would recommend for Windows generally (https://github.com/ROCm/TheRock). The runtime and dev libraries are published as Python wheels on the nightly index (`https://rocm.nightlies.amd.com/whl-multi-arch/`); point CMake at the installed `_rocm_sdk_devel` tree as `CMAKE_PREFIX_PATH` and use its `clang++` for the compilers, with the Ninja generator (not the VS toolset). The exact configure line we use, retargeted to gfx1100, where `<_rocm_sdk_devel>` is the installed TheRock devel tree:

```
cmake -S arrayfire -B build-gfx1100 -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_C_COMPILER="<_rocm_sdk_devel>/lib/llvm/bin/clang++.exe" \
  -DCMAKE_CXX_COMPILER="<_rocm_sdk_devel>/lib/llvm/bin/clang++.exe" \
  -DCMAKE_HIP_COMPILER="<_rocm_sdk_devel>/lib/llvm/bin/clang++.exe" \
  -DCMAKE_PREFIX_PATH="<_rocm_sdk_devel>" \
  -DCMAKE_HIP_ARCHITECTURES=gfx1100 \
  -DAF_BUILD_HIP=ON -DAF_BUILD_CUDA=OFF \
  -DAF_BUILD_CPU=ON -DAF_BUILD_OPENCL=OFF -DAF_BUILD_ONEAPI=OFF \
  -DAF_BUILD_UNIFIED=ON -DAF_BUILD_EXAMPLES=OFF -DAF_BUILD_FORGE=OFF \
  -DAF_WITH_CUDNN=OFF -DAF_WITH_IMAGEIO=OFF -DAF_BUILD_DOCS=OFF \
  -DAF_BUILD_TESTS=ON -DAF_STACKTRACE_TYPE=None -DAF_TEST_WITH_MTX_FILES=OFF
```

Then `ninja -C build-gfx1100`. Two of those toggles are Windows-specific workarounds rather than requirements: `AF_STACKTRACE_TYPE=None` avoids the Boost WinDbg stacktrace pulling in conflicting Windows API headers, and `AF_TEST_WITH_MTX_FILES=OFF` sidesteps the >260-char path limit hit while CMake downloads the sparse matrix files. `AF_WITH_IMAGEIO=OFF` only because we did not have FreeImage installed; turn it on if you have it.

**Subsystem linker error.** If it persists once the symbols resolve against TheRock's hipBLAS, force the console subsystem on the test/example targets (`-Xlinker /subsystem:console`); the executables are all `main()`-based.

**ROCm versions we validated.** Linux (gfx90a, gfx1100) on ROCm 7.2.1; Windows (gfx1101, gfx1201) on TheRock ROCm 7.14. The port uses the hipBLAS v2 compute-type API and the hipSPARSE generic API, so ROCm 6.0 is the floor, but the validated configurations are the ones above.

jeffdaily:
Small correction on the examples: I had `AF_BUILD_EXAMPLES=OFF` in those commands purely out of habit, not because anything is broken. I verified the examples build and run cleanly (helloworld and the C-API example both run correctly on a gfx1201 / RX 9070 XT, including FFT, scan, and sort). You can leave `AF_BUILD_EXAMPLES=ON` (the upstream default).

One thing to know if you build in-tree: because this port builds the `afcuda` library from the HIP backend with `AF_BUILD_CUDA=OFF`, the examples' `find_package(ArrayFire)` reports the CUDA backend as not found, so the per-example `*_cuda` targets are not generated in the combined build. Building the examples against an installed ArrayFire (the documented path) is unaffected and produces all of them.

villekf:
Thanks for the answers, I now managed to build the HIP version of AF. I also managed to get my own code running now. Going to test this in Linux-environment later as well. Surprisingly big differences in OpenCL vs HIP in terms of computation time.

I had to add this to the HIP backend's CMakeLists.txt though to avoid the subsystem issue:
`if(WIN32)
  target_link_options(afcuda PRIVATE "-shared")
endif()`

I think you can avoid the issues with Boost by adding: `add_compile_definitions(BOOST_USE_WINDOWS_H)`.

jeffdaily:
Thanks for the detailed follow-up, and for the two CMake fixes. I have folded both into the PR (latest commit):

- **afcuda `-shared` on Windows**, under `if(WIN32)`. The clang driver does not reliably infer the DLL subsystem here; forcing shared-library link mode fixes it without affecting the Ninja path.
- **`BOOST_USE_WINDOWS_H`**, added to the Boost compile definitions on Windows. This is the better fix: it lets Boost.Stacktrace's windbg backend include `<windows.h>` directly instead of self-declaring the Win32 API through Boost.WinAPI. I confirmed that without it the default `AF_STACKTRACE_TYPE=Windbg` fails to compile (`conflicting types for 'CreateFileA'`, `ReadFile`, etc. in `cufft.cu.obj`), and with it the full backend builds clean. So Windows users no longer need the `AF_STACKTRACE_TYPE=None` workaround I mentioned earlier; the upstream default works.

On the OpenCL vs HIP timing difference: I measured it to put numbers on it. Since you are on a gfx1100 (RDNA3), I ran it on a gfx1101 (RDNA3, Radeon PRO V710), the closest part I have - same machine, same ArrayFire build, both backends, the in-tree `examples/benchmarks`:

| Workload | HIP | OpenCL | HIP vs OpenCL |
|---|---|---|---|
| GEMM f32 (peak GFLOPS) | ~9200 | ~8000 | ~1.15x |
| GEMM f16 (peak GFLOPS) | ~52000 | ~4700 | ~11x |
| FFT 2D | - | - | ~1.3-2.8x |
| pi (Monte Carlo) | - | - | parity |
| conjugate gradient (dense) | - | - | OpenCL ~3x |

The difference is concentrated in GEMM, and it is a math-library story rather than a runtime one: the HIP backend's GEMM goes through rocBLAS/hipBLASLt and FFT through hipFFT, while the OpenCL backend uses CLBlast and clFFT. The big f16 gap is rocBLAS driving the RDNA3 WMMA path (~52 TFLOPS) where CLBlast has no matching tensor path (~4.7 TFLOPS). FFT favors HIP modestly, the JIT-bound pi kernel is parity, and dense CG is the one case OpenCL came out ahead. For reference, on a gfx1201 (RDNA4) the same f16 GEMM gap is even larger in absolute terms (hipBLASLt ~143 TFLOPS vs CLBlast ~28), but the shape is identical on both: GEMM-dominated, driven by the underlying BLAS. So the big delta you saw is expected, and most of it is the f16/f32 GEMM path.

villekf:
FYI: I tested this on Linux (Cray) environment as well. With ROCm 7.0.3 everything compiles successfully, but ROCm 6.4.4 doesn't work due to issues with hipblasDatatype_t or hipDataType, don't remember anymore exactly. So it seems that ROCm 7.x works on Linux. The Cray environment officially only supports ROCm 6.x at the moment, but with 7.0.3 the code did at least compile fine. Haven't had time to run anything yet though.

jeffdaily:
You are right, and this corrects what I wrote earlier in this thread: the floor is ROCm 7.0, not 6.0. The backend's hipblasGemmEx call takes hipDataType and hipblasComputeType_t, the hipBLAS 3 signature that ships in ROCm 7.0; ROCm 6.x has hipBLAS 2 where GemmEx takes hipblasDatatype_t, which is the compile error you hit on 6.4.4. I have stated the ROCm 7.0 minimum in the build docs (6800d5586). Thanks for the Cray data point confirming it.

melonakos:
Thank you, @jeffdaily, and thanks @villekf for the Windows and Cray testing. We are not going to carry a HIP backend in mainline releases, so the commits now live unchanged on https://github.com/arrayfire/arrayfire/tree/experimental/hip, which the README will point to; closing this PR in favour of that branch.

jeffdaily:
*Note: this reply was drafted by an AI assistant.*

Thank you, @melonakos, and thanks again @villekf for the Windows and Cray runs. Hosting the backend on `experimental/hip` works well for us, and we are glad to see it kept together with the project.

We would like to keep the branch buildable as master moves. Would you prefer we send future updates (merges of master into the HIP backend, plus fixes for new ROCm releases) as pull requests with `experimental/hip` as the base branch? If you would rather handle merges from master yourselves and only hear from us about HIP-specific fixes, that works too.

Either way, the branch this PR was opened from will stay identical to `experimental/hip`, so there is one set of commits to point at.

